PR #17, opened 2012-10-02
Move referer and user agent retrieval to if-statement header

lennartdiedrich:
Currently it is not possible to access the default Auth Component using `Shopware()->Auth()` before the `onPreDispatchBackend` of the Auth-Plugin has been executed. The instantiation of the Auth Component using `onInitResourceAuth` and `onInitResourceBackendSession` requires the `$request` property to be set, which happens in the `onPreDispatchBackend` method. 

This patch allows to use the Auth Component, at least when the referer and client checks are not executed, by retrieving the referer and user agent from the request only when the checks are enabled. 

The reason for this requirement is that we need to execute our own custom `onPreDispatchBackend` method before the default handlers `onPreDispatchBackend` method has been executed, because we adapted the backend login process to support preemptive HTTP Basic Authentication.

Maybe there's a better solution for this specific problem, but at least this pull request increases the performance a little bit by avoiding unnecessary variable instantiation.

### DIFF
--- a/engine/Shopware/Plugins/Default/Backend/Auth/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Backend/Auth/Bootstrap.php
@@ -409,8 +409,7 @@ class Shopware_Plugins_Backend_Auth_Bootstrap extends Shopware_Components_Plugin
 
         Enlight_Components_Session::start($options);
 
-        $referer = $this->request->getHeader('referer');
-        if($refererCheck && $referer !== null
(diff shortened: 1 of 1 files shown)


OliverSkroblin:
Hello lennartdiedrich,

your pull request has been merged. 

Commit: https://github.com/ShopwareAG/shopware-4/commit/1f80bc68437b11ae76bc5b01dbdd5f23257426bd

Thanks for your contribution.

best regards

Oliver Denter

PR #27, opened 2012-11-09
Update engine/core/class/sTicketSystem.php

axzx:


klarstil:
Hey @axzx,

thank you for the Pull Request. Unfortunately we couldn't merge your Pull Request in our master origin due to the fact that your Pull Requests addresses issues in a class which is marked as deprecated and will be removed in the near future. The other reason why we're denying your Pull Request is that the Ticket System is now a Shopware Premium plugin instead of a core component / module.

Thank you for your understanding.

Kind regards,
klarstil

PR #33, opened 2012-11-29
Small fix in emotion template

JostBaron:
The inclusion directive of the menu for static sites was in the wrong place.


dnoegel:
Hi,

your ticket is currently flagged for SW 4.1.3: http://jira.shopware.de/?ticket=sw-6423

PR #37, opened 2012-12-08
fixed typo in rest api

BenediktSeidl:
it was possible to import not unique email-adresses


dnoegel:
Will be released with SW 4.0.7

PR #42, opened 2013-01-11
Update engine/Shopware/Controllers/Frontend/Checkout.php

axzx:
now it works


dnoegel:
Hey @axzx,

this pull request seems to be broken: there are a few thousand changes to be merged? 

Concerning your original changeset in engine\Shopware\Controllers\Frontend\Checkout.php: Please understand, that we are not able to merge changes without a bug/feature description. We need to reproduce the behaviour, write tests etc.

Thanks in advance,

Daniel

PR #49, opened 2013-01-22
Fixed configuration form description translation issue

wdJer:
The correct translation for backend config form elements now is used instead of the original description of the form element.


dnoegel:
Hi,

I've created a ticket and internal branch for your PullRequest: http://jira.shopware.de/?ticket=sw-6422

PR #53, opened 2013-01-30
Exclude given IP's from Google Analytics

rschueler:
Added property blockIp to exclude a list of IP's from Google Analytics. For those clients the Google Analytics code will not be inclueded into the page.


dnoegel:
Hi rschueler,

thanks for your contribution. We've created a internal pull request and a ticket for this in our Tracker, you can access it here: http://jira.shopware.de/Widgets/Jira/?ticket=SW-5280.

PR #57, opened 2013-02-04
Fixed order overview page to remove the tax star (*) from tax free items

daviddavid:
Tax exempt items should not display an asterisk `*` to indicate that they include tax. This fixes that in `/templates/_default/frontend/account/order_item.tpl`


dnoegel:
Hi,

your PullRequest is now in CodeReview, the corresponding ticket can be found here: http://jira.shopware.de/?ticket=sw-6420

PR #58, opened 2013-02-08
Fixes mulitple event triggering on action forwarding.

shredding:
Fixes this issue:

https://github.com/ShopwareAG/shopware-4/issues/45
http://jira.shopware.de/Widgets/Jira/?ticket=SW-5158

The action dispatcher does not check if the `_dispatched` property of the given request has changed during the action processing.

The check is included in this pull request.

### DIFF
--- a/engine/Library/Enlight/Controller/Action.php
+++ b/engine/Library/Enlight/Controller/Action.php
@@ -136,15 +136,18 @@ abstract class Enlight_Controller_Action extends Enlight_Class implements Enligh
             }
             $this->postDispatch();
         }
+        
+        if($this->Request()->isDispatched()) {
 
-        Enlight_Application::Instance()->Events()->notify(
-            __CLASS__ . '_PostDispatch_' . $this->controller_name,
-            $args
-        );
-        Enlight_Application::Instance()->Events()->notify(
-            __CLASS__ . '_PostDispatch',
-            $args
-        );
+            Enlight_Application::Instance()->Events()->notify(
+                __CLASS__ . '_PostDispatch_' . $this->controller_name,
+                $args
+            );
+            Enlight_Application::Instance()->Events()->notify(
+                __CLASS__ . '_PostDispatch',
+                $args
+            );
+        }
     }
 
     /**
--- a/engine/Library/Enlight/Controller/Action.php
+++ b/engine/Library/Enlight/Controller/Action.php
@@ -137,8 +137,7 @@ abstract class Enlight_Controller_Action extends Enlight_Class implements Enligh
             $this->postDispatch();
         }
         
-        if($this->Request()->isDispatched()) {
-
+        if ($this->Request()->isDispatched()) {
             Enlight_Application::Instance()->Events()->notify(
                 __CLASS__ . '_PostDispatch_' . $this->controller_name,
                 $args


bcremer:
Hi shredding,

Our test suite gives green light but this change can potentially break some third party plugins, so additional reviews and tests are needed. 

I'll inform you as soon as we checked it.

engine/Library/Enlight/Controller/Action.php
```diff
@@ -136,15 +136,18 @@ public function dispatch($action)
             }
             $this->postDispatch();
         }
+        
+        if($this->Request()->isDispatched()) {
```
bcremer: Please add the missing space here, so the code is PSR-2 compliant.

bcremer:
Would't it not be easier to move the notifies into the "if"-block where the action is called:

``` php
if ($this->Request()->isDispatched() && !$this->Response()->isRedirect()) {
    $action_name = $this->Front()->Dispatcher()->getFullActionName($this->Request());

    if (!$event = Enlight_Application::Instance()->Events()->notifyUntil(
        __CLASS__ . '_' . $action_name,
        array('subject' => $this)
    )) {
        $this->$action();
    }
    $this->postDispatch();

    Enlight_Application::Instance()->Events()->notify(
        __CLASS__ . '_PostDispatch_' . $this->controller_name,
        $args
    );
    Enlight_Application::Instance()->Events()->notify(
        __CLASS__ . '_PostDispatch',
        $args
    );
}
```

engine/Library/Enlight/Controller/Action.php
shredding: Ill do that on sunday.

---

... via Phone.

shredding:
No, this wouldn't work. The bug occurs after forwarding. The forward changes the dispatch state and the action name. A forward is invoked by the action and the action is called inside this if-Block. Hence, you do have to check again after the action was executed.

Please note: the action name changes as well to the to-be-forwarded-to action. So it's not even the original action that triggers the post dispatch event (which would be wrong as well). The post dispatch event of the forward gets triggered before a single line of the action was executed.

shredding:
I have added the space :).

For those who stumble upon the error, you can workaround the bug by checking the dispatch state for yourself. Here's an example for the checkout (where I discovered it):

``` php
    /**
     * @param Enlight_Controller_EventArgs $args
     * @return void
     */
    public function onPostDispatchCheckout(Enlight_Controller_EventArgs $args)
    {
        /** @var Shopware_Controllers_Frontend_Checkout $subject */
        $subject = $args->getSubject();
        $request = $subject->Request();
        $action =  strtolower($request->getActionName());

        if ($action === 'confirm' && $request->isDispatched()) {
            // Here you go.
        }      
    }
```

bcremer:
Hi @shredding,

we've scheduled this issue for the 4.0.8 release.

We can't merge your pull request because it will break existing plugins. 

We have adopted the current behaviour from Zend Framework:
https://github.com/ShopwareAG/Zend-Library/blob/master/Controller/Action.php#L527

In 4.0.8 we will probability introduce some new "secure" postDispatch events.

Best practice for now, is to check the $request for isDispatched() as shown in your example.

A possible implementation could look like this:

``` php
// Fire "Secure"-PostDispatch-Events only if:
// - Request is Dispatched
// - Response in no Exception
// - View has template
if ($this->Request()->isDispatched()
    && !$this->Response()->isException()
    && $this->View()->hasTemplate()
) {
    Enlight_Application::Instance()->Events()->notify(
        __CLASS__ . '_PostDispatchSecure_' . $this->controller_name,
        $args
    );
    Enlight_Application::Instance()->Events()->notify(
        __CLASS__ . '_PostDispatchSecure_' . $moduleName,
        $args
    );
    Enlight_Application::Instance()->Events()->notify(
        __CLASS__ . '_PostDispatchSecure',
        $args
    );
}
```

PR #63, opened 2013-02-27
Fixed RatingStars calculation in box_article

mbarten:
In some cases the result of the used smarty statement was 9.5 or similar; resulting in showing "empty" stars on the listing page.


dnoegel:
Hi @webdiver,

we already have a ticket for that issue in our internal ticket system. It is currently planed for SW 4.0.8 => http://jira.shopware.de/Widgets/Jira/?ticket=SW-5220.

Even if the {noformat} tag does not seem to work in the public version of the bug tracker - the same solution is suggested there :)

PR #64, opened 2013-02-28
SW-5290 Update for better iFrame integration

stilhub:
Code (taken out from submitted file"gateway.tpl" from attachement to a support e-mail) submitted by technical support of Skrill (28th february 2013) in order to display the iFrame properly at the end of the checkout process without truncating it.


dnoegel:
Hi @stilhub,

thanks for your pull request. The Skrill payment plugin is delivered as a part of the default Shopware package but is developed by Skrill itself. For that reason the change will be applied with next update coming from Skrill.

Thanks for your effort.

PR #67, opened 2013-03-13
Update sBasket.php

axzx:
This patch allows adding free products.


dnoegel:
Hi @axzx,

currently we are not planing to change this behavior. We suggest using price groups with 100% discount in order to have free products.

PR #72, opened 2013-05-03
fixed host validation issue when site run under custom port

vadimeremeev:
Issue happened when I've installed site under MAMP on 8888 port. I was trying to login (front-end) but it shows me exception and stop execution. Ajax login form also not worked properly. I've checked how current validation works and seems it not checking url port so I've added fix which works for me (also works if site installed on default 80 port). 

Please review and let me know if I can explain more.


dnoegel:
Hi,

thanks for your PullRequest. Your ticket is currently flagged for SW 4.1.3: http://jira.shopware.de/?ticket=sw-6424

PR #75, opened 2013-05-15
Check blogIds to prevent sql errors and broken sitemap.xml

petigrafix:
Return from function if blog ids are not set. Otherwise an incorrect
sql statement will break the script and the closing </urlset> tag is
missing.


dnoegel:
Hi, 

I've created an internal pull request for this.

Thanks for your contribution!

PR #76, opened 2013-05-29
Fix price inserts on article import

czimnick:
the $customerPriceGroups will be overide every foreach cycle so the result in $customerPriceGroups has only the last article prices.


dnoegel:
Hi @czimnick,

thanks for your pull request. I've moved it to our internal repo and created a ticket.

PR #100, opened 2013-09-03
This fixes a namespace clash in our connector tool including Categories,...

fiesch:
... Articles and Media at once. As all other instanceof appearances in the file are with the fully qualified class name i found this to be a rather rational approach being unable to find a solution outside of the Resource code.

Basic testing yields no effect on functionality and it resolves the following namespace clash error:

Cannot use Shopware\Models\Media\Media as Media because the name is already in use in /var/www/html/shopware/engine/Shopware/Components/Api/Resource/Article.php on line 29

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Article.php
+++ b/engine/Shopware/Components/Api/Resource/Article.php
@@ -26,7 +26,6 @@ namespace Shopware\Components\Api\Resource;
 
 use Shopware\Components\Api\Exception as ApiException;
 use Shopware\Models\Article\Article as ArticleModel;
-use Shopware\Models\Media\Media;
 
 /**
  * Article API Resource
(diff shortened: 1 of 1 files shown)


dnoegel:
Hi @fiesch,

as far as I am concerned, this does not affect the shopware core itself, but your connector tool? Is there a way to reproduce this issue in a default shopware setup?

Thanks in advance,

Daniel

fiesch:
Hm i don't know for sure- it looks to me as if this should pop up whenever category, article and media models are used in the same script. I guess it could easily affect other plugins out there. 
In any case this would solve the case cleanly and bring the modified portion back in line with the rest of the file.  
I doubt that the shopware core can produce this issue without modifying code (it would have cropped up already if it was possible)

PR #104, opened 2013-09-17
unexpected value of description_long in smarty

mburamorde:


dnoegel:
Hi,

thanks for your PullRequest. We've currently flagged SW-6813 (http://jira.shopware.de/?ticket=sw-6813) for Shopware 4.1.5. Generally we are aware of this issue - but for compatibility reasons we did not resolve it, yet. If I remember right, this behaviour was implemented in SW 3.0.5.

mburamorde:
Hello,

is it possible to add 'description_long_new' in the core? Than I have no need to patch every new version.

Thanks,
Michael

dnoegel:
Hi, 

instad of patching the your, I'd recommend to have an after hook on sGetTranslations. There you'd simply check of $object == 'article' and add a new array field "my_description_long" to the return array and set it to the value of the already existing 'description_long'. 

That solution should work pretty well as sGetTranslations is called _before_ the description_long value is overwritten with the 'description' value.

dnoegel:
Hi @mmmichael,

this will be fixed with http://jira.shopware.de/?ticket=sw-5863. We are currently discussing if it will be configurable or if there will be an additional field like "description_long_full" or something like that. Feel free to subscribe to SW-5863 in order to be notified about changes.

PR #112, opened 2013-10-07
Improve laststock evaluation in article listing

stefanheppenheimer:
Currently the buy now button will not been displayed, if the laststock flag is set, even if the article is still available. This Pull-Request refines the display condition for the buy now button by putting the laststock flag in correlation to the current stock amount.


bcremer:
Thank you @stefanheppenheimer, this merge took way too long, sorry for that.

PR #114, opened 2013-10-08
Add basket events

kekster:
Adding a few more events to sBasket to modify vouchers.

### DIFF
--- a/engine/core/class/sBasket.php
+++ b/engine/core/class/sBasket.php
@@ -57,8 +57,16 @@ class sBasket
 	 * @return array
 	 */
 	public function sGetAmount(){
-		return  $this->sSYSTEM->sDB_CONNECTION->GetRow("SELECT SUM(quantity*(floor(price * 100 + .55)/100))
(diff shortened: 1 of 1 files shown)


engine/core/class/sBasket.php
```diff
@@ -682,6 +705,7 @@ public function sAddVoucher($sTicket,$BASKET=''){
         );
 		$sql = Enlight()->Events()->filter('Shopware_Modules_Basket_AddVoucher_FilterSql',$sql, array('subject'=>$this,"params"=>$params,"voucher"=>$ticketResult,"name"=>$vouchername,"shippingfree"=>$shippingfree,"tax"=>$tax));
 
+		$sql = Enlight()->Events()->filter('Shopware_Modules_Basket_AddVoucher_FilterSqlParams',$params, array('subject'=>$this,"sql"=>$sql,"voucher"=>$ticketResult,"name"=>$vouchername,"shippingfree"=>$shippingfree,"tax"=>$tax));
```
OliverSkroblin, line 708: Hi, this can't work. You pass the "params" property to the filter event and set the return value of the event to "$sql".

PR #117, opened 2013-10-09
Fix: Wrong Id/Number used for fetching premium articles

h4rvk4:
sArticles::sGetPromotionById() requires the ordernumber as thirt parameter, but in sMarketing::sGetPremiums() (line 314) the articleid is currently used.


h4rvk4:
Gut... sooo war das weniger geplant... in meiner perfekten Welt wären das 2 Pull Requests gewesen ^^;

PR #122, opened 2013-10-12
Add event to sOrder::sGetOrderNumber

kekster:
To generate order numbers for different multishops - e.g. mainshop: 20000, multishop 1: 30000, multishop 2: 40000 - we could use different ranges of numbers for each of those shops. With this new event a plugin could change the range.

### DIFF
--- a/engine/core/class/sOrder.php
+++ b/engine/core/class/sOrder.php
@@ -152,10 +152,12 @@ class sOrder
 	 */
 	public function sGetOrderNumber()
 	{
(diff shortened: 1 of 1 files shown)


dnoegel:
Hi,

i like the idea, but I am not happy with the implementation. Can you live with an replaceHook for the time being and create a ticket in jira for proper multi-shop support for the order number / invoice numbers.

=> http://jira.shopware.de

Thanks and best regards,

Daniel

PR #123, opened 2013-10-14
Better error messages for permission problems on plugin download

samuelvogel:
Currently when the permissions of the plugin folders are wrong and you try to download something from the community store you only get a message that you should change `engine/Shopware/Plugins/(Community/Default/Core)` recursively. There are two problems with that:
- Actually `engine/Shopware/Plugins` needs to be writeable as well, as `..` is also walked by the recursive iterator
- The message is really unspecific and on some webhosters it's really clumsy to change everything recursively

So this commit adds the actual path, which is not writeable, to the error message to make fixing problems like this easier.

### DIFF
--- a/engine/Shopware/Plugins/Default/Core/PluginManager/Components/Store.php
+++ b/engine/Shopware/Plugins/Default/Core/PluginManager/Components/Store.php
@@ -306,8 +306,9 @@ class CommunityStore
     {
         $target = Shopware()->AppPath('Plugins_' . $source);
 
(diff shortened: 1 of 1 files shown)


samuelvogel:
`.` and `..` are not checkend anymore since ccf9ac120aabfa01598c2b0ada1b7ce2a6c45d21, this makes the behavior a little clearer already!

bcremer:
Sorry for the late answer. It's now merged in master.

PR #128, opened 2013-10-22
Fix handling of arrays in the REST API filter

svenmuennich:
The REST API accepts filter parameters in the URL, which can contain multiple values for one property. The internal handling in the used query builder checks for arrays of filter values, to create `IN` expressions in the resulting SQL query. However, if an array is recognized, the term `IN` will be selected as the operator, but the value array will not be handled correctly. This leads to a syntax error like the following:

```
Errormesage: [Syntax Error] line 0, col 78: Error: Expected Doctrine\ORM\Query\Lexer::T_OPEN_PARENTHESIS, got ':cleared'
```

Adding parentheses, as well as one parameter key for each value in the array and replacing all these keys with their values later fixes this error.

### DIFF
--- a/engine/Shopware/Components/Model/QueryBuilder.php
+++ b/engine/Shopware/Components/Model/QueryBuilder.php
@@ -122,7 +122,19 @@ class QueryBuilder extends BaseQueryBuilder
                 }
             }
 
-            $expression = new Expr\Comparison($exprKey, $expression, $where !== null ? (':' . $parameterKey) : null);
+            $assocWhere = array();
+            if (is_array($where)) {
+                // Create one key for every value in the array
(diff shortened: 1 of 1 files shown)


bcremer:
Hey @svenmuennich, thank's for your pull request.

Would you please mind to rebase your fix to the next-branch and add a testcase to https://github.com/ShopwareAG/shopware-4/blob/next/tests/Shopware/Tests/Components/Model/QueryBuilderTest.php.

samuelvogel:
Sorry for taking that long, but we've finally created a new PR with the requested changes: #164

dnoegel:
Hey,

we've imported the linked PR #164

PR #130, opened 2013-10-24
Use absolute paths for files and directories in .gitignore

lennartdiedrich:
Since plugins could be using filenames mentioned within the gitignore file (e.g., config.php) it is advisable to use absolute paths.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,29 +16,29 @@ Thumbs.db
 
 # Internal Repositories
 internal/
-engine/Shopware/Plugins/Commercial/
+/engine/Shopware/Plugins/Commercial/
 
 # User installed plugins
-engine/Shopware/Plugins/Community/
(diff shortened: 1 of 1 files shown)


samuelvogel:
See the PAYONE plugin as an example, this file would be ignored:
https://github.com/PAYONE/shopware_4.x/blob/master/Frontend/MoptPaymentPayone/Components/Payone/Config.php

bcremer:
Hey @lennartdiedrich, thanks for your contribution.

I actually did the same fix a few days before on our next branch ;) 
See: https://github.com/ShopwareAG/shopware-4/commit/05ee36505a9ab9e2fb527b8fe0855a309897a07c

But as requested I just backported this into our master branch.

Closed with https://github.com/ShopwareAG/shopware-4/commit/5b826726966680d3727c15826c80138a7543babd

PR #135, opened 2013-11-27
Added search for supplier number in article list

jetwes:
http://jira.shopware.de/?ticket=SW-6311
added search for supplier number in backend article list


jetwes:
is still open in jira (release phase3) since october...

PR #136, opened 2013-12-11
Fix bug in Order\Shipping entity mapping leading to weird queries.

beberlei:
Fixes the bug where in some cirtumstances `Customer\Customer#shipping` contains a `Order\Shipping` instead of an `Customer\Shipping`.


dnoegel:
Hi,

thank your for your PullRequest, it was already merged and is out in the wild (4.2.0) :)

b3d7a4694bdb939bda03af14467e0c059f3db70e    13.12.13 07:44  Benjamin Eberlei    Fix bug in Order\Shipping entity mapping leading to weird queries.

PR #139, opened 2014-02-03
Add a document attributes entry to the database when creating a new document

svenmuennich:
When a new document is created using the `Document` component, a new entry in `s_order_documents` is created, but no associated entry in `s_order_documents_attributes`. Besides that the `setAttribute` method of the `Shopware\Models\Order\Document\Document` entity is broken. This pull request not only fixes this broken method, but also adds a new attributes entry when creating a document. Furthermore it is possible to add an `attributes` array to the config, which is passed to the document creation. The contained key/value pairs will be inserted or updated in the attributes entity. This fix is backwards compatible, meaning that on an update no exception will be thrown, if no attributes exist yet, but actually a new attributes entity is created and saved in the database.

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -518,6 +518,35 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
                     $this->_order->id
                 ));
 
+            if (!empty($this->_config["attributes"])) {
+                // Get the updated document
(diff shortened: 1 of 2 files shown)


svenmuennich:
I've created a new PR #167, which contains only the the insert/update of attributes using the `attributes` array found in the document config, because the fix for the broken setter method has already been merged.

PR #140, opened 2014-02-10
saveArticle returns plain message which causes php error

nlubisch:
Changed return to throw new \Exception because if this message returns as result of this method "$result->getId()" fails in line 2383 and the import hung up.


dnoegel:
Hi,

this Exception has a downside: It will break the whole import instead of e.g. skipping just the one line. Instead of throwing the exception, you might also move this check before the $result->getId() call:

```
           if (!$result instanceof \Shopware\Models\Article\Article) {
                $errors[] = $result;
                continue;
            }
```

PR #141, opened 2014-02-15
SW-7341 - Multi-image export via product feeds

hype09:
This feature enables the user to export multiple images using the product feeds.
The new Smarty syntax in the Product Feeds is:  {$sArticle.image|image:4:0}

The image urls are concatenated when fetched from the database and split/exploded in the sGetImageLink method.
So the $hash parameter that is passed to the method has the following form:
image_url01:image_position01:image_main01;image_url02:image_position02:image_main02;...

This method now accepts a parameter (in the example 0) controlling which image is returned.
If no parameter is given, the main image is returned.


hype09:
I found a bug and will fix it soon. I'll close the pull-request until then.

PR #143, opened 2014-02-26
Added cache clear commands

benboit:
Added cache-clear-commands which use the shopware cache-manager service to clear caches. Useful for deployments.

### DIFF
--- a/engine/Shopware/Commands/CacheConfigClearCommand.php
+++ b/engine/Shopware/Commands/CacheConfigClearCommand.php
@@ -1,0 +1,66 @@
+<?php
+
+/**
+ * Shopware 4
+ * Copyright © shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Commands;
+
+use Shopware\Components\CacheManager;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @category  Shopware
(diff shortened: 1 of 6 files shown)


benboit:
Not useful? The cache/clear_cache.sh ist not always the best solution.

FiveDigital:
I like the new commands :thumbsup:

jhofm:
i've been meaning to write just that very same command for the last few months, but never got around to actually do it since our shell scripts currently do the same job so there wasn't really enough pressure to do so. but still, that command means less dependencies, which i'm all about. ;) so kudos for writing this. i'll try testing this in our deployment pipeline asap.

dnoegel:
I think its quite useful, too - only question is, if this should be multiple commands or one command with multiple options, so you can clear e.g. 3 caches at once by adding --template --proxy --http or something like that. What do you think @bcremer?

bcremer:
I would prefer one command, containing options for each cache, like @dnoegel mentioned.

But I see a few problems with this command:
1. Sometimes we are not able to clear a cache from the CLI-Context. For Example APCu. The APCu cache is stored in a thread/process the CLI is not able to reach. so a `apc_clear_cache()` will have no effect. 
2. By now the `clear_cache.sh` script is some kind of undocumented feature for people that hopefully now what they are doing. When we are promoting the clear cache command as a first class citizen into the system we have to make sure that problems like (1) are fixed.
3. The command and more important the `shopware.cache_manager` need a functional shopware system to operate. Often the `clear_cache.sh` is used to recover a broken system, b/c it removes broken cache files for example

But I am not fully against including this. Maybe it's just a communication issue. So the clear cache command is used for regular use like clearing the HTTPCache on a regular basis, and the `clear_cache.sh` is some kind of recovery script.

bcremer:
To solve the APC issue we should have a look into https://github.com/Smart-Core/AcceleratorCacheBundle 

Using a temporary file and a http request the CLI context is able to communicate with the Web context to clear caches. This comes with additional complexity of course.

benboit:
I would love to see something like these commands being part of the shopware-core features. I already use these commands in a few projects in production using a separate plugin.

The naming issue should be easy to fix. Probably like

bin/console cache:clear --template --http --seo --config --search --all

Your apc-solution seems quite complex, i dont know if i understand it. They are creating a random file and check if the content of it requested using http has changed? Is that the way they try to get to know if the apc was really cleared?

jhofm:
I'm mostly interested in the deploying angle and would like to point out that in many cases it would be preferable to build the cache via cli and switch to that instantly. (via symlink for filesystems and versioning for configuration/cachey stuff in other storages). the problem with clearing is that you leave it to webserver workers to rebuild the updated caches. uncached requests take a while to handle, so you quickly get a concurrency problem with multiple processes trying to write the same data to the same place. For example I've had a memorable deployment that shipped while we were hammered by the bing bot, which caused APC to allocate over 8GBs of RAM before it stopped trying. Fortunately most commands needed to prebuild the important caches already have commands since sw4.2. but it would be nice to have a bit more grease provided by the core. ;)

dnoegel:
Hi,

@mschmaeing did a cache warmer in a hackathon once and it is intended to be shipped with shopware one day, so we are aware of that :)

PR #144, opened 2014-02-27
Fix set-cookie sent multiple times when regenerating session

matthistuff:
bildschirmfoto 2014-02-27 um 13 17 22
When logging in (account/login), Shopware somtimes will return a Set-Cokie header for the same cookie name three times (old sessionID, new sessionID, new sessionID). This might lead to unexpected behavior, as it is not clear what happens in the browser when setting the same cookie multiple times.

This is not the fix for the root problem of, but mentioned in a sidenote here: http://jira.shopware.de/?ticket=SW-7684

### DIFF
--- a/engine/core/class/sAdmin.php
+++ b/engine/core/class/sAdmin.php
@@ -1296,6 +1296,8 @@ class sAdmin
         session_regenerate_id(true);
         $newSessionId = session_id();
 
+        // clear any set-cokie header before opening a new session
+        header_remove('Set-Cookie');
         // close and restart session to make sure the db-session handler writes updates.
         session_write_close();
         session_start();


engine/core/class/sAdmin.php
```diff
@@ -1296,6 +1296,8 @@ public function regenerateSessionId()
         session_regenerate_id(true);
         $newSessionId = session_id();
 
+        // clear any set-cokie header before opening a new session
+        header_remove('Set-Cookie');
```
bcremer, line 1300: But this will also remove any other cookie that may be not session related?

matthistuff, line 1300: That is indeed true. Is there a better way for just unsetting the session-x cookies?

bcremer, line 1300: We have a ticket open for the 4.2.2 release and will further investigate into this issue.

dnoegel:
Hi,

we've a ticket for that here: jira.shopware.de/?ticket=SW-8413

We will further investigate on this and close here.

Thanks for your contribution.

PR #145, opened 2014-03-03
[imported] Set articleId when manually adding an order position

samuelvogel:
When manually adding a new order position via the backend, the `articleId` column is not set in the database. This is inconsistent, especially if one uses the autocomplete tooltip to select the article. Also the open article button is not displayed because of this for example.


dnoegel:
Thanks for your PullRequest. This feature is currently waiting for QA, we will close this PR once we merged the commit.

bcremer:
Merged as https://github.com/ShopwareAG/shopware-4/commit/c701bce91baff0540655e92dd46c488fabcc51e1

samuelvogel:
Great, thank you! :-)

PR #146, opened 2014-03-05
Allow sub shop only countries

matthistuff:
Currently you can't have a country active only in one or more sub shops without globally activating it in the main shop. One solution is to ignore the s_core_countries.active field and only use it when there is no translation for the country in the current shop or when the current shop is the main shop.


tiagojsag:
Hi,

Thank you for your PR. We have flagged this issue in the following ticket: http://jira.shopware.de/?ticket=SW-8417
Your commit will be merged and released as part of one of the following releases of Shopware.

Greetings

PR #149, opened 2014-03-17
- new and not existing options were added with a wrong name instead of t...

NiklasTeich:
While creating and updating variation options, new configurator options got wrong names (`Object`) instead of the given option names

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Variant.php
+++ b/engine/Shopware/Components/Api/Resource/Variant.php
@@ -688,9 +688,14 @@ class Variant extends Resource implements BatchInterface
             ));
 
             if (!$option) {
+
+                if (!$optionData['option']) {
+                    throw new ApiException\CustomValidationException('A new configurator option requires a name');
+                }
+
                 $option = new Option();
                 $option->setPosition(0);
-                $option->setName($option);
+                $option->setName($optionData['option']);
                 $option->setGroup($availableGroup);
                 $this->getManager()->persist($option);
             }


NiklasTeich:
This bug still exists in SW 4.2.3 ... ! Is there a chance to get a fix in 4.3? ...

bcremer:
Hey @NiklasTeich, 

we will have a look into this issue soon.

Can you provide us with a testcase in `engine/Shopware/Components/Api/Resource/Variant.php` to reproduce the failing behaviour and to test your expected behaviour?

Greeting,
Benjamin

NiklasTeich:
Hi @bcremer !

Thanks for the feedback, but is a UnitTest really necessary? ... The `setName` requires a string ... but the option `Object` is given ... 

So, if one option is new and saved, the name is "Object". If a second new option follows (for example in an update) the API crashes ...

Greetings,
Niklas

OliverSkroblin:
Hey,

thanks for your pull request. Could you write a small unit test in the Shopware_Tests_Components_Api_VariantTest class?

The class implements already different helper function to create products and variants. So it is really easy to create an unit test for this case.

Greetings,

Oliver Denter

NiklasTeich:
I'm currently unable to set up phpunit on my machine, i'm getting fatal errors `The mysql driver is not currently installed` when i try to execute the phpunit (local under xampp, windows/osx testet separately) but my test shopware environment runs ...

Obviously i'm not in the position to give you a test in the next days, but the SW 4.3 roll out is close :( I don't know what to do, i'm unable to give a test, but i need the fix in your system.

Daniel Nögel recieved 2 example CSVs from the user exx in the programming board (14-04) who also got these errors (they imported first the father article, and after that, the child articles. thats also my workflow)

Can't you help me with this? :(
Greetings ...

Niklas

OliverSkroblin:
Hi,
i added a test for you in the tests/Shopware/Tests/Components/Api/Resource/VariantTest.php.
The unit test creates an article and five variants. Each variant has a new option which will be created over the rest api resource.
The new option name is now used from the passed data array element `option`.

I close the pull request, the changes are available with the new bugfix release.

NiklasTeich:
Thank you VERY much @OliverDenter ! I REALLY appreciate your help and solution. Thank you very very much, you've made my day!

PR #150, opened 2014-03-25
update to documentation

BinaryGeometry:
The ant build requires six arguments to be entered in the command line. While the first four are self explanetory, the last two relating to the app.path and host.path could use some clarification.


dnoegel:
Hey thanks, we will improve that in the next release.

PR #155, opened 2014-04-23
[response needed] Add a method to set the variable encoderName

resslinger:
To create or set a customer via api, the variable encoderName could not be set. This variable is shown in the actual api doku (http://wiki.shopware.de/_detail_1687.html) but without function.


dnoegel:
Hi,

we think the usecase is to e.g. import customers with a given rawPassword and a specific encoder?

Can provide a unit test making use of this change, please (https://github.com/ShopwareAG/shopware-4/blob/master/tests/Shopware/Tests/Components/Api/CustomerTest.php)? Then we'd be happy to merge this.

Best regards,

Daniel

PR #156, opened 2014-04-29
CRITICAL: set doctrine-common to a more specific version to prevent lazy load bug on update

benboit:
We are using composer in production and used this composer.json als base. After we updated using "composer update" some days ago some things broke in shopware.

Creating / removing  customer-entities and even importing articles was not possible anymore.

Last known valid commit hash was: 9a7e20e779360f3b8a02c27a89d47d5a6fdce8d1
Broken for us at: b1a31ae3df2c18c24bec42e2cc32e3e4bedc07c7

Please be carefull at next build ;)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "monolog/monolog": "1.6.*",
         "ircmaxell/password-compat": "1.0.3",
         "rhumsaa/array_column": "~1.1",
-        "doctrine/common": "2.4.*",
+        "doctrine/common": "2.4.1",
         "doctrine/dbal": "2.4.1",
         "doctrine/orm": "2.4.1"
     },


bcremer:
Hey @benboit,
thanks for your contribution.

Please note that you never have to use `composer update` to install shopware.
We have `composer.lock` in our repository that contains the tested dependencies in a precise version.

To install shopware from git it's recommended to use `composer install --optimize-autoloader` to get a tested dependeny tree.

We'll have a closer look in the commit you mentioned: https://github.com/doctrine/common/commit/b1a31ae

Can you further elaborate the issues you encounter with that doctrine common version?

benboit:
I know how to use the update / install feature, but we are including other 3rd party dependencies and our own shopware-plugins using composer which force us to use "update" sometimes.

Here is the error-message/s we got:
- on creating / deleting a customer, importing articles
  "Während des Speicherns ist ein Fehler aufgetreten: Missing value for primary key id on Shopware\Models\Customer\Group"
- in the checkout
  Fatal error: Uncaught exception 'Doctrine\Common\Proxy\Exception\OutOfBoundsException' with message 'Missing value for primary key id on Shopware\Models\Partner\Partner' in /home/vagrant/src/parador/vendor/doctrine/common/lib/Doctrine/Common/Proxy/Exception/OutOfBoundsException.php on line 41 Doctrine\Common\Proxy\Exception\OutOfBoundsException: Missing value for primary key id on Shopware\Models\Partner\Partner in

bcremer:
Ok I understand your Problem.
We'll have a look into the doctrine issue and switch to more precise versions in the `composer.json` in the near future.

bcremer:
We switched to exact version numbers in the following commit: https://github.com/ShopwareAG/shopware-4/commit/43412e91127427c5c35168189e1bd3ced71e0c35

Currently this commit is available on the next branch and will be released as SW 4.3 in the near future.

PR #164, opened 2014-06-18
[imported] Fix handling of arrays in addFilter method of QueryBuilder component

samuelvogel:
The REST API accepts filter parameters in the URL, which can contain multiple values for one property. The internal handling in the used query builder checks for arrays of filter values, to create `IN` expressions in the resulting SQL query. However, if an array is recognized, the term `IN` will be selected as the operator, but the value array will not be handled correctly. This leads to a syntax error like the following:

```
Errormesage: [Syntax Error] line 0, col 78: Error: Expected Doctrine\ORM\Query\Lexer::T_OPEN_PARENTHESIS, got ':cleared'
```

Simply adding parentheses in case an array of values is supplied fixes this error.

**Updated version of PR #128:** As requested by @bcremer we rebased it on the current master (as it is even newer than next currently) and added test cases.

### DIFF
--- a/engine/Shopware/Components/Model/QueryBuilder.php
+++ b/engine/Shopware/Components/Model/QueryBuilder.php
@@ -190,7 +190,9 @@ class QueryBuilder extends BaseQueryBuilder
                 }
             }
 
-            $expression = new Expr\Comparison($exprKey, $expression, $where !== null ? (':' . $parameterKey) : null);
+            $exprParameterKey = ':' . $parameterKey;
+            if (is_array($where)) $exprParameterKey = '('.$exprParameterKey.')';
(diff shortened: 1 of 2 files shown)


bcremer:
Awesome, I'll have a look into this after the 4.3 is released.

dnoegel:
Thanks for your PullRequest. This feature is currently waiting for QA, we will close this PR once we merged the commit.

bcremer:
Merged as https://github.com/ShopwareAG/shopware-4/commit/a05b1769814e2814e86803635d566bb48d2e9a36

samuelvogel:
You should mention this in our `UPGRADE.md` for `4.3.1`.

PR #165, opened 2014-06-27
Fix handling of HTTP 'Location' headers and status codes in REST API responses to POST and PUT requests

svenmuennich:
When using `POST` and `PUT` requests with the REST API, it always explicitly sets a HTTP `Location` header, when the requests succeed. This was done using the `setHeader()` method. The problem with this approach is twofold: Firstly the location header should only be set on successful `POST` requests, because only in these cases a new resource with a new location has been created. `PUT` requests on the other hand don't create new resources and hence don't result in new locations, the requesting party could be interested in (it is still contained in the response data, though). Secondly, using the `setHeader()` method only adds a new entry to the list of headers to be sent. These headers are later sent with PHP's `header()` method, which has, according to its documentation (http://www.php.net/manual/en/function.header.php), two special-case header calls. Important for this fix is the second one:

> The second special case is the "Location:" header. Not only does it send this header back to the browser, but it also returns a REDIRECT (302) status code to the browser unless the 201 or a 3xx status code has already been set.

This behavior is responsible for the problem, that Shopware's REST API always returns a `302` in case of a successful `PUT` or `POST` request. To fix this, I replaced each `setHeader()` call for `POST` requests with the more meaningful `setRedirect()` call, which accepts a status code as an optional parameter. When passing a `201` for the status code, PHP's `header()` method will, as described in the docs, not replace the already set `201` with a `302` by setting the `Location` header. Furthermore I removed all calls to `setHeader()` for `PUT` requests completely. Now a successful `POST` request will result in a `HTTP/1.1 201 Created` with a `Location` header, while a successful `PUT` request will just result in a `HTTP/1.1 200 OK`.

### DIFF
--- a/engine/Shopware/Controllers/Api/Articles.php
+++ b/engine/Shopware/Controllers/Api/Articles.php
@@ -96,7 +96,7 @@ class Shopware_Controllers_Api_Articles extends Shopware_Controllers_Api_Rest
         );
 
         $this->View()->assign(array('success' => true, 'data' => $data));
-        $this->Response()->setHeader('Location', $location);
+        $this->Response()->setRedirect($location, 201);
     }
 
     /**
@@ -123,7 +123,6 @@ class Shopware_Controllers_Api_Articles extends Shopware_Controllers_Api_Rest
         );
 
         $this->View()->assign(array('success' => true, 'data' => $data));
-        $this->Response()->setHeader('Location', $location);
     }
 
     /**
--- a/engine/Shopware/Controllers/Api/Categories.php
+++ b/engine/Shopware/Controllers/Api/Categories.php
@@ -83,7 +83,7 @@ class Shopware_Controllers_Api_Categories extends Shopware_Controllers_Api_Rest
         );
 
         $this->View()->assign(array('success' => true, 'data' => $data));
-        $this->Response()->setHeader('Location', $location);
+        $this->Response()->setRedirect($location, 201);
     }
 
     /**
@@ -105,7 +105,6 @@ class Shopware_Controllers_Api_Categories extends Shopware_Controllers_Api_Rest
         );
 
         $this->View()->assign(array('success' => true, 'data' => $data));
-        $this->Response()->setHeader('Location', $location);
     }
 
     /**
--- a/engine/Shopware/Controllers/Api/CustomerGroups.php
+++ b/engine/Shopware/Controllers/Api/CustomerGroups.php
@@ -83,7 +83,7 @@ class Shopware_Controllers_Api_CustomerGroups extends Shopware_Controllers_Api_R
         );
 
         $this->View()->assign(array('success' => true, 'data' => $data));
-        $this->Response()->setHeader('Location', $location);
+        $this->Response()->setRedirect($location, 201);
     }
 
     /**
@@ -105,7 +105,6 @@ class Shopware_Controllers_Api_CustomerGroups extends Shopware_Controllers_Api_R
         );
 
         $this->View()->assign(array('success' => true, 'data' => $data));
-        $this->Response()->setHeader('Location', $location);
     }
 
     /**
--- a/engine/Shopware/Controllers/Api/Customers.php
+++ b/engine/Shopware/Controllers/Api/Customers.php
@@ -88,7 +88,7 @@ class Shopware_Controllers_Api_Customers extends Shopware_Controllers_Api_Rest
         );
 
         $this->View()->assign(array('success' => true, 'data' => $data));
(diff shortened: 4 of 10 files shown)


bcremer:
Hi @svenmuennich, thank you very much for you input.
We were finally able to reproduce this issue. 

The main problem is, that we are trying to set a location header with a http status code 200.
This is not strict http compatible and PHP-SAPIs and Server combinations will behave differently.
- Mod PHP leaves the `200` unaltered.
- PHP-FPM overwrites the `200` with a `302` header.
- PHP-FPM used together with Nginx preserves the payload.
- PHP-FPM used together with Nginx changes the payload with a default redirect page.

For `POST` request we could not identify any problems. The combination of location header and `201` status code is valid.
So the switch from `setHeader()` to `setRedirect()` does not seem necessary.

## Fix

To fix this issue we will remove all the location headers from the `PUT` routes.

## Example code to Reproduce

Create a file named `header.php`

```
<?php
echo "some content";
header('Location: http://www.example.com/', true, 200);
```

Apache with mod php:
`$ curl -i http://localhost/header.php`

Payload: OK
StatusCode: OK

```
HTTP/1.1 200 OK
Date: Mon, 30 Jun 2014 08:42:44 GMT
Server: Apache/2.4.9 (Unix) PHP/5.5.14
X-Powered-By: PHP/5.5.14
Location: http://www.example.com/
Content-Length: 12
Content-Type: text/html

some content
```

Apache with php-fpm:
`$ curl -i http://localhost/header.php`

Payload: WRONG
StatusCode: WRONG

```
HTTP/1.1 302 Found
Date: Mon, 30 Jun 2014 08:44:51 GMT
Server: Apache/2.2.22 (Debian)
X-Powered-By: PHP/5.4.4-14+deb7u8
Location: http://www.example.com/
Vary: Accept-Encoding
Content-Length: 292
Content-Type: text/html; charset=iso-8859-1

<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>302 Found</title>
</head><body>
<h1>Found</h1>
<p>The document has moved <a href="http://www.example.com/">here</a>.</p>
<hr>
<address>Apache/2.2.22 (Debian) Server at t1.demo.shopware.in Port 80</address>
</body></html>
```

nginx with php-fpm:
`$ curl -i http://localhost/header.php`

Payload: OK
StatusCode: WRONG

```
HTTP/1.1 302 Found
Date: Mon, 30 Jun 2014 08:42:44 GMT
Server: Apache/2.4.9 (Unix) PHP/5.5.14
X-Powered-By: PHP/5.5.14
Location: http://www.example.com/
Content-Length: 12
Content-Type: text/html

some content
```

bcremer:
Fix in commit https://github.com/ShopwareAG/shopware-4/commit/70ca0ab2069b6cc5b1204e269a37e1cb36783a5d.

svenmuennich:
Thank you very much for working on this so quickly. I think there is still a problem with `POST` requests, though.  If I see it correctly, Shopware internally sets a `201` for all successful `POST` requests. The problem still remains, that when setting a `Location` header using `setHeader()`, the `201` is replaced by a `302`. This can be fixed, using `setRedirect()` and explicitly passing a `201` to it. Of course you could also call `setHttpResponseCode()` after calling `setHeader()`, because that's exactly what `setRedirect()` does.

bcremer:
@svenmuennich, I am not able to reproduce this problem on our systems.
Can you reproduce the incorrect beaviour using our tests on your system?

You can use the [Customer Test](https://github.com/ShopwareAG/shopware-4/blob/master/tests/Shopware/Api/CustomerTest.php) for that: 

`./vendor/bin/phpunit -c tests/Shopware tests/Shopware/Api/CustomerTest.php`

svenmuennich:
When running the tests from the current `HEAD`, they all pass. So I had a closer look on the differences between all my Shopware installations (`4.1.3`, `4.2.1` and the `4.3.0-RC`/`HEAD`). In fact, the wrong behavior of automatically setting a `302` together with the `Location` header is gone since `4.2.0`, because it was a bug in Symony. They fixed it with their `v2.4.0` release (symfony/symfony@c140d4feae32d8301a8ffa0dfbbbbcd089eefd41), which was included in the update to Shopware `4.2.0`. Interestingly, on another installation I have access to, we still can reproduce the behavior with wrong `302`s, although it is on `4.2.3`. Nevertheless, I think we can close this pull request, because the behavior for `PUT` and `POST` requests seems to be correct now.

bcremer:
Awesome, thanks for the additional symfony research.

PR #166, opened 2014-07-08
[imported] Fix wrong reference in setAttribute method of order document model

svenmuennich:
This fix changes the reference, which is used to set the one-to-one relationship between the order document and its attributes. In the attribute field's [mappedBy annotation](https://github.com/VIISON/shopware-4/blob/fix-model-order-document-document-setattribute/engine/Shopware/Models/Order/Document/Document.php#L120) the name `document` is used, while in the `setAttribute()` method `orderDocument` was passed when setting the one-to-one relationship. That made it impossible to add attributes to an order document.

This pull request changes the used reference name, which is passed to `setOneToOne()`, to `document`  making it conform with the value used in the `mappedBy` annotation.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,9 @@
 # Shopware Upgrade Information
 In this document you will find a changelog of the important changes related to the code base of Shopware.
 
+## Next
+* Fixed name used as reference when setting attributes of an order document.
+
 ## 4.3.0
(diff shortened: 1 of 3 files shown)


bcremer:
Thanks for your PR. This feature is currently waiting for QA, we will close this PR once it has merged into master.

bcremer:
Merged as https://github.com/ShopwareAG/shopware-4/commit/92d52ad604258e341d20357e7f9462b7f1aca472

PR #167, opened 2014-07-15
Add attributes to a document when creating or updating it

svenmuennich:
When a new document is created using the `Document` component, a new entry in `s_order_documents` is created, but no associated entry in `s_order_documents_attributes`. This pull request adds a new attributes entry when creating or updating a document, which doesn't have attributes yet. Furthermore it is now possible to add an `attributes` array to the config, which is passed to the document creation method `initDocument()`. The contained key/value pairs will then be inserted or updated in the attributes entity. The code is backwards compatible, meaning that on an update no exception will be thrown, if no attributes exist yet, but actually a new attributes entity is created and saved in the database.

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -518,6 +518,27 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
                     $this->_order->id
                 ));
 
+            if (!empty($this->_config["attributes"])) {
+                // Get the updated document
+                $updatedDocument = Shopware()->Models()->getRepository("\Shopware\Models\Order\Document\Document")->findOneBy(array(
+                    "type" => $typID,
+                    "customerId" => $this->_order->userID,
+                    "orderId" => $this->_order->id
+                ));
+                // Check its attributes
+                if ($updatedDocument->getAttribute() === null) {
+                    // Create a new attributes entity for the document
+                    $documentAttributes = new \Shopware\Models\Attribute\Document();
+                    $updatedDocument->setAttribute($documentAttributes);
+                    // Persist the document
+                    Shopware()->Models()->flush($updatedDocument);
+                }
+                // Save all given attributes
+                $updatedDocument->getAttribute()->fromArray($this->_config["attributes"]);
+                // Persist the attributes
+                Shopware()->Models()->flush($updatedDocument->getAttribute());
+            }
+
             $rowID = $checkForExistingDocument["ID"];
             $bid = $checkForExistingDocument["docID"];
             $hash = $checkForExistingDocument["hash"];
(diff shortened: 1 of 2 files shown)


engine/Shopware/Components/Document.php
```diff
@@ -518,6 +518,32 @@ protected function saveDocument()
...
+                if ($updatedDocument->getAttribute() === null) {
+                    // Create a new attributes entity for the document
+                    $documentAttributes = new \Shopware\Models\Attribute\Document();
+                    $updatedDocument->setAttribute($documentAttributes);
+                    // Persist the document
+                    Shopware()->Models()->flush($updatedDocument);
+                }
+                // Save all given attributes
+                foreach ($this->_config["attributes"] as $key => $value) {
+                    $setter = "set" . ucfirst($key);
```
bcremer: Can't this be replaced by `$updatedDocument->getAttribute()->fromArray()`?

svenmuennich:
Before I write test cases for the provided code, I wanted to have some feedback about the basic idea of this PR. If you like it and think it should be merged, I will be more than happy to add the test cases :)

engine/Shopware/Components/Document.php
svenmuennich: @bcremer You are absolutely right! I will push an update in a minute.

bcremer:
Hi @svenmuennich, thank's for this PR.

Can you provide a simple use case in form of event callback using the new behaviour?

svenmuennich:
The simplest use case is, when creating e.g. a delivery note manually in a plugin with the following code:

``` php
$document = Shopware_Components_Document::initDocument(
    $orderId,
    $documentType,
    array(
        'netto'                   => ...,
        'bid'                     => ...,
        'voucher'                 => ...,
        'date'                    => ...,
        'delivery_date'           => ...,
        'shippingCostsAsPosition' => ...,
        '_renderer'               => ...,
        '_preview'                => ...,
        '_previewForcePagebreak'  => ...,
        '_previewSample'          => ...,
        '_compatibilityMode'      => ...,
        'docComment'              => ...,
        'forceTaxCheck'           => ....,
        // Here you can pass an array with all or some of the attribute fields
        'attributes' => array(
            'firstAttribute'          => 'some value',
            'secondAttribute'         => 'some value',
            'thirdAttribute'          => ...
        )
    )
);
$document->render();
```

Of course you can also hook e.g. the `initDocument()` method of the `Document` component to add some attributes with something like this:

``` php
public function onAfterDocumentComponentInitDocument(Enlight_Hook_HookArgs $args) {
    $document = $args->getSubject();
    $document->_config['attributes'] = array(
        'firstAttribute' => ...
    );
}
```

bcremer:
Excellent :+1: 
We may manage to get this into sw 4.3.1.
May you please add a Testcase, add a line in the UPGRADE.md and then squash your commits together into one.

svenmuennich:
While writing the test case I encountered the problem, that you somehow can't generate the attribute models using `Shopware()->Models()->generateAttributeModels(...)` more than once. This method is first called in the `TestHelper` class [when starting the kernel](https://github.com/VIISON/shopware-4/blob/master/tests/Shopware/TestHelper.php#L98). When I then add new fields to the `Document\Attributes` in my test case, the respective Doctrine files are updated in the cache but the previously generated model is already loaded.

Do you know how to overcome the loading of these generated models? Otherwise I don't see a chance to write proper tests for the Document attributes. Currently I run the tests without calling `generateAttributeModels()` in the `TestHelper`.

bcremer:
Maybe one can clear the metadatacache during runtime... but that would introduce more side effects.

I think we can leave the automated test out here and do some manual testing.

You could safe us some time if you could provide us a plugin Bootstrap.php that creates some document attributes, installes the hook and uses the new behaviour so I can handle that out to our QA ;)

Else then that put the line into UPDGRADE.md, squash and we're good to go :+1:

svenmuennich:
You can find an example plugin in [my Dropbox](https://dl.dropboxusercontent.com/u/26761731/TestOrderDocumentAttributes.zip).

bcremer:
This PR has been included in master with commit https://github.com/ShopwareAG/shopware-4/commit/06943b3e1cc8c590405a1f560b698e54e15eb8af

Thank`s for your contribution.

PR #174, opened 2014-09-03
Fix the 'removeListener' method in 'Enlight_Event_Subscriber_Config'

svenmuennich:
Currently the `removeListener` method in `Enlight_Event_Subscriber_Config` uses `array_diff` to remove a given `Enlight_Event_Handler` object from the `listeners` array. The problem is, that `array_diff` doesn't work with object arrays. This fix uses `array_search` and `array_splice` instead, to first find and then remove a given handler from the listeners.


andrewisplinghoff:
New PR: https://github.com/shopware/shopware/pull/441

PR #175, opened 2014-09-03
Add 'unsubscribeEvent' method to 'Enlight_Plugin_Bootstrap_Config'

svenmuennich:
In some plugin updates it is desirable to remove old event subscriptions, since they are no longer used. This commit adds a new method to `Enlight_Plugin_Bootstrap_Config`, from which all plugin bootstrap implementations inherit.

The `unsubscribeEvent` method requires only the event name and listener name to remove an existing subscription, just like `subscribeEvent` requires those two parameters to add a subscription in the first place.

In order to make this work, it is also necessary to adapt the `write` method of `Shopware_Components_Plugin_Namespace`, so that the removed subscribes are also removed from the database. Therefore it has now an optional parameter of type `Shopware_Components_Plugin_Bootstrap`. If it is present, that plugin's subscribes will be fully synced, meaning that its removed subscribes will be deleted from the database.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -22,6 +22,7 @@ In this document you will find a changelog of the important changes related to t
     * Exceptions are logged in a logfile since 4.2.0 (/logs)
     * The old behaviour can be restored by setting `'front' => array('showException' => true)` in the projects `config.php`
 * Hiding the country field for shipping addresses will also hide the state field. The option label in the backend was adjusted to better describe this behaviour.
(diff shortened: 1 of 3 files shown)


svenmuennich:
Please note, that `unsubscribeEvent` relies on a properly working `removeListener` method in `Enlight_Event_Subscriber_Config`. The fix for that method is provided in #174.

samuelvogel:
#452 is the new PR as this is still relevant!

PR #183, opened 2014-09-25
Update Basic.php

jetwes:
fix for php bug https://bugs.php.net/bug.php?id=41820 which is causing bad backgrounds (mixed #ffffff and #fefefe) in thumbnails

### DIFF
--- a/engine/Shopware/Components/Thumbnail/Generator/Basic.php
+++ b/engine/Shopware/Components/Thumbnail/Generator/Basic.php
@@ -95,6 +95,24 @@ class Basic implements GeneratorInterface
             $originalSize['width'],
             $originalSize['height']
         );
+        /* fix #fefefe in white backgrounds */
+        $colorWhite = imagecolorallocate($newImage,255,255,255);
+        $processHeight = $newSize['height']+0;
+        $processWidth = $newSize['width']+0;
+        for($y=0; $y<($processHeight); ++$y){
+            // Travel x axis
+            for($x=0; $x<($processWidth); ++$x){
+                // Change pixel color
+                $colorat=imagecolorat($newImage, $x, $y);
+                $r = ($colorat >> 16) & 0xFF;
+                $g = ($colorat >> 8) & 0xFF;
+                $b = $colorat & 0xFF;
+                if(($r==253 && $g == 253 && $b ==253) || ($r==254 && $g == 254 && $b ==254)) {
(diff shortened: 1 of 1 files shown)


bcremer:
Hi @jetwes, thanks for your contribution.

I was not able to reproduce this issue.
1. Created a pure white #fff 800*600px JPG
2. Uploaded as product image
3. Downloaded generated thumbnails

I tried that with a jpg and png file.

In both cases the generated thumbnails were pure #fff.

jetwes:
Hi @bcremer ,

have a look at the pictures at the shopware Demo Shop. There you can see Problem as well. 
I think a pure white image doesn't have the effect, because imagecopyresampled() has no other colors than white for resampling. If there's a bit of black in the middle you can see this. Or just try this for example (the background is pure white)
lz-16742-strickjacke-fransen-schwarz_1_260x173
lz-16742-strickjacke-fransen-schwarz_1

bcremer:
Link to bugtracker report: http://jira.shopware.de/?ticket=SW-9404

bcremer:
As far as I can see the fix originates from this blog entry dated back in 2011: http://jasuten.com/background-noise-in-magento-image-resize/

jetwes:
correct!

jetwes:
you can see similar problems in all major shop systems - all caused by the php bug i mentioned before - here is an example for oxyd: https://bugs.oxid-esales.com/view.php?id=3642 . The first fix was for magento in 2011

bcremer:
Measured performance of the patch today. 
Unfortunately the patch has a huge perfomance impact. 
The time to regenerate all thumbnails for our demodata doubled.

Without patch:

``` bash
~/www/master (master) 
$ time ./bin/console sw:thumbnail:generate -f
./bin/console sw:thumbnail:generate -f  58.38s user 1.71s system 99% cpu 1:00.28 total
```

With patch:

``` bash
~/www/master (master) 
$ time ./bin/console sw:thumbnail:generate -f
./bin/console sw:thumbnail:generate -f  125.04s user 1.54s system 99% cpu 2:06.80 total
```

Under these circumstances we cannot include this patch in the shopware core.

If you rely on perfect thumbnail generation in a project you can write a plugin that replaces the default thumbnail generator. Please have a look at https://github.com/bcremer/SwagAsyncImageGenerator to get an idea how that would look like.

PR #187, opened 2014-09-29
Don't require matching countries for tax-free company orders

samuelvogel:
When shipping to companies in a tax-free country (should be EU, since non-EU/ROW countries are completly tax-free anyways) it does not matter where the invoice is adressed to, just shipping it to a destination outside Germany (respectivly the country where the shop is run from) makes it tax-free.

For reference see the German law [§ 6a Abs. 1 UStG](http://www.gesetze-im-internet.de/ustg_1980/__6a.html):

> Eine innergemeinschaftliche Lieferung [...] liegt vor, wenn [...] Der Unternehmer oder der Abnehmer hat den Gegenstand der Lieferung in das übrige Gemeinschaftsgebiet befördert oder versendet;


samuelvogel:
I also can't find this behavior documented anywhere, so a normal Shopware admin (without source code knowledge) would be confused why the option `Steuerfrei für Unternehmen` does not work when a company from Luxembourg orders with shipping to Ireland.

PR #188, opened 2014-10-02
Include the country in the address comparison

svenmuennich:
Currently, the address comparison in the backend order details does not include the countries of the billing and shipping addresses. However, since some customers change the country to get around certain country restrictions, it is important to compare the countries, too. This PR adds this check, by comparing the `countryId`s of the two addresses.

Because this fix is located in the ExtJS backend order app, I didn't include test cases.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,10 @@
 # Shopware Upgrade Information
 In this document you will find a changelog of the important changes related to the code base of Shopware.
 
+## 4.3.2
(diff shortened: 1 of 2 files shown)


svenmuennich:
I updated my commit to include also `department`, `salutation` and `city` in the address comparison. Additionally I also made `department` and `salutation` visible in the addresses in the backend order overview. This will make the applied comparison comprehensible to the user.

PR #193, opened 2014-10-29
Update autoload.php

mktcode:
Don't blame me if I am wrong but this require_once can't work... or am i missing something?


mktcode:
hm.... this installation process is confusing to me... sry for that

PR #194, opened 2014-10-30
adding new event to modify mail model

teiling88:


bcremer:
Hi @teiling88,

do you have a special use case for this event?

I'm not sure about the placing. 
I would have added the event more at the end of `createMail()`, but that depends on what you are going to archve with that.

dnoegel:
Depending on your need, also Enlight_Components_Mail_Send might be suitable, you get an instance of the Enlight_Components_Mail there - which is basically populated by the \Shopware\Models\Mail\Mail model

teiling88:
Hey @bcremer,
I copied an older local version of this file in this pull request - sorry for that. 

My special use case is, that I develop a plugin which makes it possible to create email templates as *.tpl files in templatename/email/ for each subshop template.

@dnoegel i tried this event but i don't get the Shop information there :-(

teiling88:
i will clean up my pull request and open a new one.

PR #196, opened 2014-11-05
partnerID in s_order should be nullable

Haehnchen:
partnerID should be nullable to reflect s_order_basket and make it more doctrine orm friendlier. its a foreign key on an empty string right now, there are also some other sideeffects. "migrations" not provided by pull request...

### DIFF
--- a/_sql/migrations/317-update-order-nullable-fields.php
+++ b/_sql/migrations/317-update-order-nullable-fields.php
@@ -1,0 +1,18 @@
+<?php
+class Migrations_Migration317 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up()
+    {
+        $this->addSql("
+            ALTER TABLE `s_order` CHANGE `partnerID` `partnerID` VARCHAR( 255 ) CHARACTER SET utf8 COLLATE utf8_unicode_ci NULL;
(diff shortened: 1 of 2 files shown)


_sql/install/latest.sql
```diff
@@ -5287,7 +5287,7 @@ CREATE TABLE IF NOT EXISTS `s_order` (
   `internalcomment` mediumtext COLLATE utf8_unicode_ci NOT NULL,
   `net` int(1) NOT NULL,
   `taxfree` int(11) NOT NULL,
-  `partnerID` varchar(255) COLLATE utf8_unicode_ci NOT NULL,
+  `partnerID` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
```
bcremer: Please do not alter the `_sql/install/latest.sql` file directly.
Instead add a migration file to `_sql/migrations/` that does an alter table statement.

During build/deploy time we import `_sql/install/latest.sql` and then apply the migrations from `_sql/migrations/` on top.

To apply the migrations you can use the `build/ApplyDeltas.php` script or run `ant build-unit` inside the build directory.

Haehnchen:
@bcremer migration are in, also added more nullable fields. too crazy now, but that reflects internal usage :)

PR #198, opened 2014-11-14
Speed up risk rule loading

andrewisplinghoff:
When having many risk rules (>50), the risk management panel takes a lot of time to load the rules. With ~250 rules the panel becomes nearly unusable with a delay > 2:30 minutes. The problem is that with the current implementation, the rule fieldset gets rerendered for every single rule that exists. By hiding the fieldset prior to adding the rules and showing it only after all rules have been added, this process is immensely accelerated.


bcremer:
:+1: 

Will be merged soon.

Thanks a lot.

PR #200, opened 2014-11-18
adding function which execute risk rules

ghost:
adding a function which execute risk rules. this function can be hooked so it's possible to create custom risk rules


bcremer:
Looks like a good addition to me. 

:+1:

bcremer:
Thank you very much.

dnoegel:
If I am not mistaken this can only be replace-hooked, as you cannot modify the return value with a  before-Hook and the after-Hook will never be executed as trying to access a non-existing rule in the original method will already result in an Exception. Or did I miss something?

PR #201, opened 2014-11-20
newsletter/indexAction cleanup and passing groupID for Subscription

barbieswimcrew:
Shopware()->Modules()->Admin()->sNewsletterSubscription() as used in the indexAction() of the frontend newsletter controller just got passed the subscribed email address and a boolean flag for subscribing/unsubscribing. but the method actually has a third parameter to set the newsletter group optionally. so it isn't possible to subscribe to another newsletter group than the default group.

this feature now will pass the $groupID if existing to make it possible to subscribe to different newsletter groups.

to control the subscription this feature wraps the newsletterExtendedFields in _emotion/frontend/newsletter/index.tpl in a separate smarty block for appending or prepending an input field for setting the groupID to subscribe.

replaced Shopware()-Db() query by new doctrine model persist/flush operation.

this feature also replaces some deprecated Shopware()->System()->_POST function calls by accessing the post data via $this->Request()->getPost() in indexAction() and even doing some tiny clean ups.

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Newsletter.php
+++ b/engine/Shopware/Controllers/Frontend/Newsletter.php
@@ -120,57 +120,64 @@ class Shopware_Controllers_Frontend_Newsletter extends Enlight_Controller_Action
      */
     public function indexAction()
     {
-        $this->View()->voteConfirmed = $this->isConfirmed();
 
-        if (isset($this->request()->sUnsubscribe)) {
+        if (isset($this->Request()->sUnsubscribe)) {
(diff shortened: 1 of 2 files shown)


bcremer:
Thanks for your PR.
I'll have a closer look into it soon.
For the next time, please split you PR into separate commits. One for cleaning up, one for functional changes. That makes it easier to review for us.

barbieswimcrew:
Okay, I'm gonna pay heed to that for the next time! Thank you Benjamin!

PR #202, opened 2014-11-28
SW-9795 REST-API EAN

netaviator:
Search for EAN via REST-API

http://jira.shopware.de/?ticket=SW-9795

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Article.php
+++ b/engine/Shopware/Components/Api/Resource/Article.php
@@ -84,6 +84,43 @@ class Article extends Resource implements BatchInterface
     {
         return $this->getResource('Media');
     }
+    
+    /**
+     * Little helper function for the ...ByEan methods
+     * @param $ean
+     * @return int
+     * @throws \Shopware\Components\Api\Exception\NotFoundException
+     * @throws \Shopware\Components\Api\Exception\ParameterMissingException
+     */
+    public function getIdFromEan($ean)
+    {
+        if (empty($ean)) {
+            throw new ApiException\ParameterMissingException();
+        }
+
+        /** @var $articleDetail \Shopware\Models\Article\Detail */
+        $articleDetail = $this->getDetailRepository()->findOneBy(array('ean' => $ean));
+
+        if (!$articleDetail) {
+            throw new ApiException\NotFoundException("Article by ean {$ean} not found");
+        }
+
+        return $articleDetail
+            ->getArticle()
+            ->getId();
+    }
+    
+    /**
+     * Convenience method to get a article by ean
+     * @param string $ean
+     * @param array $options
+     * @return array|\Shopware\Models\Article\Article
+     */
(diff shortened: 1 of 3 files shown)


bcremer:
Thanks @jangilla.

I will have a closer look on monday, but looks promising on first sight. 

We were also thinkink about a way to generalize the `findOneBy` type of queries for getting a unique results.
A additional parameter like `?identifierField=ean` for example.

netaviator:
Sounds good! Thanks for your quick reply!

engine/Shopware/Controllers/Api/Variants.php
```diff
@@ -43,11 +43,16 @@ public function getAction()
...
         $id = $this->Request()->getParam('id');
         $useNumberAsId = (boolean) $this->Request()->getParam('useNumberAsId', 0);
+        $useEanAsId = (boolean) $this->Request()->getParam('useEanAsId', 0);
 
         if ($useNumberAsId) {
             $articleDetail = $this->resource->getOneByNumber($id, array(
                 'considerTaxInput' => $this->Request()->getParam('considerTaxInput')
             ));
+        } elseif ($useEanAsId) {
+            $articleDetail = $this->resource->getOneByEan($id, array(
```
bcremer, line 53: `getOneByEan` is not implemented on the variant resource.

bcremer:
Can you please commit your PR again with a real mail address and english commit message please?

engine/Shopware/Components/Api/Resource/Article.php
```diff
@@ -84,6 +84,43 @@ protected function getMediaResource()
...
+     * @throws \Shopware\Components\Api\Exception\ParameterMissingException
+     */
+    public function getIdFromEan($ean)
+    {
+        if (empty($ean)) {
+            throw new ApiException\ParameterMissingException();
+        }
+
+        /** @var $articleDetail \Shopware\Models\Article\Detail */
+        $articleDetail = $this->getDetailRepository()->findOneBy(array('ean' => $ean));
```
bcremer, line 102: `findOneBy` does not check for uniqueness, also the `ean` field is not unqiue in the database.

I put together a quick and dirty PR as an alternative, see: https://github.com/bcremer/shopware/pull/1/files.

dnoegel:
If we make this more generic, we should also implement that in the "getIdByData" method for the article resource, so that it is supported in batch-request. Furthermore: should we also think about allowing this for attributes?

bcremer:
@dnoegel regarding the attributes, yes we should.

My PR is a bit naive, it only takes fields from the variant into account.

We need a way to specifiy the entity like `a.name`, `d.ean`, `d.customAttribute` like in the filter query.
Or we skip prefixing altogether and remove the artifical split of the  article + variant and work with a whitelist of fields that we map to the matching model.

netaviator:
I'm unable to recommit this request because I deleted the repository. I'll make a new PR. Please delete the #202 and #207.

PR #203, opened 2014-12-01
Add SQL filter event to the check if an article is already in the basket

andrewisplinghoff:
With this, plugins can influence how the basket determines equality of articles when adding a new article. This filter event is needed by plugins that extend the article model in a way so that it is not enough anymore to just check if order numbers are equal.

### DIFF
--- a/engine/core/class/sBasket.php
+++ b/engine/core/class/sBasket.php
@@ -1445,19 +1445,29 @@ class sBasket
         }
 
         // Check if article is already in basket
-        $chkBasketForArticle = $this->db->fetchRow(
-            'SELECT id, quantity
-            FROM s_order_basket
-            WHERE articleID = ?
-            AND sessionID = ?
-            AND ordernumber = ?
-            AND modus != 1',
+        $builder = Shopware()->Models()->getConnection()->createQueryBuilder();
+        $builder->select('id', 'quantity')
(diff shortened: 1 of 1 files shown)


bcremer:
Thanks for you contribution.
Generally the additional event seems reasonably to me.

But can you please rewrite the query to use the DBAL query builder instead (`Shopware()->Models()->getConnection()->createQueryBuilder()`) and emit the query builder in a notify event.

We don't want to include more filterSQL events on plain SQL strings, because they are not really extendable in a sane way.

andrewisplinghoff:
Thanks for the fast feedback. Any more issues? Should I squash the commits?

bcremer:
Yes, please squash the commits.

Looks good to me, I'll do the final review tomorrow.

bcremer:
Please note that I changed the event from `filter` to a `notify` event, see example listener:

```
$this->eventManager->addListener(
    'Shopware_Modules_Basket_AddArticle_CheckBasketForArticle',
    function (Enlight_Event_EventArgs $eventArgs) {
        /** @var \Doctrine\DBAL\Query\QueryBuilder $qb */
        $qb = $eventArgs->get('queryBuilder');
        $qb->andWhere('somefoo = somebar');
    }
);
```

Thanks again for your contribution :+1:

PR #204, opened 2014-12-01
hookable functions in Shopware_Controllers_Widgets_Emotion

Haehnchen:
`Shopware_Controllers_Widgets_Emotion` doesnt allow modification of eg `getBlogEntry`, because of a private functions. in our use case we want to extend core blog widget with a category filter, but we are not able to reuse core because of missing hookable possibility. just make them protected resolves this.

### DIFF
--- a/engine/Shopware/Controllers/Widgets/Emotion.php
+++ b/engine/Shopware/Controllers/Widgets/Emotion.php
@@ -174,7 +174,7 @@ class Shopware_Controllers_Widgets_Emotion extends Enlight_Controller_Action
         $this->View()->assign('Controller', (string) $this->Request()->getParam('controllerName'));
     }
 
-    private function getArticle($data, $categoryId, $element)
+    protected function getArticle($data, $categoryId, $element)
     {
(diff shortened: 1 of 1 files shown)


OliverSkroblin:
Hey, 

i would prefer an alternativ to make all this private functions protected and hookable:

Reset the convert_function of each emotion component which you want to handle:

```
UPDATE s_library_component 
    SET convert_function = NULL
    WHERE name IN ('Blog-Artikel', '..');
```

In the `Shopware_Controllers_Widgets_Emotion_AddElement` event you can modify the data for this reseted components in your plugin.

Please restore the convert functions if your plugin will be uninstalled ;)

bcremer:
@Haehnchen is the solution proposes by @OliverDenter practicable for you? I would like to close the ticket then.

Haehnchen:
not really. in this case its easier to just create my own widgets. :)

no way to make them protected or provide an additional event? there is also an external discussion: http://forum.shopware.com/templates-und-design-f57/author-in-emotion-blog-component-ausgeben-t21494.html

PR #209, opened 2014-12-12
Exclude shipping costs from order amount when shipping costs are hidden

andrewisplinghoff:
When the document is configured to not include the shipping costs as a position, as is the case with delivery notes per default, the order amount should also not include the shipping costs, because otherwise the sum of the positions does not equal the order amount. This does normally not attract attention, but if for example the delivery note template is changed to also show the prices of the positions and the total order amount, the problem becomes obvious.

Before this change, the sum of the order positions does not equal the order amount:
selection_295

After application of this fix:
selection_294

### DIFF
--- a/engine/Shopware/Models/Document/Order.php
+++ b/engine/Shopware/Models/Document/Order.php
@@ -303,20 +303,22 @@ class
             $taxShipping = Shopware()->Config()->sTAXSHIPPING;
         }
         $taxShipping = (float) $taxShipping;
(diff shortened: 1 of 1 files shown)


samuelvogel:
@bcremer: I just discussed this with @dnoegel and he told me to just ping you again on this one :wink:

andrewisplinghoff:
I opened a new PR for this: https://github.com/shopware/shopware/pull/384

PR #211, opened 2014-12-14
Add possibility to retrieve logged in user in smarty templates

samuelvogel:
Should be useful for a lot of people, also see: http://forum.shopware.com/topic24481.html

**Disclaimer:** Not tested with active HTTP cache module, not sure if there are any implications. Needs to be checked before merging this PR!

### DIFF
--- a/engine/Library/Enlight/Template/Plugins/function.activeusername.php
+++ b/engine/Library/Enlight/Template/Plugins/function.activeusername.php
@@ -1,0 +1,10 @@
+<?php
+
+function smarty_function_activeUserName($params, $smarty)
+{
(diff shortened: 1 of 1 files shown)


bcremer:
Thanks again for the next PR, you guys seem to be on fire right now ;)
Sadly most of us are on vacation right now (me included) so I might review this starting 2015.

Also as you already mentioned this has impact to the HTTP Cache.

 We want to prevent that people accidentally show cached private information to other customers. 

So instead of a global smarty plugin I would set the username in the post dispatch event of the checkout widget.

samuelvogel:
Yeah that would be perfect! We wrote this code two years ago for a client which doesn't use the HTTP Cache and just wanted to share it because we saw the question in the forum.

PR #212, opened 2014-12-14
Clear cache after composer install/update

samuelvogel:
Clearing the cache after composer runs is generally a good idea to prevent spurious errors caused by outdated outdated cache files.

Together with #210 this prevents even more strange errors in the backend.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,5 +60,13 @@
     },
     "config": {
         "autoloader-suffix": "Shopware"
+    },
+    "scripts": {
+        "post-install-cmd": [
+            "cache/clear_cache.sh -f"
+        ],
+        "post-update-cmd": [
+            "cache/clear_cache.sh -f"
+        ]
     }
 }


bcremer:
Thanks again @samuelvogel.

bcremer:
I had to revert the PR with with commit: https://github.com/shopware/shopware/commit/c248884f4a60db4f921104292cdf4727d4578098

The commit broke the builds, see: https://travis-ci.org/shopware/shopware/builds/44228070.

I am not sure why this was not detected during the travis run on the PR: https://travis-ci.org/shopware/shopware/builds/44019631

/cc @samuelvogel

samuelvogel:
Hmm, that is too bad. It fails since it tries to connect to the database, which shouldn't be needed anyways, correct?

Can his somehow be prevented when running the `sw:generate:attributes` CLI command? Or it could silently fail, since in that case Shopware is not yet ready to use anyways.
The former would be nicer though, since we prebuild our deployment in a temporary folder and then rsync them into the productive docroot (where it is merged with the `config.php` for example). So being able to generate the attribute models in a dry environment would be nice!

bcremer:
`sw:generate:attributes` requires a connected shopware database, since the model attributes are generated using the doctrine schema manager.

PR #213, opened 2014-12-16
Fix nullable annotation of customerId in the Basket model

andrewisplinghoff:
The userID column of the s_order_basket table is set to NOT NULL (https://github.com/shopware/shopware/blob/master/_sql/install/latest.sql#L5343) and the corresponding variable of the model should have an appropriate annotation. This change should be unproblematic, because the Basket model is not being used yet (the basket logic uses the deprecated sBasket class at the moment).

### DIFF
--- a/engine/Shopware/Models/Order/Basket.php
+++ b/engine/Shopware/Models/Order/Basket.php
@@ -52,9 +52,9 @@ class Basket extends ModelEntity
     /**
      * @var integer $customerId
      *
-     * @ORM\Column(name="userID", type="integer", nullable=true)
+     * @ORM\Column(name="userID", type="integer", nullable=false)
      */
-    protected $customerId = null;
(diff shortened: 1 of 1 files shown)


bcremer:
Thanks @andrewisplinghoff for this PR.

Do you have any reproducable problems with that field or is this more a cosmetic change?
This seems to be related to https://github.com/shopware/shopware/pull/196/.

samuelvogel:
It's just cosmetically, we noticed this while debugging a problem in one of our plugins. Seems to be related to #196 yes, but just by coincidence!

In our opinion it would be better to keep `nullable=true` and make the field nullable in the database as well, but that would need changes to the rest of the beloved s_classes (`0` is used as default there) which we wanted to leave up to you guys :wink:

Haehnchen:
is really annoying to working with doctrine models here.

PR #215, opened 2014-12-17
add new replacements in sCleanupPath function

teiling88:
based on our experience on a france shop we extended this class. 
Maybe an Integration of https://github.com/Behat/Transliterator resolves future problems with this function.


bcremer:
Thanks @teiling88.

I like the idea to replace the function (`sCleanupPath`) with a better suited library.

Instead of https://github.com/Behat/Transliterator I would prefer https://github.com/cocur/slugify: 
- under active development
- usecase matches better (concentration on slugs, not general utf-8 transliterations)
- more lightweight 
- MIT Licensed instead of [Artistic License 1.0](http://opensource.org/licenses/artistic-license-1.0)

In the meantime we could add the replacement table from https://github.com/cocur/slugify/blob/master/src/Slugify.php#L29 instead of only the french replacements.

teiling88:
@bcremer thanks four your feedback. 

I merged the replacement table from slugify.

engine/core/class/sRewriteTable.php
```diff
@@ -163,28 +163,500 @@ public function sCleanupPath($path, $remove_ds = true)
...
         $replace = array(
             ' & ' => '-und-',
-            'ä' => 'ae',
-            'ö' => 'oe',
-            'ü' => 'ue',
-            'Ü' => 'Ue',
-            'Ä' => 'Ae',
-            'Ö' => 'Oe',
-            'ß' => 'ss',
-            ':' => '-',
```
bcremer, line 173: I think you removed a few shopware specific replacements here.

I'll do a separate pull request where tests and some backlinks are provided in a few minutes.

bcremer:
I would like to close this PR and do further discussions in PR #216.

PR #218, opened 2014-12-23
Prevent apache from serving .git-directory

Rokko11:
If you prefer vcs-deployment, you should protect the .git-Directory.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -19,6 +19,9 @@ RewriteRule ^(.*)$ shopware.php [PT,L,QSA]
 
 # Fix missing authorization-header on fast_cgi installations
 RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization},L]
+
+# Prevent apache from serving .git-directory
+RedirectMatch 404 /\\.git(/|$)
 </IfModule>
 
 # Staging-Rules start


.htaccess
```diff
@@ -19,6 +19,9 @@ RewriteRule ^(.*)$ shopware.php [PT,L,QSA]
 
 # Fix missing authorization-header on fast_cgi installations
 RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization},L]
+
+# Prevent apache from serving .git-directory
+RedirectMatch 404 /\\.git(/|$)
```
bcremer, line 24: Can't this be simplified to `RedirectMatch 404 /\.git`?

bcremer:
Thanks @Rokko11, we will have a closer look into it after the holidays.

.htaccess
bcremer, line 24: Stupid me, this would obviously also block access to `.gitignore` and `.gitter/` for example, not sure if this will be a real world problem but might lead to hard to debug behaviours.

bcremer:
I would like to go with the following line:

```
# Block access to VCS directories
RedirectMatch 404 /\\.(svn|git|hg|bzr|cvs)(/|$)
```

May you please update/squash your commit?

bcremer:
Thanks @Rokko11, this will be included in the SW 4.3.3 release.

PR #220, opened 2015-01-08
Fixes occaisonal rounding issues if voucher code matches total

shredding:
If the basket contains a voucher code matching the exact total of the basket, there are rare occasions where the basket size could become a very small negative value, causing the basket to be shown as empty.

Sample:

``` php
<?php

$item1 = 19.95;
$item2 = 129.95;
$voucher = -149.90;
var_dump($item1 + $item2 + $voucher);

>> double(-2.8421709430404E-14);
```

So, just round after each addition and everythings smooth again.

### DIFF
--- a/engine/core/class/sBasket.php
+++ b/engine/core/class/sBasket.php
@@ -2056,8 +2056,8 @@ class sBasket
             }
 
             $totalAmount = round($totalAmount + round($getArticles[$key]["amount"],2), 2);
-			// Needed if shop is in net-mode
-			$totalAmountWithTax = rount($totalAmountWithTax + round($getArticles[$key]["amountWithTax"],2), 2);
+	    // Needed if shop is in net-mode
+	    $totalAmountWithTax = rount($totalAmountWithTax + round($getArticles[$key]["amountWithTax"],2), 2);
 
             // Ignore vouchers and premiums by counting articles
             if (!$getArticles[$key]["modus"]) {
--- a/engine/core/class/sBasket.php
+++ b/engine/core/class/sBasket.php
@@ -2055,9 +2055,10 @@ class sBasket
                 }
             }
 
-            $totalAmount += round($getArticles[$key]["amount"], 2);
-            // Needed if shop is in net-mode
-            $totalAmountWithTax += round($getArticles[$key]["amountWithTax"], 2);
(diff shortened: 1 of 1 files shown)


engine/core/class/sBasket.php
```diff
@@ -2055,9 +2055,10 @@ private function getBasketArticles($getArticles)
                 }
             }
 
-            $totalAmount += round($getArticles[$key]["amount"], 2);
-            // Needed if shop is in net-mode
-            $totalAmountWithTax += round($getArticles[$key]["amountWithTax"], 2);
+            $totalAmount = round($totalAmount + round($getArticles[$key]["amount"],2), 2);
+			// Needed if shop is in net-mode
+			$totalAmountWithTax = rount($totalAmountWithTax + round($getArticles[$key]["amountWithTax"],2), 2);
```
bcremer: Can you please fix the whitespace here. Were using 4 spaces for indentation.

bcremer:
Thank you @shredding, we will have a look into it soon.

shredding:
Spacing is fixed.

engine/core/class/sBasket.php
```diff
@@ -2055,9 +2055,10 @@ private function getBasketArticles($getArticles)
                 }
             }
 
-            $totalAmount += round($getArticles[$key]["amount"], 2);
-            // Needed if shop is in net-mode
-            $totalAmountWithTax += round($getArticles[$key]["amountWithTax"], 2);
+            $totalAmount = round($totalAmount + round($getArticles[$key]["amount"],2), 2);
+	    // Needed if shop is in net-mode
+	    $totalAmountWithTax = rount($totalAmountWithTax + round($getArticles[$key]["amountWithTax"],2), 2);
```
samuelvogel, line 2060: Should probably be `round()`, shouldn't it?

bcremer, line 2060: Thank you @samuelvogel. Good catch.
I think we have to fix our Travis-CI configuration.

The unit tests failed but that did not trigger a failed travis build:
https://travis-ci.org/shopware/shopware/jobs/46585580#L2298

I'll check that.

Edit:
Created PR https://github.com/shopware/shopware/pull/223 for that.

PR #222, opened 2015-01-12
Make doctrine cache namespace configurable

stefanheppenheimer:
This commit adds a configuration parameter to define a custom namespace for the doctrine cache provider. This feature is e.g. necessary if one wants to run multiple Shopware instances on one server using PHP FPM chroot containers. In this case the doctine proxy path inside the chroot, on which the current namespacing relies, may be equal for the different Shopware instances, whereby caching interferences will be caused.

### DIFF
--- a/engine/Shopware/Components/Model/Configuration.php
+++ b/engine/Shopware/Components/Model/Configuration.php
@@ -56,6 +56,13 @@ class Configuration extends BaseConfiguration
     protected $fileCacheDir;
 
     /**
+     * Custom namespace for doctrine cache provider
+     *
+     * @var string
+     */
+    protected $cacheNamespace = null;
+
+    /**
      * @param array $options
      * @param \Zend_Cache_Core $cache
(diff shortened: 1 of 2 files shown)


bcremer:
Thank you @stefanheppenheimer.

I see the point in your PR. But don't we have anything that can be encoded into the namespace that is unique per instance? I am thinking about some $_SERVER var that we might use for that.

Any ideas?
I would like to prevent addional config values.

stefanheppenheimer:
I like your point of view to be sensitive about configuration parameters, but in my opinion using environmental values, like those stored within `$_SERVER`, isn't a solution on which we can 100% rely on. The suggested solution on the other side doesn't further tie a Shopware instance to it's environment and by setting a suitable default value (`null` in this case) the new parameter is completely optional.

bcremer:
Thank you @stefanheppenheimer, I am convinced ;)

PR #224, opened 2015-01-14
article restriction in voucher module

ghost:
the minimum ordernumber length is four characters (see http://wiki.shopware.com/Artikelnummern-in-Shopware_detail_642.html). the voucher module required six characters minimum string length. this is wrong


ghost:
hi @bcremer 
can you have a short look on this bugfix?

PR #229, opened 2015-01-26
Only write an error to the log when a mail attachment file is missing

andrewisplinghoff:
Previously, this condition caused an exception. Throwing an exception here is problematic, because for example in a test environment, certain media files might be missing. It should still be possible to test a shop without having to have all media files available.


samuelvogel:
Also throwing an exception in the last step of the checkout process is never a good idea for an ecommerce software, shop admins want the money :wink:

bcremer:
Thanks again guys. @andrewisplinghoff @samuelvogel.

PR #230, opened 2015-01-27
Allow generated attribute models to be referenced in custom models via a ManyToOne mapping

svenmuennich:
Sometimes it is desirable to have a `ManyToOne` mapping from a custom model in a plugin to a default Shopware model. By utilizing the attribute model generation, this commit dynamically adds properties to the attribute models, which have the inverting `OneToMany` annotation. In addition, a custom constructor for initializing the `ArrayCollection`s as well as respective getter functions are added. This is done by just adding the `ManyToOne` annotation to a property in a custom model and setting the `targetEntity` and `inversedBy` attributes accordingly. That information will then be used to determine the name of the property and getter method in the generated model.

Example:

``` php
<?php

namespace Shopware\CustomModels\UniquePrefix;

...

/**
 * @ORM\Table(name="s_test_attribute_feature")
 * @ORM\Entity
 */
class AttributeFeature extends ModelEntity
{

    ...

    /**
     * OWNING SIDE
     *
     * @ORM\ManyToOne(targetEntity="Shopware\Models\Attribute\Blog", inversedBy="uniquePrefixAttributeFeatures")
     * @ORM\JoinColumn(name="attribute_id", referencedColumnName="id")
     */
    private $attribute;

    ...

}

```

Please note that it's essential to use a unique prefix for the `inversedBy` attribute, since an exception will be thrown, if any two generated `OneToMany` properties have the same name. Additionally, only models, which are registered in the namespace `Shopware\CustomModels`, will be considered when generating the inverting `OneToMany` properties.

To create such a mapping in a plugin, just add the following steps to the plugin `Bootstrap.php`:

``` php
// 1. Register the custom models:
$this->registerCustomModels();

// 2. Re-generate the referenced ('inversedBy') attribute models
Shopware()->Models()->generateAttributeModels(array(
    's_blog_attributes'
));

// 3. Create the schema for the custom models
$em = Shopware()->Models();
$tool = new \Doctrine\ORM\Tools\SchemaTool($em);
$tool->createSchema(array(
    $em->getClassMetadata('Shopware\CustomModels\UniquePrefix\AttributeFeature')
));
```

Steps `2.` and `3.` can be performed in any order. However, the custom models having the `ManyToOne` mappings must be registered before, using `registerCustomModels()` as defined in `Shopware_Components_Plugin_Bootstrap`.

### DIFF
--- a/engine/Shopware/Components/Model/Generator.php
+++ b/engine/Shopware/Components/Model/Generator.php
@@ -65,6 +65,11 @@ class Generator
 ';
 
     /**
+     * The namespace which contains all generated attribute models.
+     */
+    const ATTRIBUTE_NAMESPACE = 'Shopware\Models\Attribute';
+
+    /**
      * Contains the required namespaces for a shopware model
      */
     const NAMESPACE_HEADER = '
@@ -108,7 +113,7 @@ class %className% extends ModelEntity
     /**
      * Definition of a standard shopware association property.
      */
-    const ASSOCIATION_PROPERTY = '
+    const ONE_TO_ONE_ASSOCIATION_PROPERTY = '
     /**
      * @var \%foreignClass%
      *
@@ -118,6 +123,30 @@ class %className% extends ModelEntity
      * })
      */
     protected $%property%;
+    ';
+
+    /**
+     * Definition of an OneToMany association property, which inverts
+     * an association property of a custom model.
+     */
+    const ONE_TO_MANY_ASSOCIATION_PROPERTY = '
+    /**
+     * INVERSE SIDE
+     *
+     * @var \Doctrine\Common\Collections\ArrayCollection
+     *
+     * @ORM\OneToMany(targetEntity="%owningClass%", mappedBy="%owningProperty%")
+     */
+    protected $%property%;
+    ';
+
+    /**
+     * Definitition of a constructor for initializing properties.
+     */
+    const CONSTRUCTOR = '
+    public function __construct() {
+        %propertyInitializations%
+    }
     ';
 
     /**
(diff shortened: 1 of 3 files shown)


bcremer:
Thank you @svenmuennich for this huge contribution. This looks indeed highly interesting.

We're quite busy right now preparing a release. But I'l go through the open pull requests in a few days.

engine/Shopware/Components/Model/Generator.php
```diff
@@ -667,6 +723,58 @@ protected function getAssociationProperty($table, $foreignKey)
     }
 
     /**
+     * Creates the source code for the doctrine association properties,
+     * which inverse the passed models. That is, models having a property
+     * that references the currently generated model in an 'inversedBy' field
+     * of a ManyToOne annotation.
+     *
+     * @param $referencingModels array
```
bcremer: The typehint should come first in the `@param` docblock.

`@param array $referencingModels`

engine/Shopware/Components/Model/Generator.php
```diff
@@ -667,6 +723,58 @@ protected function getAssociationProperty($table, $foreignKey)
...
+     * @return array
```
bcremer: When the return value is a array of a specific type like a string please use `@return string[]`.

engine/Shopware/Components/Model/Generator.php
```diff
@@ -667,6 +723,58 @@ protected function getAssociationProperty($table, $foreignKey)
...
+
+    /**
+     * Creates the source code for the custom constructor,
+     * which initializes each ManyToOne property with an empty
+     * ArrayCollection.
+     *
+     * @param $referencingModels array
+     * @return string
+     */
+    protected function getConstructor($referencingModels) {
```
bcremer: Please put the opening braces to the next line.

bcremer:
We will have a closer look onto that after the shopware 5 Release. Sorry for the delay.

PR #231, opened 2015-01-27
Add risk management functions for dunning level 1, 2 and 3

ghost:
In order to improve the risk management, it should be possible to block a payment method if there has been any kind of problem with the payment in the past. Therefor not only the payment status 16 (encashment) should be checked, but also 13, 14 and 15 (dunning level 1, 2 and 3).


bcremer:
Thank you very much @chandrakirchrath. This looks reasonable at first glance. I'll have a in depth review soon(ish).

You could also add the new risk functions to the existing testcase: https://github.com/shopware/shopware/blob/master/tests/Shopware/Tests/Core/sAdminTest.php#L2309

engine/core/class/sAdmin.php
```diff
@@ -4962,4 +4996,27 @@ private function handlePaymentMeanSurcharge($country, $payment, $currencyFactor,
...
     }
+
+    /**
+     * Convenience function to check if there is at least one order with the
+     * provided cleared status.
+     * 
+     * @param int $cleared
+     * @return boolean
+     */
+    private function riskCheckClearedLevel($cleared) {
```
bcremer: Please put the opening braces onto a new line  to adhere to [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md).

engine/core/class/sAdmin.php
```diff
@@ -4962,4 +4996,27 @@ private function handlePaymentMeanSurcharge($country, $payment, $currencyFactor,
...
+        if ($this->session->offsetGet('sUserId')) {
```
bcremer: Can you please use the condition as a guard clause and do an early return.

```
if (!$this->session->offsetGet('sUserId')) {
   return false;
}

// query here
```

ghost:
The manage testsManageRisks method of sAdminTest seems to be incomplete. I think all payment risk rules should be checked with a positive and negative order. At the moment, there is only one order with a payment status of 16.

bcremer:
Thanks for updating the pull request. Looks fine to me now.

As for the incomplete `testsManageRisks` tests:
You're absolutely right, they should all be checked for both outcomes.

Feel free to refactor the tests in a separate commit or even a separate PR. 

 Just let me now if this PR is ready to merge from your side.

ghost:
From my side this PR is ready to merge. Sorry for the delay.

bcremer:
Thank you @chandrakirchrath. We will merge this PR for the SW 4.3.3 release soon.

PR #232, opened 2015-02-02
Allow anonymous event listeners

dnoegel:
I added support for anonymous functions as event listeners. I think its pretty useful for those small event callbacks you use to have for e.g. RegisterResource-Events. 

Downsides: 
- another format to support
- currently all event listeners could easily be written to database or could be stored otherwise

### DIFF
--- a/engine/Library/Enlight/Event/EventManager.php
+++ b/engine/Library/Enlight/Event/EventManager.php
@@ -355,6 +355,10 @@ class Enlight_Event_EventManager extends Enlight_Class
         foreach ($subscriber->getSubscribedEvents() as $eventName => $params) {
             if (is_string($params)) {
                 $this->addListener($eventName, array($subscriber, $params));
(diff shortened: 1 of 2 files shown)


bcremer:
Team Neversleep \o/

But I am not quite sure if it's good practice to use callbacks in a subscriber. If you want to do it quick and dirty, why not add the callback directly to the EventManager?

Also this will break compatibility with the Symfony EventDispatcher where our `addSubscriber()` method originated from. See: https://github.com/symfony/EventDispatcher/blob/master/EventDispatcher.php#L117 

Together with the downsides you already mentioned and the additional complexity I am not convinced to merge this PR.

dnoegel:
It's not about quick and dirty - its about fancy callbacks to deal with nasty events :)

But already convinced, not sure if we should take all the disadvantages for some convenience.

PR #234, opened 2015-02-06
SW-10224 removed limitation on main shop for snippet export

benboit:
This pull-request allows the export of all existing template snippets. This removes the limitation in shopware, that only snippets from the template used in the main shop are exported.


bcremer:
Thanks ben,

the constraint to shopId 1 seems indeed very random. 

I think this PR is good to merge for the next release window.

PR #237, opened 2015-02-09
Added a path fallback for the Path()-Getter

b3nl:
Many Plugins (SwagPaymentKlarna, SwagTrustedShopsExcellence ...) use Shopware_Components_Plugin_Bootstrap::Path() before the required info property is loaded, so I inserted a fallback.

### DIFF
--- a/engine/Shopware/Components/Plugin/Bootstrap.php
+++ b/engine/Shopware/Components/Plugin/Bootstrap.php
@@ -206,7 +206,19 @@ abstract class Shopware_Components_Plugin_Bootstrap extends Enlight_Plugin_Boots
      */
     final public function Path()
     {
-        return $this->info->path;
+        $return = '';
+
+        if ($this->info instanceof Enlight_Config) {
+            $return = $this->info->path;
+        } else {
+            $reflection = new \ReflectionClass($this);
+
+            if ($fileName = $reflection->getFileName()) {
+                $return = dirname($fileName) . DIRECTORY_SEPARATOR;
+            }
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Plugin/Bootstrap.php
```diff
@@ -206,7 +206,19 @@ public function disable()
      */
     final public function Path()
     {
-        return $this->info->path;
+        $return = '';
+
+        if ($this->info instanceof Enlight_Config) {
+            $return = $this->info->path;
+        } else {
```
bcremer, line 213: Could't we just use `__DIR__` instead of reflection for the else branch?

b3nl, line 213: Nope. **DIR** http://php.net/manual/de/language.constants.predefined.php is imho the path of engine/Shopware/Components/Plugin/Bootstrap.php and not the bootstrap of the actual module.

bcremer, line 213: Stupid me. Of couse you are right. Plugin vendors should just use `__DIR__` instead of `Path()` in their plugin bootstraps... but thats another issue.

Your solution looks fine. We will merge that for the 4.3.3 release.

b3nl, line 213: +1²

bcremer:
Thanks @b3nl, it's merged and will be released in Shopware 4.3.3 soon.

PR #240, opened 2015-02-20
$setFirstVariantMain bug

netformic:
If $setFirstVariantMain is true a new variant element gets appended to $data['variants'] array which resolves in an ValidationException.

Fix by setting isMain on $data['variants'][0]


bcremer:
Thank you @netformic for this PR.
Can you provide a (previously) failing testcase in `engine/Shopware/Components/Api/Resource/Article.php` so we can reproduce this issue?

PR #241, opened 2015-02-23
fixing http header error message

teiling88:
with activated http cache and a config.php file like this

```
<?php return array (
  'db' => 
  array (
    'username' => 'root',
    'password' => 'root',
    'host' => 'localhost',
    'port' => '3306',
    'dbname' => 'shopware',
  ),'front' => array(
          'noErrorHandler' => true,
          'throwExceptions' => true,
          'useDefaultControllerAlways' => true,
          'disableOutputBuffering' => true,
          'showException' => true,
      ),
      'template' => array(
          'forceCompile' => true,
      )
);
```

you will see an fatal error message under the captcha image like this: "Fatal error: Uncaught exception 'Zend_Controller_Response_Exception' with message 'Cannot send headers; headers already sent in /var/www/wentronic.it/htdocs/engine/Shopware/Controllers/Widgets/Captcha.php, line 59' in

With this fix we set the Surrogate-Capability to false.

### DIFF
--- a/engine/Shopware/Controllers/Widgets/Captcha.php
+++ b/engine/Shopware/Controllers/Widgets/Captcha.php
@@ -35,6 +35,7 @@ class Shopware_Controllers_Widgets_Captcha extends Enlight_Controller_Action
     public function preDispatch()
     {
         $this->Front()->Plugins()->ViewRenderer()->setNoRender();
+        $this->Request()->setHeader('Surrogate-Capability', false);
     }
 
     /**


bcremer:
Thanks @teiling88, but your PR does not fix the cause of the error.

Can you confirm that this is fixed with commit https://github.com/shopware/shopware/commit/b509116c4101127cdee7547582a2c616ef3ba29e?

bestit-teiling:
Yes this commit works too :-) :+1:

teiling88:
Yes this commits works fine :-) :+1:

PR #244, opened 2015-03-12
Make order billing and shipping addresses searchable

andrewisplinghoff:
This makes it possible to search for orders with a specific address (street, city, country etc.) in the order dialog.


hlohaus:
Can you improve the performance before that ?
Because your change would make them even worse.
1. Remove:
   ->leftJoin('orders.details', 'details')
   and his selects.
2. in /Shopware/Controllers/Backend/Order.php

Replace: 

``` php
            $paginator = $this->getModelManager()->createPaginator($query);

            //returns the total count of the query
            $total = $paginator->count();

            //returns the customer data
            $orders = $paginator->getIterator()->getArrayCopy();
```

With:

``` php
            $total = Shopware()->Models()->getQueryCount($query);
            $orders = $query->getArrayResult();
```

OliverSkroblin:
Hi,
this changes decreases the performance immense. We will optimize this query in 5.3 and extend the search fields to all addresses of the order.

PR #250, opened 2015-03-31
Fix transparent search-icon

t2oh4e:
I think that this part is not necessary because the button should behave like the other buttons just next to it.
That makes this .btn:hover entry obsolete. Additionally this causes the search-button to become transparent when it's hovered or active.

Usually one doesn't notice this because the standard background behind the buttons is also white, but if you change your header to a diffrent color you can see it. Actually not that important but it may save some bytes in the CSS files as long there is no good reason to keep this entry. I didn't find one.

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/less/_modules/header.less
+++ b/themes/Frontend/Responsive/frontend/_public/src/less/_modules/header.less
@@ -135,12 +135,6 @@ More precise designations are commented inside the document.
     }
 
     .entry--search {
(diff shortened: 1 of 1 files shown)


t2oh4e:
Actually you also can reproduce this by changing the color for @btn-default-hover-bg in the theme manager, so a bit more relevant than I thought while creating this fix.

bcremer:
Thank you @t2oh4e, this will be included in the Shopware 5 RC2 release.

PR #254, opened 2015-04-03
SW-10910 - compile/compress theme LESS / CSS after plugin LESS / CSS

SvenHerrmann:
LESS- and CSS-files of your own theme should be compiled and compressed after LESS- and CSS-files of installed plugins have been processed.

So it is possible to overwrite LESS-/CSS-definitions of plugins in your own theme.

see also:
http://forum.shopware.com/themes-und-design-f101/less-files-von-plugins-uberschreiben-erweitern-t26259.html

### DIFF
--- a/engine/Shopware/Components/Theme/Compiler.php
+++ b/engine/Shopware/Components/Theme/Compiler.php
@@ -166,13 +166,15 @@ class Compiler
 
         $this->buildConfig($template, $shop);
 
-        $this->compileThemeLess($template, $shop);
+        $this->compileThemeLess($template, $shop, true, false);
 
         $this->compilePluginLess($template, $shop);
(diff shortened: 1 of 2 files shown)


bcremer:
Thanks @SvenHerrmann for the heads up.
We acknowledged that issue and fixed it in https://github.com/shopware/shopware/commit/c997ff9aa69a5207596b8a13d690b3ea6c868e6d.

Can you confirm this fixes the issue for you?

SvenHerrmann:
Works fine for me, thanks a lot!
(and your fix is much nicer than mine ;-)

bcremer:
Thank you very much for the feedback. We appreciate your involvement :+1:

PR #256, opened 2015-04-07
[FIX] spelling of length vs. lenght

htuscher:
Variable names where using the incorrect term 'lenght' instead of
'length'.
In one case of extJS backend it would cause a JS error, because lenght
if not a valid JS property.

Resolves SW-10940


bcremer:
Thanks @hhoechtl,

the fix in themes/Backend/ExtJs/backend/base/application/Shopware.model.Helper.js can be merged without any problems.

For the incorrect named config name we need a migration strategy.
We never alter the `_sql/install/latest.sql` directly.
Instead a new migration file inside `_sql/migrations/` is responsible for a UPDATE Statement to change the data.

PR #258, opened 2015-04-08
Added codesniffer for using the build/ruleset.xml

b3nl:
What do you guys think about including the codesniffer, to "enable" it out of the box without external setup?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,8 @@
         "behat/mink-extension": "@stable",
         "behat/mink-sahi-driver": "@stable",
         "behat/mink-goutte-driver": "@stable",
-        "sensiolabs/behat-page-object-extension": "*"
+        "sensiolabs/behat-page-object-extension": "*",
+        "squizlabs/php_codesniffer": "~2.3"
     },
     "autoload": {
         "psr-0": {


bcremer:
Thanks @b3nl for the input.

This not directly related to your Pull Request but as a discussion start I wrote down a few lines what bugs me about the dev-dependencies.

Actually I have some gripes with composers `require-dev`. 

### 1. Usable project dependencies get obfuscated

`composer install` pulls all dependencies into the vendor directory.

As a developer I have no clue if a certain library was pulled directly as a `require`, a dependency of a  `require`, a `require-dev`, or a dependency of a `require-dev`.

For example at this given moment a shopware developer could write some code that depends on `
symfony/process`. That will work fine until shopware is deployed using `composer install --no-dev`.

That is because `symfony/process` was never a regular `require`.

It was pulled as an indirect dependency of a `require-dev`. The continuous build did not fail because it pulled the `require-dev` dependencies because continuous build requires phpunit or php_codesniffer for example.

The following gist shows what dependencies are pulled with a `--no-dev` install vs a `--dev' install:
https://gist.github.com/bcremer/c2cb9bfdb2767b7a7750

### 2. Dev-Dependencies restrict Project-Dependencies

For example at this given moment we would not be able to require `"fabpot/goutte": "~2.0"` as a dependency for shopware.
This is because our `require-dev` dependency `behat/mink-goutte-driver` depends on `"behat/mink-goutte-driver": "~1.0.1"`.

So a dependency of a development/CI tool (Mink/Behat) that is fully autonomous from the core project is restricting what we can require as a dependency in the core product.

## Possible solutions

To get rid of the problems mentioned above the dev-dependencies have to be removed from the main composer.json.
1. Provide a (script|ant task|symfony commando) that downloads the phar files of projects like phpunit
2. Use a project like http://toolphar.org/. That's more or less a glorified version of 1.
3. Commit phar files directly to the repo
4. Separate composer.json for dev dependencies inside the `build/` directory

### Appendix
- https://phar.phpunit.de/phpunit.phar
- https://phar.phpunit.de/phpcpd.phar
- https://squizlabs.github.io/PHP_CodeSniffer/phpcs.phar
- http://static.phpmd.org/php/latest/phpmd.phar
- http://static.pdepend.org/php/latest/pdepend.phar
- https://github.com/mayflower/PHP_CodeBrowser/releases/download/1.1.1/phpcb-1.1.1.phar

b3nl:
OK, fine for me, but I would suggest, that you could add a comment to the composer.json. What do you think?

bcremer:
That don you mean by adding a comment? Unfortunately json does not support any form of comments.

PR #259, opened 2015-04-09
Enable environment in UpdateCommand

Tombro:
At the moment it is not possible to set an environment in the shopware UpdateCommand. Although it is "prepared", which means that Shopware\Recovery\Update\Console\Application expects an environment variable, there is always the default config.php loaded. This resolves in not beeing able to connect to the databases on different systems.

This patch replaces the hard coded loading of config.php with enviroment based loading via the ConfigLoader.

### DIFF
--- a/recovery/common/autoload.php
+++ b/recovery/common/autoload.php
@@ -9,4 +9,8 @@ $autoloader->addPsr4(
     SW_PATH . '/engine/Shopware/Components/Migrations/'
 );
 
+$autoloader->addClassMap(array(
+    'Shopware\\Components\\ConfigLoader' => SW_PATH . '/engine/Shopware/Components/ConfigLoader.php'
+));
+
 return $autoloader;
--- a/recovery/update/src/DependencyInjection/Container.php
+++ b/recovery/update/src/DependencyInjection/Container.php
@@ -25,6 +25,7 @@
 
 namespace Shopware\Recovery\Update\DependencyInjection;
 
+use Shopware\Components\ConfigLoader;
 use Shopware\Components\Migrations\Manager as MigrationManager;
 use Shopware\Recovery\Common\DependencyInjection\Container as BaseContainer;
 use Shopware\Recovery\Common\Dump;
@@ -43,7 +44,8 @@ class Container extends BaseContainer
         $me = $this;
 
         $pimple['db'] = function () use ($me) {
-            $conn = Utils::getConnection(SW_PATH);
+            $shopwareConfig = $me->get('shopware.config');
+            $conn = Utils::getConnection($shopwareConfig);
 
             return $conn;
         };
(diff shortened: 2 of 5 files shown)


bcremer:
Looks like an solid pull request to me, thanks @Tombro.
Added some minor review notes.

recovery/common/autoload.php
```diff
@@ -9,4 +9,8 @@
     SW_PATH . '/engine/Shopware/Components/Migrations/'
 );
 
+$autoloader->addClassMap([
```
bcremer: Why are you using `$autoloader->addClassMap` here instead of `$autoloader->addPsr4` that we already used for the migrations? 
I would prefer to keep that consistent.

recovery/update/src/DependencyInjection/Container.php
```diff
@@ -93,12 +94,25 @@ public function setup(\Pimple $pimple)
...
                 $me->get('slim.request'),
                 $me->get('slim.response'),
                 $me
             );
         };
+
+        $pimple['shopware.config'] = function () use ($me) {
+            $env = $me->hasParameter('environment') ? $me->getParameter('environment') : 'production';
+
+            $configLoader = new \Shopware\Components\ConfigLoader(
```
bcremer: Generally we try to keep dependencies to the shopware core as low as possible in the installer/updater. But the `ConfigLoader` should be stable enough so it's OK in my opinion.

recovery/common/autoload.php
Tombro: Because the ConfigLoader is located directly inside the components folder (/engine/Shopware/Components/ConfigLoader.php).

If I understand addPsr4 correctly i would have to set the namesace prefix to "Shopware\Components\" and the path to "SW_PATH . '/engine/Shopware/Components/", which would result in loading everything which is inside the components folder, right?

So i decided to load the ConfigLoader individually.

bcremer: Of course you're right regarding the components folder.
For this case then, I a would prefer a simple require:

```
require SW_PATH . '/engine/Shopware/Components/ConfigLoader.php'
```

Tombro: Hm I don't understand why you don't like the classloader :)

I mean it is already used for the migration classes and the libraries and i'm just extending the composer generated classmap (vendor/composer/autoload_classmap.php).

In addition this a nice way to maybe load more required classes in the future and have everything in one place.

Switching between classloader and require is ugly and somehow confusing in my opinion (and yours - speaking of consitence above) :)

Regards,
Thomas

bcremer: I do not dislike the classloader, it's just a personal preference to keep things simple.
But I don't resist on using require. If you like the autoloader approach more it's also fine for me.
Just replace the short array syntax if you target Shopware 4.\* (as indicated by pointing to the `master` branch).

Tombro:
Do i need to do a second pull request for this patch for shopware5 (branch next)? Or do you "forport" (is this the opposite of backport? :D) all the pull requests to shopware5?

Edit: I just checked next branch and saw that you did a lot in the UpdateCommand.. But still the config.php is loaded by default. I will take a look at this one, when Shopware5 is released and do another pull request then.

bcremer:
No need for a second pull request. We merge back changes to master into the next branch on a regular basis.

Can you just squash together both of your commits into one.
Than it's OK to merge.

Tombro:
Now it's squashed.

bcremer:
Thanks. It's merged with https://github.com/shopware/shopware/commit/8f9c53afa14726830d5583b3296164a7bc0244b9 in the master branch.

PR #260, opened 2015-04-10
bugfix: sConfiguratorSelection is ignoring lastStock = true

jochenmanz:
When working with the configurator i noticed an odd behaviour. The Main Article was set to lastStock = true, and the first variant had no stock left. Instead of displaying the second variant, the first variant was dispalyed instead.


bcremer:
Thank you @schmanz for your contribution.
We changed a lot regarding configurators in the upcoming Shopware 5 release. Can you verify whether this behavior still exists or not.

You can use the `next` branch https://github.com/shopware/shopware/tree/next or the RC2. Release Package (http://wiki.shopware.com/Major-Version-5.0.0-RC-2_detail_1829.html).

PR #261, opened 2015-04-16
updateSchema for Plugins

xetys:
This request is adding a new method to shopware's plugin bootstrap, making it possible to update the schema without killing the data inside the plugin tables, and also without killing the entire database (but keeping plugin tables), when taking use of SchemaTool::updateSchema


bcremer:
Thank you @scrippi. We appreciate your contribution.

Unfortunately we do not recommend the schema manager as a tool to create the database migrations.
An automatic schema migration can never be as safe as a handwritten migration, especially when it comes to migration of the actual user data .

We will publish a blog post about plugin update method best practices soon.

PR #265, opened 2015-04-23
[Experimental] Add configurable schema table name

bcremer:
This is supposed to make the shopware migration manager more useful and flexible for usage outside of the shopware core.

For a more detailed description see this blogpost  http://ecommerce-developer.de/shopware-migrationen-fuer-eigene-deployments-nutzen/ by @b3nl.

There is also a "fork" / Subtree-Split that implements this functionality using inheritance: https://github.com/b3nl/sw-migrations

@b3nl Would that Pull Request suffice the use cases you described in your blogpost?

### DIFF
--- a/engine/Shopware/Components/Migrations/Manager.php
+++ b/engine/Shopware/Components/Migrations/Manager.php
@@ -28,7 +28,7 @@ namespace Shopware\Components\Migrations;
  * Shopware migration manager
  *
  * <code>
- * $migrationManager = new Manager($conn, '/path/to/migrations');
+ * $migrationManager = new Manager($conn, '/path/to/migrations', 's_schema_version');
  * $migrationManager->run();
  * </code>
  *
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Migrations/Manager.php
```diff
@@ -49,14 +49,24 @@ class Manager
...
-    public function __construct(\PDO $connection, $migrationPath)
+    public function __construct(\PDO $connection, $migrationPath = null, $schemaTableName = 's_schema_version')
     {
-        $this->migrationPath = $migrationPath;
+        $this->migrationPath   = $migrationPath;
+        $this->connection      = $connection;
+        $this->schemaTableName = $schemaTableName;
 
-        $this->connection = $connection;
+        if (!$migrationPath) {
```
bcremer, line 67: Some nasty hack :( 
The new argument `$schemaTableName` should be optional to preserve backwards compatibility.
A optional argument (`$schemaTableName`) can't follow to an required argument (`$migrationPath`) in the method signature. So I had to change `$migrationPath` to an optional argument that will be validated an treated as required in the method body.

The correct way would be to change `$migrationPath` and `$schemaTableName`, but that would break B/C.

I tend to take the B/C break instead of keeping this hack.

b3nl:
Thx guys! This would be nice! I think there is just only one point missing --> customizing the scheme name and the path thru a cli param etc. We would still need to copy the build/ApplyDeltas.php to get custom deployments. But the manager itself became nicer ;)

engine/Shopware/Components/Migrations/Manager.php
b3nl, line 67: Or make the schema name just an optional setter injection like in my proposal.

PR #268, opened 2015-04-27
doc and typo fixes

orangecms:
I fixed a few typos in the documentation and some string constants as well as added documentation to an undocumented function. :)


orangecms:
Probably due to changes on the website, the build now fails. It tried to access `http://www.shopware.com/sites/default/files/assets/images/refsprite_dbahn.png`, which doesn't exist (anymore?).

bcremer:
@orangecms, we are aware of that, will be fixed soon ;)

PR #269, opened 2015-05-04

andrewisplinghoff:
Most of the sprites in extra icon set 02 do not work, because the `!important` annotation is missing. Without this annotation, the rules do not apply and the top-left icon of the icon set is shown instead of the desired icon.


bcremer:
@klarstil can you please confirm this issue.

PR #271, opened 2015-05-08
advanced menu: LESS variables for colors

SvenHerrmann:
I have introduced extra LESS variables to advanced menu.

Now you can customize the colors of advanced menu separately from the other colors of your theme. Just overwrite them in LESS file of your own theme.

Color defaults are the same as before, so nothing gets broken.


klarstil:
That's a good idea, thanks for the contribution. I checked the changes and they're all looking good for me.

@bcremer We can merge this pull request.

PR #272, opened 2015-05-29
Clear association filters before saving configuration

andrewisplinghoff:
This fixes the bug that price group discounts are removed when no customer group is chosen from the combobox before 'save' is clicked:

recorded

When a customer group is chosen before clicking 'save', the loading of an empty record to the form
(https://github.com/shopware/shopware/blob/77661288fafc1f962e7e12510988cbc5084d93d7/themes/Backend/ExtJs/backend/config/controller/form.js#L385) triggers a change event on the combobox and therefore clears the filters (https://github.com/shopware/shopware/blob/77661288fafc1f962e7e12510988cbc5084d93d7/themes/Backend/ExtJs/backend/config/controller/form.js#L244). When no customer group is chosen, this does not happen, though, because the combobox does not get changed. When an ExtJS store is synced, only its filtered results get saved. This leads to removal of the discounts because they get filtered out.

### DIFF
--- a/themes/Backend/ExtJs/backend/config/controller/form.js
+++ b/themes/Backend/ExtJs/backend/config/controller/form.js
@@ -384,6 +384,11 @@ Ext.define('Shopware.apps.Config.controller.Form', {
         formPanel.disable();
         formPanel.loadRecord();
 
+		record.associations.each(function(association) {
+			var store = record[association.name]();
+			store.clearFilter(true);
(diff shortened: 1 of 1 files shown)


bcremer:
That one is merged on 5.0 in this commit: https://github.com/shopware/shopware/commit/5288b2d6ea50be48999991a30297a117bc776b18

PR #275, opened 2015-06-02
Primitive typehint in Topseller component removed

jhofm:
Caused "catchable" fatal errors.


bcremer:
Makes perfect sense. Thank you @jhofm :+1:

PR #277, opened 2015-06-05
Fix for #SW-11616 - Click area on article detail page

OliverZachau:
The click area for the lightbox is based on the image width and height. When the width and height is smaller then the wrapper class itself the result is a click area which is not directly placed on top of the image but aligned to the left. Making the click on the image not getting recognized.

By changing the width of the click area to auto the area has the needed width to make sure that every click is getting recognized.

### DIFF
--- a/templates/_emotion/frontend/_resources/javascript/jquery.shopware.js
+++ b/templates/_emotion/frontend/_resources/javascript/jquery.shopware.js
@@ -4398,7 +4398,7 @@ jQuery.effects||function(a,b){function c(b){var c;return b&&b.constructor==Array
              We need the dummy background image as IE does not trap mouse events on
              transparent parts of a div.
              */
-            $mouseTrap = jWin.parent().append($.format("<div class='mousetrap' style='z-index:999;position:absolute;width:auto;height:%1px;left:%2px;top:%3px;\'></div>", sImg.outerHeight(), 0, 0)).find(':last');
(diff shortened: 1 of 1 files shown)


bcremer:
Thanks for your pull request.

Please have a look into our [Contribution Guidlines](https://github.com/shopware/shopware/blob/5.0/CONTRIBUTING.md). 

The pull request should be in English. Also please remove the imgur.com link from the commit message since the link may be no longer valid in the feature.

bcremer:
@shopware/shopware-webdev Can you please confirm the bug and test the provided fix.

klarstil:
Alright, I checked out the pull request and reviewed it. I'm sorry, but the modification doesn't work as intended. The mousetrap has `position: absolute` and no content, therefore the mousetrap would have a width of `0px`. The other issue I spotted was the arguments in the string. The placeholder for the first argument in the string should be `%0` but in the pull request the first placeholder is `%1`, which results in a broken `top` and  `left` values:

Chrome DevTools

I've marked the pull request as "declined". @OliverZachau, you still have the ability to change the pull request and fix the issues.

OliverZachau:
I fixed the errors and provided a new fix that should now work as intended.

Thank you for the feedback!

klarstil:
@OliverZachau Thanks for the quick response. The changes looking good and the code is working as intended.

@bcremer The pull requests looks good :) You can merge it

bcremer:
Merged in commit https://github.com/shopware/shopware/commit/ea40624f7bf74a3ab81b6dcdf0c36beed545fd10.

Thanks for your help.

PR #279, opened 2015-06-08
Add Filter in sGetNotes

wesolowski:
i need to extend the profuctinfo in the wishlist, 
for example Variantinformations. Therefore i need the ListProduct-Object. 
I could create it again but it would be faster if i use the same object.


hlohaus:
A wishlist-service would be nice, or?

bcremer:
@wesolowski Thanks for you contribution :+1:

PR #280, opened 2015-06-08
Add $targetAction in saveBillingAction

wesolowski:
Currently there is no query for for the targetAction in the saveBilling. 
I would like to have an other possibility to redirect except indexAction.

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Account.php
+++ b/engine/Shopware/Controllers/Frontend/Account.php
@@ -467,7 +467,12 @@ class Shopware_Controllers_Frontend_Account extends Enlight_Controller_Action
         if (!$target = $this->Request()->getParam('sTarget')) {
             $target = 'account';
         }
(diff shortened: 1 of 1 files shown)


hlohaus:
No, problem. Use the postDispatch-Event :)

bcremer:
Thank you @wesolowski for the contribution.

May you please further describe what you are trying to archive.
As @hlohaus already hinted this should be possible using the existing postDispatch event on that route: 

``` php
$this->subscribeEvent(
    'Enlight_Controller_Action_PostDispatchSecure_Frontend_Account',
    'onPostDispatchFrontendAccount'
);

public function onPostDispatchFrontendAccount(Enlight_Controller_ActionEventArgs $args)
{
    if ($args->getRequest()->getActionName() !== "saveBilling") {
        return;
    }

    // your code
}
```

bcremer:
@wesolowski the solution of @hlohaus should solve your problem.

If that's not the case feel free to reopen the issue and leave us a comment.

Thanks for your involvment.

PR #282, opened 2015-06-11
load new snippets to db after updating from upstream

SvenHerrmann:
After updating from upstream I got some error message due to missing snippets.
So I added the command for loading new snippets.


bcremer:
Merged in commit https://github.com/shopware/shopware/commit/70bb491568b943b95325fdcfdeaa3e5fdd2f94ed.

Thank you @SvenHerrmann.

PR #284, opened 2015-06-12
Refine structure of 'build-database' target in 'build.xml'

svenmuennich:
Previously the `build-database` command always built the main database, applied the update deltas and then deployed the demo data. This patch refines the structure of this command to have three dependencies:
- `build-database-foundation`
- `build-database-apply-deltas`
- `build-database-deploy-demodata`

While calling `ant build-database` has the exact same behaviour as before, it is now also possible e.g. to build the database incl. deltas, but without installing the demo data, by calling `ant build-database-foundation build-database-apply-deltas`.


bcremer:
Thank you @svenmuennich :+1:

svenmuennich:
Thank you for the quick merge!

PR #286, opened 2015-06-15
Can't unset filter in old templates

OliverZachau:
When using the old templates with the old style to filter you can't use the "Show all" Link to remove the filter already set.

The underlying problem is the introduction of the QueryAlias "f" for filtering, which is not set in the getFacetProperties function. By setting also the new short version to the correct value it is also possible to remove a filter when there is only one left.


bcremer:
Thank you @OliverZachau. It's merged and will be released in SW 5.0.2 soon.

PR #287, opened 2015-06-17
[BUGFIX] Fixed manufacturer attribute join condition

FiveDigital:
Fixed manufacturer attribute join condition in ListProductGateway to use supplierID instead of id. This bugs leads to wrong manufacturer attributes when the supplierID differs from the id in s_articles_supplier_attributes.


bcremer:
Good catch! Thanks for your help :+1:

PR #288, opened 2015-06-18
unnecessary array assignment by key

jochenmanz:
Hallo,

i am currently working on an extended search bundle for a customer. We want to display certain variants of a product side by side to normal products in the list view. Unfortunately it wasnt enough to override the QueryBuilderFactory and change the query to reflect our needs. At some point down the road the results get preprocessed. At every other location where products get preprocessed (search for convertListProductStruct) these array assignments by key are not present, so i removed it from there and our changes just worked. Please find attached my small PR.

Best regards,
Jochen

### DIFF
--- a/engine/Shopware/Core/sArticles.php
+++ b/engine/Shopware/Core/sArticles.php
@@ -2267,7 +2267,7 @@ class sArticles
             }
             $article['description_long'] = $this->sOptimizeText($article['description_long']);
 
-            $articles[$article['articleID']] = $article;
+            $articles[] = $article;
         }
 
         $pageSizes = explode("|", $this->config->get('numberArticlesToShow'));


bcremer:
Thank you @schmanz for your contribution to shopware :+1: 

We are aware of this issue. We will check what BC breaks that would implicate.

/cc @OliverSkroblin

bcremer:
@schmanz unfortunately that would introduce a BC break for anybody hooking into the following methods:
- `\Shopware_Controllers_Frontend_Listing::manufacturerAction`
- `\Shopware_Controllers_Frontend_Listing::indexAction`
- `\Shopware_Controllers_Widgets_Listing::ajaxListingAction`
- `\sArticles::sGetArticlesByCategory`

For that reason we cannot consider this PR for Shopware 5.0.*.

I created Ticket [SW11781](http://jira.shopware.de/?ticket=SW-11781) to track that issue and will do my best to get that into consideration for Shopware 5.1. Please vote for that ticket to increase the chances of an implementation and to promote the need of this change.

In the meantime I recommend to use a replace hook to replace `\sArticles::sGetArticlesByCategory` with your own implementation. Just copy the content of `\sArticles::sGetArticlesByCategory` and `\sArticles::getListing()` into your own plugin and do the necessary modifications. Please note that this might break compatibility with other plugins hooking the above mentioned methods.

For Shopware 5.1 we might change the assignment to a `ordernumber` based array key:

```
$articles[$article['ordernumber']] = $article;
```

Please leave feedback if the described procedure works for you so I can close this PR.

jochenmanz:
@bcremer thanks for the quick feedback on that matter. Please close this PR; i will fix it as you described in the meantime.

PR #290, opened 2015-06-23
[FEATURE] add filter events to collect javascript and less methods

Tombro:
At the moment is it not possible to remove javascript or less/css files which are added by plugins or even the core. These filter events solve this lack. In addition to removing files you can also reorder files with this events. This is pretty useful for javascript files.

### DIFF
--- a/engine/Shopware/Components/Theme/Compiler.php
+++ b/engine/Shopware/Components/Theme/Compiler.php
@@ -321,6 +321,12 @@ class Compiler
             $this->collectInheritanceCss($inheritances['custom'])
         );
 
+        $definitions = $this->eventManager->filter(
+            'Theme_Compiler_Collect_Less_Definitions_FilterResult',
+            $definitions,
+            array('subject' => $this)
+        );
+
         return $definitions;
     }
 
@@ -375,6 +381,12 @@ class Compiler
         $files = array_merge(
             $files,
             $this->collectInheritanceJavascript($inheritances['custom'])
+        );
+
+        $files = $this->eventManager->filter(
+            'Theme_Compiler_Collect_Javascript_Files_FilterResult',
+            $files,
+            array('subject' => $this)
         );
 
         return $files;


bcremer:
Hi @Tombro, thanks for you contribution.

Can you elaborate your usecase a bit further.

This new event would expose some internal details that might change in the feature.

As of Shopware 5.0.2 we will also provide a way to modify jquery plugins, please see here: https://developers.shopware.com/designers-guide/modify-jquery-plugins/

Tombro:
Hey @bcremer,
we are using the plugin swag_advanced_cart which adds this javascript file:

```
"engine/Shopware/Plugins/Community/Frontend/SwagAdvancedCart/Views/responsive/frontend/_public/src/js/jquery.autocomplete.js"
```

This is used on the wishlist for adding articles.
Unfortunately the name "autocomplete" conflicts with jquery ui autocomplete, which we are using as well. 
So i wanted to get rid of this plugin.

After this there is still the wrong call in the file:

```
"engine/Shopware/Plugins/Community/Frontend/SwagAdvancedCart/Views/responsive/frontend/_public/src/js/swag_advanced_cart.js"
```

```
        /**
         * Plugin constructor
         */
        init: function () {
            var me = this;

            me.applyDataAttributes();
            me.registerEvents();
            me.initSlider();

            // Account Article Autocomplete
            $(".add-article--text-field").autocomplete(jsUrlObject.search, {
                selectFirst: true,
                remoteDataType: 'json',
                processData: function(data) {
                    var i, processed = [];
                    for (i=0; i < data.length; i++) {
                        processed.push(data[i][1] + ' (' + data[i][0] + ')');
                    }
                    return processed;
                }
            });
        }
```

So i came up with my own solution for extending JS plugins, where i was going to do a second pull request. Obviously I dont need to do that anymore, because you are working on such a solution (very nice btw, we need that! when is 5.0.2 going to be released?)

This was my solution:

```
    $.plugin = function (name, plugin) {
        if ($['plugin_' + name] != undefined) {
            plugin = $.extend($['plugin_' + name], plugin);
        }
        $['plugin_' + name] = plugin;

        $.fn[name] = function (options) {
            return this.each(function () {
                var element = this,
                    pluginData = $.data(element, 'plugin_' + name);

                if (!pluginData) {
                    if (typeof plugin === 'function') {
                        pluginData = new plugin();
                    } else {
                        var Plugin = function () {
                            PluginBase.call(this, name, element, options);
                        };

                        Plugin.prototype = $.extend(Object.create(PluginBase.prototype), { constructor: Plugin }, plugin);

                        pluginData = new Plugin();
                    }

                    $.data(element, 'plugin_' + name, pluginData);
                }
            });
        };
    };
```

Its just the first 4 lines. I am saving the plugin objects globally in jQuery and extend it if defined. (Probably not the most beautiful solution)

So back to topic:
As you can see, an ability to extend SW js plugins ($.plugin) does not solve my problem completly, as I want to get rid of 3rd party libraries.

bcremer:
Thank you @Tombro for your very detailed response :+1: 
- Glad you like the upcoming jquery plugin extensions - 5.0.2 RC will be released soon
-  The duplicated name `autocomplete` hints to the bigger underlying problem: missing vendor namespace for (shopware-core-)jquery plugins. Don't know if that can be fixed (/cc @shopware/shopware-webdev)
- Nevertheless your usecase is valid - we have to find a solution for that

We might need a few days to think about a solution for that. We might use a filter-event as you proposed.

Tombro:
Hey @bcremer,
thanks for your reply.
I dont think you can do anything about this specific autocomplete name conflict, too. This is a jquery issue. For shopware plugins you could establish a convention to prefix them with a vendor. Like the core plugin imageSlider would become swagImageSlider or something.

Another approach to solve the lack of removing javascripts could be a second array in the Theme.php like it is now with $javascript, but only in reverse like $removeJavascript. But this is not as flexible as the events because you have to have a custom theme and you cant manipulate the ordering of the scripts.

Would be really nice if you could find a solution for this in 5.0.2 :)

Thanks, Thomas

bcremer:
Hey @Tombro.

We have two tickets in the current sprint covering this problem.
- In ticket [SW-11821](http://jira.shopware.de/?ticket=SW-11821) we will vendor prefix the jquery plugins. During the 5.0.\* time there will be an alias to the old name. That alias will be removed in SW 5.1.0
- In ticket [SW-11750](http://jira.shopware.de/?ticket=SW-11750) we will refactor the StateManager Plugin Initialization so it should be possible to remove jquery plugins before they are getting executed

The new event is sill under consideration so I will leave this pull request open for the time being.

bcremer:
Shopware 5.0.2:
- jQuery plugins prefixed, aliases added https://github.com/shopware/shopware/commit/bcd105d45104e1ecbacc2194cd9ae790f97fbea0
- jQuery Plugin/Statemanager refactored https://github.com/shopware/shopware/commit/470b5f9587e365ffab3e3e61bd98c788ad533a1d

Shopware 5.1:
- jQuery deprecated aliases removed https://github.com/shopware/shopware/commit/c9ece3a5902ab88fc9e1efbc9249d7ad2b601dfa

@Tombro 
Is the new event stil relevant for you or will the above mentioned commits solve your problems?

bcremer:
I am closing this for now. If this did not solve your problem feel free to reopen. Thanks.

Tombro:
Hey @bcremer,
sorry i was on vacation.
I will check if these events could still be necessary and reopen this request in that case.

Thanks!

Tombro:
Hey folks,
i am reopening this one because i ran into another case in 5.1.4 which make these events necessary.

SwagAdvancedCart includes the JS lib jquery-ui.min.js but only some components of the lib. I need some other components. 

Obviously StateManager.removePlugin() doesnt help here, because its no shopware-jquery-plugin.

Sure I could just include the new ui-widgets in my plugin and use the core of advancedcart, but this will crash if advancedcart will be uninstalled.

I still think these two events are really helpful to enable full control of JS-Files.

Tombro:
Hmm, i am just realizing, that its impossible to reopen a PR?
So I would make a new one (with new branches as well) if you consider merging it.
Please let me know

bcremer:
Hi @Tombro, sorry for my late response.
Please go ahead an open a updated pull request regarding this issue.

PR #291, opened 2015-06-23
Fix subshop specific order count

andrewisplinghoff:
There was an error in the SQL query so that for every subshop either the total order number for all subshops or 0 was returned, but not the correct order count for the subshop.

To reproduce there need to be multiple subshops and each needs to have orders so that the bug becomes apparent.


bcremer:
Thank you @andrewisplinghoff. PR looks good, I will put it into into QA tomorrow :+1:

bcremer:
Thanks @andrewisplinghoff.

PR #293, opened 2015-06-24
Fix ordernumber length in s_order_basket

aragon999:
ordernumber in s_articles_detail is varchar(40) so when trying to add an article with ordernumber longer than 30 chars it is not added to the basket


bcremer:
Hi @aragon999,

we can't accept this Pull Request for the following reasons:
- Schema changes have to be done using migration files (See: https://github.com/shopware/shopware/blob/5.0/_sql/migrations/101-add-extended-editor-field.php as an example)
- The `ordernumber` is stored in multiple tables that also have to be updated. For example `s_order_basket`

Thank you for your involvement. Please open a new pull request once the above mentioned problems are fixed.

Greetings.

PR #294, opened 2015-06-24
change uses of $value in getBasketArticles function to $getArticles

teiling88:
with the event "Shopware_Modules_Basket_GetBasket_FilterItemStart" i can't change the modus for a specific entry of the array $getArticles.

In our case, we use this event to prevent a voucher update.

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2022,7 +2022,7 @@ class sBasket
         $totalAmountNet = 0;
         $totalCount = 0;
 
-        foreach ($getArticles as $key => $value) {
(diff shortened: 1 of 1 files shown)


dnoegel:
Looks good; the only thing one might discuss is the unused `$value` variable that actually should not be used for extensibility reasons (as this PR shows). So an `foreach(array_keys($getArticles) as $key) { }` or something like that might also do the trick.

teiling88:
thanks for your feedback - i will fix it

teiling88:
the fixed version as promised :-)

bcremer:
@teiling88 Thank you for getting involved  :+1:

PR #295, opened 2015-06-24
fixing inconsistent ordernumber schema

teiling88:
Based on this pull request: https://github.com/shopware/shopware/pull/293

I hope i found every possible article ordernumber in the db schema :-)

### DIFF
--- a/_sql/migrations/504-fixing-ordernumber-schema.php
+++ b/_sql/migrations/504-fixing-ordernumber-schema.php
@@ -1,0 +1,50 @@
+<?php
+
+class Migrations_Migration504 Extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        // fixing ordernumber for s_addon_premiums
+        $sql = <<<'EOD'
+            ALTER TABLE `s_addon_premiums`
+            MODIFY COLUMN `ordernumber` varchar(40),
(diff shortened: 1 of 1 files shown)


dnoegel:
Thanks @teiling88, looks good for me. Question is, if we shouldn't make it VARCHAR(255) anyway @bcremer? I guess there is no technical reason for the limitation - except the fact, that its a bad index - but that also applies for VARCHAR(40).  What do you think?

bcremer:
I also opt for `VARCHAR(255)`.

But if we are going to change the field we should do some performance research first. Because we are using this fields often times as an index there might be a benefit in changing the type into `CHAR`. See:   http://dba.stackexchange.com/a/1915.

ahoehne:
@bremer habt ihr schon char/varchar Tests gefahren? Wäre da echt an einem Ergebnis interessiert, gerade in Bezug auf InnoDB...
Caching dürfte nach der Logik ja eher suboptimal sein (mehr Speicheverbrauch = weniger Platz zum cachen), dass hat interessanterweise noch keiner in dem Thread angesprochen.

teiling88:
thanks @bcremer

bcremer:
This was fixed with commit https://github.com/shopware/shopware/commit/ac3ddaa6208.

Thanks @teiling88

teiling88:
\o/
Thanks!

PR #297, opened 2015-06-25
prevent infinite recursion for invalid group ids

jhofm:
the patch fixes an inifinite recursion when calling a detail page with invalid group parameters, which floods the database until the script timeout kicks in. the affected core component is removed in 5.0.

### DIFF
--- a/engine/core/class/sConfigurator.php
+++ b/engine/core/class/sConfigurator.php
@@ -243,12 +243,10 @@ class sConfigurator
         if (!empty($selectedItems) && empty($selected)) {
             if ($settings['type'] == self::TYPE_STANDARD) {
                 unset($this->sSYSTEM->_POST["group"]);
-                return $this->getArticleConfigurator($id, $articleData, true);
-
             } elseif ($settings['type'] == self::TYPE_SELECTION) {
-                $group = $this->sSYSTEM->_POST["group"];
-                array_pop($group);
-                $this->sSYSTEM->_POST["group"] = $group;
+                array_pop($this->sSYSTEM->_POST["group"]);
+            }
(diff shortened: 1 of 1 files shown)


jhofm:
i should probalby clarify that the infintie recursion only happens on articles where the configurator settings diverged from the article's details. which should not happen, but totally can for lack of transactional safety in many parts of the core (@4.3). so this ist not an instant sw4.3 DOS via GET request for everyone. still annoying when it happens.

engine/core/class/sConfigurator.php
```diff
@@ -243,12 +243,10 @@ public function getArticleConfigurator($id, $articleData, $recursiveCall = false
...
                 unset($this->sSYSTEM->_POST["group"]);
-                return $this->getArticleConfigurator($id, $articleData, true);
-
             } elseif ($settings['type'] == self::TYPE_SELECTION) {
-                $group = $this->sSYSTEM->_POST["group"];
-                array_pop($group);
-                $this->sSYSTEM->_POST["group"] = $group;
+                array_pop($this->sSYSTEM->_POST["group"]);
+            }
+            if(sizeof($this->sSYSTEM->_POST["group"])) {
```
bcremer: Can you please use `count` instead of `sizeof` here. `sizeof` is just an alias to `count` which is far more common.
Also please add a space between `if` and the opening parenthesis: `if (`.

bcremer:
@jhofm The fix looks good, thank you for your contribution.
Can you please fix the minor coding style issues and we are good to merge.

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/ed5fa7c35bb77b47a463f060cabb864bc5b38126.

Thank you @jhofm :+1:

PR #298, opened 2015-06-25
Update gitignore with excludes for .gitkeeps and .htaccess

b3nl:
Without these excludes, your .gitkeeps and .htaccess are dropped out of remote repos, so i excluded them.


bcremer:
Thank you @b3nl :+1:

PR #299, opened 2015-07-01
Patch 1

chuscazo:
Hello,
I think there are a bug in compile. If Bare Theme hat none javascript, the javascript from Plugin are loaded before "jquery.min.js" and then there are jquery errors.

Example: 
- I copy theme Responsive and make the rename changes in theme.php
- Compile works , but in browser there are javascript error.
- Debug from compile say that javascripts from Pluging are first included then jquery.min.js

array(53) {
  [0]=>
  string(113) "/html/engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Views/frontend/_public/src/js/jquery.advanced-menu.js"
  [1]=>
  string(135) "/html/engine/Shopware/Plugins/Community/Frontend/SwagLiveShopping/Views/responsive/frontend/_public/src/js/jquery.swag_live_shopping.js"
  [2]=>
  string(76) "/html/themes/Frontend/Minol/frontend/_public/vendors/js/jquery/jquery.min.js"
.....
.....

thanks.
David Lopez


stefan-b314:
Hi,

i can confirm this behavior / bug.

I also have a theme that inherits from Bare instead of Responsive and when enabling the "Advanced Menu" plugin, there will be jQuery errors.

The problem is indeed the order of includes. Bare has no JS files, then all plugins' JS files will be included and after that the JS files from the custom theme are included as you can see in [engine/Shopware/Components/Theme/Compiler.php#L368](https://github.com/shopware/shopware/blob/5.0/engine/Shopware/Components/Theme/Compiler.php#L368) and following.

The reason this works for the Responsive theme is that someone has been so clever to define the "bare" property to either a theme that has no parent or an instance of ResponsiveTheme in [engine/Shopware/Components/Theme/Inheritance.php#L109](https://github.com/shopware/shopware/blob/5.0/engine/Shopware/Components/Theme/Inheritance.php#L109). So when you have enabled the Responsive Theme, the $files variable in [engine/Shopware/Components/Theme/Compiler.php#L368](https://github.com/shopware/shopware/blob/5.0/engine/Shopware/Components/Theme/Compiler.php#L368) will already include the JavaScripts of the Responsive theme, then it will add the scripts of the plugins and after that merge again the file list of responsive theme.

Hope this helps evaluating this bug and fix it soon!

Kind Regards,
Stefan

stefan-b314:
I've created an issue in JIRA for this: http://jira.shopware.de/?ticket=SW-12272

klarstil:
@stefanullinger I closed the issue cause it's a duplicate with the following ticket: http://jira.shopware.de/?ticket=SW-12782

bcremer:
You can find the fix for this issue in the following commit: https://github.com/shopware/shopware/commit/a689775b5646062db4e3b8f4d392982215e81fe2

chuscazo:
The fix still in version 5.1.4
in Theory was solved in version 5.1.2 
or im wrong?
Where is this new configuration property?

PR #303, opened 2015-07-09
show newsletter sent-date instead of create-date in newsletter archiv…

SvenHerrmann:
…e listing

imho it makes more sense to show the date when the newsletter was sent instead of when you started to work on it


janbuecker:
Thanks for your contribution @SvenHerrmann.

We decided to close this PR because of time constraints and limited overall benefit to the product. If this change is still important to you, please create a new issue on https://issues.shopware.com so people can upvote this feature.

PR #304, opened 2015-07-10
Smarty plugin {link} does not work if Plugins/Community is a symlink

SvenHerrmann:
We deploy our shops with Capistrano and use the Plugins/Community-folder as shared folder, so it is a symlink. 
The Smarty plugin {link} cannot handle this symlink, so that e.g. the PayPal logos are not shown.

In the patch we check if the real path of the Community folder differs from the expected path and rewrite it if necessary.

### DIFF
--- a/engine/Library/Enlight/Template/Plugins/function.flink.php
+++ b/engine/Library/Enlight/Template/Plugins/function.flink.php
@@ -63,6 +63,15 @@ function smarty_function_flink($params, $template)
             $docPath = getcwd() . DIRECTORY_SEPARATOR;
         }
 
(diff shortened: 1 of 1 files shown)


arvatis:
👍

FiveDigital:
Shouldn't this be used for Local plugins, too? We manage our Local plugins via composer but I can imagine that somebodys symlinks them, too.

bcremer:
Hi @SvenHerrmann, thanks for your contribution. We will close this PR as it is incomplete. 
Please have a look into https://github.com/shopware/shopware/pull/829. 
Also please open a ticket in the issue tracker so people can vote for this feature.

PR #305, opened 2015-07-11
typo in default value for style-parameter of transform-style mixin

SvenHerrmann:
There was a typo in default value for style-parameter of transform-style mixin.

The mixin had not been used without parameter, so this has no impact on existing code.


bcremer:
Thank you @SvenHerrmann. This will be merged right after SW 5.0.2 is released.

PR #311, opened 2015-07-16
BugFix: Support readOnly-Setting in TinyMCE form field

mirkobeine:
The editor configuration of tinyMCE expects the setting "readonly" to be in lower case, otherwise it is ignored.


bcremer:
Cherry Picked to the 5.0 branch: https://github.com/shopware/shopware/commit/a81d8fc73a337b3a890e43bb2fe441a0f768cf29

Thank you @mirkobeine :+1:

PR #312, opened 2015-07-18
Added createMediaField to Emotion Component

shyim:
Adds a method to the emotion component to add media selection fields


bcremer:
Thank you @Shyim!

PR #313, opened 2015-07-21
Support for batch user deleting on Backend UserManager

shyim:
This commit adds the ability to batch user deleting on Backend UserManager


bcremer:
Thanks @Shyim :+1:

PR #314, opened 2015-07-23
5.0

SDJeff:
At the engine/Shopware/Models/Order/Repository.php  protected function getOrdersQuery() your given expressions aren't accepted at the filter params.

At Line 501 of the protected function filterListQuery()

it should be sth. like:

`$builder->andWhere($filter['property'] . " " . $filter['expression'] . "'" . $filter['value'] . "'");`

instead of :

`$builder->andWhere($expr->eq($filter['property'], $filter['value']));`

Otherwise everthing which runs throw the switch statement is by default interpreted as `$expr->eq`

But this isn't what you expected if you fill the filter params array as documented with 'expression' => '>' or  others.


bcremer:
Hi @SDJeff, your pull request looks messed up. Can you please fix that.

SDJeff:
Any hint?

bcremer:
I am not sure what you intention is.
If you have a commit with a bugfix that we can merge, please open a new Pull Request as described here: https://help.github.com/articles/creating-a-pull-request/

For reporting bugs please use our bug tracking system that can be found here: http://jira.shopware.de/

PR #317, opened 2015-08-05
Add missing use instruction for ConversionException

andrewisplinghoff:
The ConversionException is used here: https://github.com/shopware/shopware/blob/ed505da037607918f31d756dd18bccc07539791f/engine/Shopware/Components/Model/DBAL/Types/DateStringType.php#L72

Without the use instruction, the following error comes up when an invalid string is encountered:
`Fatal error: Class 'Shopware\Components\Model\DBAL\Types\ConversionException' not found in    /home/andre/viison/EisenschmidtShop/engine/Shopware/Components/Model/DBAL/Types/DateStringType.php on line 72`


bcremer:
Thanks @andrewisplinghoff

PR #318, opened 2015-08-07
Breaks if there is a category without parent

solidevolution:
I got an error while importing parentless categories from xtc with SwagMigration, fixed it with !empty check


bcremer:
A category in shopware can not be without a parent. A exception is the root category with id `1`.
This fix might allow the import, but this will break other things in shopware that rely on a parent id.

The fix should be made directly in https://github.com/shopwareLabs/SwagMigration. So no categories without parents should be imported.

janbuecker:
Thanks @solidevolution 

We decided to close this PR because we cannot reproduce this issue. If this change is still important to you, feel free to create a new pull request or create an issue on https://issues.shopware.com.

PR #319, opened 2015-08-10
Hackathon search optimization

mitelg:
in our hackathon project @MarcelSchmaeing and I implemented an option to connect search terms with AND logic. so the more words you type in, the more precise the resutl should be


mitelg:
ups sorry, I think we did our changes on base of the 5.0 branch. will fix that...

PR #321, opened 2015-08-11
[TASK] prevent empty property sets from causing non object errors

Tombro:
Hey folks,
if you assign a property set to an article in the backend but do not put any values to the properties the whole frontend listings crash. The reason is this error:

```
Fatal error: Call to a member function getGroups() on a non-object in /engine/Shopware/Components/Compatibility/LegacyStructConverter.php on line 594
503 Service Unavailable
```

I know it does not make sense to assign empty properties, but this happened to our customer editors two times now.

This simple instanceof condition solves that issue.

Regards,
Thomas Christiansen

### DIFF
--- a/engine/Shopware/Components/Compatibility/LegacyStructConverter.php
+++ b/engine/Shopware/Components/Compatibility/LegacyStructConverter.php
@@ -591,23 +591,26 @@ class LegacyStructConverter
     public function convertPropertySetStruct(StoreFrontBundle\Struct\Property\Set $set)
     {
         $result = [];
-        foreach ($set->getGroups() as $group) {
-            $values = array_map(
-                function (StoreFrontBundle\Struct\Property\Option $option) {
-                    return $option->getName();
-                },
-                $group->getOptions()
-            );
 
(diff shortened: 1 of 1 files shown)


hlohaus:
thanks, my ticket: http://jira.shopware.de/?ticket=SW-11983 :)

Tombro:
Ah I didnt find any ticket to this issue, but good to know :)

bcremer:
Hey @Tombro, thanks for the Pull Request.

Your fix gave a hint in the right direction. The root cause is `sArticles::sGetArticleProperties()`.

https://github.com/shopware/shopware/blob/5.0/engine/Shopware/Core/sArticles.php#L326

``` patch
+        if (!$set) {
+            return [];
+        }
```

I will push the fix after internal code review and QA.

bcremer:
Fixed with commit: https://github.com/shopware/shopware/commit/284f35422344b50d687e6123325bcd385c45ed75

@Tombro Please feel free to open another Pull Request if that did not fix your issue.

Thanks for your contribution :+1:

PR #323, opened 2015-08-12
Remove database references to non-existing article with id 133

andrewisplinghoff:
The current demo data contains inconsistent data that references the non-existing article with id 133. This removes the affected detail rows as well as prices, images and configuration relations.


bcremer:
Awesome, thanks @andrewisplinghoff :+1:

PR #324, opened 2015-08-14
Remove database references to non-existing article with id 138

andrewisplinghoff:
I found some more inconsistencies in the demo data (compare #323). There is also no article with the id 138, so we remove references to this article. This removes the affected detail rows as well as prices, images and configuration relations.


bcremer:
Another easy mergeable pull request - thanks again @andrewisplinghoff.

PR #325, opened 2015-08-19
Product Detail Page: Image check fails

jhit:
Even if no image is set for the product, the $sArticle.image variable
will be initialized with a empty array. The check always returns true
and the image zoom will try to zoom a empty image.
Better check if the $sArticle.image.id variable is set.

### DIFF
--- a/themes/Frontend/Bare/frontend/detail/index.tpl
+++ b/themes/Frontend/Bare/frontend/detail/index.tpl
@@ -96,8 +96,8 @@
         <div class="product--detail-upper block-group">
             {* Product image *}
             {block name='frontend_detail_index_image_container'}
-                <div class="product--image-container image-slider{if $sArticle.image && {config name=sUSEZOOMPLUS}} product--image-zoom{/if}"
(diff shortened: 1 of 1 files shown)


bcremer:
Hi @jhit, thanks for your contribution.

Unfortunately I am not able to reproduce this error:
1.  If no image is set for the product, `$sArticle.image` is `null` instead of an empty array
2. Even with an hardcoded empty array in `$sArticle.image` the condition will cast the empty array to `false` and the zoom is disabled.

Can you please verify the contents of `$sArticle.image` in your template using `{$sArticle.image|var_dump}`. Maybe a broken plugin is putting some garbage values in there.

jhit:
Hi Benjamin. I did this var_dump with this result:

```
 array(2) {
    ["description"]=> NULL
    ["relations"]=> NULL
}
```

Plugins-Installed: 
image

Seems the Büroware Connector  might cause the issue. 
Sorry for bothering you.

bcremer:
No worries. I am glad you found the cause.

PR #326, opened 2015-08-20
regarding customer group discounts and tax rules for product exports

jemmel:
the product export does not regard any configured customer group discounts and/or tax rules.
this leads to some wrong product information (especially the gross price calculation).


janbuecker:
Thanks @jemmel 

PR #330, opened 2015-08-24
add double quotes after title tag

ottscho:
Wenn mir jemand sagt wie ich statt ottscho meinen realen Namen rein bekomme, mache ich das gerne
@bcremer


bcremer:
Thank you @ottscho, this PR looks fine.
But can you please redo the commit using your real name as git author. This makes it easier for us to get into contact with contributors later on.

Also please use `5.0` as a target branch. We will merge fixes for the `5.0` branch up into the `5.1` branch on a regular basis.

t2oh4e:
I think this should be merged as soon as possible as it destroys the whole cart when there is an item without image.

bcremer:
Thank you @ottscho, the fix is included in commit https://github.com/shopware/shopware/commit/02cab1236df99de43204baed474f41b46c25b409.

PR #302, opened 2015-07-09
Adding media association to sproperties

t2oh4e:
This will add a new entry media in the sPropeties Array. Useful if you want to display the images associated to the properties in the frontend.

PR #331, opened 2015-08-25
Adding media association to sproperties

t2oh4e:
This will add a new entry media in the sPropeties Array. Useful if you want to display the images associated to the properties in the frontend.

Same change as in #302 which I messed up.


bcremer:
Thanks @t2oh4e.

PR #332, opened 2015-08-25
add double quotes after title tag

ottscho:
add double quotes after title tag


bcremer:
Duplicates https://github.com/shopware/shopware/pull/330.

PR #333, opened 2015-08-28
Not able to generate multishop css with grunt

solidevolution:
Currently it's not possible to do a

grunt --shopId=2

This will be overwritten by the grunt.option.init, so grunt always creates the css for the first shop

Occurs with

grunt-cli v0.1.13
grunt v0.4.5

### DIFF
--- a/themes/Gruntfile.js
+++ b/themes/Gruntfile.js
@@ -1,7 +1,9 @@
 module.exports = function (grunt) {
-    grunt.option.init({
-        shopId: 1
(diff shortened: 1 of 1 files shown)


bcremer:
This should be fixed with commit: https://github.com/shopware/shopware/commit/43a8ddcd6e9219797fc238ad0122d4f3dc80c2ce

If the problem still exists feel free to reopen the PR.
Thanks.

PR #334, opened 2015-08-28
don't set the article cache invalidate by postUpdate or postPersists …

resslinger:
…on variants

If an item with 120 variants is updated, the cache of the item will be set invalidate over 120 times.
An article update requires considerable time and performance.

We test it with the Plentymarkets Plugin. In SW 4.3.6, before the variants were included in httpcache, an import of an article with 120 variants was finished in 20 seconds. And now with SW 5.0.3 a import is finished with the same article in over 2-15 minutes.


dnoegel:
I think there is an issue with the API and invalidation anyway - if you update multiple products (batch request), the article invalidation might still be quite slow.
I'd rather collect all invalidate "requests" of a given request until the end of the request and then invalidate all channged products alltogether. 
So in this case, when updating multiple variants or using the batch resource, Shopware should not invalidate until the request is processed. That might increase the performance massively.

janbuecker:
Thanks @k-30 

PR #335, opened 2015-08-31
fix fatal error on OSX System update process

wrongspot:
without this fix, every update process stops with a fatal error on osx filesystems


janbuecker:
Hi @wrongspot, thanks for your PR.

We cannot reproduce your issue and we should not implement this special use case.

If this is still an issue, feel free to open a new PR but with a better list of files to skip.

PR #336, opened 2015-09-02
AdvancedMenu new feature: Menu open delay on hover.

oskude:
Add option hoverDelay to set delay in milliseconds before opening the frontend menu when mouse cursor hovers over it.

The open menu got constantly in my way during development, specifically when the menu is tall and browser inspector is open at the bottom of the window. And as a client i would find it less annoying for the menu not opening when i just want to move the mouse cursor over/past it.


klarstil:
Hey @oskude,

thanks for your PR. The feature makes totally sense for me.

@bcremer What you think about it? Maybe it's something for SW 5.1?

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/08e7f4525a26808d3f9160da4a63cf1db03eaf1f.

Thank you @oskude.

PR #341, opened 2015-09-05
Add Grunt Notify

ChristopherDosin:

PR #342, opened 2015-09-05
Added notification task

ChristopherDosin:
Show a notification if the less compile task is completed

### DIFF
--- a/themes/Gruntfile.js
+++ b/themes/Gruntfile.js
@@ -32,6 +32,14 @@ module.exports = function (grunt) {
     }
 
     grunt.initConfig({
+        notify: {
+          watch: {
+            options: {
+              title: 'Task complete',
+               message: 'Less files were compiled'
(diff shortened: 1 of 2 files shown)


bcremer:
Thanks @kayyyy for your contribution.
Please include PR #341 in this one so it's one PR we can merge.

@shopware/shopware-webdev can you please review? I am also not sure about adding more dependencies. Any opinion on this?

ChristopherDosin:
@bcremer done

bcremer:
Thank you @kayyyy for your work on this PR.

Unfortunately we decided to not include the notify dependency.
The is due to the fact that we want to keep the requirements low so the script can also be used on production servers etc. 
Also the notifications might be disruptive for some users.

Instead we will think of an extension solution, so the customer can extend the grunt config with own tasks or custom overrides to customize it for their own needs.

PR #345, opened 2015-09-11
typos in method name and phpdoc

SvenHerrmann:


bcremer:
`assertVersionGreaterThen()` is already deprecated in favor of `assertMinimumVersion()`.
We did not alone to fix the than/then-typo, but also to be more semantic correct. 

`assertVersionGreaterThen()` should have been named `assertVersionEqualsOrGreaterThan()` from the beginning.

I cannot see a need for another alias to the new `assertMinimumVersion()`.

Any reasons why the new method should be introduced?

tiagojsag:
We should, nonetheless, fix the typo in the `deprecated` tag, as to make the deprecation more visible (ie. PHPStorm automatically highlights deprecated methods, which is not currently true for this method due to the typo). 

As for the wrapper method, I agree with @bcremer that we should not introduce it. The typo problem will be fixed once we remove the deprecated method.

SvenHerrmann:
Ok, you're right. But at least the typo in phpdoc should be fixed, so one can see in PhpStorm, that the method is deprecated.

bcremer:
Thank you @SvenHerrmann. I included your commit into the 5.1 branch.

PR #346, opened 2015-09-11
Fix for wrong error message

uehler:
When the compiler checks, if the JS files exist, the error message says there is no CSS file instead of JS files


bcremer:
Thank you @uehler, your fix is included in the https://github.com/shopware/shopware/tree/5.1 branch.

PR #347, opened 2015-09-14
smarty block correction

ottscho:


bcremer:
Hi @ottscho,

we fixed that bug in commit https://github.com/shopware/shopware/commit/16733edcac958ca617c3edc45861fe512aee31eb.
This bugfix will be released with in Shopware 5.0.4 which is currently available as a [Release Candidate ](https://github.com/shopware/shopware/releases/tag/v5.0.4-RC1).

The block `frontend_index_header_title` is already defined later in the same file so it cannot be simply renamed.

Thanks for your contribution. If you think the commit provides not the best way to fix this issue feel free to reopen this PR.

PR #348, opened 2015-09-19
Updated Codemirror and improved modes support

shyim:
Example Plugin to test all modes: https://github.com/shyim/sw-codemirror-test


shyim:
It would be nice, when the bug is fixed in 5.2 :(

PR #350, opened 2015-09-28
reversed js-file order because setting inheritanceConfig to false cre…

benboit:
At the moment there seems to be a problem with template-inheritance and the console-command sw:theme:dump:configuration. When i have a theme and i dont want to have inheritanceConfig enabled the order of js-files in the dumped config is reversed.

This means, that plugin js-files are concatted first, which leads to a failing dependency with jquery. Probably im doing it wrong, but this looks like a bug to me.

### DIFF
--- a/engine/Shopware/Components/Theme/Compiler.php
+++ b/engine/Shopware/Components/Theme/Compiler.php
@@ -369,14 +369,14 @@ class Compiler
 
         $files = array_merge(
             $files,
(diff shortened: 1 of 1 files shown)


bcremer:
@benboit, I closed this PR because we are tracking this issue as ticket http://jira.shopware.de/?ticket=SW-12782 in our bug-tracker.
A bugfix is already in QA and should be merged to the 5.1 branch soon.

Thanks for your help.

PR #351, opened 2015-09-28
removed default shop-id from gruntfile which overrides external argum…

benboit:
This option shouldn't be set there. Developers should run Grunt like this:

grunt production --shopId=2

Currently every shop-id parameter is overriden by these lines.

### DIFF
--- a/themes/Gruntfile.js
+++ b/themes/Gruntfile.js
@@ -1,8 +1,4 @@
 module.exports = function (grunt) {
-    grunt.option.init({
-        shopId: 1
-    });
-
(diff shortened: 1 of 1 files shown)


bcremer:
@shopware/shopware-webdev can you please verify this?

klarstil:
Thanks for the PR, @benboit The code looks fine, I verified the behavior and the PR makes sense :) @bcremer Would you be so kind to merge the PR?

bcremer:
Thank you @benboit :+1:

PR #352, opened 2015-09-29
Add backend self healing for attribute models

svenmuennich:
This PR expands the 'self healing' to the backend to automatically fix problems with corrupted attribute models. This is useful e.g. if a plugin installation fails after changing certain attribute fields, but before regenerating the respective model files. If you end up receiving exceptions when loading backend applications, you currently have to visit the shop frontend to trigger the self healing. However, this will not necessarily trigger it, because the exception might occur only in the backend. With this PR, the same self healing logic is also executed for exceptions in backend requests.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -3,6 +3,7 @@ In this document you will find a changelog of the important changes related to t
 
 ## 5.1.0 RC3
 * Activated media fallback by default so that old media paths get resolved to the new location
+* Add backend self healing for attribute models
 
 ## 5.1.0 RC2
 * Update ongr/elasticsearch-dsl to version 1.0.0-RC1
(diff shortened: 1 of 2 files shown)


bcremer:
Hi @svenmuennich, sorry for the delayed answer. Can you evaluate if the following PR would fix your problem: https://github.com/shopware/shopware/pull/432

svenmuennich:
#432 looks good. This PR will probably be obsolete soon anyway, because it just extends the [AttributesSubscriber](https://github.com/shopware/shopware/blob/ac749bdbc042345882a6314e1cb635cc0160eab7/engine/Shopware/Components/AttributeSubscriber.php), which you mentioned in #432 as potentially obsolete.

bcremer:
Thanks for you feedback.

PR #353, opened 2015-09-29
Allow clearing caches after plugin update

svenmuennich:
Currently the response of a request to the `PluginInstaller` is evaluated in all plugin install/uninstall related actions except for an update. Hence it is neither possible to prompt the user to clear certain caches after updating a plugin, nor to display a sticky notification with a custom message. This PR fixes this behaviour by invoking the common `handleCrudResponse` handler before the passed callback function.


bcremer:
Merged with commit https://github.com/shopware/shopware/commit/978900cd866184b521045eb5867bd85cf0fcaed5.

Thank you @svenmuennich and sorry for the delay.

PR #354, opened 2015-09-29
only cache the files directory from Composer's cache

xabbuh:
See the explanation in https://github.com/puli/manager/pull/9#discussion_r40499480 on how the Composer cache works.


bcremer:
Thanks @xabbuh.

PR #357, opened 2015-10-02
Apply select box replacement after reloading shipping payment form

svenmuennich:
When selecting a different payment method during the checkout process, the payment form is reloaded synchronously and injected into the DOM. In case the now active payment method contains a select box in its form, it is not replaced with a fancy box. This PR calls the select box replacement plugin again on all respective select boxes, as soon as the new payment form is loaded.


bcremer:
@shopware/shopware-webdev can you please review this, thanks.

klarstil:
@bcremer the code looks good for me, the plugin base prevents multiple instances on the same element. I guess we haven't had a payment method with select boxes in the core, therefore we haven't seen the need to add it.

bcremer:
Good to merge then. Thanks @svenmuennich :+1:

PR #359, opened 2015-10-03
Fix newsletter unsubscribe link

kekster:
The newsletter controller will automatically preset the dropdown menu to "unsubscribe" instead of "subscribe"


Phil23:
Thank you @kekster for the PR.

We removed two commits, which had nothing to do with the actual fix. Please create seperate pull requests for each individual fix and add a description for each change.

Thank you for your work!

PR #360, opened 2015-10-05
Remove plugin manager timeouts

svenmuennich:
Currently the plugin manager uses the default timeout settings when installing/updating/uninstalling plugins. That is, the default PHP timeout (e.g. 30 seconds) for the operation itself, the Ext JS default timeout for AJAX requests (30 seconds) as well as the default timeout of 15 seconds for hiding the loading animation. When running e.g. a plugin installation/update/uninstallation with a duration of > 30 seconds, currently the following happens:
1. After 15 seconds: the loading animation disappears, although the request is still pending.
2. After 30 seconds: The AJAX requests gets cancelled, although the operation on the server is still running.
3. After PHPs timeout: The operation gets cancelled.

This behaviour is not desirable for several reasons:
1. The user thinks e.g. the plugin installation is finished, as soon as the loading mask is hidden (after max. 15 seconds). However, the plugin is still marked as _not installed_. Clicking the _install_ button again might cause problems because of two concurrent installation operations.
2. Success messages are not displayed, because the AJAX request is cancelled.
3. Cancelling the install operation itself might result in e.g. a corrupt database.

This PR fixes the described behaviour by
1. Removing the timeouts of PHP operations for the respective actions
2. Disabling the `autoTimeout` of the loading mask
3. Seting the timeout of the respective AJAX request to a reasonable long time of 5 minutes


cjuner:
Please close this PR in favor of PR #568, which is a rebase on top of the 5.2 branch.

mitelg:
Closed in favour of https://github.com/shopware/shopware/pull/568

PR #361, opened 2015-10-06
using ssl in light backend modules

benboit:
Currently there is a problem with the new light backend-modules, when using ssl in backend. The module iframe will be generated with a http-url, even though the backend is opened via https. Browsers will block this unsafe request.

I dont know if this is a proper solution, because shopware-backends with ssl not enabled will probably break.


bcremer:
This was fixed in commit https://github.com/shopware/shopware/commit/74d89327153 and will be released in Shopware 5.1.2 soon.

Sorry for not including your commit, this was fixed using a ticket from our issue tracker.

PR #362, opened 2015-10-09
SW-12761

freedmo:
I change both files, so forward slash will replaced with space and not merge the two substrings.
So searchterms, are same as search keywords in db.


bcremer:
Thank you @freedmo, the PR looks good. This will be merged into the 5.1.2 Release soon.

PR #364, opened 2015-10-13
Workaround for overlay in Firefox

aragon999:
When opening the overlay (for example in the offcanvas cart) several times at some point Firefox 41.0.1 under Windows 7 has an issue that the overlay still "takes" all the clicks and scroll events. Therefore no button can be clicked and scrolling does not work either. 
I think this is an Firefox issue since when resizing the Window or changing one CSS value (disable and enable it with Firebug), everything works again as expected. 
This PR fixes this issue by disabling all pointer events in the closed overlay and enable them again in the opened overlay.

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/less/_components/overlay.less
+++ b/themes/Frontend/Responsive/frontend/_public/src/less/_components/overlay.less
@@ -31,11 +31,13 @@ default color:`#000000` default opacity (alpha):`0.7`
     top: 0;
     z-index: @zindex-overlay;
     visibility: hidden;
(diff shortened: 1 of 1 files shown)


MarcelSchmaeing:
Hi,

I had a look at your PR again. I can't reproduce your problem with the latest firefox version. Do you have more information for me so I can reproduce the problem more reliably.

Greetings,

Marcel

MarcelSchmaeing:
Hi,

I will close this PR. Please create a new one if you have still the problem in the future.

Greetings,

Marcel

PR #247, opened 2015-03-25
Make proxy file permissions configurable

FiveDigital:
The hardcoded file permissions 0664 can now be configured in the
config.php like:

'hook' => array(
'proxyChmod' => 0660
)

The default file permissions remain untouched at 0644. With the option
to configure the proxy file permissions it is e.g. possible to add
group write rights to the files which is often needed.

PR #366, opened 2015-10-21
Make proxy file permissions configurable

FiveDigital:
This PR ist the new version of #247.

The hardcoded file permissions 0664 can now be configured in the
config.php like:

'hook' => array(
'proxyFilePermissions' => 0660
)

The default file permissions remain untouched at 0644. With the option
to configure the proxy file permissions it is e.g. possible to add
group write rights to the files which is often needed.

### DIFF
--- a/engine/Library/Enlight/Hook/HookManager.php
+++ b/engine/Library/Enlight/Hook/HookManager.php
@@ -84,7 +84,8 @@ class Enlight_Hook_HookManager extends Enlight_Class
         $this->proxyFactory = new Enlight_Hook_ProxyFactory(
             $this,
             $options['proxyNamespace'],
-            $options['proxyDir']
+            $options['proxyDir'],
+            $options['proxyFilePermissions']
         );
 
         $loader->registerNamespace(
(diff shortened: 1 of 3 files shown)


roman-1983:
+1

bcremer:
Thank you @FiveDigital, so far the PR looks good, even so I would like to look further into this problem scope because of this: http://www.doctrine-project.org/2015/08/31/security_misconfiguration_vulnerability_in_various_doctrine_projects.html

Maybe we can even further improve the permissions of the generated cache files in shopware.

FiveDigital:
+1 for having the possibility to configure the chmod of all cache files

In our scenario we need chmod 660 for files and 770 for directories because the webserver user and our CLI users share the same group. Whenever there are files with 640 or directories with 750 we have caching problems wether on the CLI or the web, depending on where the cache was created the last time.

+100 for having the possibility to put the cache files into a Redis db to share it between all webservers ;-) We successfully used this in many Magento Setups in the past to avoid these problems.

FiveDigital:
@bcremer Sorry I messed up this PR :( Should I create a new one?

FiveDigital:
How about a merge for 5.1.2? :+1:

FiveDigital:
ping

bcremer:
Internal Issue: https://issues.shopware.com/#/issues/SW-13731

bcremer:
Closed this Pull Request and moved the Diskussion to this PR: https://github.com/shopware/shopware/pull/410

PR #367, opened 2015-10-22
Fixed birthday check

leehant:
Fixed birthday check which prevents dates before 01.01.1970

### DIFF
--- a/engine/core/class/sAdmin.php
+++ b/engine/core/class/sAdmin.php
@@ -686,7 +686,7 @@ class sAdmin
                 (int) $postData['birthday'],
                 (int) $postData['birthyear']
             );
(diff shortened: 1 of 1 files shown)


jhofm:
by the way, this error is still present in shopware 5.1.
https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Core/sAdmin.php:491

janbuecker:
Thanks for your contribution @leehant. The if-condition is unnecessary since all arguments to `mktime()` are casted. Therefore it will never be `false`. We need to validate the day/month/year parameters correctly.

I've reopened a ticket which will handle this problem. You can track it's progress [here](http://jira.shopware.de/?ticket=SW-11025).

PR #368, opened 2015-10-22
Duplicate block name

beongmbh:
Duplicate Block name as in backend/article/view/esd/detail.js


bcremer:
Thank you @beongmbh, will be released in SW 5.1.1 soon.

PR #369, opened 2015-10-29
left-join for "joinDefaultPrices"

jppeter:
We have some customer-environments without prices for the customer group "EK".
The "defaultPrice"-fallback-routine fails in this case and destroy price-features in product-listings: 
Price-filter has no values and price-sorting is not possible.
With the commited left-join everything is fine :)


p3king:
+1

OliverSkroblin:
Hi,
it is required that only products with standard prices will be selected, otherwise the pagination can not be determined clean. Products without standard prices would be removed by the ListProductService.

bcremer:
@jppeter For the reasons given by @OliverSkroblin I will close this PR for now. Please feel free to reopen or comment here if you think this needs to be changed.

PR #370, opened 2015-10-30
[BUGFIX] Write plugin changes to database only if it is a string

FiveDigital:
I found this by trying to install the latest update of the "FutiFAQManager" plugin which raised an error:

`bin/console sw:plugin:update FutiFAQManager`

`PHP Catchable fatal error:  Object of class Enlight_Config could not be converted to string engine/Library/Zend/Db/Statement/Pdo.php on line 228`

After digging in the code I found out that the Plugin Namespace Component tries to write an Enlight_Config object to the database.

This PR checks that the given info is a string. Even though I think writing the changes to the database isn't needed anymore. Looking into our production database there are only 2 entries both with the value "Object"...


janbuecker:
HI @FiveDigital, thanks for your PR.

Please rebase your PR. We'll then verify it again.

FiveDigital:
Rebasing done @janbuecker

PR #371, opened 2015-10-31
Event for custom password encoders

ghost:
with this change it's possible to override the default password encoder. so you can easilty validate user logins with third party tools like ldap / ms ad ds, facebook etc.. additional i add a getter and setter for the password encoder in the customer model to modify the user password Encoder.

### DIFF
--- a/engine/Shopware/Components/Password/Manager.php
+++ b/engine/Shopware/Components/Password/Manager.php
@@ -126,6 +126,14 @@ class Manager
             }
         }
 
+        $encoderName = $this->eventManager->filter(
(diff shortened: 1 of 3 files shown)


bcremer:
Hi @bestit-tkellner, can you please elaborate your Use Case?

It is possible to add Custom Password Encoders to the system as documented [here](https://developers.shopware.com/developers-guide/password-encoder/).
The custom password encoder will also be available in the settings and can be set as the default one.

bcremer:
Closed due to the lack of response. Feel free to reopen.

PR #372, opened 2015-11-03
Same Voucher for subshop

wesolowski:
Hello, i got an issue while creating Vouchercodes. 

If you create two vouchercodes with an identically name in two different subshops and you limit their accessibility to their own subshob the second vouchercode wont be accepted.  

Its  because the Sql-request asks just for the name and is limited to one.  Only the first vouchercode will be found, wich is limited to the first subshop.


janbuecker:
Thanks @wesolowski 

PR #373, opened 2015-11-03
Add Filter for Voucher Params

wesolowski:
Hello, 

i would like to modify the parameters of vouchers which will be written in a table.  At the moment im just able to modify the querry. Could you please create a filter for that.


janbuecker:
Thanks @wesolowski 

PR #374, opened 2015-11-03
Fix slow VoucherCode-Query in Frontend

wesolowski:
Hello, i found an issue while searching for vouchers.

When you got a large number of vouchers (e.g. s_emarketing_voucher_codes  > 1GB) it takes a long time so serach for them (~2-5sec). Its much faster when you split up the join in two separated Sql-requests.


jhofm:
oh, someone else fixed this. ^^
https://github.com/shopware/shopware/pull/400/files
i just lost a holiday because of that query. again.

db

that's our db cpu before and after applying the fix.

bcremer:
@wesolowski @jhofm merged with commit https://github.com/shopware/shopware/commit/03a06be77eecfb26eb9563310e5291b17f1a7a24.

Thanks for your contribution :+1:

PR #379, opened 2015-11-10
Add listing ID to body

ChristopherDosin:
Adds the ID of the current listing/category page to the body element.

### DIFF
--- a/themes/Frontend/Bare/frontend/index/index.tpl
+++ b/themes/Frontend/Bare/frontend/index/index.tpl
@@ -11,7 +11,7 @@
 	{include file='frontend/index/header.tpl'}
 {/block}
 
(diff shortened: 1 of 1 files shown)


bcremer:
Can you please elaborate the use-case for this change.

Also this class can be added to the `body` element using the `frontend_index_body_classes` smarty block and a simple plugin.

ChristopherDosin:
Because the use-case: 
http://forum.shopware.com/themes-und-design-f101/less-fur-bestimmte-kategorie-t32552.html#p141609

http://forum.shopware.com/themes-und-design-f101/less-fur-bestimmte-kategorie-t32552-20.html#p142463

Of course there is the smart block ... didnt noticed this block ... :)

But maybe it should be just a core function and not only for listings so the user can set styles for each specific site.

janbuecker:
Thanks @kayyyy 

PR #382, opened 2015-11-12
Fix sorting by Position for numerical values

MBDealer:
http://imgur.com/a/A3zhp

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/ProductPropertyGateway.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/ProductPropertyGateway.php
@@ -182,6 +182,7 @@ class ProductPropertyGateway implements Gateway\ProductPropertyGatewayInterface
         $query->orderBy('filterArticles.articleID')
             ->addOrderBy('relations.position')
             ->addOrderBy('propertyGroup.name')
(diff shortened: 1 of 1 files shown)


bcremer:
Can you please give a little more information what the problem is, and how the fix solves that issue for you, thanks.

MBDealer:
Please, look in your IRC client for more information. I have sent you a private message.

bcremer:
Thanks for your message @MBDealer, I also did some tests:

Without patch: 
sort
With patch:
sort-fixed

There is still a problem with sorting values with prepended numbers as you can see in above screenshot (test_prefix).

Turns out, that natural sorting is not that easy in mysql. I will do some further research.

Without patch:
alpha-sort

With patch:
alpha-sort-fixed

bcremer:
It turns out that natural sorting is a hard problem for MySQL so I put together a quick proof of concept for a PHP based sorting of the properties: https://github.com/shopware/shopware/commit/8cf35921603fe5d41e2f249846aedb7a9a805f5b

Can you please try if that patch works for your usecase?
Sorting in PHP also has the benefit that we can sort the translated properties.

bcremer:
We created a ticket for this issue in our bugtracker: http://jira.shopware.de/?ticket=SW-13147
I will close this PR for now. You can track progress for this issue using the above link.

Thanks for your input @MBDealer :+1:

PR #384, opened 2015-11-17
Exclude shipping costs from order amount when shipping costs are hidden

andrewisplinghoff:
When the document is configured to not include the shipping costs as a position, as is the case with delivery notes per default, the order amount should also not include the shipping costs, because otherwise the sum of the positions does not equal the order amount.

To reproduce this problem, the prices have to be made visible in the delivery note, e.g. by removing the following lines from the `themes/Frontend/Bare/documents/index_ls.tpl` template:

```
{block name="document_index_table_price"}
{/block}
{block name="document_index_amount"}
{/block}
```

Before the PR, the order amount is wrong:

lieferschein1

After the PR, the order amount is correct:

lieferschein2

### DIFF
--- a/engine/Shopware/Models/Document/Order.php
+++ b/engine/Shopware/Models/Document/Order.php
@@ -301,19 +301,20 @@ class Shopware_Models_Document_Order extends Enlight_Class implements Enlight_Ho
         }
 
         $taxShipping = (float) $taxShipping;
-        if ($this->_order["taxfree"]) {
-            $this->_amountNetto =  $this->_amountNetto + $this->_order["invoice_shipping"];
-        } else {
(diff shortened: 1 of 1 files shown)


andrewisplinghoff:
Please review this, it is also important for <a href='http://www.pickware.de'>Pickware</a> (we support to create partial cancellation documents without including the shipping costs and the same problem occurs there).

bcremer:
The Pull Request is scheduled for Shopware 5.1.2 and currently in internal QA.

Sorry for the delay.

samuelvogel:
Maybe include it with the upcoming `5.1.4` sprint? :wink:

janbuecker:
PR has been merged with 3c8ce497a6094ea3a60f39ac06767c353ecf2259.

samuelvogel:
Nice, thank you guys! 👍

PR #385, opened 2015-11-17
Enable multiple categories for emotions

PortaltechGithub:
Hey folks,
with this minor change it is possible to assign an emotion to multiple categories. This was already implemented for landingpages. All I had to do was to move the landingpage categories field out of the landingpage specific fieldset. I put it at the first place in general fieldset because I think it is one of the most important fields.

multiplecategoriesforemotions

Regards,
Thomas

### DIFF
--- a/engine/Shopware/Controllers/Backend/Emotion.php
+++ b/engine/Shopware/Controllers/Backend/Emotion.php
@@ -167,8 +167,6 @@ class Shopware_Controllers_Backend_Emotion extends Shopware_Controllers_Backend_
 
         if (!empty($emotion["isLandingPage"])) {
             $emotion["link"] = "shopware.php?sViewport=campaign&emotionId=".$emotion["id"];
-        } else {
-            $emotion["categoryId"] = !empty($emotion["categories"][0]["id"]) ? $emotion["categories"][0]["id"] : 0;
         }
 
         $validFrom = $emotion['validFrom'];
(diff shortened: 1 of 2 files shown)


FiveDigital:
:+1:

bcremer:
Looks fine to me. The pull request is currently in internal QA.

bcremer:
@PortaltechGithub 

QA discovered some bugs that need to be fixed before we can merge: 
- For "Slave" landingpages the categories field should not be visible
- This fix is not compatible with `SwagEmotionAdvanced`. As soon as an Emotion is Promoted to type "Storytelling" the categories field disappears. 

I think the most easy way fixing that is keeping the categories selection in the category fieldset.

PortaltechGithub:
Hey @bcremer,
thanks for your feedback.

I had many problems with creating two categories fields (one for the category fieldset and one for the landingpage fieldset). This was resulting in duplicate entries in the mm-table and some more errors.
And I also think it is not pretty to have two fields which do the same.

The next thing I tried was to put the one categories field in both extjs containers, but it seems like the last container wins and the field is only inserted there.

So I initially create the categories field in the category fieldset and if the landingpage checkbox is clicked, I move the field to the landingpage fieldset. This way it is hidden if a master landingpage is selected. In addition it does not disappear for Storytelling emotions.

Regards,
Thomas

tokopherol:
👍

bcremer:
Thank you Thomas. Ticket is back in QA. I'll keep you posted.

bcremer:
@PortaltechGithub Green light by our QA. Merged with commit https://github.com/shopware/shopware/commit/35488698913bf938418cdc07657f417d142d7752. This will be included in Shopware 5.1.2 that will be relased soon.

Thank you, great work :+1:

PR #376, opened 2015-11-03
Bugfix SW-12289

rgroli:
The hydrator `HYDRATE_ARRAY` does not support nested sets. 
Accessing `$site['children']` does not fetch the dependent models.
With `$site->getChildren()` it works.

PR #387, opened 2015-11-18
Support for more than 2 level nesting of sites on backend site applic…

shyim:
The Application has problems to display more than 2 level nesting sites. This PR will fix this.
Before: http://i.imgur.com/cYcfh6N.png
After: http://i.imgur.com/HpLGOqM.png


shyim:
It the same as https://github.com/shopware/shopware/pull/376/commits

PR #388, opened 2015-11-19
[FEATURE] add notifyUntil event to executeRiskRule

PortaltechGithub:
Hey folks,
with this new event it is possible to add custom risk rules. The Extjs part is already prepared to add new rules:

```
...

data: [
    { description: '{s name=risks_store/comboBox/startValue}Please choose{/s}', value: "" },
    //{block name="backend/risk_management/store/risk/data"}{/block}
    { description: '{s name=risks_store/comboBox/orderValueGt}Ordervalue >={/s}', value: 'ORDERVALUEMORE' },

...
```

from themes/Backend/ExtJs/backend/risk_management/store/risks.js

But the sAdmin class does not offer any events to check those new risks.

I added the `$rule` into the event name to prevent accidentally overwriting all rules.

Regards,
Thomas

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -2592,6 +2592,19 @@ class sAdmin
      */
     public function executeRiskRule($rule, $user, $basket, $value)
     {
+        if ($event = $this->eventManager->notifyUntil(
+            'Shopware_Modules_Admin_Execute_Risk_Rule_' . $rule,
+            array(
+                'subject' => $this,
+                'rule' => $rule,
+                'user' => $user,
+                'post' => $basket,
+                'value' => $value
+            )
+        )) {
+            return $event->getReturn();
+        }
+
         return $this->$rule($user, $basket, $value);
     }
 


engine/Shopware/Core/sAdmin.php
```diff
@@ -2592,6 +2592,19 @@ public function sManageRisks($paymentID, $basket, $user)
      */
     public function executeRiskRule($rule, $user, $basket, $value)
     {
+        if ($event = $this->eventManager->notifyUntil(
+            'Shopware_Modules_Admin_Execute_Risk_Rule_' . $rule,
+            array(
+                'subject' => $this,
```
bcremer, line 2598: Can you please remove `subject` from the array?

We did that in older versions of shopware but omit a reference to `this` in newer events.
This is to decouple the emitted event from the event-emitter.

bcremer:
@PortaltechGithub  The Pull Request looks fine to me. One commit comment that needs to be fixed before merge.

engine/Shopware/Core/sAdmin.php
```diff
@@ -2592,6 +2592,19 @@ public function sManageRisks($paymentID, $basket, $user)
...
+                'rule' => $rule,
+                'user' => $user,
+                'post' => $basket,
```
bcremer, line 2601: Any reason for this `'post' => $basket`?
Please rename to `'basket' => $basket`.

bcremer:
Merged the PR with commit https://github.com/shopware/shopware/commit/0aecbfa02999b831d0c27846503ca5bc2b1ba1ac.

Also rebased to current 5.1 branch and did some minor fixes.

Thanks @PortaltechGithub :+1:

bcremer:
Also added a section to our developer documentation: https://developers.shopware.com/developers-guide/risk-rules/

PR #389, opened 2015-11-27
[FEATURE] Add attributes to propery options

FiveDigital:
This PR adds a new table `s_filter_options_attributes` and an attribute relation to the `PropertyOption` model to be able to add attributes to this model.

### DIFF
--- a/_sql/migrations/628-add-propery-option-attributes.php
+++ b/_sql/migrations/628-add-propery-option-attributes.php
@@ -1,0 +1,23 @@
+<?php
+
+class Migrations_Migration628 Extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $sql = <<<EOD
+CREATE TABLE IF NOT EXISTS `s_filter_options_attributes` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `filterOptionID` int(11) DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `filterOptionID` (`filterOptionID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci AUTO_INCREMENT=1 ;
+EOD;
+        $this->addSql($sql);
+
+        $sql = <<<EOD
+ALTER TABLE `s_filter_options_attributes` ADD CONSTRAINT `s_filter_otpions_attributes_ibfk_1` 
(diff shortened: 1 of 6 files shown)


FiveDigital:
ping

bcremer:
We are currently building a new module to manage attributes in the Shopware backend. The `s_filter_options_attributes` table will also be added in this process.
Please keep an eye on the `5.2` branch.

I will close this PR for now.

FiveDigital:
That sounds like an really good idea :+1:

FiveDigital:
I'm still playing around with this. Adding the new table and association in the model is not everything we need. We also need the Attribute joined in `Shopware\Models\Property\Repository::Shopware\Models\Property()` and associated (to a yet missing group attribute model) in the ExtJS model `Shopware.apps.Property.model.Group`. Will you add these, too @bcremer?

Furthermore it is necessary to add this code snippet in `Shopware_Controllers_Backend_Property::updateGroupAction()` on line 322 to save the attributes:

```
$data = $this->Request()->getPost();

if (isset($data['attribute'][0])) {
    $data['attribute'] = $data['attribute'][0];
}

$option->fromArray($data);
```

bcremer:
@OliverSkroblin @janbuecker Can you please take care of this in the new Attribute Module.

FiveDigital:
Since we have the code alreay, I could update this PR.

bcremer:
Feel free to update the PR :+1:

FiveDigital:
I have added the modified JS files which now contain a new `Shopware.apps.Property.model.GroupAttribute` model and a relation to the new model in the `Shopware.apps.Property.model.Group` model. I have also modified the `Shopware_Controllers_Backend_Property` controller to save the attributes when creating or updating a property group. 

Please merge this changes with the new attributes feature in the 5.2 release. We will use this already in our current customer's project. To give some background information: We need this because we are importing the properties from a PIM system which is a quite common scenario in enterprise projects. To match the PIM values against the Shopware properties we need an unique identifier in Shopware. We tried to add this as an attribute to the property groups and found out that this was not possible.

<offtopci>The big picture might be to have a "code" on every model in core which can be used for exporting/importing things to backend system. We always have to add this to the payment and shipping models for instance.</offtopic>

janbuecker:
As we are releasing Shopware 5.2 within the next weeks, please make use of the new attribute management announced with Shopware 5.2.

Feel free to create a new PR including your changes. Please select branch `5.2` as your target branch.

PR #390, opened 2015-11-30
Attribute fix for backend dispatch module

shyim:


janbuecker:
Thank you @shyim.

This PR is identical to #421. As written there, we won't implement this fix due to a whole new attribute management in Shopware 5.2. For more informations refer to PR #421.

PR #391, opened 2015-12-01
Don't override memory limit if a larger one is set

hashworks:
While rebuilding my search index I got into memory limits after 512 megabytes. Setting a larger limit in my php.ini didn't change that, this is a fix for this.


hashworks:
What would be a good place for the setMinimumMemoryLimit helper function?

bcremer:
Implemented that feature with commit https://github.com/shopware/shopware/commit/e0f4ac7ec568c828f5ef792d8c4a5707dacd7df9.

Thank you @hashworks.

PR #393, opened 2015-12-10
fix jquery.register.js to perform the ajax validations

resslinger:
Without this fix, the ajax validations doesn't work on the register page. The fix was tested on 5.0.4, 5.1.1 and works fine. In your demoshop, the ajax validation doesn't work, too.


bcremer:

PR #394, opened 2015-12-14
change sql condition, wrong condition for this db field type

resslinger:
The column ordernumber in the db table "s_order" has the type varchar. With this change, ordernumbers as a string can be filter correct now.


janbuecker:
Merged your PR with commit 709749fd2ffbd8410dbd986141afd39f3ab08759.

Thank you @k-30

PR #395, opened 2015-12-15
Fixed manufacturer link output in emotion slider

ascheider:
Manufacturer slider builds url's always for the listing.
So if you defines url 'Link to manufacturer website' in shopware backend
image
it has no affection to the output. With this patch you will be able to overwrite the standard listing url.


janbuecker:
The current behaviour is like intended to be. We won't fix this issue. Closing this PR.

PR #396, opened 2015-12-15
defuse voucher code query timebomb

jhofm:


jhofm:
Okay, that might have been a tad drastic. Still, this needs to be adressed somehow.

jhofm:
jesus, i intended to close my other pull request. i'll restore this tomorrow.

PR #397, opened 2015-12-15
remove detachment of shop instances for fun and profit

jhofm:
not expecting to get this approved; honestly i only want to see if the tests pass when shop instances are not generally detached from the ORM or if the world does actually explode. which would be about the only thing worse than having multiple main shops all of the sudden because somebody sent an email and the entity manager was flushed afterwards.


jhofm:
Okay, this was a bit drastic. Still, those detached shop instances floating around everywhere are timebombs with all those \Shopware()->Models()->flush() calls about. ;) This should be adressed somehow; it is basically my main concern in upgrading prod towards Shopware 5 currently.

PR #398, opened 2015-12-15
Vouchercode query fix

jhofm:
Restored pull request after i accidentally closed the original one. Sorry about the additional carbon footprint.


jhofm:
wrong branch, apologies. multitasking, late hours and code do not mix. see you tomorrow. -.-

PR #400, opened 2015-12-16
individual voucher code query performance fix

jhofm:
Restored pull request after multiple failed attempts. Sorry about the additional carbon footprint.

The function is marked as deprecated but still seems to be in use as of 5.2.

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -616,22 +616,21 @@ class sBasket
         } else {
             // If we don't have voucher details yet, need to check if its a one-time code
             $voucherDetails = $this->db->fetchRow(
-                'SELECT s_emarketing_voucher_codes.id AS id, s_emarketing_voucher_codes.code AS vouchercode,
-                    description, numberofunits, customergroup, value, restrictarticles,
-                    minimumcharge, shippingfree, bindtosupplier, taxconfig, valid_from,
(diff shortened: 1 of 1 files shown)


jhofm:
thanks to our upgrade to shopware 5.1 this week this query raped our database a second time. we didn't notice it had reappeared because the path also changed. i'd really look into this problem if you care about marketing your shop as scalable.

jhofm:
by the way: the soft lock part was bullshit. that is not what matters. what does is that the query builds a huge table in environments with lots of stored voucher codes without any way for innodb to limit the rows it has to scan, least because of the function used in the where criteria. and one of these babies hits once per page request for any user with a voucher in the basket, since vouchers are recalculated on every basket->get. this query is death.

bcremer:
Thank you @jhofm, we will look into this after the Shopware Community Day.

bcremer:
Related to https://github.com/shopware/shopware/pull/374.

jhofm:
thanks for looking into it. sorry for the harsh tone, i lost quite a bit of sleep at the time; which makes me grumpy. i've abandoned my plans to smuggle a club into the shopware community day by now. ;)

bcremer:
@jhofm I will close this as a duplicate of https://github.com/shopware/shopware/pull/374.

If you still have performance issues feel free to open a new PR.

Thank you.

PR #402, opened 2016-01-08
add fuzziness parameter to elasticsearch query

sunchairs-deploy:
Add fuzziness parameter to retrieve fuzzy search results when using elasticsearch


bcremer:
Hi @nicoanders, thank`s for contribution.

We would like to keep the Elasticsearch search query as simple as possible. The `fuzziness` parameter may lead to way to many results for some customers.

Customization and advanced parameters should be added using a plugin.
Please have a look at the following documentation to extends or replace the `SearchTermQueryBuilder` in a plugin: https://developers.shopware.com/developers-guide/elasticsearch/#extend-product-search-query

PR #403, opened 2016-01-08
Allow manyToOne associations to Shopware.apps.Property.model.Set

jeboehm:
While following your [ManyToOne associations guide](http://community.shopware.com/Shopware-4-models_detail_1252.html#OneToMany_.2F_ManyToOne_annotations), I tried to set up an ManyToOne association to the Shopware.apps.Property.model.Set ExtJS model.

Now, if I open up the detail window which was created as described in the documentation, the function `createModelField` in class `Shopware.model.Container` calls `fieldModel.getConfig('field')` which doesn't exist and leads to an `Uncaught TypeError`.

This PR fixes that.

### DIFF
--- a/themes/Backend/ExtJs/backend/property/model/set.js
+++ b/themes/Backend/ExtJs/backend/property/model/set.js
@@ -34,10 +34,10 @@
 Ext.define('Shopware.apps.Property.model.Set', {
 
     /**
-     * Extends the standard ExtJS 4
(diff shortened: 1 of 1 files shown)


dnoegel:
@bcremer, @OliverSkroblin: This should be good to go if I am not mistaken?

bcremer:
Looks good. Thank you @jeboehm :+1:

PR #404, opened 2016-01-09
Shopware Debug plugin

wesolowski:
Hello,

I wrote a plugin for shopware, which makes the [Debug-Plugin](https://github.com/shopware/shopware/tree/5.1/engine/Shopware/Plugins/Default/Core/Debug) available for [Clockwork](https://github.com/itsgoingd/clockwork-chrome) integration. 
All relevant informations can now be displayed in a chrome-tab which provides all features from the FirePHP-version in a more readable overlay. In addition there are some timelines for informations like how long did it take to load an event or to render the site.

For this I need:
- inside the collectors class the attribute 'utils' (protected) to have it available by extending the class
- two "filter"-events in the Bootstrap.php of the Debug-plugin
  - Shopware_Plugins_Core_Debug_Bootstrap_FilterCollectors in the "registerCollectors"-Methode to change the Collector
  - Shopware_Plugins_Core_Debug_Bootstrap_FilterHandlerRegister in the "getHandlers"-Methode to add the ClockworkHandlers

### DIFF
--- a/engine/Shopware/Plugins/Default/Core/Debug/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Core/Debug/Bootstrap.php
@@ -236,7 +236,7 @@ class Shopware_Plugins_Core_Debug_Bootstrap extends Shopware_Components_Plugin_B
 
         $handlers = array();
         foreach ($handlerRegister as $handler) {
-            if ($handler instanceof HandlerInterface && $handler->acceptsRequest($request)) {
+            if ($handler instanceof HandlerInterface) {
                 $handlers[] = $handler;
             }
         }
--- a/engine/Shopware/Plugins/Default/Core/Debug/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Core/Debug/Bootstrap.php
@@ -141,43 +141,55 @@ class Shopware_Plugins_Core_Debug_Bootstrap extends Shopware_Components_Plugin_B
     {
         $this->get('loader')->registerNamespace('Shopware\Plugin\Debug', __DIR__ . '/');
 
+        $collectors = array();
         $eventManager = $this->get('events');
         $utils = new Utils();
         $errorHandler = $this->Collection()->get('ErrorHandler');
 
         if ($this->Config()->get('logTemplateVars')) {
-            $this->pushCollector(new TemplateVarCollector($eventManager));
+            $collectors[] = new TemplateVarCollector($eventManager);
         }
 
         if ($this->Config()->get('logErrors')) {
-            $this->pushCollector(new ErrorCollector($errorHandler, $utils));
+            $collectors[] = new ErrorCollector($errorHandler, $utils);
         }
 
         if ($this->Config()->get('logExceptions')) {
-            $this->pushCollector(new ExceptionCollector($eventManager, $utils));
+            $collectors[] = new ExceptionCollector($eventManager, $utils);
         }
 
         if ($this->Config()->get('logDb')) {
-            $this->pushCollector(new DatabaseCollector($this->get('db')));
+            $collectors[] = new DatabaseCollector($this->get('db'));
         }
 
         if ($this->Config()->get('logModel')) {
-            $this->pushCollector(new DbalCollector($this->get('modelconfig')));
+            $collectors[] = new DbalCollector($this->get('modelconfig'));
         }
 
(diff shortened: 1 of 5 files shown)


engine/Shopware/Plugins/Default/Core/Debug/Bootstrap.php
```diff
@@ -141,43 +141,55 @@ public function registerCollectors()
...
         if ($this->Config()->get('logEvents')) {
-            $this->pushCollector(new EventCollector($eventManager, $utils));
+            $collectors[] = new EventCollector($eventManager, $utils);
         }
 
-        foreach ($this->collectors as $collector) {
+        $collectors = Enlight()->Events()->filter(
+            'Shopware_Plugins_Core_Debug_Bootstrap_FilterCollectors',
+            $collectors,
+            array(
```
bcremer: Please remove `eventManager`, `utils` and `errorHandler` from the event args. If a custom Collector needs them as a dependency just pull them from the DI-Container.

engine/Shopware/Plugins/Default/Core/Debug/Bootstrap.php
```diff
@@ -141,43 +141,55 @@ public function registerCollectors()
...
 
         if ($this->Config()->get('logController')) {
-            $this->pushCollector(new ControllerCollector($eventManager, $utils));
+            $collectors[] = new ControllerCollector($eventManager, $utils);
         }
 
         if ($this->Config()->get('logEvents')) {
-            $this->pushCollector(new EventCollector($eventManager, $utils));
+            $collectors[] = new EventCollector($eventManager, $utils);
         }
```
bcremer, line 181: Please use the `collect()` Event type like this:

```
$collection = new ArrayCollection($collectors);
$collection = Enlight()->Events()->collect('Shopware_Plugins_Core_Debug_Bootstrap_Collect_DebugCollector', $collection);
$collectors = $collection->toArray();
```

engine/Shopware/Plugins/Default/Core/Debug/Bootstrap.php
```diff
@@ -217,9 +229,16 @@ public function onStartDispatch(\Enlight_Event_EventArgs $args)
      */
     public function getHandlers(\Enlight_Controller_Request_Request $request)
     {
+        $handlerRegister = Enlight()->Events()->filter(
```
bcremer: Please use the collect() Event type. See comment above.

hlohaus:
Nice :) :+1:

engine/Shopware/Plugins/Default/Core/Debug/Bootstrap.php
wesolowski, line 181: I want to change some Collectors.
With the Collect-Event i can only add Collectors, but i cant replace them.
Here is an example how i´d like to use it: [Debug-Subscriber](https://github.com/wesolowski/shopware-clockwork/blob/master/Subscriber/Debug.php#L40)

bcremer, line 181: @wesolowski you are of course right. Please keep the `filter()` event type.

engine/Shopware/Plugins/Default/Core/Debug/Bootstrap.php
```diff
@@ -217,9 +225,20 @@ public function onStartDispatch(\Enlight_Event_EventArgs $args)
...
+        $handlerRegister = $this->get('events')->collect(
+            'Shopware_Plugins_Core_Debug_Bootstrap_FilterHandlerRegister',
+            $handlerRegister
+        );
+
         $handlers = array();
-        if ($this->get('monolog.handler.firephp')->acceptsRequest($request)) {
-            $handlers[] = $this->get('monolog.handler.firephp');
+        foreach ($handlerRegister as $handler) {
+            if ($handler instanceof HandlerInterface && $handler->acceptsRequest($request)) {
```
bcremer: `acceptsRequest()` is not part of the `HandlerInterface` and should not be checked here.
`acceptsRequest()` is only available on the `monolog.handler.firephp` and should be checked only on that handler.

wesolowski: You are right, that is not necessary. FirePhpHandler check acceptsRequest in setUp method

janbuecker:
Thanks @wesolowski 

PR #405, opened 2016-01-11
Remove @covers annotation for not existing method

jeboehm:
Hi,

running PHPUnit with coverage throws the error `Trying to @cover or @use not existing method "sAdmin::sRiskBILLINGPREGSTREET"`. The method doesn't exist anymore, so I removed the annotation.


bcremer:
Thank you :+1:

PR #406, opened 2016-01-11
Remove duplicate test

jeboehm:
Hi,

like in #405, the method used in the `@covers`annotation doesn't exist anymore. Because the `testRewriteLink` test does the same as `testsRewriteLink`, I removed it completely.


bcremer:
Thank you @jeboehm.

PR #407, opened 2016-01-12
Removed duplicate template assignment

shyim:
Removes the duplicate template assignment in ajaxAmount


bcremer:
Thank you @shyim.

PR #408, opened 2016-01-12
Switch Grunt to relativeUrls to unify the paths to less.php

shyim:
We have some problems in the projects with the less paths. Grunt currently doesnt modifiy the urls and the less.php makes relative to the files and this annoying the development. This PR should fix the issue and make the same paths like less.php.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,9 @@
 # Shopware Upgrade Information
 In this document you will find a changelog of the important changes related to the code base of Shopware.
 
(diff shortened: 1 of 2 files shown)


bcremer:
@shopware/shopware-webdev can you please review? Thanks.

klarstil:
@bcremer I checked the PR. It makes sense to let grunt rewrite the file paths instead of overwriting LESS variables. If you don't have any concerns, we can merge the PR.

PR #409, opened 2016-01-14
[TASK] Add payment and dispatch attributes to OrderQueryBuilder

FiveDigital:
With this PR the attributes for payment and dispatch will be included in the REST API. We need this because we added codes to both models for our customers ERP. I think there is no reason to exclude the attributes.


bcremer:
Merged with commit https://github.com/shopware/shopware/commit/1ebf8127d195c3119c7d17237a296b17f7162c43.

Thank you @FiveDigital :+1:

PR #410, opened 2016-01-14
SW-13731 - Respect umask for all created cache files

bcremer:
Alternative for Pull Request https://github.com/shopware/shopware/pull/366 by @FiveDigital
Internal Issue: https://issues.shopware.com/#/issues/SW-13731

**Problem**
Instead of a dozen different settings for file and directory permissions I would like Shopware to respect the [`umask`](http://php.net/manual/en/function.umask.php) of the PHP Process for all generated caches.

**Example:**
Your desired permissions for files of `660` and `770` for directories can be archived by setting the `umask` to `007`.

This can be either done in the Apache configuration/startup script, or during early Shopware boot in the PHP Process itself. For that add `umask(0007);` to your `config.php`.

**Implementation:**
Most caches already did respect the `umask`. 
This pull request fixes the remaining template and the proxy caches.

The only hardcoded permission of `0664` and `0775` are the generated Doctrine Proxies. This is due to the recent changes regarding a possible Security Misconfiguration Vulnerability in Doctrine: http://www.doctrine-project.org/2015/08/31/security_misconfiguration_vulnerability_in_various_doctrine_projects.html

I already reported that issue to Doctrine Upstream and provided a pull request to fix that. See:
https://github.com/doctrine/doctrine2/pull/5602

@FiveDigital @roman-1983 
Can you please verify that this pull request will fix your permission issues?

### DIFF
--- a/engine/Library/Enlight/Hook/ProxyFactory.php
+++ b/engine/Library/Enlight/Hook/ProxyFactory.php
@@ -103,7 +103,7 @@ class <namespace>_<proxyClassName> extends <className> implements Enlight_Hook_P
         $this->proxyNamespace = $proxyNamespace;
 
         if (!is_dir($proxyDir)) {
-            if (false === @mkdir($proxyDir, 0777, true)) {
+            if (false === @mkdir($proxyDir, 0777, true) && !is_dir($proxyDir)) {
                 throw new \RuntimeException(sprintf("Unable to create the %s directory (%s)\n", "Proxy", $proxyDir));
             }
         } elseif (!is_writable($proxyDir)) {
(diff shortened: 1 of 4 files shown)


FiveDigital:
I will pull this PR into our current project's repository an evaluate the changes. We always have problems like `Unable to create the cache directory (/var/www/htdocs/var/cache/production_201601120732)` on either the CLI or in the web. In our scenario the webserver user and the CLI users share the same group and all directories have the sticky bit set. This is a common scenario since you won't use the webserver user on CLI. But since there are directories and files without write permissions for the group we often run into problems.

bcremer:
@FiveDigital any feedback on this? Our internal tests are are positive so I am going to merge this PR during the day.

FiveDigital:
I will have a look at this today :+1: The solution itself is of course better than my initial approach. We will test this in our scenarios on the CLI and web and provide you with feedback asap. Does the creation of the cache folder itself `production_201601120732` already respect the umask? We had problems with that folder and with the files under `web/cache`.

bcremer:
The cache directory already respected the umask.

See: https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Kernel.php#L454

```
if (false === @mkdir($dir, 0777, true) && !is_dir($dir)) {
    throw new \RuntimeException(sprintf("Unable to create the %s directory (%s)\n", $name, $dir));
}
```

`mkdir()` applies the the umask by default.

bcremer:
Was merged with commit: https://github.com/shopware/shopware/commit/dfd3b21d587023f7d257a78db77f1040067b823b

PR #412, opened 2016-01-16
Add aggregation size to prevent doc_count_error_upper_bound

xreal:
When you got a large amount of propertys you ran fast in the "doc_count_error_upper_bound" trap.
To prevent that, set good default. I think 5000 is enough for the most environments.


bcremer:
Looks good to me. 
@OliverSkroblin can you also review please.

bcremer:
Thank you @xreal :+1:

PR #413, opened 2016-01-20
remove legacy media url in index controller

teiling88:
if you use an category image on start page you get athe old media url structure.


bcremer:
Thank you @teiling88  :+1:

PR #414, opened 2016-01-21
introduced smarty blocks for footer headlines

SvenHerrmann:


klarstil:
@bcremer The PR looks good for me, it seems that the code indention is not correct. Can you fix it?

bcremer:
Thank you @SvenHerrmann :+1:

PR #418, opened 2016-01-25
Fixed output of '0' in xml convert

shyim:
I am currently building a export of orders, but all fields where the value is 0 will be droped. This Pull Request will solve the issue

### DIFF
--- a/engine/Shopware/Components/Convert/Xml.php
+++ b/engine/Shopware/Components/Convert/Xml.php
@@ -73,7 +73,7 @@ class Shopware_Components_Convert_Xml
                     unset($item['_attributes'], $item['_value']);
                 }
             }
-            if (empty($item)) {
+            if (empty($item) && $item != 0) {
                 $ret .= "$pad<$key$attributes></$key>{$this->sSettings['newline']}";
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Convert/Xml.php
```diff
@@ -73,7 +73,7 @@ public function _encode($array, $pos = 0, $ekey = "")
                     unset($item['_attributes'], $item['_value']);
                 }
             }
-            if (empty($item)) {
+            if (empty($item) && $item != 0) {
```
roman-1983, line 76: Why don't you use `isset()` instead?

roman-1983 (changes requested):
use `isset()` instead

engine/Shopware/Components/Convert/Xml.php
shyim: We have 2017 i can't say why.
I guess this pr will never be merged. This class is Deprecated in 5.3

janbuecker:
We decided to close this PR because of time constraints and limited overall benefit to the product.

PR #419, opened 2016-01-25
[TASK] Added order and payment state constants

Tombro:
This commit adds states from s_core_states as constants to the Status model. This is way prettier than magic numbers like:

```
$builder->andWhere($builder->expr()->notIn('orders.status', array('-1')));
```

(from engine/Shopware/Models/Order/Repository.php:223)

Regards,
Thomas


janbuecker:
Merged your PR with commit b36f8919cfb2d235165ce40d7e009d93809d02c5.

Thank you @Tombro

PR #421, opened 2016-01-26
Save dispatch attributes in backend controller

FiveDigital:
This is needed when adding new dispatch attributes. Other backend controllers like payment already have this.

### DIFF
--- a/engine/Shopware/Controllers/Backend/Shipping.php
+++ b/engine/Shopware/Controllers/Backend/Shipping.php
@@ -161,6 +161,8 @@ class Shopware_Controllers_Backend_Shipping extends Shopware_Controllers_Backend
             $params['bindTimeTo'] = null;
         }
 
+        $params['attribute'] = $params['attribute'][0];
+
         // convert params to model
         $dispatchModel->fromArray($params);
 


janbuecker:
This change introduces some other problems. If the attribute is already created, things get saved correctly. If you don't have an attribute for the dispatch yet, it will result in endless new attribute entries. This is because the data sent from the server does not match with the mapping in ExtJS - in fact, the answer loses the attributeId and dispatchId.

Since we are implementing a whole new attribute management in Shopware 5.2, we won't fix this issue.

FiveDigital:
If the new attribute system enables saving of dispatch attributes it is okay for us :+1: Otherwise I could fix this PR. Since you are doing this in the PaymentController, too it might not be too expensive.

See: https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Controllers/Backend/Payment.php#L183 and https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Controllers/Backend/Payment.php#L241

janbuecker:
Yup, it will. :-)

Keeping this PR closed. Thank you

FiveDigital:
I'm really excited about the new attribute management and appreciate platform changes more than new marketing features :+1:

PR #423, opened 2016-01-27
Update Property Group Model to fix bugs in the API

t2oh4e:
Removed array conversion in the Property Group model. Should be more consistent to the other models. This should fix #SW-13924. As far as I see there are no sideeffects with other parts that rely on an array but I'm not sure about that. Please verify.


janbuecker:
Thanks @t2oh4e 

As you already mentioned, the ticket SW-13924 will keep track of this issue. We'll close your PR for now.

PR #426, opened 2016-01-27
controller and action name in lowercase

shyim:
All css sectors using controller name in lower case like is--ctl-account, but when somebody request now the controller with upper case this will break some css.

Example:
http://sw51.shyim.de/newsletter
http://sw51.shyim.de/Newsletter

http://i.imgur.com/bd01oQP.png
http://i.imgur.com/cTRR6Ij.png

### DIFF
--- a/themes/Frontend/Bare/frontend/checkout/items/premium-product.tpl
+++ b/themes/Frontend/Bare/frontend/checkout/items/premium-product.tpl
@@ -15,7 +15,7 @@
                                 <div class="table--media-outer">
                                     <div class="table--media-inner">
                                         <a href="{$sBasketItem.linkDetails}" title="{$sBasketItem.articlename|strip_tags}" class="table--media-link"
-                                            {if {config name=detailmodal} && {controllerAction} === 'confirm'}
+                                            {if {config name=detailmodal} && {controllerAction|lower} === 'confirm'}
                                                 data-modalbox="true"
                                                 data-content="{url controller="detail" action="productQuickView" ordernumber="{$sBasketItem.ordernumber}" fullPath}"
                                                 data-mode="ajax"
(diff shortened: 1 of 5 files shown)


bcremer:
Thank you @shyim.

The fix looks ok for me. But please also replace all other occurrences of `{controllerName}` and `{controllerAction}` with the lower case version.

The following files contain occurences:
- themes/Frontend/Bare/frontend/index/logo-container.tpl
- themes/Frontend/Bare/frontend/index/index.tpl
- themes/Frontend/Bare/frontend/index/shop-navigation.tpl
- themes/Frontend/Bare/frontend/checkout/items/product.tpl
- themes/Frontend/Bare/frontend/checkout/items/premium-product.tpl

shyim:
@bcremer 
I think the best method should be add strtolower to the smarty function which handles the {controllerAction} and {controllerName} or can this make problems?

bcremer:
@shyim 

I also thought about fixing that in `smarty_function_controllerName` and `smarty_function_controllerAction` or even directly in `Enlight_Controller_Request_RequestHttp` but opted against it.

Changing `$request->getActionName()` and `$request->getControllerName()` to always lowercase would introduce a very subtile BC break that is hard to debug.

Putting the lowercase conversion into `smarty_function_controllerName` seems elegant but introduces a "magic" and hard to find difference in the request object and in the template variable.

Using the `|lower` modifier is the most obvious and will most likely not break BC.

shyim:
@bcremer 
Okay its done rebased :)

bcremer:
Fixed with commit https://github.com/shopware/shopware/commit/32ca47a5f12329381c45dcdbef37e5c19ba4f23d, merged with commit https://github.com/shopware/shopware/commit/e22544ac1c5f0944112ea3bcdf51a2bea44f43e7.

Thank you @shyim.

PR #427, opened 2016-01-29
SEO optimization: Show always h1 in category listing, including fallb…

schliesser:
SEO optimization: Show always h1 in category listing, including fallback "$sCategoryContent.name"

For good SEO every page should have a h1.
In Line 5 we removed:
`{if $sCategoryContent.cmsheadline || $sCategoryContent.cmstext}`
and added category description as Fallback:
`{else}<h1 class="hero--headline panel--title">{$sCategoryContent.name}</h1>`

Also we put the container "hero--text" inside the {if}:
`{block name="frontend_listing_text_content"}
            {if $sCategoryContent.cmstext}
                <div class="hero--text panel--body is--wide">`

### DIFF
--- a/themes/Frontend/Bare/frontend/listing/text.tpl
+++ b/themes/Frontend/Bare/frontend/listing/text.tpl
@@ -2,60 +2,60 @@
 
 {* Categorie headline *}
 {block name="frontend_listing_text"}
-    {if $sCategoryContent.cmsheadline || $sCategoryContent.cmstext}
-        <div class="hero-unit category--teaser panel has--border is--rounded">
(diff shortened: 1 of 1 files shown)


bcremer:
@shopware/shopware-webdev can you please review? Thanks.

bcremer:
@schliesser Thanks for your effort. Unfortunately we can't merge your pull request as it would break compatibility with a lot of already released themes. Also such an SEO optimization should be done in a custom theme that is adjusted 
per customer as it depends on the structure of the shop. Please see https://developers.shopware.com/designers-guide/theme-startup-guide/ for custom templates.

PR #428, opened 2016-02-03
Remove active header Class

Liverson:
Remove active header class if option closeSiblings is true


bcremer:
Thank you @Liverson :+1:

PR #431, opened 2016-02-09
Update pocketgrid.less

mario-fehr:
Caused an css error. http://i.imgur.com/XX2ByVU.png

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/vendors/less/pocketgrid/pocketgrid.less
+++ b/themes/Frontend/Responsive/frontend/_public/vendors/less/pocketgrid/pocketgrid.less
@@ -11,7 +11,7 @@
 
 /* Clearfix */
 .block-group {
-    *zoom: 1;
+    zoom: 1;
 }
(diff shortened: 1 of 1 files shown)


klarstil:
@mfehr94 Good catch, thank you. We're using `npm` (see [Gruntfile.js](https://github.com/shopware/shopware/blob/5.2/themes/Frontend/Responsive/Gruntfile.js)) to upgrade our frontend dependencies. I'll port your changes to the original branch (https://github.com/klarstil/pocketgrid-less) and will update the `pocketgrid.less` file in the shopware repository.

I'll let you know when the changes are pushed to the shopware repo.

bcremer:
This bug was also fixed in the upstream project from @klarstil. See: https://github.com/klarstil/pocketgrid-less/pull/1.

@klarstil can you please release a new version on npm and update the lib in the core please.

bcremer:
Thank you @mfehr94, this patch is included in the upcoming SW 5.4.1 release.

PR #432, opened 2016-02-09
SW-13999 - Create attribute models on model manager initialization

bcremer:
Create attribute models during model manager initialization. This could replace 
https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Commands/GenerateAttributesCommand.php and https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Components/AttributeSubscriber.php.


bcremer:
Merged with commit https://github.com/shopware/shopware/commit/dc9a9ec8f61921fb721f1365ba5135bc926b189e.

PR #433, opened 2016-02-12
Bugfix article seo urls for shops with fallback translation

martinweinmayr:
If more than one subshops do have the same language (e.g. english) and you intend to get the article translations inherited from one subshop to another, a bug occured. The newly generated seo url then will be translated in the language that has been set as default (e.g. german).


bcremer:
Thank you @martinweinmayr. This will be released in Shopware 5.1.4 soon.

PR #434, opened 2016-02-13
fixed BLOG Pagination + SEO relevant Data

christopherholst:
- Updated SEO rel Links
- Updated SEO Pagination more then actual 1 Link // now 4+ and 4-
- Updated SEO title snippet for Page from to CounterAllPages


MarcelSchmaeing:
Hi,

I had a more detailed look at your pull request. Seems like it is not fully responsive. 
Would be great if you can change the following things so we can accept your pull request:

1.) Please have a look at pagings in other listings. Your change should work and look the same so we have it consistent.
2.) We don't work with the rel attributes because the destination page declares the meta information already.

Thanks again for your contribution.

Greetings,

Marcel

MarcelSchmaeing:
Hi,

I will now close this PR. Please create a new one and also notice my comments.

Greetings,

Marcel

PR #436, opened 2016-02-16
Fix articleId on backend order position updates

samuelvogel:
Before #145 alls changes to order positions in the backend would not update the `articleId` correctly. #145 fixed this for newly added positions but altering the article of an order position (by using the article search dropdown) would not update `articleId` yet. This PR fixes this and also makes the PHP code of the original PR irrelevant and is much shorter!

**tl;dr** The icon to jump to the article from an order position now always opens the correct article as the `articleId` in the database is correct.


bcremer:
Thank you @samuelvogel.

PR #356, opened 2015-09-30
Fix shop duplication bug

andrewisplinghoff:
This is a new pull request for the shop duplication bug previously discussed here: https://github.com/shopware/shopware/pull/169

The proposed fix is based on @hlohaus changes as mentioned in
https://github.com/shopware/shopware/pull/169#issuecomment-49440259

The changes to engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php are necessary so that real subshops created with the SwagMultiShop plugin still work after the fix. Without these changes, the user is always redirected to the main shop. I successfully tested that language subshops still work without these lines.

/cc @jhit

PR #169, opened 2014-07-18
Fix shop duplication bug

stefanheppenheimer:
This pull request is a hotfix for the the shop duplication bug, which may occur in case that the associated shop entity of an order entity is processed by the `fixActive()` method of the shop repository and the order entity is flushed to the database thereafter. This behaviour is caused by [detaching the shop entity](https://github.com/ShopwareAG/shopware-4/blob/2dc9f4a33ef2ce247364428a0ad412d5f93e06b7/engine/Shopware/Models/Shop/Repository.php#L360) within the mentioned method in conjunction with the [cascading configuration of the order model](https://github.com/ShopwareAG/shopware-4/blob/2dc9f4a33ef2ce247364428a0ad412d5f93e06b7/engine/Shopware/Models/Order/Order.php#L253).

When the order entity is flushed to the database, doctrine evaluates the associated shop entity as `STATE_NEW` and processes an insert action for it (`cascade={"persist"}`), which leads to the duplication of the shop entity within the database.

Instead of detaching the shop entity, this hotfix marks the shop entity as read-only (no change-tracking) to prevent changes made to it from being flushed to the database. Nevertheless this hotfix is not a general solution. An implicit change of an entities behaviour (detatchment / mode change) is bad practice in general. The `fixActive()` method transfers missing properties to language subshops from the main shop. It should be removed to no longer change the behaviour of the shop entity. One approach to replace it would be to move the property transferring to the corresponding getters of the shop model like this:

```
    /**
     * @return string
     */
    public function getHost()
    {
        if ($this->getMain() !== null) {
            return $this->getMain()->getHost();
        }

        return $this->host;
    }

```

Remark: Because the main shop is not changed at all within the `fixActive()` method it is not necessary to detach it.

Run testcase:

```
vendor/bin/phpunit -c tests/Hotfix
```

PR #437, opened 2016-02-16
Fix shop duplication bug

samuelvogel:
This is a new pull request for the shop duplication bug previously discussed in #169 and #356 (based on 5.0).

The proposed fix is based on @hlohaus changes as mentioned in https://github.com/shopware/shopware/pull/169#issuecomment-49440259.

This is a big issue for a many plugin vendors as a lot of constellations as simple as calling `createStatusMail()` at the "wrong" time can lead to duplicated shop entities and the only workaround currently is to not use any Doctrine in these cases.

The changes to `engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php` are necessary so that real subshops created with the SwagMultiShop plugin still work after the fix. Without these changes, the user is always redirected to the main shop. I successfully tested that language subshops still work without these lines.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -9,6 +9,7 @@ In this document you will find a changelog of the important changes related to t
 * Fixed `AND` search logic for search terms which not exist in the s_articles table.
 * Added order and payment state constants in `\Shopware\Models\Order\Status`
 * change email validation to a simple regex: `/^.+\@\S+\.\S+$/`. You can implement your own email validation by implementing the `EmailValidatorInterface`. 
+* Deprecated `Shopware\Models\Shop\Repository::fixActive`
 
 ## 5.1.3
 * Switch Grunt to relativeUrls to unify the paths to less.php
--- a/engine/Shopware/Models/Shop/Repository.php
+++ b/engine/Shopware/Models/Shop/Repository.php
@@ -288,10 +288,6 @@ class Repository extends ModelRepository
         $builder->setParameter('shopId', $id);
         $shop = $builder->getQuery()->getOneOrNullResult();
 
-        if ($shop !== null) {
-            $this->fixActive($shop);
-        }
-
         return $shop;
     }
 
@@ -305,10 +301,6 @@ class Repository extends ModelRepository
         $builder = $this->getActiveQueryBuilder();
         $builder->where('shop.default = 1');
         $shop = $builder->getQuery()->getOneOrNullResult();
-
-        if ($shop !== null) {
-            $this->fixActive($shop);
-        }
 
         return $shop;
     }
@@ -365,10 +357,6 @@ class Repository extends ModelRepository
         /** @var $shops \Shopware\Models\Shop\Shop[] */
         $shops = $builder->getQuery()->getResult();
 
-        foreach ($shops as $currentShop) {
-            $this->fixActive($currentShop);
-        }
-
         //returns the right shop depending on the url
         $shop = $this->getShopByRequest($shops, $requestPath);
 
@@ -384,10 +372,6 @@ class Repository extends ModelRepository
 
         $shop = $builder->getQuery()->getOneOrNullResult();
 
-        if ($shop !== null) {
-            $this->fixActive($shop);
-        }
-
         return $shop;
     }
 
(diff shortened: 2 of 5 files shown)


NiklasTeich:
I solved this issue by simple calling `Shopware()->Models()->clear();` after the `createStatusMail()` / `sendStatusMail()` but yeah, i would also appreciate this fix ;)

samuelvogel:
Yes, but this makes code where you want to persist stuff using Doctrine before and after such a call very complicated.

jhit:
Please fix this bug once and for all. This is a major risk for agencies using third party plugins. We cannot verify every plugin author knows about this and we cannot check the plugins because the code is "ioncubed". Each time we add a new third-party plugin to Shopware we do not know if this bug bites us or not.

samuelvogel:
@bcremer: https://github.com/shopware/shopware/pull/356#issuecomment-145218306 was fixed by https://github.com/VIISON/shopware/commit/df5923e0fbad7de384cf1e530960aa56e662c794. So this should finally be ready to merge!

bcremer:
@samuelvogel, thank you for providing a updated pull request for this ongoing issue.
This pull request is scheduled and will be included in our next scrum sprint that starts next week.

bcremer:
Can you please verify if the following PR will fix the shop duplication issue for you: https://github.com/shopware/shopware/pull/443

The proposed fix in this PR introduced other problems.
 `Shopware()->Shop()` is no longer detached, so any change on this entity will be persisted to the database on the next `flush()`.

In PR #443 a new `DetachedShop` object will be introduced that is not managed by Doctrines Entity Manager so there is no risk of accidentally written changes on the Shop Object.

samuelvogel:
I just added a test case to the PR (rebased on upstream) which reproduces the shop duplication with three simple steps every plugin vendor had to fear before:
1. Modify an order model
2. Send status mail to customer (thanks @Haehnchen, see https://github.com/shopware/shopware/pull/169#issuecomment-50042300)
3. Flush order model

#443 fixes this as well, but it seems to make the shop handling even more complicated by adding another level of indirection with the `DetachedShop` on the already broken design of `Shopware\Models\Shop\Repository::fixActive()`.

Can you clarify why `Shopware()->Shop()` not being detached anymore is an issue? Yes, changes to it will now be persisted on `flush()`, but before this led to the shop duplication which makes your shop unusable anyways. So this should not be a backwards compatibility break, as it broke your shop before and therefore should not be in any production code-base in the wild (I hope)!

jhit:
> Yes, changes to it will now be persisted on flush(), but before this led to the shop duplication which makes your shop unusable anyways. So this should not be a backwards compatibility break, as it broke your shop before and therefore should not be in any production code-base in the wild (I hope)!

@samuelvogel To assume that nobody triggers the `flush()` on the shop Model is very dangerous. Actually there were plug-ins that triggered the bug in the past and probably there will be plug-ins that `flush()`in the future. The whole point of fixing this issue is to prevent shops  getting damaged. If the flush() call will overwrite settings of the store with empty or wrong data, this would be worse than the bug you try to fix here, because restoring the shop would be far more difficult than deleting a ghost-store entry in the configuration.

bcremer:
@samuelvogel I second the argumentation of @jhit.

Take for example the following code in `\Shopware_Plugins_Core_Router_Bootstrap::onRouteStartup()` (https://github.com/shopware/shopware/blob/7d29e5efe6e3ae4d95ba8d5a7becb8d421289a58/engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php#L79).

This runtime changes of the `$shop` object will get persisted into the database on the next `flush()` - that most likely not what you want.

Also my patch should also not introduce any real BC breaks.

Using `Shopware()->Shop()` as an entity for an relation `$order->setLanguageSubShop(Shopware()->Shop())` created a broken entry in `s_core_shops` before, now it will throw an appropriate exception as `Shopware()->Shop()` is not an managed/attached entity that should be used for assignments.

bcremer:
I am closing this PR here. Please continue discussions in #443.

Thank you @samuelvogel for you input on this issue.

samuelvogel:
The `engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php` part of this was merged with f206b27888cebf5e4b6fdf6d820c7955f916ef33.

PR #438, opened 2016-02-17
made php-paths in ant configurable

kirtixs:
Currently it is not possible to set the path to php and php-related tools such as phpcs, phpcpd ... which may be absent from the current $PATH.

This PR fixes this issue by introducing the following config options in `build.properties.dist`:

```
script.php      = php
script.phpcs    = phpcs
script.phpmd    = phpmd
script.phpcpd   = phpcpd
script.phploc   = phploc
script.pdepend  = pdepend
script.phpcb    = phpcb
```

The default values are the currently used ones and therefore this patch is save to BC-breaks.
This way `script.php` can be set to paths like `/usr/local/bin/php55/bin/php`, `/usr/local/bin/php70/bin/php` without editing the $PATH variable.

The best,
Simon

### DIFF
--- a/build/build.properties.dist
+++ b/build/build.properties.dist
@@ -21,6 +21,13 @@ app.dir   = ${basedir}/engine/Shopware
 
 karma.config = ${basedir}/themes/Frontend/Responsive/karma.conf.js
 
+script.php      = php
+script.phpcs    = phpcs
+script.phpmd    = phpmd
+script.phpcpd   = phpcpd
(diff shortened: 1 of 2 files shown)


bcremer:
Looks good to me. This can be scheduled.

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/64124b642c5b1b2ca7ec063ed9301b272ef921b8.

Thank you @redshark1802 :+1:

PR #439, opened 2016-02-18
Update Article.php

martinkouba:
Fix for SW-13812
Update article number in configurator template so that generated variants can use the updated number.


OliverSkroblin:
With this fix, the order number field in the backend configurator template mask would be useless. The fix overwrites each time the provided $configuratorTemplate['number'].

martinkouba:
Do you have any suggestions as to how to fix the bug in another way?
Without this fix when you change the article number and then update the variants the number appears to be arbitrary and creates confusion.

PR #441, opened 2016-02-22
Fix the 'removeListener' method in 'Enlight_Event_Subscriber_Config'

andrewisplinghoff:
This fix uses 'array_search' and 'array_splice' to actually find and remove a given handler from the listeners. This is necessary, because the previously used method 'array_diff' doesn't work with objects.

This is a new version of PR https://github.com/shopware/shopware/pull/174 against the current branch.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,7 @@ In this document you will find a changelog of the important changes related to t
 * Added index on `s_article_img_mapping_rules`.`mapping_id` and `s_article_img_mapping_rules`.`option_id`
 * Fixed `AND` search logic for search terms which not exist in the s_articles table.
 * Added order and payment state constants in `\Shopware\Models\Order\Status`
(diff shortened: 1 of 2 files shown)


bcremer:
Thank you @andrewisplinghoff, can you add a small usage example how to reproduce the broken behavior and to verify the fix?

andrewisplinghoff:
Find a simple example plugin below. To reproduce:
1. Install the plugin
2. Change the version number in `getVersion` to 1.0.1
3. Update the plugin

**Expected behaviour:**
The hook that has been created during the initial installation is removed and document generation will work like normal.

**Actual behaviour before the fix:**
The hook does not get removed and document generation is broken.

https://gist.github.com/andrewisplinghoff/63ec05f9e0a5df04a035

bcremer:
This PR has been replaced by https://github.com/shopware/shopware/pull/451.

PR #443, opened 2016-02-23
SW-14164 - Introduce DetachedShop to prevent shop duplication issues

bcremer:
Alternative fix for https://github.com/shopware/shopware/pull/437.

Issue https://issues.shopware.com/#/issues/SW-14164.

This PR introduces a new `\Shopware\Models\Shop\DetachedShop` object that will be returned for `Shopware()->Shop()` , `$shopRepository->getActiveByRequest()` and similar methods.

 `\Shopware\Models\Shop\DetachedShop`  extends `\Shopware\Models\Shop\Shop` to preserve BC with typehints.

This new object is not managed/attached to Doctrines Entity manager.

### DIFF
--- a/engine/Shopware/Components/Model/ModelRepository.php
+++ b/engine/Shopware/Components/Model/ModelRepository.php
@@ -25,7 +25,6 @@
 namespace Shopware\Components\Model;
 
 use Doctrine\ORM\EntityRepository;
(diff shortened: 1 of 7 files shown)


FiveDigital:
I have tested this fix in a scenario where we were having the shop duplication bug and it seems fixed :+1:

PR #444, opened 2016-02-23
SW-14293 - Fix deletion of first subshop in config module

bcremer:
Fix for https://issues.shopware.com/#/issues/SW-14293
See: https://twitter.com/msslovi0/status/702099567291068416

It is not possible to delete the first subhop / language shop in the shop section of the config module in the shopware backend.

### DIFF
--- a/themes/Backend/ExtJs/backend/config/model/form/shop.js
+++ b/themes/Backend/ExtJs/backend/config/model/form/shop.js
@@ -82,7 +82,7 @@ Ext.define('Shopware.apps.Config.model.form.Shop', {
         { name: 'fallbackId', convert: function(v, record) {
             return v || record.raw && record.raw.fallback && record.raw.fallback.id;
         }, useNull: true },
-        { name: 'deletable', type: 'boolean', convert: function(v, r) { return r.data.id > 2; } }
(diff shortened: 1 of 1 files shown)


dnoegel:
Perhaps this was done for some "s_core_translation" reason? Do other tables reference the shop instead of the localeId?

Actually we have similar logic in themes/Backend/ExtJs/backend/config/model/form/locale.js:38 - in that case for a good reason, I guess.

msslovi0:
Messed up my VM database later today and had to re-setup both subshops. However, this time both can be deleted.

bcremer:
@msslovi0 looking at the sourcecode and the proposed fix this makes sense, since your newly created subshops have IDs > 2 because of the auto_increment value of the database table.

msslovi0:
@bcremer That's indeed the case. I just wondered because I dropped all tables after the messup and did a clean install. 

But here's maybe the reason why the first subshop is on ID 3 instead of 2 now:
I clicked 'Create' and figured that the MultiShop plugin is installed but not activated. So I canceled the creation, activated the plugin and after that created both subshops. Looks like the create/cancel incremented the auto increment. 

So the fix is still needed. Thanks for taking care of this.

PR #445, opened 2016-02-24
Improved subshop inheritance

samuelvogel:
At first, thanks for merging a fix for the shop duplication with #443 /  0cadd8fb63eb28d92dea65c59786c16c196562d3. For plugins vendors and support teams this is a day to celebrate :wink:

We would like to resubmit our proposed solution though, as we still think it's easier and also doesn't break backwards compatibility, continuing the discussion from #437.

@bcremer: The write to the shop entity you mention in https://github.com/shopware/shopware/pull/437#issuecomment-188151805 is not really relevant, we've removed it and more with 18fce6e93cbf9324295359a68cc1f40f84492738. They where only executed in the special case when the property was an empty string and then the value was just set to the one already set before by `Shopware\Models\Shop\Repository::fixActive()` anyways, so it never had an effect.

**tl;dr** We think there is a better way for subshop inheritance which formerly produced the shop duplication and want to discuss it because it could clean up some very important shopware code.


samuelvogel:
@jhit: I don't think you are correct in https://github.com/shopware/shopware/pull/437#issuecomment-188135588. Before a `flush()` on any entity somehow connected/referencing a (unchanged) shop model after doing something that inherently calls `Shopware\Models\Shop\Repository::fixActive()` would trigger the shop duplication which resulted in a broken shop. Now these cases will work as expected (entity persisted, shop not changed). The problem was not with intentional writes to the shop model, these where never done on the detached entity produced by `Shopware\Models\Shop\Repository::fixActive()` because that would have duplicated the shop as well instead of updating it.

So in my opinion it's save to assume that plugins triggering the bug in the past where not intentionally writing the shop model (as it didn't work anyways but instead broke the shop) but instead where `flush()`ing some entity just referencing an actual unchanged shop (this was the case with our plugins and the others we saw producing this bug in the wild).

> The whole point of fixing this issue is to prevent shops getting damaged. If the flush() call will overwrite settings of the store with empty or wrong data, this would be worse than the bug you try to fix here, because restoring the shop would be far more difficult than deleting a ghost-store entry in the configuration.

Could you elaborate why you think that empty shops would be saved? The PR makes subshop models fall back to their parents settings transparently without touching the database (it's all done just on the reading side in the `getXYZ()` methods), where the former `Shopware\Models\Shop\Repository::fixActive()` way detached the shop entity and wrote to it's fields (producing a tainted entity which must never be saved to the database).

bcremer:
@samuelvogel thanks for keeping the discussion open. I will give you are more detailed answer tomorrow. 

> [...] as we still think it's easier and also doesn't break backwards compatibility, continuing the discussion from #437.

How exactly does the current patch break backwards compatibility?

samuelvogel:
Sorry, I did not mean to say that your PR #443 does break BC, I just meant, that this new one shouldn't either.

jhit:
@samuelvogel My argument was just based on my understanding of the comments from you and @bcremer. If there is absolutely no risk of flushing partially empty shop models to the database, I'm happy with either solution.

samuelvogel:
Ok, thanks for the clarification! Empty properties of subshops are perfectly valid and lead to a fallback on the main shop (example from the shopware default dataset): 
screenshot 2016-02-25 09 46 53

bcremer:
@samuelvogel I am going to close this PR. PR #443 is tested and verified and solves the problem for now and will be included in the upcoming 5.1.4 release.

While I can clearly see the benefits of your PR in terms of greatly reduced complexity in some parts of the application it still leads to an attached Shop-Entity that can be accidentally modified by plugins.

For example:
As a plugin I might temporary modify the current shop for testing purposes, sending a mail or whatever.
A plugin or process later in the shopware stack might issue a `Shopware()->Models()->Flush()` so this change gets persisted into the database and leads to an broken shop:

```
Shopware()->Shop()->setBasePath('somePath');
// many many lines later
Shopware()->Models()->Flush()
```

In SW 5.1.3 and with my patch this will just work like before. With the patch provided by you this will persist the basePath.

This is a BC break that may lead to broken shops so it's not safe for us to include.
I was able to reproduce this problems with some plugins out there in the shopware store.

I really appreciate your work on this issue but given above mentioned problems I'll close this issue.

PR #446, opened 2016-02-24
Update PHPDoc Blocks for functions install, update, uninstall

heyjorgedev:
For instance on the install function you can of course do a "return true" to say to shopware that the Plugin was installed successfully or you can return an array like "array('success' => true, 'invalidateCache' => array('frontend'));" for saying that the Plugin was installed successfully but also to ask Shopware to clear the cache.


bcremer:
Thank you very much @jorgemurta.

PR #447, opened 2016-02-25
Fixed W3C Error - div as ul child

iLuHa91:
W3C Error "Element div not allowed as child of element ul" fixed.

### DIFF
--- a/themes/Frontend/Bare/frontend/index/sidebar.tpl
+++ b/themes/Frontend/Bare/frontend/index/sidebar.tpl
@@ -18,14 +18,13 @@
 						</a>
 					</li>
 				{/block}
-
-				{* Switches for currency and language on mobile devices *}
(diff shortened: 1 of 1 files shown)


bcremer:

klarstil:
@iLuHa91 Thanks for your contribution. We have to double check your changes cause it can break the layout of the mobile switches.

@bcremer We should import this as a new ticket.

iLuHa91:
Have fixed this in my custom Theme like that. Nothing is broken ;-) Just for your information.
Would be nice if this would be fixed in core.

Phil23:
Thank you for your work.
Changes will be released with SW 5.1.4

iLuHa91:
I'm missing this change in Shopware 5.2.X and also in Shopware 5.3.X

PR #448, opened 2016-02-25
Fixed W3C Error - empty action

iLuHa91:
W3C Error "Bad value <space> for attribute action on element form: Must be non-empty." fixed.


bcremer:

klarstil:
@bcremer The change makes sense and is a low hanging fruit. Let's create a ticket for it, so we can work on it in the next sprint.

bcremer:
Fixed with commit https://github.com/shopware/shopware/commit/dc18145793c27e8b66407cd8c6ae0bacadb8aa19.
I also removed all other occurrences of empty action attributes.

Thank you @iLuHa91.

PR #449, opened 2016-02-25
[BUGFIX] use plain sql query instead of querybuilder in sAdmin

Tombro:
Hey folks,
although it physically hurts me to replace the querybuilder with a plain sql query, this is a nessecary bugfix.

The problem:
We are adding some custom fields to the billing address attributes.
The registration uses sAdmin functions to validate and persist fields, which themselves all use plain sql queries. Thats why its nessecary to have underscored values for name and value attributes in the template (as the db column names).

Eg: `<input name="register[billing][foo_bar_baz]" value="{$form_data.foo_bar_baz}" type="text" />`

But in the Account settings (/account/billing) which use the same template, the values are not set. This is because the querybuilder returns lowerCamelCased results which do not match the template variables.

For the default attributes this is not a problem, because text1 etc is of course the same lowerCamelCased an underscored.

Regards,
Thomas

**EDIT**: added a new patch which includes same logic for shipping address attributes.

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -2479,21 +2479,17 @@ class sAdmin
     /**
      * Returns the given user's shipping address attributes
      *
-     * @param $userId User id
+     * @param $shipping Shipping Address
      * @return array The given user's shipping address attributes
      */
-    private function getUserShippingAddressAttributes($userId)
+    private function getUserShippingAddressAttributes($shipping)
     {
-        $builder = Shopware()->Models()->createQueryBuilder();
-        $attributes = $builder->select(array('attributes'))
-            ->from('Shopware\Models\Attribute\CustomerShipping', 'attributes')
-            ->innerJoin('attributes.customerShipping', 'shipping')
-            ->where('shipping.customerId = :userId')
-            ->setParameter('userId', $userId)
-            ->setFirstResult(0)
-            ->setMaxResults(1)
-            ->getQuery()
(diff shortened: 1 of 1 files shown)


Tombro:
A little more information:

The name and value attribute of the input field must use the same spelling like
`<input name="register[billing][foo_bar]" value="{$form_data.foo_bar}" />`

The position in the core which requires this is in the Register-Controller in the saveBillingAction() line 306:
`$this->View()->register->billing->form_data = new ArrayObject($this->post['billing'], ArrayObject::ARRAY_AS_PROPS);`

Otherwise the fields would be empty if there are errors in the validation.

Now you could still say that you can use lowerCased values for both, all you need to do is map the lowerCased post values to the underscored column names in the Event listener for `Shopware_Modules_Admin_SaveRegisterBillingAttributes_FilterSql`.

But there is a second position which require that the spelling is underscored. It is in the sAdmin method filterBillingAttributeData() line 4990 which is called by AccountConrollter->saveBillingAction()->sUpdateBilling().

Here you do an sql query to filter the allowed post keys. This query returns underscored values which means that the post values have to be underscored as well.

3stadt:
Hello @Tombro 

we are sorry we are not able to merge your pull request. Currently we are in the process of rewriting the address management as well as the attribute management.
Both will be released in 5.2 and your problem will most likely be fixed by the new overhauled code.

Since your pull request would imply a major break and a fix will be released with 5.2 we will decline it, but keep an eye out for the 5.2 branch and the rewritten address/attribute management which will be introduced in the near future.

Thank you for your good work though, we appreciate it very much!

PR #451, opened 2016-02-28
Fix enlight event subscriber remove listener

samuelvogel:
Currently the `removeListener` method in `Enlight_Event_Subscriber_Config`, `Enlight_Event_Subscriber_Array` and `Enlight_Event_EventManager` use `array_diff()` to remove a given `Enlight_Event_Handler` object from the `listeners` array. The problem is, that `array_diff()` doesn't work with object arrays and instead removed all listeners. This fix uses `array_search()` and `array_splice()` instead, to first find and then remove just the given handler from the listeners.

This is a new version of PR #174 against the current branch and includes more fixes than #441 and also brings tests.

Please merge this fix, it's really a pity that this never worked and the fix was first submitted shortly after `4.3.0` was released.


bcremer:
Thanks @samuelvogel,
I closed #441 and reassigned this pull request to issue https://issues.shopware.com/#/issues/SW-14281.

bcremer:
Included with merge commit https://github.com/shopware/shopware/commit/963a30c028f145565370f96ff0332ff65afa383a.

Thank´s team VIISON :+1:

PR #452, opened 2016-02-28
Add unsubscribeEvent() method for plugins

samuelvogel:
In some plugin updates it is desirable to remove old event subscriptions, since they are no longer used. This commit adds a new method to `Enlight_Plugin_Bootstrap_Config`, from which all plugin bootstrap implementations inherit.

The new `unsubscribeEvent()` method requires only the event name and listener name to remove an existing subscription, just like `subscribeEvent()` requires those two parameters to add a subscription in the first place.

In order to make this work, it is also necessary to adapt the `write()` method of `Shopware_Components_Plugin_Namespace`, so that the removed subscribes are also removed from the database. Therefore it has now an optional parameter of type `Shopware_Components_Plugin_Bootstrap`. If it is present, that plugin's subscribes will be fully synced, meaning that its removed subscribes will be deleted from the database.

Please note, that `unsubscribeEvent()` relies on a properly working `removeListener()` method in `Enlight_Event_Subscriber_Config` which was fixed by #451. Until that is merged, the newly added test will fail!

This is an updated version of #175.


cjuner:
Re-created on top of the 5.2 branch in PR #565. Please close this old PR.

janbuecker:
To discuss, please refer to #565. Closing..

PR #455, opened 2016-03-02
Remove smarty var sCategoryInfo (duplicate of sCategoryContent)

schliesser:
Smarty variable sCategoryInfo is duplicate of sCategoryContent. It's used in listing and blog-listing.
In detail there is also sCategoryInfo, but this one provides other informations!

Since 5.2 is a major update and old emotion templates using this listing smarty var, the duplicate should be removed. It may breaks some plugins depending on sCategoryContent, but I think the update to 5.2 will break many plugins itself.

Maybe I'm wrong using sContentCategory, but searching the code showed this many times more.

Thx André Schließer
zeroseven design studios


bcremer:
This pull request was merged with commit https://github.com/shopware/shopware/commit/76a442f4fab75ff19553727d123f3e3102b8c52f.

Thanks @schliesser :+1:

PR #456, opened 2016-03-02
changed the stepcount to reflect the actual prices

ghost:
Without this fix its currently possible to select for example 18€ as a starting price without it actually starting at 18€. In my example, the maximum price was 38€. With the stepcount of 100, it was possible to visually choose 18€, but it was in fact 17.85€. 

It should not break anything except that the slider now works a bit differently.


bcremer:
@arvatis The proposed fix will introduce usability and performance problems. Increasing the stepcount to the facets maximum value might introduce very high stepcounts that are way to fine grained to be usable in the storefront.

Also the range slides should be fixed globally in the jquery plugin instead of modifing the input parameters: https://github.com/shopware/shopware/blob/5.1/themes/Frontend/Responsive/frontend/_public/src/js/jquery.range-slider.js

We appreciate your contribution. Please feel to reopen a new pull request if you find a fix to the above mentioned problems.

PR #457, opened 2016-03-02
Added opacity polyfills

svenfinke:
The opacity mixin was pretty useless before, as it only included a simple 'opacity: @opacity;'. I included IE fallbacks.


bcremer:
I don't think this is needed.
Every supported browser understands the native `opacity` keyword without vendor prefix or hacks as far as I now.
See: http://caniuse.com/#search=opacity

/cc @shopware/shopware-webdev

Phil23:
Yes, @bcremer is right, the opacity property is supported in all current browsers.

This PR can be closed.

PR #458, opened 2016-03-03
Add support for category-parameter on ajax-calls.

giginos:
Actually the ajax-handling for variants does not support the optional category-parameter (?c=32), which is passed in the url. This pull-request extends the ajax-call to pass this parameter, too.

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.ajax-variant.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.ajax-variant.js
@@ -74,6 +74,8 @@
             var me = this,
                 location;
 
+            var categoryid = this.getUrlParameter('c');
+
             // `location.origin` isn't available in IE 11, so we have to create it
             if (!window.location.origin) {
                 window.location.origin = window.location.protocol + "//" + window.location.hostname + (window.location.port ? ':' + window.location.port: '');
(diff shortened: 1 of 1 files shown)


dnoegel:
@giginos: Thanks - will discuss it tomorrow with the others :)

themes/Frontend/Responsive/frontend/_public/src/js/jquery.ajax-variant.js
```diff
@@ -186,6 +193,21 @@
             $.publish('plugin/swAjaxVariant/onChange', [ me, values, $target ]);
 
             me.requestData(values, true);
+        },
+        // Private function to get existing url params
+        getUrlParameter: function getUrlParameter(sParam) {
```
bcremer, line 198: @shopware/shopware-webdev is the function safe? Do we have a similar function already in scope?

klarstil, line 198: @bcremer We're using a similar functionality for the filter but the function is marked as private and can't be used by other plugins.

We usually using anonymous functions and just apply them to the configuration object of the plugin. I would like to have this consistent across all of our default jQuery plugins.

klarstil:
@giginos I was working on the ajax variants for the latest shopware release cause we had a small issue with the history management and fixed your issue as well.

You can see the changes here: https://github.com/shopware/shopware/commit/6dcc838b81c0f96b4a996b39a01762c96f28fce3

PR #460, opened 2016-03-04
Fix 'from' filter to include orders created at 00:00:00

andrewisplinghoff:
If an order has been created at midnight (e.g. exactly '2016-02-01 00:00:00'), it should be included in the order list dialog when setting the 'from' filter to the respective date ('2016-02-01' in the example). This was previously not the case because of the use of the '>' operator instead of '>='.


bcremer:
Merged your PR with commit https://github.com/shopware/shopware/commit/b76fa590e8d5698642d7a5960f602b3325f4526d.

Thanks @andrewisplinghoff :+1:

PR #462, opened 2016-03-04
added the missing private $requestHandlers

jochenmanz:
Small PR, adding a missing class property


bcremer:
Thank you @jochenmanz :+1:

PR #463, opened 2016-03-05
convertMediaStruct() retinaSource null value

plugware:
The convertMediaStruct function doesn't include the single retinaSource value properly into thumbnails array.


janbuecker:
Merged with commit 158c8745c1eb8d7641b0b08b4f93e42099754d53.

Thank you, @plugware.

PR #466, opened 2016-03-08
Make Symfony Request available from Container

shyim:
The Symfony Request has more functions like getting files as the Enlight Request and we should place it in the container.

### DIFF
--- a/engine/Shopware/Kernel.php
+++ b/engine/Shopware/Kernel.php
@@ -145,6 +145,7 @@ class Kernel implements HttpKernelInterface
 
         /** @var $front \Enlight_Controller_Front **/
         $front = $this->container->get('front');
+        $this->container->set('request', $request);
 
         $request = $this->transformSymfonyRequestToEnlightRequest($request);
 


bcremer:
I don't think that is a good idea to put the Symfony request object directly in the container.

Putting the request object in the container has some serious problems thats why the Symfony project introduced the Request Stack in Symfony 2.4 and removed the request from container in Symfony 3.
Reasons for that can be found in their blog post: [New in Symfony 2.4: The Request Stack](http://symfony.com/blog/new-in-symfony-2-4-the-request-stack).
Also see [Accessing the Request in a Service](http://symfony.com/doc/current/book/service_container.html#accessing-the-request-in-a-service) in the Symfony Documentation.

In Shopware the current request can be accessed using the frontcontroller `Shopware()->Front()->Request();`

To address your original intent, accessing symfony request object inside the shopware scope, I am also not sure if we should allow that.
It may confuse developers to have two kinds of request objects. Also changes to one of the objects will not be reflected into the other one.

To solve your problem in a more pragmatic way you can create a new Symfony Request object just where you need it from the Superglobals like:

```
$request = Symfony\Component\HttpFoundation\Request::createFromGlobals();
```

Please be aware that this might not work when caching or testing is involved since it depends on the Shopware Request objects behavior to modify the superglobals.

PR #467, opened 2016-03-08
made sAdmin::loginUser public

kirtixs:
Currently it is not possible, or very complicated, to overwrite `sAdmin::sLogin` and the checks in it, this PR fixes this by making `sAdmin::loginUser` public.

With this fix it is possible to do the following:
1. replace `sAdmin::sLogin` using `$this->subscribeEvent('sAdmin::sLogin::replace',           'replaceSLogin');`
2. implement custom logic to verify that the user has entered the right credentials in `replaceSLogin`
3. trigger `sAdmin::loginUser` using `$arguments->getSubject()->executeParent(              'loginUser', $getUser, $email, $password, $isPreHashed, $encoderName, $plaintext, $hash);`

This wont be a BC break.

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -4129,7 +4129,7 @@ class sAdmin
      * @param $plaintext
      * @param $hash
      */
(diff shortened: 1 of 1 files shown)


janbuecker:
What is your exact intent with this change? Can you describe it a little bit further?

PR #468, opened 2016-03-10
Hide compare entry after onDeleteCompare

PatrickBauer:
I'd like to propose that after deleting a comparison, the menu item should be re-hid again. 
Inside jquery.product-compare-add.js inside onAddArticleCompare() the class is removed, if a new compare item is added. Hiding it after deleting the comparison would be consistent action to do. 

```
if (compareMenu.hasClass(me.opts.hiddenCls)) {
    compareMenu.removeClass(me.opts.hiddenCls);
}
```

I'm using custom styling on my top-bar elements and after deleting the comparison, the element stays visible. Thats how I noticed the missing feature.


janbuecker:
Thanks, merged with 9d06a041e28f25885f6e742afb0dc9e3fb2953f1.

PR #469, opened 2016-03-11
Add {block} tags for all .js files in the PluginManager

edwint88:
I'd like to propose that the Plugin Manager to have the {block} tags in order to be extended by other plugins.

I've tried to override a backend plugin and I encountered this problem. More: 
[http://forum.shopware.com/discussion/35522/backend-plugins-manager-cannot-be-extended#latest](url)

### DIFF
--- a/engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/app.js
+++ b/engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/app.js
@@ -1,4 +1,33 @@
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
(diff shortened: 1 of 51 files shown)


bcremer:
Merged this PR with commit https://github.com/shopware/shopware/commit/e06524e892f24de697afb4d64c9e5469ba69bbd4.

Thank you very much @edwint88, good work :+1:

dnoegel:
@edwint88 Great, thank you! I just realized, that the "blocks" for the models "fields" are still missing, so you won't be able to extend those. For comparison see here: https://github.com/shopware/shopware/blob/5.1/themes/Backend/ExtJs/backend/article/model/article.js#L50

PR #470, opened 2016-03-11
Use sPerPage instead of sPage in LegacySearchSubscriber request values

hashworks:
This was an insane expenditure to create this pull request in comparisation to the change. Hell, with all the travis and Scrutinizer stuff going on now I bet this wasted a huge amount of brown coal. 

@bcremer, @SimonBaeumer actually you two are now offically responsible for the climate Change.


janbuecker:
Good catch :D I scheduled this PR.

janbuecker:
This file is not available in 5.2 any more. Closing this PR.

PR #471, opened 2016-03-11
Fix createSnippetAction

AlexanderSchikowsky:
If someone edits an existing Snippet, the createSnippetAction creates empty snippets for each subshop which is not filled with content.

This causes that the subshop snippets will not Inherit content from their parents.

We fixed this with a simple if condition which checks if the snippet value is empty.


janbuecker:
This PR was closed due to outdated code changes. This issue may have already been resolved in the past.

PR #474, opened 2016-03-16
fixing Smarty fetch error for reset password

yesitsfebreeze:
There was a Smarty fetch error when you tried to display the account reset form with an error message.

The error occurs when you leave the reset input field empty or with false data.
https://adamhall.dev/shop/de-de/account/resetPassword


janbuecker:
Hi @hvlmnns and thanks for your PR.

We've refactored the registration and account area and outsourced some validation rules into classes. I've created a ticket, which will make use of the new validations in here.

You can track the progress here https://issues.shopware.com/#/issues/SW-15109

I'm closing this PR in favour of new ticket.

Edit by @mitelg fixed link :wink:

PR #478, opened 2016-03-18
PHP Warning on Component createField() method

heyjorgedev:
Me and my work colleges we were having an issue with ansible installing a shopping worlds plugin.

The problem is that the PHP function max() raises a warning when you pass as an argument an empty array, I'm aware that Shopware is expecting it because of this line:
- $this->maxFieldPositionValue = max($positions) ? : 0;
  but the problem is that PHP Warnings are unnecessary. Our suggestion  is not raising warnings.

Thanks

### DIFF
--- a/engine/Shopware/Models/Emotion/Library/Component.php
+++ b/engine/Shopware/Models/Emotion/Library/Component.php
@@ -715,7 +715,8 @@ class Component extends ModelEntity
                 function ($field) {return $field->getPosition();},
                 $this->getFields()->toArray()
             );
-            $this->maxFieldPositionValue = max($positions) ? : 0;
+            
+            $this->maxFieldPositionValue = !empty($positions) ? max($positions) : 0;
         }
 
         return $this->maxFieldPositionValue;


engine/Shopware/Models/Emotion/Library/Component.php
```diff
@@ -715,7 +715,8 @@ public function getMaxPositionValue()
                 function ($field) {return $field->getPosition();},
                 $this->getFields()->toArray()
             );
-            $this->maxFieldPositionValue = max($positions) ? : 0;
```
heyjorgedev, line 718: This lines gives a PHP Warning

engine/Shopware/Models/Emotion/Library/Component.php
```diff
@@ -715,7 +715,8 @@ public function getMaxPositionValue()
...
+            
+            $this->maxFieldPositionValue = !empty($positions) ? max($positions) : 0;
```
heyjorgedev, line 719: This one will suppress the PHP Warning by Validating the "Empty" array

heyjorgedev:
@bcremer

dnoegel:
Hi @jorgemurta,

thanks for the contribution, for me it looks good. The GitHub tickets are also part of the core development's sprints, so it might take a short while sometimes. If @bcremer didn't nit-pick, yet, it is usually a good sign :)

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/3536b3f3c9cfd29727d689098066ca84648e0ab0.

PR #479, opened 2016-03-18
adding password field-type to plugin configuration

teiling88:
With this small extension - it is possible to use the config type password instead of text for passwords which are needed for APIs etc.

Here a example of an installation() Menthod:

```
    public function install()
    {
        $form = $this->Form();
        $form->setElement(
            'text',
            'username',
            array(
                'label' => 'Username',
                'value' => null
            )
        );

        $form->setElement(
            'password',
            'text',
            array(
                'label' => 'Hidden Password',
                'value' => null
            )
        );

        return true;
    }
```

### DIFF
--- a/themes/Backend/ExtJs/backend/base/component/Shopware.form.PluginPanel.js
+++ b/themes/Backend/ExtJs/backend/base/component/Shopware.form.PluginPanel.js
@@ -280,6 +280,10 @@ Ext.define('Shopware.form.PluginPanel',
                     field.value = new Date(field.value);
                     field.value = new Date((field.value.getTime() + (field.value.getTimezoneOffset() * 60 * 1000)));
                 }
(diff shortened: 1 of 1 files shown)


shyim:
Its already possible, you can define the type in the options :)
Example:

```
$form->setElement(
            'text',
            'password',
            array(
                'label' => 'Password',
                'value' => null,
                'inputType' => 'password'
            )
        );

```

teiling88:
Ups - good solution :+1:

bcremer:
Thank you @shyim for pointing that out #communityworks :+1:

bcremer:
@teiling88 @shyim 

For you information. In the upcoming plugin config, password fields will be supported as first-class citizens 

``` xml
<element type="password">
    <name>password</name>
    <label>Your Password</label>
</element>
```

:+1:

PR #482, opened 2016-03-21
Double-Check ordernumbers for uniqueness

hashworks:
A customer told me they experienced two orders with the same ordernumber. I guess this happens because of some racing stuff. Someone on IRC told me that this happens quite often, so here is a pull request.

If you have a better idea to solve this, feel free to tell me.

Note that this won't work if users decide to prefix / suffix the ordernumber in `Shopware_Modules_Order_GetOrdernumber_FilterOrdernumber`. We can't use `%:number%` here through (for obvious reasons). Basically this should be checked _after_ the event call, but if it isn't unique we would need to call the event again, which isn't good as well, right?

``` php
 $doCount = 1;
do {
    $number = $this->db->fetchOne(
        "/*NO LIMIT*/ SELECT number FROM s_order_number WHERE name='invoice' FOR UPDATE"
    );

    $this->db->executeUpdate(
        "UPDATE s_order_number SET number = number + 1 WHERE name='invoice'"
    );
    $number += 1;

    $number = $this->eventManager->filter(
        'Shopware_Modules_Order_GetOrdernumber_FilterOrdernumber',
        $number,
        array('subject'=>$this)
    );

    $doubleTap = $this->db->select()
        ->from('s_order', 'ordernumber')
        ->where('ordernumber = :number')
        ->limit(1);
    $doubleTap = $this->db->executeQuery($doubleTap, ['number' => $number]);

    if ($doubleTap->rowCount() == 0 || $doCount > 3) {
        break;
    }

    $doCount++;
} while(true);

return $number;
```

### DIFF
--- a/engine/Shopware/Core/sOrder.php
+++ b/engine/Shopware/Core/sOrder.php
@@ -224,10 +224,24 @@ class sOrder
         $number = $this->db->fetchOne(
             "/*NO LIMIT*/ SELECT number FROM s_order_number WHERE name='invoice' FOR UPDATE"
         );
-        $this->db->executeUpdate(
-            "UPDATE s_order_number SET number = number + 1 WHERE name='invoice'"
-        );
-        $number += 1;
+
+        $doCount = 1;
+        do {
+            $this->db->executeUpdate(
+                "UPDATE s_order_number SET number = number + 1 WHERE name='invoice'"
+            );
+            $number += 1;
+
+            $doubleTap = $this->db()->select()
+                ->from('s_order', 'ordernumber')
+                ->where('ordernumber = :number')
+                ->limit(1);
+            $doubleTap = $this->db()->executeQuery($doubleTap, ['number' => $number]);
+            if ($doubleTap->rowCount() == 0 || $doCount > 3) {
+                break;
+            }
+            $doCount++;
+        } while(true);
 
         $number = $this->eventManager->filter(
             'Shopware_Modules_Order_GetOrdernumber_FilterOrdernumber',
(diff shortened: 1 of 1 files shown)


bcremer:
@hashworks Thanks for your PR :+1: 

This issue is already tracked with ticket https://issues.shopware.com/#/issues/SW-13498 (it's also assigned for the current sprint).
In this ticket the following solution is proposed: 

``` php
try {
    $number = $this->db->fetchOne(
        "/*NO LIMIT*/ SELECT number FROM s_order_number WHERE name='invoice' FOR UPDATE"
    );
    $this->db->executeUpdate(
        "UPDATE s_order_number SET number = number + 1 WHERE name='invoice'"
    );
    $number += 1;
    $this->db->commit();
} catch (Exception $e) {
    $this->db->rollBack();
    throw $e;
}
```

This is due to a change in the `FOR UPDATE` keyword as of MySQL 5.5

> Locking of rows for update using SELECT FOR UPDATE only applies when autocommit is disabled (either by beginning transaction with START TRANSACTION or by setting autocommit to 0. If autocommit is enabled, the rows matching the specification are not locked.
> Source: http://dev.mysql.com/doc/refman/5.7/en/innodb-locking-reads.html

Can you please verify if that also fixes the issue for you. Thank you very much.

hashworks:
Ah, that solution seems way cleaner. Haven't thought of that, I should read it up.
I'll implement that and give feedback if I get into any problems.

bcremer:
Fixed that with commit https://github.com/shopware/shopware/commit/b8a05549f4a75c791c17c1b8b1e7adb429b5e73a.

PR #483, opened 2016-03-22
SW-14601 - close detail window after blog article is saved

uehler:
If you added content to the attribute fields of a new blog entry and save it twice without closing the detail window the connection between the attribute fields and the blog entry gets lost. The blog_id in s_blog_attributes will be null.

This commit closes the detail window of the blog entry when you save it to get rid of this behaviour


uehler:
i did something wrong
i will try it again

PR #484, opened 2016-03-23
Add order detail window to footer buttons

shyim:
Its very annoying, to search every time where the order detail is =)
We should add the window to the footer buttons (footer window dinger(™) @t2oh4e)


Phil23:
Thanks @shyim, changes are merged for next release.

PR #487, opened 2016-03-29
Added smarty block to Shopware.apps.Config.model.CountryAttribute

FiveDigital:
This block is missing and needed for adding new attributes to the country model.

### DIFF
--- a/themes/Backend/ExtJs/backend/config/model/form/country.js
+++ b/themes/Backend/ExtJs/backend/config/model/form/country.js
@@ -71,6 +71,7 @@ Ext.define('Shopware.apps.Config.model.CountryAttribute', {
     extend: 'Ext.data.Model',
 
     fields : [
+        //{block name="backend/config/model/form/country_attribute/fields"}{/block}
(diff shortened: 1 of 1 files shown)


shyim:
I had the same problem last week.
The Attributes in Config.Model are useless, because he uses the attributes from Base.Model.
https://github.com/FiveDigital/shopware/blob/feature/countryAttributes/themes/Backend/ExtJs/backend/config/model/form/country.js#L62
:)

FiveDigital:
I found that issue, too. I tried to fix it but wasn't successful. So I think the whole country attribute thing is broken and will hopefully be fixed with the new attributes feature in 5.2 that has been announced quite a few times.

janbuecker:
Hi @FiveDigital and thanks for your PR.

With Shopware 5.2, we've removed the attribute models in the backend due to the new attribute management.

FiveDigital:
We will check the new attribute system when upgrading to 5.2 in our customer's project. If it doesn't fit our needs I will create a new PR for 5.2

PR #489, opened 2016-03-29
add block definition for Shopware.component.Preloader

sensus-media:
We added a block definition for Shopware.component.Preloader.js. We need to modify / override the default preloaders behavior for plugin development purposes.


janbuecker:
Thanks for your PR, @sensus-media.

Merged with 9fb859a1d7e7e0276da67111bde5a1239441b465.

PR #490, opened 2016-03-30
Update composer.json

SvenRtbg:
~v2.0.0-rc1 is an illegal version constraint.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         "behat/mink-extension": "~2.1",
         "behat/mink-goutte-driver": "~1.2",
         "behat/mink-selenium2-driver": "~1.3",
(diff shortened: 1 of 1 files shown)


andrewisplinghoff:
I had this issue today, too, but it seems to only affect older composer versions. With the newest composer version, it is not an issue, but with the version from apt-get (Ubuntu Wily), it is.

bcremer:
Thank you very much @SvenRtbg, seems like this caused some confusion in the composer world. See: https://twitter.com/b_viguier/status/715108800290820096

Merged this commit with https://github.com/shopware/shopware/commit/10ef3c21c44df8f8ad44a1cc61f09d77765d7e26.

PR #491, opened 2016-03-31
Property doesn't get found if ID is not integer

Gugiman:
In some cases $value is array but $value["id"] is string.
So the propertyValueRepository can't find the correct property.

In my spedific case I wanted to change a property value of an article but after saving the value falls back to the previous. After this little changes everything works just fine.


Gugiman:
fixed otherwise

PR #492, opened 2016-04-02
Fix first-line issue when plugin JS files get merged

ckilb:
This will fix an issue when several plugin javascript files are merged into a single compiled template file by Smarty.

Because the last line of the app.js is //{/block}, the first two characters for starting a comment will be in front of the first line of the next plugin JS file. Because comments will get removed afterwards, the first line of a plugin JS file will be completely removed.
So currently plugin developers always need to add a useless first line into their javascript files (most likely a comment) to make the plugin work.
This commit will fix this by adding a new line character between the merged files.


dnoegel:
Hi,

thanks, @ckdot, that's a nasty bug!

bcremer:
This was merged with https://github.com/shopware/shopware/commit/91fb872da14.

PR #493, opened 2016-04-04
[FEATURE] Add possibility to load services.xml from plugins

Tombro:
Hey folks,
with this patch it is possible to register a services.xml file directly from the plugin directory.
The solution with services_local.xml which is already implemented will not work for any plugins which should be in the store.

Regards,
Thomas


FiveDigital:
:+1: And please add the possibility to add a custom Compiler Pass, too!

bcremer:
Thanks for your contribution.

We a currently working on exactly that problem. 

Your implementation is a bit too naiv. For example it does not take into account if a plugin is enabled.
We will release some concepts for a new plugin system in the following days so stay tuned.

I am going to close this PR for now because it's not in a mergeable state.

PR #494, opened 2016-04-05
Always display document number when available

samuelvogel:
Currently showing the document number (column `docID` in `s_order_documents`) in the order documents tab in the backend is specifically disabled for document type `4` (correction of invoice). Shopware doesn't even generate it by default for this type, but a plugin like [Pickware](http://www.pickware.de) might want to.
With this PR it is simply displayed when it's present, which doesn't change Shopwares default behavior but allows to generate/display it if wanted.

This reopens #181.


janbuecker:
Thanks, merged with 0205c5f008186046e71c92a7ac72816b368dffbb.

PR #495, opened 2016-04-05
Update sExport.php

freedmo:
Change following line '$baseprice' to '$baseprice as baseprice,' which fix problems when you want access to baseprice in feeds.
Without it is baseprice register in SQL-Selecet as '$baseprice = "IFNULL(p2.baseprice, p.baseprice)";', in case when you don't use the default customergroup('EK').


janbuecker:
hi @freedmo and thanks for your PR.

The `baseprice` has been removed in Shopware 5.2. Therefore, this PR is no longer necessary.

PR #496, opened 2016-04-06
Update icon-set.less - fix styletile docs, rename progress-icon

screeny05:
The Styletile Docs have been updated to correctly indicate the names of certain icons.

Also the icon with the name `icon--brogress-1` has been renamed to `icon--progress-1` while preserving a copy of `brogress` for backwards-compatibility.


janbuecker:
Good catch, thanks! Merged with c3fabf9bb548a778169f3429b23d5a8703f22954.

PR #498, opened 2016-04-06
Article Statistic: Calculate only Order Items with modus 0

shyim:
When a voucher has the same id as a article, it will be shown in the article statistics. That makes the revenue sum wrong. This pr should fix the problem


janbuecker:
Hi @shyim and thanks for your PR.

This issue has been addressed in a recently created PR (#555) which fixed another occurrence too.

I'm closing your PR in favour of the more complete one.

PR #499, opened 2016-04-06
Update jquery.state-manager.js to throw informative error

screeny05:
Checks whether a plugin which should be added actually exists.

Currently when a plugin is missing the javascript-execution halts with a cryptic:
`Uncaught TypeError: element[pluginName] is not a function` (in chrome) which doesn't say what plugin is missing. It actually reads this way.

This way there's also an error thrown which halts javascript-execution, but it reads like this:
`Uncaught Error: Plugin <pluginName> is not a valid jQuery-plugin!` Where `<pluginName>` is the name of the plugin missing.

This will help developers to better debug what plugin is missing.


janbuecker:
Thanks for your PR. I changed it to soft-fail on error. So the change has been merged with 96afc383e3a74eb7a785bc8cae1f944b1648c053.

PR #500, opened 2016-04-06
Add Event "Shopware_Modules_Admin_regenerateSessionId_Start"

shyim:
I am currently buildung a Plugin, which replaces the Zend_Session and uses the Session Class from Symfony HttpFoundation. But the regenerateSessionId function breaks my sessionId from Symfony Session. This PR adds a new NotifyUntil Event, which allows to cancel this action.


DennisGarding:
Hello,

thanks for your contribution!

The branch is merged.

PR #501, opened 2016-04-07
fix wrong doctrine column definition of article id field

ghost:
the id column of article table is defined as signed in doctrine model.

this is not correct because this column is created as unsigned (see: https://github.com/shopware/shopware/blob/5.2/_sql/install/latest.sql#L30)

### DIFF
--- a/engine/Shopware/Models/Article/Article.php
+++ b/engine/Shopware/Models/Article/Article.php
@@ -45,7 +45,7 @@ class Article extends ModelEntity
     /**
      * @var integer $id
      *
-     * @ORM\Column(name="id", type="integer", nullable=false)
+     * @ORM\Column(name="id", type="integer", nullable=false, options={"unsigned"=true})
      * @ORM\Id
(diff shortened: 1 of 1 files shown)


janbuecker:
Hi @bestit-tkellner and thanks for your PR.

Do you have any issues without this change or is it just cosmetic?

ghost:
Hi @janbuecker,
yes there is an issue. 

If you create a custom model which contains a field, which is related to the id field of the article model, the created table has a wrong table definition. This cause problems with foreign keys.

Example:

```
<?php
/**
 * Class Example
 * @ORM\Entity
 * @ORM\Table(
 *              name="test"
 * )
 */
class ExampleModel extends ModelEntity
{
    /**
     * @ORM\OneToOne(targetEntity="Shopware\Models\Article\Article")
     * @ORM\JoinColumn(name="related_article_id", referencedColumnName="id")
     */
    private $relatedArticleId;
}
```

janbuecker:
Thanks @bestit-tkellner.

PR #505, opened 2016-04-12
Sort manufacturers from elasticsearch

xrogers:
Thanks to this pull request display of the manufacturers facet with ES backend should be consistent with the DBAL one.


bcremer:
Good catch. I've scheduled your PR to be included in one of the next sprints.

bcremer:
Thank you @xrogers, this was merged with commit https://github.com/shopware/shopware/commit/778b8f00211.

PR #506, opened 2016-04-12
Add orderDetailsId to Event Shopware_Modules_Order_SaveOrderAttribute…

shyim:
orderDetailsId would be a very helpful unique identifier for a basketItem


janbuecker:
Hi @shyim, thanks for your PR. Looks good to me, should be included in the next version.

Cheers!

OliverSkroblin:
Hi @shyim, we refactored the attribute system with 5.2 and also the store front implementation in sSaveOrder.

See: https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Core/sOrder.php#L718

Order attributes are now saved over the data persister of the attribute bundle, so i close the pull request.

PR #507, opened 2016-04-13
Update Border Color of Fancy-Select Trigger

screeny05:
The trigger has to have the same border-color as the fancy-select itself. In order to prevent different colors when changing `@btn-default-border-color`.


janbuecker:
Thanks @screeny05 

We're planning to remove this whole fancy select thing with Shopware 5.3. We've already merged the changes into 5.3.

PR #508, opened 2016-04-13
Added a Partner Filter to Backend Order View

shyim:


engine/Shopware/Controllers/Backend/Order.php
```diff
@@ -1156,6 +1156,15 @@ public function openPdfAction()
     }
 
     /**
+     * Returns filterable partners
+     */
+    public function getPartnersAction()
+    {
+        $this->View()->success = true;
+        $this->View()->data = Shopware()->Db()->fetchAll('SELECT IFNULL((SELECT company FROM s_emarketing_partner WHERE idcode = partnerID), partnerID) as name, partnerID as `value` FROM s_order WHERE partnerID IS NOT NULL AND partnerID != "" GROUP BY partnerID');
```
janbuecker: Could rewrite this to `Shopware()->Container()->get('dbal_connection')` please? I know we're using this everywhere else too but we don't want to in the future.

mitelg: edit to the above: you can use the short hand `$this->get('dbal_connection')` :wink:

janbuecker:
Thanks, merged with 7bc7ce6b8928f5ea72c940de680619da263f331a.

PR #509, opened 2016-04-13
SwagUpdate: Throw a Exception, if curl is missing

shyim:
We need a Exception to display a error on extjs =)

### DIFF
--- a/engine/Shopware/Plugins/Default/Backend/SwagUpdate/Components/Download.php
+++ b/engine/Shopware/Plugins/Default/Backend/SwagUpdate/Components/Download.php
@@ -156,7 +156,7 @@ class Download
 
         $range = $size . '-' . $totalSize;
 
-        if(!function_exists('curl_init')) {
+        if (!function_exists('curl_init')) {
             throw new \Exception('PHP Extension "curl" is required to download a file');
         }
 
--- a/engine/Shopware/Plugins/Default/Backend/SwagUpdate/Components/Download.php
+++ b/engine/Shopware/Plugins/Default/Backend/SwagUpdate/Components/Download.php
@@ -156,6 +156,10 @@ class Download
 
         $range = $size . '-' . $totalSize;
 
+        if(!function_exists('curl_init')) {
+            throw new \Exception('PHP Extension "curl" is required to download a file');
+        }
+
         // Configuration of curl
         $ch = curl_init();
 
--- a/engine/Shopware/Plugins/Default/Backend/SwagUpdate/Views/backend/swag_update/controller/progress.js
+++ b/engine/Shopware/Plugins/Default/Backend/SwagUpdate/Views/backend/swag_update/controller/progress.js
@@ -144,7 +144,8 @@ Ext.define('Shopware.apps.SwagUpdate.controller.Progress', {
 
                 try {
                     json = JSON.parse(response.responseText)
-		    Shopware.Msg.createStickyGrowlMessage({
+                    
+                    Shopware.Msg.createStickyGrowlMessage({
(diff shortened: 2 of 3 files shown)


engine/Shopware/Plugins/Default/Backend/SwagUpdate/Components/Download.php
```diff
@@ -156,6 +156,10 @@ public function downloadFile($sourceUri, $destinationUri, $totalSize, $hash)
 
         $range = $size . '-' . $totalSize;
 
+        if(!function_exists('curl_init')) {
```
bcremer: Please add a space `if(!function_exists('curl_init')) {`

bcremer:
I am not sure if we should merge this.

The `curl` extension is mentioned and checked multiple times as required dependency.
- https://github.com/shopware/shopware#shopware-server-requirements
- https://github.com/shopware/shopware/blob/v5.1.5/engine/Shopware/Components/Check/Data/System.xml#L38
- https://github.com/shopware/shopware/blob/5.1/composer.json#L19

We usually don't verify required dependencies during runtime. 

I'll schedule this ticket anyway for further discussion.

shyim:
It's checked as requirement, but shopware is installable without curl, because the error flag in the system.XML is missing (diff with zip extension). 
I guess we should amend the error flag in system.xml to solve this.

janbuecker:
This exception here is useless because you'll receive an error message regarding the batch size. https://github.com/shyim/shopware/blob/ffb8a0fe125eb9b46cf00806f9144f04e2809700/engine/Shopware/Plugins/Default/Backend/SwagUpdate/Views/backend/swag_update/controller/progress.js#L143-146

shyim:
Done

engine/Shopware/Plugins/Default/Backend/SwagUpdate/Views/backend/swag_update/controller/progress.js
```diff
@@ -140,10 +140,25 @@ Ext.define('Shopware.apps.SwagUpdate.controller.Progress', {
...
-                });
+                var json = {};
+
+                try {
+                    json = JSON.parse(response.responseText)
+                } catch (err) {
+
+                }
+
+                if (typeof json.message != 'undefined') {
```
janbuecker: i think it would be more readable if you put the given error message into the try {} block and the default into the catch. there is no need for another if-condition here

engine/Shopware/Plugins/Default/Backend/SwagUpdate/Views/backend/swag_update/controller/progress.js
```diff
@@ -140,10 +140,20 @@ Ext.define('Shopware.apps.SwagUpdate.controller.Progress', {
...
             failure: function (response) {
-                Shopware.Msg.createStickyGrowlMessage({
-                    title: '{s name=progress/timeOutTitle}An error occured{/s}',
-                    text: "{s name=progress/timeOut}The server could not handle the request. Please choose a smaller batch size.{/s}"
-                });
+                var json = {};
+
+                try {
+                    json = JSON.parse(response.responseText)
+		    Shopware.Msg.createStickyGrowlMessage({
```
janbuecker: please correct the line indent

engine/Shopware/Plugins/Default/Backend/SwagUpdate/Components/Download.php
janbuecker: still unchanged @shyim

janbuecker:
Merged with 049f1f206a71a518c72d1932dfd2cab3a8bd4595.

PR #510, opened 2016-04-14
fix wrong database column type of paymentInstance->createdAt

larsbo:
changed doctrine annotation database column type of created_at to datetime to match the model type.

### DIFF
--- a/engine/Shopware/Models/Payment/PaymentInstance.php
+++ b/engine/Shopware/Models/Payment/PaymentInstance.php
@@ -156,7 +156,7 @@ class PaymentInstance extends ModelEntity
     /**
      * @var \DateTime $createdAt
      *
-     * @ORM\Column(name="created_at", type="date", nullable=false)
+     * @ORM\Column(name="created_at", type="datetime", nullable=false)
      */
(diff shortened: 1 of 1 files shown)


larsbo:
I'm not sure why the test failed. Looks like it still uses `date` instead of `datetime`. Can someone look into this please? :smiley: 
`[exec] PHP Fatal error:  Uncaught exception 'Doctrine\DBAL\Types\ConversionException' with message 'Could not convert database value "2016-04-14" to Doctrine Type datetime. Expected format: Y-m-d H:i:s' in /home/travis/build/shopware/shopware/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ConversionException.php:63`

dnoegel:
Hi,

the database tables are not created from the doctrine schema but from an SQL base image + the migrations. For this kind of change you probably need to add a delta, which also changes the field type in the table s_core_payment_instance.

janbuecker:
Thanks @larsbo.

We decided to close this PR because the change is incomplete. If this change is still important to you, feel free to create a new pull request or create an issue on https://issues.shopware.com.

PR #512, opened 2016-04-15
Fix re-persisting all emotion elements when saving via backend

buddhaCode:
I'm currently working on a plugin for a customer, where i need to link elements placed on an emotion canvas and save this mapping to the database. This is so save the connection between an original emotion (element) and the corresponding translated emotion (element). A while ago i realized, that the mapping suddenly gets out of sync, because the IDs of the element entities somehow changed. Reading through the code, i found out, that every time an emotion is save via the backend, all element and their data associations will be purged and re-persisted. Obviously, this will result in a new ID and my element mapping will break.

I was lucky to see, that the API call to save the elements has their old IDs with it. So i re-wrote the saveAction to update existing element entities, persist new ones and delete the removed ones. The data entities will still get re-persisted with a new id, because there is no ID with them.

I would love to see the changes in the core, so i don't need the replace hook in my plugin.

### DIFF
--- a/engine/Shopware/Controllers/Backend/Emotion.php
+++ b/engine/Shopware/Controllers/Backend/Emotion.php
@@ -460,7 +460,7 @@ class Shopware_Controllers_Backend_Emotion extends Shopware_Controllers_Backend_
             }
         }
 
-        return $emotion->getElements()->toArray();
+        return $emotion->getElements() ?: array();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


reConNico:
+1

janbuecker:
Hi @buddhaCode and thanks for your PR.

This behaviour has changed in Shopware 5.2 to exactly what you need. Since we are releasing Shopware 5.2 in some weeks, there is no need to back-port this behaviour to 5.1.

buddhaCode:
Great. Love the new code!

PR #514, opened 2016-04-15
[BUGFIX] add event to filter shippingaddress attributes params

Tombro:
Hey folks,
this event is missing if you want to save custom shipping attributes. There is an event to modify the sql, but not to modify the params. Now there is.

Regards,
Thomas


janbuecker:
Hi @Tombro and thanks for your PR.

This method will be removed in Shopware 5.2. You now can pass all custom attributes to the new address service using the entity directly.

I will close this PR.

PR #517, opened 2016-04-18
Add service for retrieving next number from range (s_order_number)

svenmuennich:
Previously there existed three different ways to get the next number of a number range (`s_order_number`) from the database and save it back. However, two of them were not safe and might resulted in using the same number twice due to a lack of locking the `s_order_number` table. This PR adds a new service `shopware.number_range_manager`, which can be used to safely retrieve the current and next number of a range by just passing its name. Furthermore it makes use of that service in three places, which previously used a separate (potentially unsafe) implementation to get the next number:
- Customer creation (via shop frontend)
- Order creation
- Order document creation

**Note:** The new `shopware.number_range_manager` service can only be used in places, where it's possible to create a new transaction. Hence it doesn't work in Doctrine hooks, e.g. [Shopware\Models\Customer\Billing::onSave()](https://github.com/shopware/shopware/blob/6c6fdcdb11cd8c9a30076d0e394bc48118b4fdee/engine/Shopware/Models/Customer/Billing.php#L580-L597). That said, the code that requires this specific hook (most likely the customer creation via the Shopware backend) should probably be refactored to make `Shopware\Models\Customer\Billing::onSave()` obsolete.

### DIFF
--- a/engine/Shopware/Components/NumberRangeManager.php
+++ b/engine/Shopware/Components/NumberRangeManager.php
@@ -1,0 +1,108 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Components;
+
+use Shopware\Components\Model\ModelManager;
+
+/**
+ * @category  Shopware
+ * @package   Shopware\Components\NumberRangeManager
+ * @copyright Copyright (c) shopware AG (http://www.shopware.de)
+ */
+class NumberRangeManager
+{
+    /**
+     * @var ModelManager The entity manager
+     */
+    private $em;
+
+    /**
+     * @param ModelManager $em
+     */
+    public function __construct(ModelManager $em)
+    {
+        $this->em = $em;
+    }
+
+    /**
+     * @param string $name
+     * @return int
+     * @throws \Exception
+     */
+    public function getCurrentNumber($name)
+    {
+        $numberRange = $this->em->getRepository('Shopware\Models\Order\Number')->findOneBy([
+            'name' => $name
+        ]);
+        if (!$numberRange) {
+            throw new \Exception('Number range with name "' . $name . '" does not exist.');
+        }
+
+        return $numberRange->getNumber();
+    }
+
+    /**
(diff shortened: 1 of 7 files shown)


engine/Shopware/Components/NumberRangeManager.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        $this->em = $em;
+    }
+
+    /**
+     * @param string $name
+     * @return int
+     * @throws \Exception
+     */
+    public function getCurrentNumber($name)
```
bcremer, line 54: `getCurrentNumber()` is not used anywhere but in the tests. The public API should be as narrow as possible. Please remove that public method.

engine/Shopware/Components/NumberRangeManager.php
```diff
@@ -0,0 +1,108 @@
...
+namespace Shopware\Components;
+
+use Shopware\Components\Model\ModelManager;
+
+/**
+ * @category  Shopware
+ * @package   Shopware\Components\NumberRangeManager
+ * @copyright Copyright (c) shopware AG (http://www.shopware.de)
+ */
+class NumberRangeManager
```
hlohaus, line 34: It should implement an interface . Then you can replace it, if necessary.

bcremer:
@svenmuennich Thank you very much for this pull request. Nice Work.

I pushed a rebased version of this PR that includes some changes like using `dbal_connection` instead of the `em`. Also I removed the superflous `getCurrentNumber()` method.

https://github.com/shopware/shopware/tree/sw-14817/5.2/introduce-service-for-retrieving-next-number-from-range

Please let me now if these changes are OK for you. I am open for discussion before merge.

bcremer:
The new service should also be used for the following methods:
- `\Shopware_Controllers_Backend_Article::getNewArticleData()`
- `\Shopware\Models\Customer\Billing::onSave()`

Together with the proposed interface @hlohaus suggested, it should be possible to completely change the implementation.

svenmuennich:
Thanks you for the feedback.

I wasn't sure whether you preferred a solution based on the `ModelManager` or the `DBAL\Connection`, so I went with the former. I know that `getCurrentNumber()` is currently not used anywhere and is just a shortcut for `Shopware()->Models()->getRepository('Shopware\Models\Order\Number')->findOneBy(['name' => $name])`, so I don't mind that it's been removed.

> The new service should also be used for the following methods:
> - `\Shopware_Controllers_Backend_Article::getNewArticleData()`
> - `\Shopware\Models\Customer\Billing::onSave()`

I don't know whether the new DBAL connection based implementation allows using it for `\Shopware\Models\Customer\Billing::onSave()`. However, `\Shopware_Controllers_Backend_Article::getNewArticleData()` does not update `s_order_number`, but only determines a new number to suggest for the article creation. So I don't think that method needs changing.

svenmuennich:
I guess the goal is to eliminate all usage of `s_order_number` in the code, so `\Shopware_Controllers_Backend_Article::getNewArticleData()` could use the service's `getCurrentNumber()`.

hlohaus:
Yes, i think it's good idea. But an interface would be nice. ;)

svenmuennich:
Are you going to take care of the remaining changes or do you want me to merge [sw-14817/5.2/introduce-service-for-retrieving-next-number-from-range](https://github.com/shopware/shopware/tree/sw-14817/5.2/introduce-service-for-retrieving-next-number-from-range) back in [pickware/add-service-for-retrieving-next-number-from-range](https://github.com/VIISON/shopware/tree/pickware/add-service-for-retrieving-next-number-from-range) and take care of it?

dnoegel:
Hi @svenmuennich,

@OliverSkroblin is taking care of it currently.

svenmuennich:
Cool, thank you!

bcremer:
Merged with https://github.com/shopware/shopware/commit/8295eb119e5f55be53da1e3cb29d16a7b5b7364b.

Thanks to all collaborators :+1:

PR #518, opened 2016-04-18
Fix initialization of attribute fields having default values

svenmuennich:
It has always been possible to add a default value to attribute fields. However, previously these fields were not initialized with the respective default value when creating a new entity instance. Hence, when adding an attribute field that used a combination of a not NULL constraint and a default value, one always had to explicitly initialize the fields when creating the instance to avoid integrity constraint violations. This commit fixes this behaviour by adding a constructor to the generated attribute models, which explicitly initializes all properties with a configured default value.


bcremer:
Merged with commit https://github.com/shopware/shopware/compare/da85ff802315...83b224e1b17c.

Thank you @svenmuennich :+1:

PR #519, opened 2016-04-18
Validate order number when creating variant using REST API

svenmuennich:
When previously trying to create a variant using the REST API and passing a `number` that already exists in the database, the response was a `500 Internal Server Error` with the JSON message being the SQL error. This PR adds an explicit validation of the passed order number that checks whether it is already used by another existing variant. If already used, a `CustomValidationException` with a meaningful message is thrown, which results in a `400 Bad Request` response.


svenmuennich:
I overlooked that `Shopware\Components\Api\Resource\Variant::prepareData()` is also used when updating an existing variant. In that case, the `number` must not be checked, since it most likely will already exist. Hence I added an [extra condition](https://github.com/shopware/shopware/pull/519/files#diff-f2ef9262de522304c5efc59028234efaR462) to only check the number, if it is different to the number of the given `Shopware\Models\Article\Detail` instance. This fixes all previously failing unit tests. I also replaced the originally provided API unit test with a test in `tests/Functional/Components/Api/VariantTest.php`.

svenmuennich:
Looks like this was merged via bc6f19233c58d36ec2ecdf9ec46eb8e07d4c1127. Thank you!

PR #520, opened 2016-04-19
Allow uploading file when creating media using the REST API

svenmuennich:
Previously, creating a new `Media` instance using the REST API was only possible, if the respective file was already available on the server. However, a common use case for a a `media` REST API is to upload new files and register them with the media manager by creating a `Media` instance. By this PR, the REST API's `postAction()` checks for a POSTed file named `file` and, if present, uploads it and sets the respective path in the request parameters. Furthermore the `Media` resource is adjusted, to accept an instance of `\Symfony\Component\HttpFoundation\File\File` as the `file` parameter. To prevent API users from uploading harmful files, the `MediaManager`'s file blacklist is used to block certain file types.

**Note:** I added some test cases to `tests/Api/MediaTest.php`, which is not part of the automatically run tests anymore (seems to have changed on the `5.2` branch). Any plans on including the API tests again?


janbuecker:
FYI: I've improved your PR a little bit by using the `fileBag` earlier and increasing the max length of media names / paths.

You can review the changes here 41c66ac0e6e75654b092f0b15729123f5d2606d5.

PR #521, opened 2016-04-19
Grunt: use chokidar instead of grunt-contrib-watch

screeny05:
grunt-contrib-watch has some serious flaws, especially when using vagrant-boxes. Developers should not have to work for their tools, the tools should work for them!

_[grunt-contrib-watch](https://github.com/gruntjs/grunt-contrib-watch) is based on [gaze](https://github.com/shama/gaze/)_
_[grunt-chokidar](https://github.com/JimRobs/grunt-chokidar) is a fork of grunt-contrib-watch (pretty much compatible) based on [chokidar](https://github.com/paulmillr/chokidar)_
## What this change does:
- Use [grunt-chokidar](https://github.com/JimRobs/grunt-chokidar) instead of [grunt-contrib-watch](https://github.com/gruntjs/grunt-contrib-watch)
- Work around problems like not catching file-added events.
- Use the option `spawn: false` to speed up tasks, by eliminating the ~500ms delay, which comes when spawning tasks instead of just running them. I have tested this for a whole day of frontend-work without issues. Notable speed-up!
- Rename the chokidar-task to watch, so no compatibility-issues arise
## Why chokidar over gaze:

_stolen from the chokidar project-page_

> Node.js `fs.watch`:
> - Doesn't report filenames on OS X.
> - Doesn't report events at all when using editors like Sublime on OS X.
> - Often reports events twice.
> - Emits most changes as `rename`.
> - Has [a lot of other issues](https://github.com/joyent/node/search?q=fs.watch&type=Issues)
> - Does not provide an easy way to recursively watch file trees.
> 
> Node.js `fs.watchFile`:
> - Almost as bad at event handling.
> - Also does not provide any recursive watching.
> - Results in high CPU utilization.
> 
> Chokidar resolves these problems.
## What will be broken:

**Nothing**

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -201,7 +201,7 @@ In this document you will find a changelog of the important changes related to t
 * `\Shopware\Bundle\StoreFrontBundle\Gateway\GraduatedPricesGatewayInterface` requires now a provided `ShopContextInterface`
 * Categories of `Shopware\Components\Api\Resource\Article::getArticleCategories($articleId)` are no longer indexed by category id
 * Moved `<form>` element in checkout confirm outside the agreement box to wrap around address and payment boxes
-
+* Grunt now using [grunt-chokidar](https://github.com/JimRobs/grunt-chokidar) instead of [grunt-contrib-watch](https://github.com/gruntjs/grunt-contrib-watch) for watching files, in order to catch events like file-adds. Also speed-up of watch-task.
 
 ## 5.1.5
 * The smarty variable `sCategoryInfo` in Listing and Blog controllers is now deprecated and will be removed soon. Use `sCategoryContent` instead, it's a drop in replacement. 
--- a/themes/Gruntfile.js
+++ b/themes/Gruntfile.js
@@ -70,14 +70,20 @@ module.exports = function (grunt) {
                     '../engine/Shopware/Plugins/**/*.less',
                     '../themes/Frontend/**/*.less'
                 ],
-                tasks: ['less:development']
+                tasks: ['less:development'],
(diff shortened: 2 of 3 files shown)


Phil23:
Thank you for your PR.

We decided to decline this pull request, because there are not enough benefits for replacing the existing task runner in the standard. In first hand the grunt watcher is just a recommendation. All developers should feel free to use the task runner of their choice, either if it's grunt, gulp or whatever.

screeny05:
Hello Philipp, please re-review this PR. This does not change the Task-Runner. This PR just replaces the standard file-watcher with a file-watcher which is more reliable.
The only thing which changes in terms of functionality:

_Reliable Less- & JS-Compilation on file-creation and deletion_
- This is in no way a replacement for grunt.
- This changes no workflows.
- This improves developers life by not having to restart grunt each time a file is deleted or created.

Phil23:
Hello,

we have to do some further tests and come back to this PR in a future sprint.

Phil23:
Hey @screeny05,

testing chokidar for some time now. Your main argument about the auto detection of new files is nonsense, because you have to add the files to the system before they get used. New JS and CSS files have to be added to the Theme.php file and new LESS files are added by importing them in the all.less file. Tracking new files which are not used by the system makes no sense at all.

One advantage I have to admit is the performance increase. The chokidar does the task up to 2 seconds faster. That is defenetly an advantage. I did the tests now on Ubuntu (Vagrant), Ubuntu (VM) and Windows. I will let some of our Mac guys test the thing too. If everything works fine it could be added to the next patch release.

PR #522, opened 2016-04-23
Updated Forum Link

shyim:


bcremer:
Thank you @shyim, good catch.

Can you please revert the change in `_sql/install/latest.sql`.
This file should be considered read only and only modified using the migration files.

shyim:
Its reverted

janbuecker:
Thanks, merged with e2d705d7b96d0afeaec2596f60b345c8d0274169.

PR #523, opened 2016-04-25
introduce variables for column count in masonry emotions

SvenHerrmann:
Make column count in masonry emotion worlds configurable, so you can define the column count by setting a LESS variable in your own theme.

No BC breaks as the default values are the old ones.

see also: http://forum.shopware.com/discussion/34740/override-einkaufswelt-spalten-anpassen-fuer-masonry-in-emotion-less


bcremer:
@Phil23 might have a look?

Phil23:
Hi @SvenHerrmann,

thank you for your commit. The change is a good idea, but we are refactoring the emotion worlds for an upcoming release, so this is no longer relevant. I think the new features will be merged to the 5.2 branch in the next weeks. Stay tuned!

PR #525, opened 2016-04-27
Allow usage of 'base' stores in config combo boxes

svenmuennich:
This commits improves the base combo box `Shopware.apps.Base.view.element.Select`, which is used e.g. for the plugin configuration forms, to allow passing a store ID as the `store` parameter. This is escpecially usefull when providing selections of e.g. payment methods (`base.Payment`) or dispatch methods (`base.Dispatch`). In order to avoid side effects of other backend modules using the same store instance (and e.g. setting filters on it), the combo box does not use the instance that is registered by the passed store ID, but creates and uses a copy of that store instead. This copy uses the same model, URL and proxy settings, but has its own ID and its solely purpose is to be used as a store for base combo boxes.

This commit does not introduce any breaking changes.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -206,6 +206,7 @@ In this document you will find a changelog of the important changes related to t
 * Removed `sAdmin::sUpdateAccount()`
 * Removed `saveAccount()` in `Controllers/Frontend/Account.php`
 * Moved field `birthday` from billing address to customer
+* Added support for loading stores by ID in the base combo box `Shopware.apps.Base.view.element.Select`
 
 ## 5.1.5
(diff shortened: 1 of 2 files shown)


OliverSkroblin:
Hi,
i rewrited your implementation. Please take a look to this commit:
https://github.com/shopware/shopware/commit/3fc607dd19ba4c7092a4b42f2aa632c153a7f67d

Please let me now if it solves your problem.
An example usage is documented in the extjs function

svenmuennich:
This works for me. Thank you for "merging" 👍

svenmuennich:
@OliverSkroblin FYI: I just opened a new PR #600, which applies the same changes as done in your commit 3fc607dd19ba4c7092a4b42f2aa632c153a7f67d to `Shopware.apps.Config.view.element.Select`. This is necessary, because the plugin config can be viewed via the settings module as well and therefore that field must support loading the stores too.

PR #526, opened 2016-04-27
Ability to clear the cache, after plugin config save

shyim:
We should give the plugin developer the choice, to clear the cache, after plugin config saves.
The reasons could be like less variables in the plugin config etc.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -207,7 +207,7 @@ In this document you will find a changelog of the important changes related to t
 * Removed `saveAccount()` in `Controllers/Frontend/Account.php`
 * Moved field `birthday` from billing address to customer
 * Added validation of order number to `Shopware\Components\Api\Resource\Variant::prepareData()` to respond with meaningful error message for duplicate order numbers
-
+* Added `saveConfig` method to `Shopware_Components_Plugin_Bootstrap` to handle clearing caches, after plugin form submit
 
 ## 5.1.5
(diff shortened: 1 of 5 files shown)


dnoegel:
You could archive the same by registering to Shopware_Controllers_Backend_Config_After_Save_Config_Element and \Shopware\Components\Api\Resource\Cache::delete('config') if I am not mistaken. For install() and update() it already works.

I'd rather clear the config cache after _any_ plugin config saving automatically. But due to HTTP Cache and other caches, it will not have affect automatically anyways… so I am afraid, this seems a bit tricky and needs more consideration if we don't want to bloat the Bootstrap even more.

shyim:
What would you think about a filter event at end of the function before assigning the values to the view, with the plugin as param. And the plugin developer could subscribe it and could add the invalidateCache array for the backend confirmation. That should reach it and i can remove the function from the bootstrap.

I could subscribe to Shopware_Controllers_Backend_Config_After_Save_Config_Element  and clear the config cache, but when i have a color picker in my plugin configuration, which is using the less compiler, its needs to generate the theme cache after a change. Thats the point, that i would handle that with the default crud message after install/uninstall/update.

// sorry for my english :)

shyim:
@dnoegel @bcremer 
What would you think about a node "cache-invaldation" in the new pluginsystem to provide this actions. So we don't bloat the "old" bootstrap.php and it would be easy accessible from the plugin developer.

Example:

``` xml
<cache-invalidation>
<invalidate on="enable">theme</invalidate> 
<invalidate on="saveConfig">config</invalidate> 
</cache-invalidation>
```

bcremer:
@shyim We decided to close this PR for now. We will reconsider to clear the config cache for any config change automatically.

PR #527, opened 2016-04-28
Deprecate price range specific 'basePrice' and add single 'purchasePrice' as replacement

svenmuennich:
This PR deprecates the field `basePrice` of `Shopware\Models\Article\Price` and replaces it with a new field `purchasePrice` in `Shopware\Models\Article\Detail`.

This change fixes two problems:
1. The term 'base price' is used ambiguously throughout the Shopware code base. On one hand it is used for the purchase price of an article (previously `basePrice` in `Shopware\Models\Article\Price`). On the other hand it is used for the price with respect to the size and unit of the product, e.g. 'Euros per liter'. While the former is stored in the database, the latter is always calculate on-the-fly when necessary. At one point, a variable `$basePrice` is even used to store the default price of an article.
2. The base price, meaning purchase price, is stored for each entry in `s_articles_prices` (`Shopware\Models\Article\Price`). Hence it is possible to have different purchase prices for different quantity ranges (fields `from` and `to`) of the same article detail. However, these ranges are meant to control the price the shop customers are charged, not the purchase price.

By adding a new field `purchasePrice` to `Shopware\Models\Article\Detail`, the purchase price is only stored once per article detail and by changing the name the ambiguity is removed. Here's an overview of the changes made by this PR:
- Removed field `basePrice` from `Shopware\Models\Article\Configurator\Template\Price`
- Added field `purchasePrice` to `Shopware\Models\Article\Detail`
- Changed `Shopware\Models\Article\Price::setBasePrice()` to set the new base price also in `Shopware\Models\Article\Detail::purchasePrice`
- Add Doctrine `preUpdate` hooks to `Shopware\Models\Article\Detail` and `Shopware\Models\Article\Price`, which make sure that changes made to either `Shopware\Models\Article\Detail::purchasePrice` or `Shopware\Models\Article\Price::basePrice` are synced across both models
- Removed the 'base price' column from the price grids in the backend article and variant details modules
- Added 'purchase price' number field to the base forms of the backend article and variant details modules
- Added a database migration, which copies the old `baseprice`s of `s_articles_prices` to the new `purchaseprice` in `s_articles_details`, by using the `baseprice` of the price row having `from = 1` and belonging to the default shop's customer group (e.g. `EK`), and writes the same `baseprice` to all price rows of an article detail
- Added export of `s_articles_details.purchaseprice` as `baseprice` to `sExport` (since the class is marked as `deprecated`, I decided to use the old name `baseprice`)
- Removed the usage of the variable `$basePrice` in `sArticles`, to remove ambiguity
- Added parsing of `purchasePrice` to `Shopware\Components\Api\Resource\Variant`
- Added export of `s_articles_details.purchaseprice` to `\Shopware_Controllers_Backend_ImportExport`
- Added import of `baseprice` into `s_articles_details.purchaseprice` to `\Shopware_Controllers_Backend_ImportExport`

This PR does not introduce any breaking changes, since `Shopware\Models\Article\Price::basePrice` is only deprecated and always kept in sync with the new `Shopware\Models\Article\Detail::purchasePrice`. However, both `basePrice` and `purchasePrice` must only be changed using the respective setters, to keep them in sync.

Final comment: The old `basePrice` (purchase price) is not really used anywhere in Shopware, except for the price grid in the backend article and variant details modules, where it can be edited by the user. However, _grepping_ for `baseprice` will result in many matches, because of the ambigous use of the word.


svenmuennich:
I changed my initial PR to just deprecate `Shopware\Models\Article\Price::basePrice` instead of removing it. See the (updated) description for details.

_sql/install/latest.sql
```diff
@@ -163,6 +163,7 @@ CREATE TABLE IF NOT EXISTS `s_articles_details` (
   `releasedate` date DEFAULT NULL,
   `shippingfree` int(1) unsigned NOT NULL DEFAULT '0',
   `shippingtime` varchar(11) COLLATE utf8_unicode_ci DEFAULT NULL,
+  `purchaseprice` double NOT NULL DEFAULT '0',
```
bcremer: Please don't change `_sql/install/latest.sql`. All changes should be done in the migration.

engine/Shopware/Models/Article/Detail.php
```diff
@@ -898,4 +936,28 @@ public function setImages($images)
...
     }
+
+    /**
+     * Checks the change set for the 'purchasePrice' and, if present, makes sure that all associated
+     * price entities is flushed too. To break flush-cycles, the prices' change sets are checked
+     * and they are only flushed, if their 'basePrice' is not already part of the change set.
+     *
+     * @ORM\PreUpdate
+     */
+    public function beforeUpdate()
```
bcremer: Please use a Doctrine Event Listener or Subscriber instead of Lifecycle-Callback:
See for example: https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Components/Model/CategorySubscriber.php#L41

http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/reference/events.html#listening-and-subscribing-to-lifecycle-events

bcremer:
Hi @svenmuennich, after some discussions in the team we decided the following:
- Add field `purchaseprice` to Table `s_articles_details` to and  the Entity `Shopware\Models\Article\Detail`
- Migrate data from `s_articles_prices.baseprice` into `s_articles_details.purchaseprice`
- Deprecate but not remove  `s_articles_prices.baseprice` from the database
- Remove all getters/setters and other access to `s_articles_prices.baseprice` from the shopware core
- Do not sync writes from  `s_articles_details.purchaseprice` back to `s_articles_prices.baseprice` - The data will just stay in the database to prevent data loss in case people saved data in different price ranges
- Also please add purchaseprice to the configurator template

Keeping the data in sync using events is error prone so we decided to take the hard cut. This should simplifie the implementation a lot.

Thank you.
Greetings, Benjamin

svenmuennich:
I changed everything as described. The field `s_articles_prices.baseprice` / `Shopware\Models\Article\Price::basePrice` still exists, but is not used anymore anywhere.

bcremer:
Great, thank you @svenmuennich :+1: 
I'll test and review in depth tomorrow.

bcremer:
Finished another review. Code wise it looks fine. 
But I am missing a few functionalities:
- When new variants are created the purchasePrice of the mainDetail is not adopted.
- It's not possible to set a  `purchasePrice` in a configurator template (Article-Module > Variants > Configuration > Define configurator template)
- It's not possible to apply the purchasePrice from the mainDetail using the "Apply standard data" button (Article-Module > Variants > Available variants > Apply standard data)

svenmuennich:
Thank you for the feedback @bcremer. I'm going to take care of this today.

bcremer:
Thanks @svenmuennich, please just add it as a new commit on top because I already did some changes to this branch locally and will cherry-pick your commit.

svenmuennich:
The `purchasePrice` can now be set in the configurator and is also copied from the main detail, when creating variants. I decided to add a new option to the configurator _mapping_ window, so that it is possible to selectively copy the purchase price when applying the data. I was struggling though to increase the height of the _mapping_ window and the only way I could make it work was to [set the height when the window is shown](https://github.com/shopware/shopware/pull/527/commits/5b0294b840c70bf9753e72b49f1bab889477bc5e#diff-0b60aa9b849b281f0045839866eebc3dR127). Otherwise the window would always keep a height of `300`.

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/04c18d76e9762f526bc441e83cc7b55082c7c6bb.

Thanks for your work @svenmuennich :dancers:

svenmuennich:
Great! Thank you for merging 👍 👍

PR #528, opened 2016-04-29
Removed pseudo second argument on Shopware\Models\Menu\Repository::fi…

shyim:
On PHP 7 is impossible to use the second argument, because the typehinting will throw a exception.
Example:  ->findOneBy('name', 'Einstellungen')

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -212,9 +212,12 @@ In this document you will find a changelog of the important changes related to t
     * `sAdmin::assignCustomerNumber()`
     * `sOrder::sGetOrderNumber()`
     * `Shopware_Components_Document::saveDocument()`
+* Removed pseudo second argument on `Shopware\Models\Menu\Repository::findOneBy()` and deprecated methodes:
+    * `addItem`
+    * `save`
 
 ## 5.1.5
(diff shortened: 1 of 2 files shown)


hrombach:
Just a suggestion, couldn't you just remove the typehint instead of removing the function altogether?

shyim:
@hrombach 
It would be the same as the extendes class, thats the reason, why i removed it :)
See: http://www.doctrine-project.org/api/orm/2.2/class-Doctrine.ORM.EntityRepository.html

hrombach:
Looks to me like the parent function only takes an array as parameter. The point of the override is allowing implementations without an array wrap

shyim:
Yes it takes only a array.
Only the MenuRepository has overriden the parent and ajusted the option, we should take it equal to the other model repositorys and a array as argument is default doctrine :)

hrombach:
You're right about that, i see your point

bcremer:
Implemented that change on the 5.2 branch: https://github.com/shopware/shopware/commit/096ac260d910c6a76595df8349183246cc7a500c.

Thank you @shyim :+1:

PR #530, opened 2016-05-02
Allow changing the dispatch method of existing orders

svenmuennich:
This PR adds a new `Dispatch data` fieldset to the backend order details `Details` tab, containing a single combobox, which allows to change the dispatch method of the order to any available dispatch method. Analogous to changing the payment method of an order, changing the dispatch method does not have any side effects on the order. That is, e.g. its shipping costs are not changed according to the newly selected dispatch method.


janbuecker:
Thanks for your PR.

I'd like you to use the `Shopware.form.field.SingleSelection` introduced with 5.2. You can view some examples in `themes/Backend/ExtJs/backend/base/attribute/field`.

janbuecker:
@svenmuennich can you update your PR? Otherwise it will be closed within the next days.

svenmuennich:
Sorry, but after one hour of searching the shopware code for any explicit example I couldn't come up with anything that would work. When I look e.g. at `Shopware.form.field.CustomerSingleSelection`, it is only used in `Shopware.attribute.CustomerFieldHandler`, which is only used in `Shopware.attribute.Form`. However, the attribute form is not what we need here. Do you use these kind of fields anywhere else than for the attribute handling?

janbuecker:
Well, I've evaluated the use of a single selection and it seems not to fit here that good. So I think, that your changes are ok and I'll finish this now.

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/897318b610fe72f74fbb1bfb7574fce0badc6c55.
Thank you!

PR #531, opened 2016-05-02
Allow empty strings and booleans as attribute default values

svenmuennich:
This PR changes the creation of custom attribute fields to allow default values such as empty strings or booleans. When previously passing an empty string as the default value of an attribute field it was replaced by `NULL`. However, it is perfectly valid to use an empty string as the default value for a table field and sometimes desirable. This commit removes the check for an empty string to allow using empty strings as default. Furthermore passing `false` as the default for a boolean (`tinyint`) attribute field previously resulted in an exception. Therefore this commit also adds a check of the passed `$default` for being a boolean type and translates it to `0` or `1` for `true` and `false` respectively.


svenmuennich:
@OliverSkroblin It looks like you already merged this one via c882929830b1574b4ff90e19bfe320108ea1a73f. Hence I go ahead and close it.

Thank you 👍

PR #532, opened 2016-05-03
Fix dumb snippet for search-breadcrumbs

screeny05:
When using the search, the breadcrumb gets set to `Search results for Foobar`.

The problem is, when the User didn't enter a search and instead just pressed enter, so the breadcrumb will now be `Search results for` when instead it should be just `Search results`.

This PR fixes this issue. It also provides a German & English Text for the new Snippet `SearchResultsEmpty`.


janbuecker:
Merged with efa67131168b1db54414b33b8015fe77e89f0ddb.

PR #533, opened 2016-05-03
Update box-shadow.less for browser compatibility.

rip3rs:
I added to the .box-shadow mixing to have cross browser compatibility.
I was having cross browser issue for a client and I believe that this should be in the bare shopware code.


screeny05:
-1, to keep the css as bloat-free as possible, there is no need for vendor-prefixes for box-shadow. The browser-support is exeptionally good.

[Caniuse](http://caniuse.com/#feat=css-boxshadow)

The global support for the unprefixed version stands in at 89.2% while the support with prefixes is at 89.46%. That means, there is a global total usage of 0.26% of browsers which need vendor-prefixes.

Phil23:
I agree with @screeny05. The prefixes are not necessary in our supported browsers.

PR #535, opened 2016-05-04
Fix wrong initial 'from' value at price model

migo315:
A default value of "0" causes problems at listing (price will not be shown). Initial value should be 1.

### DIFF
--- a/engine/Shopware/Models/Article/Price.php
+++ b/engine/Shopware/Models/Article/Price.php
@@ -68,7 +68,7 @@ class Price extends LazyFetchModelEntity
      *
      * @ORM\Column(name="`from`", type="integer", nullable=false)
      */
-    private $from = 0;
+    private $from = 1;
 
     /**
      * @var integer $to


teiling88:
+1 very triky issue in combination with customergroup prices on listing page if you use the doctrine model for customer group prices.

janbuecker:
Hi, thanks for your PR.

I cannot reproduce your issue. Could you explain it in more detail?

migo315:
Sorry for the delay. 

If you save a price like this  ...

```
$price->setArticle($article);
$price->setDetail($detail);
$price->setCustomerGroup($customerGroup);
$price->setPrice(12);
$this->models->persist($price);
```

... will the price never be shown at frontend. Because, the default value of the `from` property is 0. Actually, you have to set every time `$price->setFrom(1);`. IMHO, the default value of the property should be 1.

Does the explanation help?

soebbing:
Hi chowanski,

your direct problem should have been fixed by another change in the end of last year, nevertheless did your change help with some other smaller issues. Thank you very much for your PR!

Have a nice day!
Hendrik

PR #536, opened 2016-05-04
Add the possibility to make shop-independent sessions

jeboehm:
Hi,

this PR adds the possibility to make sessions independent of the selected (sub-) shop, just by adding a name parameter to Shopware's session configuration.

Before the change, it was always overwritten, so you had to completely override the Session-Factory class.

See support ticket #56586 for your reference.


janbuecker:
Thanks @jeboehm 

PR #537, opened 2016-05-04
bugfix for external adapters with plain strategy

martinweinmayr:
also return it with the media url, otherwise you won't get files from external services


janbuecker:
Your fix would break calls which are providing an already built url to an image. e.g. `http://cdn.shop.com/media/image/foo.jpg` would result in `http://cdn.shop.com/http://cdn.shop.com/media/image/foo.jpg`. I'll discuss this internally and maybe enhance your fix.

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/aae238a553d82123427c673156195df11d191f5d.

Thanks @martinweinmayr :+1:

PR #538, opened 2016-05-04
Fix Javascript error when opening an order without a shipping address

andrewisplinghoff:
It was not possible before to open an order in the Shopware backend if it was missing its shipping address, because a Javascript error came up. The cause is that the condition (!billingStore == null) never resolves to true because !billingStore is either true or false, but never null.

selection_100


OliverSkroblin:
Hi,
The described problem is with shopware 5.2 no longer reproducible. Backend order detail pages can be opened with and without any billing and/or shipping address.

PR #539, opened 2016-05-05
SW-14601 - close detail window after blog article is saved

uehler:
If you added content to the attribute fields of a new blog entry and save it twice without closing the detail window the connection between the attribute fields and the blog entry gets lost. The blog_id in s_blog_attributes will be null.

This commit closes the detail window of the blog entry when you save it to get rid of this behaviour


OliverSkroblin:
Hi @uehler,
we refactored the attribute system with shopware 5.2, also the implementation in the backend blog module. 
The described problem is no longer reproducible.
I created a new blog entry, filled the attributes at the same time and saved multiple times, but the reference still works. Same behavior with existing blog entries. 
If the problem still exists please reopen the request.

PR #540, opened 2016-05-05
Fixed misnamed parameters in sBasket

uehler:
I renamed the parameter $id of sBasket::sAddArticle() and sBasket::getArticleForAddArticle() to $ordernumber because both methods expect an article ordernumber and not an article ID.

I didn't change the name of the value in the events of those method because i didn't want to break plugins which rely on those events


dnoegel:
Hi,

unfortunately this change will also introduce breaks for plugins accessing those parameters in before/after/replace hooks. That's one of the drawbacks of hooks - and the reason why we didn't rename those params already. As we a currently planning to rewrite the basket anyway in the future, I think we should keep this inconsistency for the time being.

janbuecker:
Hi @uehler and thanks for your PR.

I agree with @dnoegel and we should keep this inconsistency until the new basket has been finished.

Closing this PR.

PR #541, opened 2016-05-06
API-Bug: mp3-files are imported as mpga

schliesser:
We noticed that the API uses the "File" class to guess the file extension. This guesser method fails on mp3 files. Jpgs and pdfs guessed correctly. There's also an class called "UploadedFile" which works properly on mp3, pdf,... but not for jpgs. So we need to check if the file extension is mp3 and use there the proper File class.

May you can provide a better workaround?

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Media.php
+++ b/engine/Shopware/Components/Api/Resource/Media.php
@@ -29,6 +29,7 @@ use Shopware\Components\Thumbnail\Manager;
 use Shopware\Models\Media\Album;
 use Shopware\Models\Media\Media as MediaModel;
 use Symfony\Component\HttpFoundation\File\File;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
 
 /**
  * Media API Resource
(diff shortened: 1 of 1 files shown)


3stadt:
Hello @schliesser,

we will not merge your PR in its current form.

The behaviour you are describing is not neccessarily a bug, the underlying symfony `MimeTypeExtensionGuesser` lists `mpga` as Extension for `audio/mpeg` (see: [symfony](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/File/MimeType/MimeTypeExtensionGuesser.php#L623)). The reason for this is, symfony is using the [apache curated list of file extensions](http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types), which lists mpga as first choice for this mime type.

So a better solution would be to keep the file extension if the file is recognized as `audio/mpeg` and has one of `mpga mp2 mp2a mp3 m2a m3a` as extension. This way, all other possible problems regarding this topic would be solved as well.

Additionally, your PR only covers a part of the file upload possibilities, e.g. a direct upload via POST to `api/media` would currently still show the `mpga` extension.

Please update your code to reflect the aforementioned changes and create a new pull request.

Greetings, 3stadt

PR #542, opened 2016-05-09
[FEATURE] Set "snippet_resource" service to public

barbieswimcrew:
As discussed with Martin Weinmayr (CEO dasistweb GmbH) with and agreed by Elke Klein-Ridder it is necessary to remove the public="false" property in the "snippet_resource" service definition to create the opportunity to override the Snippet Manager for special needs.

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/services.xml
+++ b/engine/Shopware/Components/DependencyInjection/services.xml
@@ -23,8 +23,7 @@
         </service>
 
         <service id="snippet_resource"
-                 class="Enlight_Components_Snippet_Resource"
-                 public="false">
+                 class="Enlight_Components_Snippet_Resource">
             <argument type="service" id="snippets"/>
             <argument>%shopware.snippet.showSnippetPlaceholder%</argument>
         </service>


OliverSkroblin:
Hi, 
what is your intention to publish the `snippet_resource` service? 
The \Enlight_Components_Snippet_Resource is responsible to compile the smarty snippet tags.

Read and write accesses of snippets are handled by the `\Shopware_Components_Snippet_Manager` class, service: `snippets`.

If this isn't your intention you can also overwrite the `snippet` resource over the smarty engine, by registering your own `snippet` resource. This overwrites the existing and you can compile the snippet tags by your own.

barbieswimcrew:
Hi Oliver,

it's our intention to fully override the existing snippets service to avoid having multiple snippet resources at once. Respectively to even change the way of how snippets are being managed (we'd like to do this via xml in some custom logic) I think it's necessary to get the service declared public.

Hope this will help you. If there are more questions regarding this please feel free to contact me.

Kind regards!

bcremer:
Resolved with commit: https://github.com/shopware/shopware/commit/7bc44e3c613e386a345bfca2a4976d452f9029b0

Thanks :+1:

PR #543, opened 2016-05-10
[BUGFIX] Load basket products with sArticles::sGetArticleById()

FiveDigital:
Load basket products with sArticles::sGetArticleById() instead of sArticles::sGetProductByOrdernumber() in sBasket::getBasketArticlesUncached(). This loads the products using the ProductService instead of the ListProductsService. This fixes a problem that the data in the "additional_data" field has always been loaded from the main variant even if you have a specific variant in the cart.

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2085,7 +2085,7 @@ class sBasket
 
             // Get additional basket meta data for each product
             if ($getArticles[$key]["modus"] == 0) {
-                $tempArticle = $this->moduleManager->Articles()->sGetProductByOrdernumber($getArticles[$key]['ordernumber']);
+                $tempArticle = $this->moduleManager->Articles()->sGetArticleById(0, null, $getArticles[$key]['ordernumber']);
 
                 if (empty($tempArticle)) {
                     $getArticles[$key]["additional_details"] = array("properties" => array());


teiling88:
It is possible, that this change causes side effects. In older Shopware Versions the function sGetArticleById sets in some cases a new sCategoryId and changes the category array structure for the view.
In which case do you need more information based on the sGetArticleById method?

FiveDigital:
We found this issue while trying to display the reference price in the cart. After hours of debugging through the services and the struct converter and the mess in the core classes I found out that there is no bug in all of that but a bug in the logic: The list product service is not the right service for the cart because you want the information of the variant you have chosen in the cart. But using the list product service all information in `addtional_data` will be fetched from the main variant. This behavior is correct in lists since you cannot show variants except the main variant there. 

If this PR does not suit for 5.1 perhaps you can merge it in the 5.2 branch? We plan to update our customers installation on time.

teiling88:
I'm not a shopware crew member :-) but i know that the shopware checkout will be refactored in the future daniel mentioned it in https://github.com/shopware/shopware/pull/540

FiveDigital:
I'm pretty sure this won't happen before Shopware 6.0 :-) And providing wrong product data to the cart/checkout is a bug which has to be fixed asap imho.

teiling88:
But a fix with new bugs is not a good fix ;-)

FiveDigital:
Any feedback @bcremer?

bcremer:
@FiveDigital we are busy with the Shopware Community Day atm. We will work down the github queue soon.

FiveDigital:
Okay, cu there :-)

PR #544, opened 2016-05-10
Codemirror Update and Improved Modes Support

shyim:
Example Plugin to test all modes: https://github.com/shyim/sw-codemirror-test
Its the updated pr of #348 for 5.2 branch :)


bcremer:
@shopware/shopware-webdev I would like to like to include that for 5.2. Can you give it a quick try?

PR #545, opened 2016-05-10
[BUGFIX] wrong usage of unitize mixin

schliflo:
This fixes wrong used .unitize() mixins.
Line 46 was obsolete as it had never an effect due to the wrong mixin usage.
Line 47 was refactored to be applied correctly now as it probably was intended.


3stadt:
Hi @schliflo,

good catch! Your PR is merged and will be included in the next release.
Thank you! :)

3stadt

bcremer:
Merged with commit: https://github.com/shopware/shopware/commit/92f39b3fbb66527b72f5616544acd1ed7269d7ae

PR #547, opened 2016-05-13
Added stock and purchase related information to order items in account

FiveDigital:
This missing infos are needed if you want to add a "Add to cart" button on each order item in the account order history without reloading the product structs. Since other information like `purchaseunit` are already copied from the product to, this PR should be easy to merge.


OliverSkroblin:
Hi,

i changed the implementation an provide the full article as array element:
https://github.com/shopware/shopware/commit/4f3eb9f6cca3ef54ebf0c833312b9dfa967a0336

PR #548, opened 2016-05-13
[BUGFIX] wrong action names in if-statement

mjosef89:
This fixes wrong action names in if-statement.


janbuecker:
Hi @mjosef89 and thanks for your PR.

We've got your intend and improved it by moving the `setNoRender()` method call into the actions. This approach is more straight forward.

Merged with 4c1081f1b9b68e856dba551be95650968a50d397.

PR #550, opened 2016-05-17
Fix adding article to note listing dont send sUniqueID cookie on non existing session id

Haehnchen:
We are adding article to note list via api request. But initial sAddNote call dont sends `sUniqueID` because we have not yet a session. Only after calling the page again session is set and and we get the Set-Cookie header. See guzzle testing which pipes to sAddNote:

``` php
        // force session error ;)
        $guzzle->get('/');

        $guzzle->post('shop-api/note', ['body' => json_encode([
            'action' => 'addLineItem',
            'productId' => 'SW10043',
        ])]);

        $response = $guzzle->post('shop-api/note', ['body' => json_encode([
            'action' => 'addLineItem',
            'productId' => 'SW10150.5',
        ])])->json();
```

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -1149,10 +1149,9 @@ class sBasket
      */
     public function sAddNote($articleID, $articleName, $articleOrderNumber)
     {
-        $cookieData = $this->front->Request()->getCookie();
         $uniqueId = $this->front->Request()->getCookie('sUniqueID');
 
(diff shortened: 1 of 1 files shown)


janbuecker:
Could you please provide a test scenario so we can verify this bug?

janbuecker:
@Haehnchen reminder. Otherwise, this PR will be closed within the next days.

Haehnchen:
not able to provide a tests. guzzle client and testing request workflow included if u want to fix it later on.
so closed on me side

PR #551, opened 2016-05-17
Time function in sAddNote should be lowercase

Haehnchen:


janbuecker:
Thanks for your PR. Merged with 377dd1e7d8b.

PR #552, opened 2016-05-18
SW-15039 fixing url builder replacement of ampersand to german 'und'-word

itinance:
sReplaceTable will replace every "- & -" of a name of an entity with "-und-", which makes no sense on SEO-side and also will not fit to articles and categories of language sub shops other than german.

A category called "Home & Garden" will get the URL "home-und-garden". The word "und" is german and should be replaced with "and". But, from seo-side, this is a stop word and has no relevance on SEO-side. The recommendation of various SEO-experts is to use a hyphen instead of any translations.

With my commit the url would be "home-garden". From seo side these are two separate words, while underscore would be interpreted as a concatenated word. Some good source about this topic is [this one](http://www.ecreativeim.com/blog/2011/03/seo-basics-hyphen-or-underscore-for-seo-urls/).

### DIFF
--- a/engine/Shopware/Core/sRewriteTable.php
+++ b/engine/Shopware/Core/sRewriteTable.php
@@ -541,7 +541,7 @@ class sRewriteTable
         'Ỷ' => 'Y',
         'Ỹ' => 'Y',
         // shopware rules
-        ' & ' => '-und-',
+        ' & ' => '-',
         ':' => '-',
         ',' => '-',
         "'" => '-',
--- a/tests/Shopware/Tests/Core/sRewriteTest.php
+++ b/tests/Shopware/Tests/Core/sRewriteTest.php
@@ -58,7 +58,7 @@ class sRewriteTest extends PHPUnit_Framework_TestCase
             array('Hello,World', 'Hello-World'),
             array('Hello;World', 'HelloWorld'),
             array('Hello&World', 'HelloWorld'),
-            array('Hello & World', 'Hello-und-World'),
+            array('Hello & World', 'Hello-World'),
             array('Nguyễn Đăng Khoa', 'Nguyen-Dang-Khoa'),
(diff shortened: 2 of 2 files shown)


itinance:
Hm... this fix would break a lot of existing german URLs. Maybe we should think about another solution?

bcremer:
Makes sense.

But we need to think about the migration of the existing URLs as you already mentioned.

We will look into this after the Shopware Community Day.

teiling88:
Based on this [sInsertUrl](https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Core/sRewriteTable.php#L1149) function: the actual valid link get an main=0 and redirects via 301 on the new valid link with main=1

So in my opinion nothing breaks :-)

itinance:
thx @teiling88, you're right! I tried it already out. The old url will be redirected to the new one

dnoegel:
I agree that nothing will break due to the redirect system; but I am not sure, if one should simply "change" the canonical URL of all existing shops having an "und" in their URLs.  Perhaps we should make the "$replaceRules" array at least easier to modify and provide an "compatibility" plugin which restores the old behavior. 

Generally this hardcoded german snippet should be removed - the question is just, how to deal with those "canonical url" changes.

bcremer:
`\sRewriteTable::sCleanupPath()` should delegate to a new interfaced service that is responsible for the URL cleanup. Something like https://github.com/cocur/slugify/blob/master/src/SlugifyInterface.php.

So this service can be replaced by the customer/developer using their very own replacement rules.

FiveDigital:
👍

bcremer:
See also: https://issues.shopware.com/#/issues/SW-14378

janbuecker:
Thanks for your PR @itinance.

We've discussed our current implementation of creating seo urls and decided to implement the whole `cocur/slugify` library as `shopware.slug`. You can then decorate this service to get into the generation process.

Make sure to keep an eye on https://issues.shopware.com/#/issues/SW-15039 as this will introduce the new feature.

PR #553, opened 2016-05-20
Disable xdebug for travis build

alfredbez:
see https://docs.travis-ci.com/user/languages/php#Disabling-preinstalled-PHP-extensions and https://getcomposer.org/doc/articles/troubleshooting.md#xdebug-impact-on-composer

> Compared to a cli command run with "xdebug" enabled a speed improvement by a factor of up to 3 is not uncommon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ cache:
    directories:
      - $HOME/.composer/cache/files
 
+before_script:
+    - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then phpenv config-rm xdebug.ini; fi
+
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,6 +13,9 @@ cache:
    directories:
      - $HOME/.composer/cache/files
 
+before_script:
+    - phpenv config-rm xdebug.ini
```
xabbuh: You don't want to execute this on HHVM (see the failure):

``` yaml
before_script:
    - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then phpenv config-rm xdebug.ini; fi
```

alfredbez: thanks for the hint, I fixed it, build looks ok now

PR #554, opened 2016-05-20
Prevent model generator to generate invalide php files

GM-Alex:
If there are third party attribute tables at the database which are not covered by the createTableMapping function the generator creates corrupted php class files.


janbuecker:
Thanks for your PR. Could you please cleanup your commits and only commit the changes related to this issue?

bcremer:
Merged with commit: https://github.com/shopware/shopware/commit/dcf30e7f458fc36161d9d4dd4b9b9d3ee0e200fa
Thank you @GM-Alex :+1:

PR #555, opened 2016-05-21
Fix error in article statistics due to mixup of article and voucher ids

cjuner:
Re-created for rebase against 5.2.

Please see the original PR for more information: https://github.com/shopware/shopware/pull/221#issue-54067005

Original PR text:

> We found that article statistics erroneously show redeemed vouchers that have nothing to do with the respective article:
> 
>  651d535a-9a77-11e4-85bc-79614ceeac49
> 
> The problem arises in the statistics tab of the article that has the same id as the respective voucher.
> Technical details for the proposed fix:
> 
> The articleID column of the s_order_details table only represents a real article id when modus is 0. When modus is 2, for example, the articleID column contains a voucher id instead.


samuelvogel:
This was merged with https://github.com/shopware/shopware/commit/2903670bbe1a60d9ee278f767e5b61b198b038bb.

PR #558, opened 2016-05-22
Codemirror Update and Improved Modes Support 

shyim:
#544 on Rebase to solve the conflicht i broke the pr. here is the new :)


xabbuh:
I think the mode changes (added executable bits) should be reverted.

shyim:
@xabbuh 
What did you mean with "executable bits" i use the default files from codemirror repo.

PR #559, opened 2016-05-22
add "console.command" tag support which registers commands in a more Symfony way

Haehnchen:
see http://symfony.com/doc/current/cookbook/console/commands_as_services.html

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -308,6 +308,7 @@ In this document you will find a changelog of the important changes related to t
 * Moved s_user_billingaddress.customernumber to s_user table
 * Removed \Shopware\Models\Customer\Billing::number property
 * Removed method `Shopware\Bundle\PluginInstallerBundle\Service\InstallerService::getPluginBootstrap()`
+* Add support for Symfony `console.command` service tag to register commands directly inside service container
 
 ## 5.1.6
 * The interface `Enlight_Components_Cron_Adapter` in `engine/Library/Enlight/Components/Cron/Adapter.php` got a new method `getJobByAction`. For default implementation see `engine/Library/Enlight/Components/Cron/Adapter/DBAL.php`.
--- a/engine/Shopware/Components/Console/Application.php
+++ b/engine/Shopware/Components/Console/Application.php
@@ -150,6 +150,7 @@ class Application extends BaseApplication
         }
 
         $this->registerFilesystemCommands();
+        $this->registerTaggedServiceIds();
     }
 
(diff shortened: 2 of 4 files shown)


xabbuh:
Just out of curiosity: Is there a special reason for Shopware to duplicate compiler passes provided by Symfony? I mean you now have to take care of fixing all bugs yourself that have been fixed upstream.

bcremer:
@xabbuh Good Question.

Most Compiler Passes like the `AddConsoleCommandPass` are located in the `symfony/framework-bundle` that we currently not require.

Pulling in `symfony/framework-bundle` would also add a lot unnecessary transitive dependencies:

```
  - Installing symfony/templating (v3.0.8)
    Downloading: 100%         

  - Installing symfony/stopwatch (v3.0.8)
    Downloading: 100%         

  - Installing symfony/polyfill-util (v1.2.0)
    Loading from cache

  - Installing symfony/polyfill-php56 (v1.2.0)
    Downloading: 100%         

  - Installing symfony/security-core (v3.0.8)
    Downloading: 100%         

  - Installing symfony/security-csrf (v3.0.8)
    Downloading: 100%         

  - Installing symfony/routing (v3.0.8)
    Downloading: 100%         

  - Installing symfony/asset (v3.0.8)
    Downloading: 100%         
```

I am not saying requiring `symfony/framework-bundle`  is a blocker for us but "just" for one or two compiler passes it's too much right now.

I will put the inclusion of ``symfony/framework-bundle`  under discussion again as we could probably use some more components like the Commands.

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/f22d881beba7c8242a69e89bb01f75d8b4479585.

Thanks @Haehnchen :+1:

xabbuh:
@bcremer Maybe we can discuss moving some compiler passes and/or commands to other components if that better fits (e.g. we already did that for some commands in the past).

bcremer:
@xabbuh Would it be an option to move `Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler\AddConsoleCommandPass` into the  `symfony/console` namespace?
That would introduce coupling to the framework in the `symfony/console` component.
Alternatively framework code could be moved to a extra framework repository (`symfony/console-framework`) just for code that is coupled to the framework like compiler passes and commands etc.

Haehnchen:
@xabbuh not my first code duplication. simplest way to use dic and console component. maybe moving CompilerPass to console is a good idea.

PR #560, opened 2016-05-23
Make order billing and shipping addresses searchable

cjuner:
Re-created [PR 224](https://github.com/shopware/shopware/pull/244) for a rebase on top of 5.2.

Original PR text:

> This makes it possible to search for orders with a specific address (street, city, country etc.) in the order dialog.

This patch adds joins to `orders.shipping` and `shipping.country`. In the original PR, potential performance issues [were raised](https://github.com/shopware/shopware/pull/244#issuecomment-78919892). However, the proposed changes, i.e. not joining `orders.details`, are orthogonal to this commit, which does not touch anything `orders.details` related. While the actually added additional joins may incur a performance cost, I propose to deal with performance improvements in a separate issue and rather add this functionality for now.


OliverSkroblin:
Hello,
thank you for creating this pull request. Your extension decreases the listing performance of the module, even if no search term added.
We will refactor the listing query in 5.3 to optimize performance and filter opportunities.

PR #562, opened 2016-05-23
Fix bug in price group management by automatically preselect first customer group

cjuner:
Re-created PR #127 for rebase on top of 5.2.

Orignal PR text:

> Currently discounts, which are associated to a price group, are lazy-loaded, when the user selects a customer group. Since the save action for price groups simply overwrites the data stored in the database with the data received from the client, this behavior leads to the deletion of associated discounts, if the user only alters the name of a price group without "lazy-loading" the discounts before he clicks the save button. The automatic preselection of the first customer group ensures, that the discounts are always been loaded, when the user triggers the save action.


janbuecker:
Thanks. Merged with ec6c28c6e16da2b2e85ae5e008e4f770c2cbf804.

PR #563, opened 2016-05-23

cjuner:
Re-created PR #269 for rebase on top of branch 5.2.

Original PR text:

> Most of the sprites in extra icon set 02 do not work, because the !important annotation is missing. Without this annotation, the rules do not apply and the top-left icon of the icon set is shown instead of the desired icon.


cjuner:
Closing, incorrect target branch selected. This is meant for 5.2, not 5.1.

PR #564, opened 2016-05-23

cjuner:
Re-created PR #269 for rebase on top of the 5.2 branch.

Original PR text:

> Most of the sprites in extra icon set 02 do not work, because the !important annotation is missing. Without this annotation, the rules do not apply and the top-left icon of the icon set is shown instead of the desired icon.


janbuecker:
Thanks, merged with cdf2a188badab18bec144e24d2a6e9d5fd54ac50.

PR #565, opened 2016-05-23
Add unsubscribeEvent() method for plugins

cjuner:
Re-created PR #452 for rebase on top of the 5.2 branch.

Original PR text:

> In some plugin updates it is desirable to remove old event subscriptions, since they are no longer used. This commit adds a new method to Enlight_Plugin_Bootstrap_Config, from which all plugin bootstrap implementations inherit.
> 
> The new unsubscribeEvent() method requires only the event name and listener name to remove an existing subscription, just like subscribeEvent() requires those two parameters to add a subscription in the first place.
> 
> In order to make this work, it is also necessary to adapt the write() method of Shopware_Components_Plugin_Namespace, so that the removed subscribes are also removed from the database. Therefore it has now an optional parameter of type Shopware_Components_Plugin_Bootstrap. If it is present, that plugin's subscribes will be fully synced, meaning that its removed subscribes will be deleted from the database.
> 
> Please note, that unsubscribeEvent() relies on a properly working removeListener() method in Enlight_Event_Subscriber_Config which was fixed by #451. Until that is merged, the newly added test will fail!
> 
> This is an updated version of #175.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -316,6 +316,7 @@ In this document you will find a changelog of the important changes related to t
 * Include the departments, salutations, cities and countries in the address comparison of the backend order details
 * Display the departments in the backend order details overview
 * Added new API resource 'Country' and respective REST API controller 'countries'
+* Add a new `unsubscribeEvent` method to `Enlight_Plugin_Bootstrap_Config` for removing existing event subscriptions
 
 ## 5.1.6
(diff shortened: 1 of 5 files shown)


samuelvogel:
The fix/PR this relies on (#451) is already merged in core since release `v5.1.4`.

janbuecker:
Thanks @cjuner 

We do not plan to implement this in the old plugin system. This issue is not present in the 5.2 plugin system.

PR #566, opened 2016-05-23
inconsistent net and gross amounts in basket for modus = 4

SvenHerrmann:
our use case:
In a shop with net mode, we need quantified shipping surcharges which we add via a plugin (i.e. we add basket items with modus = 4 and quantities > 1).
Net and gross amount calculations for such positions differ:
net amount is correct: quantity \* single net price
but gross amount is always 1 \* single net price \* tax_factor.

This is inconsistent.
Either quantity should be ignored for net and gross or quantity should be respected for both (which I would prefer ;-)

Corresponding lines in sBasket.php: 2201 (gross amount calculation) and 2208 (net amount calculation)


janbuecker:

PR #568, opened 2016-05-23
Remove plugin manager timeouts

cjuner:
Re-created PR #360 for a rebase on top of the 5.2 branch.

Original PR text:

> Currently the plugin manager uses the default timeout settings when installing/updating/uninstalling plugins. That is, the default PHP timeout (e.g. 30 seconds) for the operation itself, the Ext JS default timeout for AJAX requests (30 seconds) as well as the default timeout of 15 seconds for hiding the loading animation. When running e.g. a plugin installation/update/uninstallation with a duration of > 30 seconds, currently the following happens:
> - After 15 seconds: the loading animation disappears, although the request is still pending.
> - After 30 seconds: The AJAX requests gets cancelled, although the operation on the server is still running.
> - After PHPs timeout: The operation gets cancelled.
> 
> This behaviour is not desirable for several reasons:
> - The user thinks e.g. the plugin installation is finished, as soon as the loading mask is hidden (after max. 15 seconds). However, the plugin is still marked as not installed. Clicking the install button again might cause problems because of two concurrent installation operations.
> - Success messages are not displayed, because the AJAX request is cancelled.
> - Cancelling the install operation itself might result in e.g. a corrupt database.
> 
> This PR fixes the described behaviour by
> - Removing the timeouts of PHP operations for the respective actions
> - Disabling the autoTimeout of the loading mask
> - Seting the timeout of the respective AJAX request to a reasonable long time of 5 minutes

### DIFF
--- a/engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/controller/plugin.js
+++ b/engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/controller/plugin.js
@@ -157,7 +157,7 @@ Ext.define('Shopware.apps.PluginManager.controller.Plugin', {
 
         me.authenticateForUpdate(plugin, function() {
             me.startPluginDownload(plugin, function() {
-                me.displayLoadingMask(plugin, '{s name=execute_update}Plugin is being updated{/s}');
+                me.displayLoadingMask(plugin, '{s name=execute_update}Plugin is being updated{/s}', false);
                 me.executePluginUpdate(plugin, function() {
 
                     Shopware.app.Application.fireEvent('load-update-listing', function() {
@@ -434,7 +434,7 @@ Ext.define('Shopware.apps.PluginManager.controller.Plugin', {
 
     authenticateForUpdate: function(plugin, callback) {
         var me = this;
-        
+
         if (plugin.flaggedAsDummyPlugin()) {
             callback();
         } else {
@@ -450,10 +450,10 @@ Ext.define('Shopware.apps.PluginManager.controller.Plugin', {
             { technicalName: plugin.get('technicalName') },
             function(response) {
                 me.handleCrudResponse(response, plugin);
-                if (typeof callback == 'function') {
-                    callback(response);
-                }
-            }
+                callback(response);
+            },
+            null,
+            300000
         );
     },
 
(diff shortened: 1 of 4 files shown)


bcremer:
@cjuner, please don't recreate Pull requests that are still open. We will rebase the open PRs as needed.

samuelvogel:
Sorry, this was a misunderstanding on our side, we actually just wanted to reopen some old PRs from Shopware 4 times, which where closed with the branch switch to Shopware 5.

bcremer:
Nevermind :)

Can you please close #360 then.

samuelvogel:
Can you do this as admin? @svenmuennich is not available right now.

mitelg:
@samuelvogel done

samuelvogel:
Thanks! The `UPGRADE.md` was still outdated, I just fixed that.

engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/controller/plugin.js
```diff
@@ -450,10 +450,10 @@ Ext.define('Shopware.apps.PluginManager.controller.Plugin', {
             { technicalName: plugin.get('technicalName') },
             function(response) {
                 me.handleCrudResponse(response, plugin);
-                if (typeof callback == 'function') {
```
janbuecker, line 453: why did you remove the type check condition?

svenmuennich, line 453: It's been a while since I originally changed this, but I think I removed it since all calls to `authenticateForUpdate()` provide a callback. Furthermore [there's no type check, if the passed plugin is a dummy plugin](https://github.com/shopware/shopware/pull/568/files/02bb72de862a31eeb9900203d445f93ce345c647#diff-2e1215078200965f1d09e9a415518fc6L439).

janbuecker, line 453: That doesn't mean that it is best practise to don't do it. ;-) But I think it won't break anything since this would be the only call which has this check. I just wanted to know if you had a specific purpose for this removal.

janbuecker:
Merged with 64de514edc43c37aedc5a84debb48e84fd519825.

PR #569, opened 2016-05-24
Change sCategories::convertCategory() from private to public

FiveDigital:
Change sCategories::convertCategory() from private to public to be hookable


FiveDigital:
ping @OliverSkroblin @bcremer

PR #571, opened 2016-05-24
Created additional Smarty-block for configs

agreif84:
I needed an easier way to change the configs in this template-file.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,9 @@
 # Shopware Upgrade Information
 In this document you will find a changelog of the important changes related to the code base of Shopware.
 
+## 5.1.7
+* Added new config-block in `themes/Frontend/Bare/widgets/emotion/index.tpl`.
+
 ## 5.1.6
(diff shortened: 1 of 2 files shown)


screeny05:
this is rather similiar: #497, but i like your solution more.

shyim:
We should mention it on the UPGRADE.md to inform the template developers about the new block.

agreif84:
Thanks! I upgraded the file as mentioned.
@screeny05 My solution is for the 5.1-branch, yours is for the newer (5.2). That template-file has been refactored since then.

screeny05:
I see, i'll close my pr, as shopware already implemented this change by themselves.

[5.2/themes/Frontend/Bare/widgets/emotion/index.tpl](https://github.com/shopware/shopware/blob/5.2/themes/Frontend/Bare/widgets/emotion/index.tpl#L7)

janbuecker:
Thanks for your PR.
Like @screeny05 already mentioned, this has already been implemented recently and will be available as of Shopware 5.2 BETA 2.

PR #572, opened 2016-05-24
Skip stats folder on update

ChristopherDosin:
In ISPConfig there is always a `stats` folder. Because this folder does not belong to Shopware, Shopware is blocking Updates everytime, so you have to change this line manually in the Core file.

It would be good to merge this pull request to skip this folder in future updates.


janbuecker:
Thanks @kayyyy 

PR #573, opened 2016-05-25
Updated add product to compare click event target

lukiwolski:
I've run into an issue where adding article to compare list doesnt work when parent dom with data-product-compare-add="true" has some children dom elements. In your demo page this is observable when clicking icon, not the text of compare button.
This simple change checks for attribute on the event source, instead of clicked children


htkassner:
Thank you for your contribution. Your repository is not available anymore, so I couldn´t import your code. I will close this PR now. Is this still an issue for you in 5.2? Maybe you would like to contribute to 5.2 if this is still relevant.

PR #574, opened 2016-05-25
Check if vouchers are available on cart_footer.tpl

stevehempel:
With the settings you can say, that no additional options should be shown on checkout. This means that for example no voucher shoul be possible. When you change this setting to "no" vouchers will be shown on cart. But I think that is not correct. Actual you have to disable vouchers by editing or extending cart_footer.tpl
To change this behaviour I added the "if" from confirm to cart_footer.tpl


janbuecker:
Thanks @stevehempel 

PR #576, opened 2016-05-25
Uninstall and Cache Invalidation Options for new plugin system

shyim:
I added in this Pull Request a new method uninstall in the plugin class. This is needed in some requirements like drop tables on uninstall, mail templates etc...

My second point was to add a cache invalidation in the plugin.xml.
Example:

``` xml
<cache-invalidation>
        <invalidate on="install">
            <cache>config</cache>
            <cache>proxy</cache>
        </invalidate>
        <invalidate on="activate">
            <cache>theme</cache>
        </invalidate>
        <invalidate on="uninstall">
            <cache>config</cache>
            <cache>proxy</cache>
        </invalidate>
        <invalidate on="saveConfig">
            <cache>config</cache>
        </invalidate>
    </cache-invalidation>
```

Special Thanks to @t2oh4e which helps me with xsd.......

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
@@ -81,7 +81,10 @@ class PluginInstaller implements PluginInstallerInterface
             $this->em->flush($plugin);
         });
 
-        return true;
+        return [
+            'success' => true,
+            'invalidateCache' => $this->getCacheActions($pluginBootstrap, 'install')
+        ];
     }
 
     /**
@@ -129,7 +132,13 @@ class PluginInstaller implements PluginInstallerInterface
         $this->removeFormsAndElements($pluginId);
         $this->removeEmotionComponents($pluginId);
 
-        return $this->getPluginByName($plugin->getName())->uninstall();
+        $pluginBootstrap = $this->getPluginByName($plugin->getName());
+        $pluginBootstrap->uninstall();
+
+        return [
+            'success' => true,
+            'invalidateCache' => $this->getCacheActions($pluginBootstrap, 'uninstall')
+        ];
     }
 
     /**
@@ -160,6 +169,11 @@ class PluginInstaller implements PluginInstallerInterface
 
             $this->em->flush($plugin);
         });
+
+        return [
+            'success' => true,
+            'invalidateCache' => $this->getCacheActions($pluginBootstrap, 'update')
+        ];
     }
 
     /**
@@ -170,7 +184,12 @@ class PluginInstaller implements PluginInstallerInterface
         $plugin->setActive(true);
         $this->em->flush($plugin);
 
-        return true;
+        $pluginBootstrap = $this->getPluginByName($plugin->getName());
+
+        return [
+            'success' => true,
+            'invalidateCache' => $this->getCacheActions($pluginBootstrap, 'activate')
+        ];
     }
 
     /**
@@ -181,7 +200,12 @@ class PluginInstaller implements PluginInstallerInterface
         $plugin->setActive(false);
         $this->em->flush($plugin);
 
-        return true;
+        $pluginBootstrap = $this->getPluginByName($plugin->getName());
+
+        return [
+            'success' => true,
+            'invalidateCache' => $this->getCacheActions($pluginBootstrap, 'deactivate')
+        ];
     }
 
     /**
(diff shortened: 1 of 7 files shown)


bcremer:
@OliverSkroblin @dnoegel What is your opinion on that strategy?

I remember one argument of us was that you might want to invalidate some caches only for some update steps but not for all.

FiveDigital:
From my point of view the cache invalidation for updates should be handled by the migrations. That way each migration can invalidate the caches needed for whatever is done in the migration itself e.g. setting configuration, maniplating something.

The one method for all updates approach should be dropped imho in flavour of the migrations. That's the way we have done it with the update classes in our ReplyBase plugin: The `update()` methods checks for new update classes (or migrations in the new plugin system) and applies them. Every update tells which caches has to be invalidated.

Edit: And I think the three nested `foreach` loops can be replaced with a handy xpath query.

shyim:
Migrations is a beautiful idea, but I guess that is not planed ( I questioned it last week on irc) My idea were two new attribute fromVersion and toVersion as optional in XML to handle the caches on plugin update.

dnoegel:
Hey @shyim and @FiveDigital,

currently we are discussing to have migrations / update files for the 5.2 final. It depends a bit on the workload - but at least from my side it is a "must have". On the other hand, there is no hurry bloating the system - one could still add it with a 5.2.1 as a non-breaking change. 

Either way: We already discussed a system with install/update context objects, that will be injected to the update/install methods OR the migrations (if implemented). Those context objects will allow you to schedule cache clear operations conveniently and also to make it version dependend (e.g.: if SW 52 && Plugin-Version 2.0. && $MOON_PHASE == "full moon": clear http cache).

I really appreciate your work @shyim, but currently I am not convinced about the XML clear-cache thingy…

dnoegel:
Oh, but we shouldn't forget this "uninstalll" thing. Looks quite useful for me :)

FiveDigital:
👍  For migrations
👍  For context objects

Can't wait to see you implementation since we already planned to build this for our ReplyBase plugin. We had some discussions about topics like how to name the migration files etc. and will provide feedback once the code is on GitHub.

shyim:
@dnoegel Why are you not convinced with xml clear-cache :D? Because its xml :P?

dnoegel:
@shyim: "Clear the cache" is another kind of issue then "My menu item is called 'test'" or "My plugin is named 'foo'". The first described behavior, the latter a state. As Thomas already stated, one wants to _control_ when / how caches are cleared and in order to do that, you'd need to add all kind of attributes like "fromShopwareVersion", "toShopwareVersion", "fromOldPluginVersion", "toOldPluginVersion" etc. The cache invalidation might even depend on other plugins being installed or not?

Don't get me wrong, I don't want to emphasize this behavior vs state difference too much and if there are good arguments for it, I will happily encode those things into XML as well. But if the situation is, that we need to have some sort of proper cache-invalidation handling in the install/update/migration system anyway, then I don't see why we should start with the XML cache-invalidation in the first place.

Is there anything that XML cache-invalidation can do for us, that programmatic cache invalidation can't?

FiveDigital:
I also want to add that the XML files will become very large when you have many plugin updates. In our internal plugins we often have more than 50 update files and when the new migration system hopefully offers the ability to set store specific shop configuration it will become much more.

OliverSkroblin:
Hey,

we discussed if it makes sense that the plugin developer has to decide which different kind of caches has to be cleared.
So we took a look to our caches and which are interesting for plugin developers:
- config cache => has no performance impact to clear each time
- proxy cache => same here, no performance impact
- router => only interesting if you have a "live" generation
- theme => makes only sense if HTTP and TEMPLATE are cleared too
- http => this is critical for a live system, but the customer has to decide this at the end
- template => as stand alone cache, only intersting for backend

With this list of caches we had the idea that config, proxy, router and template caches have to be cleared every time on plugin crud operations and the plugin developer can define if the http and theme, the `frontend` cache (the critical for live systems), has to be cleared too.

What do you think about this solution?

shyim:
It is a great idea, but how we can define that http and theme cache should be cleared?
In the xml way or @dnoegel said programmatic way :)
It would be interesting for my pr

FiveDigital:
As a developer I always like being in full control of what happens. So in my opinion: 
- it's okay to clear the caches @OliverSkroblin mentioned by default in every migration
- it should be possible to define which caches should be cleared in every migration optionally
- that enables me to say: don't clear any cache in this migration or: rebuild the theme and clear the http cache afterwards

Thinking in code I have an `abstract class Migration` that defines the caches to be cleared by default in perhaps a `protected $cachesToClear = ['config', 'proxy', 'router'];` member that I can override in my concrete migration class that extends the `Migration`.

Edit: I also really appreciate having the possibility to talk with you about the architecture here 👍

OliverSkroblin:
Hey, we implemented a context class which are provided to the activate/deactivate/install/uninstall function of each plugin. The context contains a collection of scheduled tasks like clearCache or display message.
You can see the implementation here:
https://github.com/shopware/shopware/commit/b93312fa61005f080eb628f4e997097299a8ea5e

OliverSkroblin:
An example could looks like this:

```
<?php

namespace SwagTest;

use Shopware\Components\Plugin\PluginContext;

class SwagTest extends Plugin
{
    public function install(PluginContext $context)
    {
        $context->scheduleClearCache(PluginContext::CACHE_LIST_DEFAULT);
    }

    public function activate(PluginContext $context)
    {
        $context->scheduleClearCache(PluginContext::CACHE_LIST_FRONTEND);
    }

    public function uninstall(PluginContext $context, $keepUserData)
    {
        parent::uninstall($context, $keepUserData);
    }
}
```

You can completely overwrite each function and handle the cache invalidation by your self, or you can call the parent function which clears in default the `PluginContext::CACHE_LIST_DEFAULT` cache list.

PR #577, opened 2016-05-27
Stronger Cron Security Rules

shyim:
Most Customers doesnt define a Cronkey. That is a big vulnerability for informations like Email-Addresses or something else, which is printed by the crons. Every User can go to a shop and request the page /backend/cron.
This PR disables the webcron access, when all security fields (allowed key, allowed ip, secured by account) are empty.


janbuecker:
Thanks for your PR. 

I like the idea! +1

janbuecker:
This PR will introduce a major break and should only be activated for install instances. I believe, that update customers will get into trouble because their cronjobs are not running anymore and they don't know why.

/cc @bcremer

shyim:
@janbuecker i guess the easiest method is a unvisible plugin config element to avoid the trouble with old installations.
I also changed arrays to short syntax and changed the base branch to 5.3

janbuecker:
This PR was closed because your change implements a feature which does not fit to our roadmap and therefore won't be implement in the near future.

PR #581, opened 2016-06-01
Make order and basket 'partnerID' nullable

svenmuennich:
This PR makes the fields `s_order.partnerID` and `s_order_basket.partnerID` nullable and sets their default values to `NULL`. Previuosly the `partnerID` was set to an empty string by default, which makes not much sense, since it mapps a foreign key relation to `s_emarketing_partner`. Besides making the field nullabel, I also changed the code to not cast `partnerID`s to strings, but pass the value as it is, when creating an order, which preserves `NULL` values upon `INSERT`ing.

_Note: This change was first proposed in PR #196, but never merged._


janbuecker:
Thanks for your PR.

Could you please remove your changes to the `latest.sql` file since we do not update this file.

svenmuennich:
Just to make sure: I changed `_sql/demo/latest.sql`, [which is changed frequently](https://github.com/shopware/shopware/commits/5.2/_sql/demo/latest.sql). I thought only `_sql/install/latest.sql` is not supposed to change, but the demo data should be adjusted.

@janbuecker Can you please confirm, that I should revert my changes in `_sql/demo/latest.sql`?

janbuecker:
Whoops, my fault. Misread that, you're correct :) 👍

svenmuennich:
Okay, thank you for clarifying 👍

svenmuennich:
Please note that the error reported in [my first comment](https://github.com/shopware/shopware/pull/581#issue-157825275) was [caused by a corrupt composer `autoloader.php`](https://github.com/shopware/shopware/commit/9138d7ef7d00ec7ee6b61e80f904de2cc653c30d#commitcomment-17944954) in my shopware system. Nevertheless I think it makes sense to make the `partnerID` `NULL`able.

janbuecker:
I think we should split up the migrations into 2 files because of process timeouts. We aim for 30 seconds or less per migration. Shops with a large amount of orders may run into issues.

What do you think @bcremer?

janbuecker:
After reviewing, is there any reason why you left out the `s_order_basket` table? there is a `partnerID` too.

svenmuennich:
I guess I just missed that one.

janbuecker:
Then you know what to do :)

janbuecker:
@svenmuennich can you update this PR please? Otherwise it will be closed within the next days.

svenmuennich:
Sorry for the delay @janbuecker, but I've been quite busy for the last days. I added the following changes:
- Rebase the branch to the current HEAD of 5.2 (39212c31d0167b76ca64bcdc32324f29662694fb)
- Make `s_order_basket.partnerID` `NULL`able
- Split up the migration in three files:
  1. Update the fields `s_order.partnerID` and `s_order_basket.partnerID` to be nullable and `NULL` by default
  2. Replace empty string values in `s_order.partnerID` with `NULL`
  3. Replace empty string values in `s_order_basket.partnerID` with `NULL`
- Fixed an analytics query condition
- Add changes to `UPGRADE-5.2.md`

janbuecker:
Our tests with elasticsearch failed with your changes as well as the mink tests. Can you please describe, why do you want to introduce this change?

svenmuennich:
Initially I thought that making `partnerID` nullable fixes an exception, which occurred when saving an order in the backend. But as it turned out, the exception was caused by something else. I think in general it is just more correct to `NULL` the field by default, making it easier to create a new order using Doctrine. E.g. the [Backend orders plugin](https://github.com/shopware/sw-swagbackendorder) currently [works around this problem](https://github.com/shopware/sw-swagbackendorder/blob/00d9afb33adca517563eaa07d81453dd49851420/Components/CreateBackendOrder.php#L34-L38).

3stadt:
Hi, since this PR doesn't fix your problem and the benefit is rather small while breaking the ES testsuite, we will close it.

samuelvogel:
This change was now partly adopted by 386215a6007c086366f16d336b29e57d37d25665.

PR #582, opened 2016-06-01
Fix recalculation of invoice amount for taxfree orders

andrewisplinghoff:
This fixes a bug concerning tax-free orders, to reproduce proceed as follow:
- Generate a tax-free order. The invoice amount will be equal to the net invoice amount, which makes sense since no taxes apply.
- Make changes to the order positions, e.g. add a new order position and save it
- **Expected behaviour:** the order amount still is a net amount without taxes
- **Actual behaviour:** the order amount is now a gross amount including taxes, which does not make sense, since this is a tax-free order

This PR also makes the code easier to understand, since previously the variable `$invoiceAmountNet` was used to sum up gross amounts in case of a net order, which is quite confusing.

### DIFF
--- a/engine/Shopware/Models/Order/Order.php
+++ b/engine/Shopware/Models/Order/Order.php
@@ -1039,7 +1039,12 @@ class Order extends ModelEntity
         //iterate order details to recalculate the amount.
         /**@var $detail Detail*/
         foreach ($this->getDetails() as $detail) {
-            $invoiceAmount += $detail->getPrice() * $detail->getQuantity();
+            // For net invoices, the price of the order details is the net price
+            if ($this->net) {
+                $invoiceAmountNet += $detail->getPrice() * $detail->getQuantity();
+            } else {
+                $invoiceAmount += $detail->getPrice() * $detail->getQuantity();
+            }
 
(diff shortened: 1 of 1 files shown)


hlohaus:
Hey,
internally there is already a pull request for the ticket SW-14351.

`if ($this->taxFree) {`
`$this->invoiceAmountNet = $invoiceAmount + $this->invoiceShippingNet;`
`$this->invoiceAmount = $this->invoiceAmountNet;`
`} elseif ($this->net) {`
`$this->invoiceAmountNet = $invoiceAmount + $this->invoiceShippingNet;`
`$this->invoiceAmount = $invoiceAmountNet + $this->invoiceShipping;`
`}`...

andrewisplinghoff:
As far as I can see, both modifications are equivalent with regards to functionality. I would still propose to merge this PR as it fixes the confusing variable naming.

samuelvogel:
This PR is obsolete now since https://github.com/shopware/shopware/commit/bdc1c83a3b52ada60503dacdfdaced39e8d393e3 was merged.

bcremer:
Thank you @samuelvogel, closed this PR now.

PR #583, opened 2016-06-02
Check relations before translation them in article resource

Haehnchen:
Following error occurs using article resource with translation option, because article relation array are not fully validated.

```
            $article = \Shopware\Components\Api\Manager::getResource('Article')
                ->getOneByNumber($id, ['language' => 2]);
```

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Shopware\Components\Api\Resource\Article::translateAssociation() must be of the type array, null given
```


janbuecker:
Thanks, merged with 2b17692efc5070b6a8df71a2820d81669753ce19.

PR #584, opened 2016-06-02
[BUGFIX] No files as plugin

xf-:
Don't allow files like .gitkeep to be tested as plugin. isDir and isLink should be OK

Avoid errors like

``` php
Warning: is_file(): open_basedir restriction in effect. File(/var/www/plugins/.gitkeep/.gitkeep.php) is not within the allowed path(s): (/var/www:[MORE]) in /var/www/engine/Shopware/Kernel.php on line 311
```

### DIFF
--- a/engine/Shopware/Kernel.php
+++ b/engine/Shopware/Kernel.php
@@ -312,7 +312,7 @@ class Kernel implements HttpKernelInterface
 
         $pluginRoot = $this->getRootDir().'/custom/plugins';
         foreach (new \DirectoryIterator($pluginRoot) as $pluginDir) {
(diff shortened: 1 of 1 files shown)


janbuecker:
Seems to be related to the bug described here https://issues.shopware.com/#/issues/SW-15145

xf-:
Changed title to BUGFIX (is listed as bug) and added related to in message

bcremer:
This should be fixed with commit: https://github.com/shopware/shopware/commit/7bbe3a99faf79708bfc0a5726b2272e127becbc4

PR #586, opened 2016-06-03
Add description to "lfm" (linear metre) unit of measure.

oktupol:
Added a description for the 'lfm'-unit in the Database-Table `s_core_units`, as well as an english translation to the unit and the abbreviation.


janbuecker:
Has been merged into 5.2

PR #588, opened 2016-06-06
translate replacement of '&' in SEO URLs

uehler:
If you have a non-german shop with categories like "Foo & Bar" then the seo url looks like "/foo-und-bar". With this request you can translate the replacement of the ampersand for each shop, so the url for an english shop would look like "/foo-and-bar".

Feel free to change the namespace of the snippet if you are not happy with it.

There is also a ticket in your issue tracker: https://issues.shopware.com/#/issues/SW-14378


bcremer:
This is related to: https://github.com/shopware/shopware/pull/552

uehler:
Oh you're right, i didn't see that request.
But in the other solution the "&" would just be removed instead of replaced.

bcremer:
Closed as duplicate.
For the Shopware 5.2.4 release we will introduce a new interface which can be decorated/replaced. Please see this comment  https://github.com/shopware/shopware/pull/552#issuecomment-234969957 and accordingly this issue https://issues.shopware.com/#/issues/SW-15039.

PR #590, opened 2016-06-07
Fix ExtJs grid selection on row update

svenmuennich:
Previously, when using a grid panel with both a selection model and e.g. the row editor plugin, the appearance of a row might have not reflected the underlying selection model, when updating the row. That is, prior to ExtJs `4.1.2` a selected row would appear as not selected, after updating it using e.g. the row editing plugin, although the row is still contained in the selection. This is a known issue and was also discuss in [Sencha's support forum](http://www.sencha.com/forum/showthread.php?244974-Grid-loses-%28visible%29-selection-after-store-record-edit-and-after-comit). This PR adds an ExtJs override, which fixes the problem by using the solution provided in the linked support forum topic. The respective lines are marked with a 'Fix:' comment.


Phil23:
Thx @svenmuennich,

the fix will be merged in the next 5.2 patch release.

PR #591, opened 2016-06-08
[FEATURE] Add filter event for getSearchTerm in Search controller

PortaltechGithub:
We need to modify the searchTerm when productNumberSearch is active. I hope the naming is correct.

Background:
We are prefixing the article number in the frontend with a string. So we need a filter here to easyliy modify the search term.

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Search.php
+++ b/engine/Shopware/Controllers/Frontend/Search.php
@@ -148,12 +148,9 @@ class Shopware_Controllers_Frontend_Search extends Enlight_Controller_Action
         //we have to strip the / otherwise broken urls would be created e.g. wrong pager urls
         $term = str_replace("/", " ", $term);
 
-        $term = \Shopware()->Container()->get('events')->filter(
+        $term = Shopware()->Container()->get('events')->filter(
             'Shopware_Controllers_Frontend_Search_getSearchTerm_Term',
-            $term,
-            [
-                'subject'      => $this
-            ]
+            $term
         );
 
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Frontend/Search.php
```diff
@@ -148,6 +148,14 @@ private function getSearchTerm()
         //we have to strip the / otherwise broken urls would be created e.g. wrong pager urls
         $term = str_replace("/", " ", $term);
 
+        $term = \Shopware()->Container()->get('events')->filter(
+            'Shopware_Controllers_Frontend_Search_getSearchTerm_Term',
+            $term,
+            [
+                'subject'      => $this
```
OliverSkroblin: please remove the subject event parameter, so it is easier to move this event at a later moment to another class without breaking parameters.

janbuecker:
Hi @PortaltechGithub, thanks for your PR.

Unfortunately, your change does not affect the search since the conditions are added in `Shopware\Bundle\SearchBundle\CriteriaRequestHandler\CoreCriteriaRequestHandler::addSearchCondition`.

You should rework this to make it work with the store front bundle.

bcremer:
This has been resolved with the following merge: https://github.com/shopware/shopware/commit/f83675c10a9.

This merge introduced a new interface
`\Shopware\Bundle\SearchBundle\SearchTermPreProcessorInterface` (Id: `shopware_search.search_term_pre_processor`) that can be replaced or decorated.

PR #592, opened 2016-06-09
[TASK] htmlencode snippet translation emptyText

Tombro:
Hey folks,
having some HTML as a Snippet value breaks the translations in the backend.
See Screenshot:

snippettranslationbroken

This is because the value `<strong>bold example text</strong>` is used as the emptyText of the translations Textarea, which will become the placeholder attribute of the element. HTML in the placeholder attributes breaks the DOM. Encoding the value solves this issue.

Unfortunally there is a bug when you focus in and out a textarea the raw html-entities will be displayed. I think this is an ExtJs Bug as this behavior is reproducible in a Sencha Fiddle even with the newest ExtJs Version. But anyway a broken placeholder is way better than a broken value.

By the way the fieldStyle property fixes a chrome bug which would resize the textareas on focus.

Regards,
Thomas

### DIFF
--- a/themes/Backend/ExtJs/backend/snippet/view/main/translate_form.js
+++ b/themes/Backend/ExtJs/backend/snippet/view/main/translate_form.js
@@ -103,8 +103,9 @@ Ext.define('Shopware.apps.Snippet.view.main.TranslateForm', {
             fieldSetItems.push({
                 fieldLabel: locale.get('displayName'),
                 name: translation.internalId,
-                emptyText: translation.get('defaultValue'),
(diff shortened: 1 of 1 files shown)


Tombro:
Ping.

bcremer:
@Tombro please do not ping. This just generates useless noise in our inboxes. As you can see we currently have 102 open pull requests that we are trying to merge as fast as we can.

janbuecker:
Thanks @Tombro. Fixed with 56443027b469aa25b5888042ca53650edabb61b2.

PR #596, opened 2016-06-13
Allow viewing core and plugin log files in Backend log module

cjuner:
# Why:
- Shopware administrators should have an easy option to review logs
- Shopware administrators should be able to easily copy log entries to communicate them to
  plugin authors.
# What:
- All **logging** functionality based on the `monolog.handler.rotating_file.plugin` service, including `corelogger` and `pluginlogger`, use a **structured JSON format** for output in their log files.
- There are now multiple tabs in the Backend's log viewer:
  shopware ___version___ ___version_text___ rev ___revision___ - backend c shopware ag - chromium_023
- `corelogger` log entries are now visible in the Backend's log viewer:
  shopware ___version___ ___version_text___ rev ___revision___ - backend c shopware ag - chromium_024
- `pluginlogger` log entries are now visible in the Backend's log viewer. 
- The Backend's log viewer can display both the new JSON format and the legacy line-based format.
- Exceptions related to core and plugin log entries and their causes (previous exceptions) can be seen in a detail view: shopware ___version___ ___version_text___ rev ___revision___ - backend c shopware ag - chromium_022
- A log entry can be copied with the click of a button.
# Breaking something?
- Backend API end point renamed: `/log/getLogs` -> `/log/getBackendLogs`
- Backend API end point renamed: `/log/deleteLogs` -> `/log/deleteBackendLogs`


bcremer:
I appreciate your work, that looks like a very useful feature to me.

My biggest gripe with this PR is the changed output format:

```
# old
[2016-07-14 07:47:13] core.ERROR: This a test {"foo":"bar"} {"uid":"3a8f227"}

# new
{"level":"ERROR","timestamp":"2016-07-14T05:47:03+0000","message":"This a test","context":{"foo":"bar"},"uid":"94ca7d8"}
```

This is a breaking change that will break existing tooling. Also I find it much harder to read as a human while `tail`ing the logfile.

Is the JSON format really necessary? 

Also please consider to extract some methods from `engine/Shopware/Controllers/Backend/Log.php` into a separate log-parser that can be unit tested.

3stadt:
Awesome feature! :+1: 

One thing is missing, you should escape the content.  
When displaying plugin and core logs, we have no control what messages are shown to the backend user.

If a plugin logs raw user content, e.g. input from somewhere, malicious content would be executed in the backend atm.

Example: Create  a (legacy) log message which has the following message string.

`Stuff happened: <span onClick="alert('Foobar');">Foobar</span>`

When clicking on the span in log view, JS is executed.

Please change your PR so JS/HTML is removed before display.

janbuecker:
@svenmuennich @cjuner what's the status here? 

Can you update this PR soon, please? Otherwise, it will be closed within the next days without action.

svenmuennich:
Sorry @janbuecker, I haven't got any notifications until you _@mentioned_ me. I will do the requested changes in the next days.

bcremer:
@svenmuennich please also see my comment: https://github.com/shopware/shopware/pull/596#issuecomment-232564498.

I don't think we should change the log format to json in a minor version.
So for 5.2 I would like to go just with the new backend module.

svenmuennich:
I did all the requested changes:
- Remove the custom `JsonFormatter` and use a customized `LineFormatter` instead (this allows to show better error information)
- Refactor the parsing of log entries into two separate classes `Shopware\Components\Log\Parser\FileReader` and `Shopware\Components\Log\Parser\LineFormatParser`
- Escape the log entry string before it's parsed to make it safer to display its content in the shopware backend
- Add unit tests for `LineFormatParser`

@bcremer @janbuecker Shall I squash all the commits as usual?

engine/Shopware/Components/Log/Parser/LineFormatParser.php
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * Match 'extra' log field JSON. We match any string beginning with '{' and preceding
+     * whitespace and ending with '}'. This preceding whitespace condition prevents us from
+     * matching inner objects inside the complete "extra" JSON object. Within the object, we
+     * do allow '{' characters, but only if they are not preceded by whitespace. This means
+     * we do not allow JSON containing ' {' in a string. An alternative would be to find all
+     * '\s{' backwards and try parsing each potential match as JSON until we find valid JSON.
+     *
+     * @param &string $line
```
bcremer: Please remove the preceding ampersand.

bcremer:
@svenmuennich nice work. Yes, please go ahead an squash into logical commits. I will do a in depth code review next because the ticket was move out of the current sprint but on first sight it looks fine for me.

svenmuennich:
@bcremer What do you think about 1d288e7c417f23c3b4c04207ff0457a2e69b3b25? I noticed during my testing that _many_ exceptions were not logged correctly, because they had been thrown outside the dispatch loop. That is, exceptions thrown in subscribes on `Enlight_Controller_Front_StartDispatch`, `Enlight_Controller_Front_RouteStartup` etc. were not logged.

tests/Unit/Components/Log/Parser/LineFormatParserTest.php
```diff
@@ -0,0 +1,276 @@
...
+        // Extra
+        $this->assertArrayHasKey('uid', $result['extra']);
+        $this->assertArrayHasKey('some', $result['extra']);
+        $this->assertArrayHasKey('id', $result);
+        $this->assertEquals($parts['extra']['uid'], $result['id']);
+    }
+
+    /**
+     * @param array $expected
+     * @param array $actal
```
bcremer: Typo: actal -> actual

engine/Shopware/Components/Log/Parser/LineFormatParser.php
```diff
@@ -0,0 +1,201 @@
...
+        if (is_array($logEntry['extra']) && isset($logEntry['extra']['uid'])) {
+            $logEntry['id'] = $logEntry['extra']['uid'];
+        }
+
+        return $logEntry;
+    }
+
+    /**
+     * @param string $line
+     * @return array
```
bcremer: The return type hint should be `\DateTime`

engine/Shopware/Components/Log/Parser/LineFormatParser.php
```diff
@@ -0,0 +1,201 @@
...
+        ];
+
+        try {
+            // Parse the components
+            $logEntry['timestamp'] = $this->parseTimestamp($line);
+            $logEntry['level'] = $this->parseLevel($line);
+            $logEntry['extra'] = $this->parseExtra($line);
+            $logEntry['message'] = $this->parseMessage($line);
+            $logEntry['context'] = $this->parseContext($line);
+        } catch (ParseException $e) {
```
bcremer: `\Shopware\Components\Log\Parser\ParseException` should be removed.
It's only used inside the private methods of `LineFormatParser` and always catched in the only public method `parseLine`.

Should be replaced with a inbuilt `InvalidArgumentException`.

engine/Shopware/Controllers/Backend/Log.php
```diff
@@ -165,4 +167,64 @@ public function createLogAction()
...
+    /**
+     * @param string $type
+     * @param int $offset
+     * @param int $limit
+     * @param boolean $sortAscending
+     * @return array
+     */
+    private function parseLog($type, $offset, $limit, $sortAscending)
+    {
+        $logReader = new FileReader(
```
bcremer: The log reader instantiation should happen in the DI-Container. Please move the log file reader into the `logger.xml` di configuration. See https://github.com/shopware/shopware/pull/596/files#r73322385.

engine/Shopware/Components/Log/Parser/FileReader.php
```diff
@@ -0,0 +1,114 @@
...
+    private $logFiles = [];
+
+    /**
+     * @param string $logsPath
+     * @param string $environment
+     * @param string $loggerType
+     */
+    public function __construct($logsPath, $environment, $loggerType)
+    {
+        $this->lineParser = new LineFormatParser();
```
bcremer: `LineFormatParser` should be injected in the constructor. Can be a private service in `logger.xml`.

engine/Shopware/Components/Log/Parser/FileReader.php
```diff
@@ -0,0 +1,114 @@
...
+ */
+class FileReader
+{
+    /**
+     * @var LineFormatParser $lineParser
+     */
+    private $lineParser;
+
+    /**
+     * @var array $logFiles
```
bcremer: `array` doctypes should be replaced with a more specific type where possible.
In this case it should be `string[] $logFiles`.

engine/Shopware/Components/Log/Parser/FileReader.php
```diff
@@ -0,0 +1,114 @@
...
+        $logFiles = array_filter($files, function ($fileName) use ($pattern) {
+            return preg_match($pattern, $fileName) === 1;
+        });
+        $this->logFiles = array_map(function ($fileName) use ($logsPath) {
+            return $logsPath . '/' . $fileName;
+        }, $logFiles);
+    }
+
+    /**
+     * @param int $offset (optional)
```
bcremer: Please remove the `(optional)` comment. We usually don't use that convention.

engine/Shopware/Components/Log/Parser/LineFormatParser.php
svenmuennich: GitHub somehow doesn't recognize it, but I fixed that comment in 9653f0537585c0df9271b5db11bb34799e220c2c.

bcremer:
@svenmuennich In regards of https://github.com/shopware/shopware/commit/1d288e7c417f23c3b4c04207ff0457a2e69b3b25.

This looks like an good idea. I'll have to check what happens if the `logger` throws an exception itself. That may swallow the original exception.

tests/Unit/Components/Log/Parser/LineFormatParserTest.php
```diff
@@ -0,0 +1,276 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Tests\Components\Escaper;
```
bcremer: Copy'n'Past error.

engine/Shopware/Components/Log/Parser/LineFormatParser.php
```diff
@@ -0,0 +1,201 @@
...
+     * If a component cannot be parsed successfully, the log entry is assumed to be malformatted
+     * and anything that has been parsed from the line so far is returned.
+     *
+     * @param string $line
+     * @return array
+     */
+    public function parseLine($line)
+    {
+        // Escape HTML tags to make it safe to display the parsed content
+        $line = htmlspecialchars($line, ENT_NOQUOTES);
```
bcremer: Escaping belongs to the view layer. `LineFormatParser` can not now how the output has to be escaped. For example if the LineFormatParser is used in a CLI context HTML output is perfectly fine.

The escaping should be done in the extjs module.

engine/Shopware/Components/DependencyInjection/logger.xml
```diff
@@ -72,5 +79,22 @@
...
             <argument>debug</argument>
         </service>
+
+        <!-- Core log file reader -->
+        <service id="log.line_format_parser" class="Shopware\Components\Log\Parser\LineFormatParser" public="false" />
+        <service id="log.file_reader.core" class="Shopware\Components\Log\Parser\FileReader">
+            <argument type="service" id="log.line_format_parser" />
+            <argument>%kernel.logs_dir%</argument>
+            <argument>%kernel.environment%</argument>
+            <argument>core</argument>
```
bcremer: Core/Plugin are runtime arguments and should be passed to the `readEntries` method. This way we only need one service here.

engine/Shopware/Components/Log/Formatter/LineFormatter.php
```diff
@@ -0,0 +1,73 @@
...
+ * @copyright Copyright (c) shopware AG (http://www.shopware.de)
+ */
+class LineFormatter extends MonologLineFormatter
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function format(array $record)
+    {
+        if (isset($record['context']) && isset($record['context']['exception'])
```
bcremer: This `isset($record['context']) && isset($record['context']['exception']` can be shortened to `isset($record['context']['exception']`.

Shout out to @Haehnchen who taught me that in https://github.com/shopware/shopware/pull/619 :)

engine/Shopware/Components/Log/Parser/LineFormatParser.php
```diff
@@ -0,0 +1,201 @@
...
+            if (empty($logEntry['message'])) {
+                // Use 'rawLine' as fallback message
+                $logEntry['message'] = $logEntry['rawLine'];
+            }
+
+            return $logEntry;
+        }
+
+        // Move exceptions to the top level if available
+        if (is_array($logEntry['context']) && isset($logEntry['context']['exception'])) {
```
bcremer: Can be shortened, see: https://github.com/shopware/shopware/pull/596/files#r73374446

engine/Shopware/Controllers/Backend/Log.php
```diff
@@ -165,4 +167,46 @@ public function createLogAction()
...
+        $limit = $this->Request()->getParam('limit', 1000);
+        $sort = $this->Request()->getParam('sort', []);
+
+        // Parse log files
+        $sortAscending = !empty($sort) && isset($sort[0]['direction']) && $sort[0]['direction'] === 'ASC';
+        $result = $this->get('log.file_reader.plugin')->readEntries(intval($start), intval($limit), $sortAscending);
+
+        // Check result entries' context for plugin name
+        foreach ($result['data'] as &$entry) {
+            if (isset($entry['context']) && isset($entry['context']['plugin'])) {
```
bcremer: Can be shortened, see: https://github.com/shopware/shopware/pull/596/files#r73374446

engine/Shopware/Components/Log/Formatter/LineFormatter.php
```diff
@@ -0,0 +1,73 @@
...
+            && ($record['context']['exception'] instanceof \Exception || $record['context']['exception'] instanceof \Throwable)) {
```
Haehnchen, line 42: Throwable === PHP7 and also no user class is able to implement it. i see no real use case for upper user exception and finally Exception also includes Throwable  ;)

engine/Shopware/Components/Log/Formatter/LineFormatter.php
Haehnchen: :thumbsup:

engine/Shopware/Components/Log/Formatter/LineFormatter.php
bcremer, line 42: @Haehnchen Every `Exception` is a `Throwable` but not every `Throwable` is a `Exception`.
One could use this formatter format a previously caught [`Error`](https://secure.php.net/manual/en/class.error.php)

engine/Shopware/Components/Log/Formatter/LineFormatter.php
svenmuennich: Thanks for the tip @bcremer and @Haehnchen 👍

engine/Shopware/Components/Log/Formatter/LineFormatter.php
svenmuennich, line 42: I copied the check from [`Monolog\Formatter\NormalizerFormatter`](https://github.com/Seldaek/monolog/blob/5f56ed5212dc509c8dc8caeba2715732abb32dbf/src/Monolog/Formatter/NormalizerFormatter.php#L119-L121). There's also an issue about that in Seldaek/monolog#718. I think we should stick to the current check for both `Exception` and `Throwable`.

engine/Shopware/Components/Log/Parser/LineFormatParser.php
svenmuennich: I decided to escape only `<` and `>` in the view layer to keep the logs as readable as possible (274784b281bee49a34608a1b0331a5008df29e43).

svenmuennich:
@bcremer Good point. I just tested this and it is exactly as you expected: If the logger throws an exception, the original exception is swallowed. Hence I wrapped `$this->container->get('corelogger')->error($e);` in another `try/catch`, whose exception is ingored: 81cbe4bd62c6e9defa94a57956716c2c29a01ca3.

I refactored everything according to your feedback. I will go ahead an squash it into a few logical commits.

engine/Shopware/Components/Log/Parser/FileReader.php
```diff
@@ -0,0 +1,133 @@
...
+        $entries = [];
+        $logFiles = $this->findLogFiles($logType, $sortAscending);
+        foreach ($logFiles as $filePath) {
+            // Read file line by line
+            $handle = fopen($filePath, 'r');
+            if (!$handle) {
+                continue;
+            }
+            $lines = [];
+            while (($line = fgets($handle)) !== false) {
```
bcremer, line 81: Please note that this code is very (memory) inefficient.

First all files are read into memory, afterwards lines will be discarded according to limit and offset.
Given a logfile can grow multiple megabytes this should be fixed in the future.

I implemented a simple library to read file forwards and backwards here: https://github.com/bcremer/LineFileReader.
I will probably include the library after we merged this PR and replace the current code with it.

FiveDigital, line 81: This is a good point! The logfiles of our customers grow around 30-40MB per day (when debugging is enabled).

svenmuennich: @bcremer Should I include your `LineFileReader` in this PR to get rid of this last issue and get this PR merged soon?

bcremer: @svenmuennich I have this PR scheduled for the current sprint that starts on monday. I am looking forward to merge this soon. I`ll do any necessary performance enhancements later on.

bcremer:
@svenmuennich I spend some good time trying to get this merged. Please see my current status in this branch: https://github.com/shopware/shopware/tree/sw-15680/5.2/allow-viewing-core-and-plugin-log-files-in-backend-log-module. 

I ran into multiple problems that prevent me from merging this at it currently is: 
- The changes to the formatter are generation log output that is way to verbose, for an exception thrown in a controller the log line contains the whole html body in the output
- The LineFormatParser is not able to parse log lines that were generated with the old formatter

I really like the log module and in order to get this merged I propose to strip this module down to a minimum so that it acts more like an UI-Access to the files and just does minimal parsing of the log lines (maybe just the date and log level).

Once this is merged we can try to get proper exception presentation and parsing on top of that.

Example how to generate a currently non-parsable log entry:

``` php
    public function indexAction()
    {
        $e = new \InvalidArgumentException('Some Inner Exception', 9999);
        throw new \LogicException('Some Outer Exception', $e->getCode(), $e);
    }
```

This generates the following log line:

```
[2016-10-11 10:33:10] core.ERROR: InvalidArgumentException: Some Inner Exception in /home/bcremer/www/5.2/engine/Shopware/Controllers/Frontend/Detail.php:71 Stack trace: #0 /home/bcremer/www/5.2/engine/Library/Enlight/Controller/Action.php(159): Shopware_Controllers_Frontend_Detail->indexAction() #1 /home/bcremer/www/5.2/engine/Library/Enlight/Controller/Dispatcher/Default.php(523): Enlight_Controller_Action->dispatch('indexAction') #2 /home/bcremer/www/5.2/engine/Library/Enlight/Controller/Front.php(223): Enlight_Controller_Dispatcher_Default->dispatch(Object(Enlight_Controller_Request_RequestHttp), Object(Enlight_Controller_Response_ResponseHttp)) #3 /home/bcremer/www/5.2/engine/Shopware/Kernel.php(177): Enlight_Controller_Front->dispatch() #4 /home/bcremer/www/5.2/vendor/symfony/http-kernel/HttpCache/HttpCache.php(487): Shopware\Kernel->handle(Object(Enlight_Controller_Request_RequestHttp), 1, true) #5 /home/bcremer/www/5.2/engine/Shopware/Components/HttpCache/AppCache.php(255): Symfony\Component\HttpKernel\HttpCache\HttpCache->forward(Object(Symfony\Component\HttpFoundation\Request), true, NULL) #6 /home/bcremer/www/5.2/vendor/symfony/http-kernel/HttpCache/HttpCache.php(444): Shopware\Components\HttpCache\AppCache->forward(Object(Symfony\Component\HttpFoundation\Request), true) #7 /home/bcremer/www/5.2/vendor/symfony/http-kernel/HttpCache/HttpCache.php(344): Symfony\Component\HttpKernel\HttpCache\HttpCache->fetch(Object(Symfony\Component\HttpFoundation\Request), true) #8 /home/bcremer/www/5.2/engine/Shopware/Components/HttpCache/AppCache.php(178): Symfony\Component\HttpKernel\HttpCache\HttpCache->lookup(Object(Symfony\Component\HttpFoundation\Request), true) #9 /home/bcremer/www/5.2/vendor/symfony/http-kernel/HttpCache/HttpCache.php(210): Shopware\Components\HttpCache\AppCache->lookup(Object(Symfony\Component\HttpFoundation\Request), true) #10 /home/bcremer/www/5.2/engine/Shopware/Components/HttpCache/AppCache.php(114): Symfony\Component\HttpKernel\HttpCache\HttpCache->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true) #11 /home/bcremer/www/5.2/shopware.php(113): Shopware\Components\HttpCache\AppCache->handle(Object(Symfony\Component\HttpFoundation\Request)) #12 {main}  Next LogicException: Some Outer Exception in /home/bcremer/www/5.2/engine/Shopware/Controllers/Frontend/Detail.php:72 Stack trace: #0 /home/bcremer/www/5.2/engine/Library/Enlight/Controller/Action.php(159): Shopware_Controllers_Frontend_Detail->indexAction() #1 /home/bcremer/www/5.2/engine/Library/Enlight/Controller/Dispatcher/Default.php(523): Enlight_Controller_Action->dispatch('indexAction') #2 /home/bcremer/www/5.2/engine/Library/Enlight/Controller/Front.php(223): Enlight_Controller_Dispatcher_Default->dispatch(Object(Enlight_Controller_Request_RequestHttp), Object(Enlight_Controller_Response_ResponseHttp)) #3 /home/bcremer/www/5.2/engine/Shopware/Kernel.php(177): Enlight_Controller_Front->dispatch() #4 /home/bcremer/www/5.2/vendor/symfony/http-kernel/HttpCache/HttpCache.php(487): Shopware\Kernel->handle(Object(Enlight_Controller_Request_RequestHttp), 1, true) #5 /home/bcremer/www/5.2/engine/Shopware/Components/HttpCache/AppCache.php(255): Symfony\Component\HttpKernel\HttpCache\HttpCache->forward(Object(Symfony\Component\HttpFoundation\Request), true, NULL) #6 /home/bcremer/www/5.2/vendor/symfony/http-kernel/HttpCache/HttpCache.php(444): Shopware\Components\HttpCache\AppCache->forward(Object(Symfony\Component\HttpFoundation\Request), true) #7 /home/bcremer/www/5.2/vendor/symfony/http-kernel/HttpCache/HttpCache.php(344): Symfony\Component\HttpKernel\HttpCache\HttpCache->fetch(Object(Symfony\Component\HttpFoundation\Request), true) #8 /home/bcremer/www/5.2/engine/Shopware/Components/HttpCache/AppCache.php(178): Symfony\Component\HttpKernel\HttpCache\HttpCache->lookup(Object(Symfony\Component\HttpFoundation\Request), true) #9 /home/bcremer/www/5.2/vendor/symfony/http-kernel/HttpCache/HttpCache.php(210): Shopware\Components\HttpCache\AppCache->lookup(Object(Symfony\Component\HttpFoundation\Request), true) #10 /home/bcremer/www/5.2/engine/Shopware/Components/HttpCache/AppCache.php(114): Symfony\Component\HttpKernel\HttpCache\HttpCache->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true) #11 /home/bcremer/www/5.2/shopware.php(113): Shopware\Components\HttpCache\AppCache->handle(Object(Symfony\Component\HttpFoundation\Request)) #12 {main} [] {"uid":"313f6a2"}
```

bcremer:
We merged a very simplified version of this module with commit https://github.com/shopware/shopware/commit/a0fb109d4001d885cece82edb42c39d046d785b0.

It's by far not as complete as your solution but it's a starting point to further evolve.

Thank you very much for proposing this module @cjuner  /cc @svenmuennich 

I'm closing this PR for now. Please feel welcome to build more features upon the current module.

PR #597, opened 2016-06-13
Log order number of saved order in Backend

cjuner:
# Why:
- Shopware administrators should be able to see which orders have been edited for review purposes.
# What:
- Several Backend logging strings have been adjusted to include the order number of a modified order.
# Breaking something?

No.


bcremer:
Looks good for me.

Generally we are using `[0]` as a placeholder instead of `%s`. Is there a reason why you choose `%s`?

svenmuennich:
I just refactored it to use `Ext.String.format` and the placeholder `[0]`.

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/2190d8858d47667514f2c4c760b644d0991fc119.
Thanks @cjuner annd @svenmuennich :+1:

PR #599, opened 2016-06-14
Fix jquery.tab-menu acknowledging whitespace-only tabs as having content

screeny05:
If the Tab contains a tab which has only whitespace in it, the plugin thinks it has content and adds the class `.has--content` which leads to tabs being shown which have no content.


bcremer:
This has been fixed as a side effect with https://github.com/shopware/shopware/commit/5f9e882e017.
See: https://github.com/shopware/shopware/blob/v5.2.16/themes/Frontend/Responsive/frontend/_public/src/js/jquery.tab-menu.js#L132

PR #600, opened 2016-06-15
Ensure usage of unique store instance in config combo boxes

svenmuennich:
This PR improves the base combo box `Shopware.apps.Config.view.element.Select`, which is used e.g. when viewing plugin configuration forms via the settings module, to allow passing a store ID as the `store` parameter. This is escpecially usefull when providing selections of e.g. payment methods (`Shopware.apps.Base.store.Payment`) or dispatch methods (`Shopware.apps.Base.store.Dispatch`). In order to avoid side effects of other backend modules using the same store instance (and e.g. setting filters on it), the combo box does not use the instance that is registered by the passed store ID, but creates and uses a new instance of that store instead.

See also PR #525 for the original change of the `base` combo box, which is used when viewing plugin config forms in the plugin manager. This PR fixes the viewing of the same plugin config forms in the settings module. Furthermore it already includes the latest fix to support old storeIDs (22f5cacc17358161d7d0109de679973cba2c8906).

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -371,6 +371,7 @@ In this document you will find a changelog of the important changes related to t
 * Removed wrong parameter usage of `Shopware\Models\Menu\Repository::findOneBy`, which allows to provide two strings as criteria instead of array.
 * Updated composer dependency elasticsearch/elasticsearch to version 2.2.0
 * Changed default labelWidth for emotion component fields in `Shopware.apps.Emotion.view.components.Base` to 170 pixels
+* Added support for loading a new store instance by ID in the config combo box `Shopware.apps.Config.view.element.Select`
 
 ## 5.1.6
 * The interface `Enlight_Components_Cron_Adapter` in `engine/Library/Enlight/Components/Cron/Adapter.php` got a new method `getJobByAction`. For default implementation see `engine/Library/Enlight/Components/Cron/Adapter/DBAL.php`.
--- a/themes/Backend/ExtJs/backend/config/view/element/select.js
+++ b/themes/Backend/ExtJs/backend/config/view/element/select.js
@@ -62,6 +62,8 @@ Ext.define('Shopware.apps.Config.view.element.Select', {
             eval('me.store = ' + me.store + ';');
             // Remove value field for reasons of compatibility
             me.valueField = me.displayField;
+        } else if (typeof(me.store) === 'string' && me.store.substring(0, 5) !== 'base.') {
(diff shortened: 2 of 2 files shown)


janbuecker:
Hi & thanks for your PR.

Currently, it is possible to use base store id's like this:

``` php
$form->setElement('combo', 'customer_groups', 
    array(
        'label' => 'Kundengruppen',
        'store' => 'base.Payment',
        'multiSelect' => true,
        'scope' => Shopware\Models\Config\Element::SCOPE_SHOP,
        'value' => '',
    )
);
```

svenmuennich:
Yes, but the current implementation has side effects, e.g. when random plugins add filters to a `base` store, since these stores are singletons. Please see also [my comment](https://github.com/shopware/shopware/pull/525#issue-151290059) in the already merged PR #525

svenmuennich:
This PR is actually more a fix, since using the new store capabilities added in 3fc607dd19ba4c7092a4b42f2aa632c153a7f67d causes an error in ExtJs when opening the plugin config via the settings module: `Uncaught TypeError: Cannot read property 'on' of undefined`

I forgot to dass this to the description.

janbuecker:
I do not encounter any problems neither in the plugin config nor in the base configuration. It is necessary to define the store with this kind of id `'store' => 'base.Payment'` and not `'store' => 'Shopware.apps.Base.store.Payment'`.

svenmuennich:
Yes, it still works using e.g. `base.Payment`. However, in that case the store is initialised using `getStore()`, which might be an instance of e.g. the `payment` store, which some plugin might have added filters to. We encountered this issue several times in our plugins, where e.g. not all payment methods would show up for selection in the config. Hence we opened PR #525 to fix this, by ensuring that a new instance of the store is created, when passing the full name of a store (e.g. `Shopware.apps.Base.store.Payment`).

The point of this PR is to achieve compatibility of the `Shopware.apps.Config.view.element.Select` for the new store loading. You can reproduce this error I mentioned [here](https://github.com/shopware/shopware/pull/600#issuecomment-227091327) by replacing `base.Payment` with `Shopware.apps.Base.store.Payment` in one of your config elements. You will still be able to select payment methods when opening the plugin config via the plugin manager, but get an error when trying to do the same via the settings module.

janbuecker:
I've gone a step further and refactored the whole component since it's nearly an 1:1 copy of the base combo box. 

Merged with c301e70

svenmuennich:
Great, thank you 👍

PR #601, opened 2016-06-15
thumbnail generator performance optimization

jochenmanz:
**Why:**

The current implementation will load every associated media object from the database. This will take a significant amount of time before the acutal generation can start properly.

**What:**

Optimized the ThumbnailGeneratorCommand to use the Doctrine pagination and a loop.

**Breaking something?**

No.

### DIFF
--- a/engine/Shopware/Commands/ThumbnailGenerateCommand.php
+++ b/engine/Shopware/Commands/ThumbnailGenerateCommand.php
@@ -24,12 +24,12 @@
 
 namespace Shopware\Commands;
 
+use Exception;
 use Shopware\Components\Model\ModelManager;
 use Shopware\Models\Media\Album;
 use Shopware\Models\Media\Media;
 use Shopware\Models\Media\Repository;
-use Symfony\Component\Console\Helper\ProgressHelper;
-use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


engine/Shopware/Commands/ThumbnailGenerateCommand.php
```diff
@@ -106,11 +107,6 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $albumId             = (int)$input->getOption('albumid');
 
         foreach ($this->getMediaAlbums($albumId) as $album) {
-            //no size configured or no media object? continue
-            if ($this->hasNoThumbnails($album)) {
```
bcremer, line 110: The method `hasNoThumbnails()` is not in use any longer and should be removed.

engine/Shopware/Commands/ThumbnailGenerateCommand.php
```diff
@@ -125,27 +121,60 @@ private function createAlbumThumbnails(Album $album)
     {
         $this->output->writeln("Generating Thumbnails for Album {$album->getName()} (ID: {$album->getId()})");
 
-        /** @var ProgressHelper $progress */
+        /**
+         * @var ProgressHelper $progress
+         */
         $progress = $this->getHelperSet()->get('progress');
```
bcremer: The [`ProgressHelper`](http://symfony.com/doc/current/components/console/helpers/progresshelper.html) is deprecated. Please consider rewriting it to the newer `ProgressBar`.

engine/Shopware/Commands/ThumbnailGenerateCommand.php
```diff
@@ -27,6 +27,7 @@
 use Shopware\Components\Model\ModelManager;
 use Shopware\Models\Media\Album;
 use Shopware\Models\Media\Media;
+use Shopware\Models\Media\Repository;
 use Symfony\Component\Console\Helper\ProgressHelper;
 use Symfony\Component\Console\Input\InputArgument;
```
bcremer: `InputArgument` is not in use and should be removed.

bcremer:
There is another potential performance win possible in `createMediaThumbnails()`:

``` diff
-            if ($this->thumbnailExists($path) && !($this->force)) {
+            if (!$this->force && $this->thumbnailExists($path)) {
```

If the `force` parameter is given the call to `thumbnailExists()` will be skipped.

bcremer:
Except for the comments I added the PR looks good to me. I could measure a positive performance impact.

Thank you @jochenmanz, good catch.

jochenmanz:
Hi,

i've just refactored the internals of the command a bit more. I was able to remove the loop completly. Performance impact should be as positive as before.

Best regards,
Jochen

PR #605, opened 2016-06-21
Duplicate conditional expression in Store.php

kilida:
`!is_dir(dirname($path))` is repeated

### DIFF
--- a/engine/Shopware/Components/HttpCache/Store.php
+++ b/engine/Shopware/Components/HttpCache/Store.php
@@ -298,7 +298,7 @@ class Store extends BaseStore
     private function save($key, $data)
     {
         $path = $this->getPath($key);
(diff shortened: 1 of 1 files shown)


dnoegel:
Hi,
it is not repeated - IIRC it is the recommended way to prevent race conditions / other errors when creating directories as there is no "atomic" way to handle it. Symfony handles it in a similar manner: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/HttpCache/Store.php#L41

So basically in words this is: if directory does not exists, try to create it. If it still does not exist: Throw exception. 

Is there any issue with that method or is it just confusing to read?

bcremer:
As @dnoegel correctly stated it's about preventing race conditions. Please see the original issue in the symfony project: https://github.com/symfony/symfony/issues/16258

PR #607, opened 2016-06-23
Voucher Optimization

oktupol:
We reworked the way vouchers are added and updated in the basket.

Previously, vouchers were removed every time, when sBasket::sGetBasket was called and then re-added to check if the voucher is still applicable for the basket. However, this is not very performant.

We have changed the order in which way sAddVoucher checks, if the voucher is applicable;
extracted the checks into a new function (prepareVoucher) and used this new function in the sUpdateVoucher to check if the voucher has to be removed.

We also optimized the query to fetch a voucher by an individual code, so it works with joins instead of a select from two tables, as well as minor SQL optimizations in other methods.

Our first tests show major execution time improvements of up to 100%.


janbuecker:
Thanks @oktupol 

PR #611, opened 2016-06-30
BUGFIXING clear cache folders only if dir exists

barbieswimcrew:
Wrapping the clearDirectory($dir) methods (in CacheManager and Compiler classes) logic in IF statement to only let the RecursiveDirectoryIterator delete the cache directories if the given directory already exists. This will fix a bug occuring when trying to regenerate the cache by console command. 

After calling `$ var/cache/clear_cache.sh` and `bin/console sw:theme:cache:generate` an exception will be thrown if the folder "cache/.../html" doesn't exist.

### DIFF
--- a/engine/Shopware/Components/CacheManager.php
+++ b/engine/Shopware/Components/CacheManager.php
@@ -399,26 +399,28 @@ class CacheManager
      */
     private function clearDirectory($dir)
     {
-        if (file_exists($dir)) {
-            $iterator = new \RecursiveIteratorIterator(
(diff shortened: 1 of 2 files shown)


bcremer:
Please rewrite your code to use an early return like this:

```
 if (!file_exists($dir)) {
    return;
}
```

This is to keep the indentation depth low and has the nice side effect of an easy to view diff.

barbieswimcrew:
@bcremer Did it, thanks!

PR #612, opened 2016-06-30
Fix "ml" in demo SQL

hanneswernery:
Exchange "ml" and "Milliliter" in demo data, since _unit_ should be the abbreviation and _description_ the full name.


hanneswernery:
This should also be fixed in SW 5.2. Should I create a second pull request, or will you handle it?

mitelg:
Hi @hanneswernery ,
thanks for your PR. could you change the destination branch to 5.2 ?

bcremer:
One cannot change the destination branch of a PR.
We'll apply it to the correct branch when we merge it. No need to change.

bcremer:
Thank you @hanneswernery, I merged this into 5.2 with commit https://github.com/shopware/shopware/commit/b2e1e498db452535348a0b05df1a7fb6a69921ad.

PR #524, opened 2016-04-26
Add 'countries' API endpoint and resource

svenmuennich:
This PR adds a new API resource `Country` that allows fetching, creating, updating and deleting instances of `Shopware\Models\Country\Country`. Furthermore it adds an API controller, which uses the new resource to provide all its operations via the REST API.

This PR does not introduce any breaking changes.

PR #613, opened 2016-07-01
Fix 'countries' REST API response data and header

svenmuennich:
This PR fixes two problems in the `countries` REST API:
1. The structure of the response data of `GET /api/countries` previously contained a nested `data` field inside the top-level `data` field. This PR fixes the structure to contain the country data directly in the top-level `data` field.
2. When sending a `POST` request to `/api/countries` the response was previously missing the `Location` header, which should be set according to REST best practices. This PR adds the missing header with the correct location.

This PR does not introduce any breaking changes, since the REST API `countries` endpoint is only introduced in shopware 5.2. It was added via PR #524. Therefore I think this PR must be merged before the final release of shopware 5.2.

Pinging @janbuecker since he merged the original PR.


janbuecker:
Merged with 4f71faa7ada132219315d05b7437044c27c3d7a9.

PR #614, opened 2016-07-01
Fix exception in account bundle register service

svenmuennich:
This commit fixes a `PDOException` inside the `RegisterService` of the `AccountBundle`, which is raised because of creating a nested transaction. The source of the exception is the `number range incrementer`, which I introduced in PR #517 for shopware 5.2. The incrementer uses a transaction to ensure a safe increment of the numbers. However, the `RegisterService` starts a transaction itself and then uses the `number range incrementer` within this transaction.

This commit fixes the exception by moving the check, whether a customer number must be created, as well as the increment of the number just before starting the transaction and then passing the number to the `saveCustomer()` method.

This commit does not introduce any breaking changes.


svenmuennich:
I just debugged this further, since I noticed that the exception only occurs when creating a customer via the REST API. I found that the problem lies within Doctrine's `DABL\Connection`, whose underlying connection is different when using the REST API than when using the shopware frontend or backend. I will investigate this further and create a new PR, if I manage to fix the underlying problem.

PR #615, opened 2016-07-01
making sure the elasticsearch index we are operating on is green

LarsFronius:
It is not safe to close and open an index in yellow state and actually broke indexing for us with a high number of replicas being initialised and left the cluster in a red state. Fortunately we can check whether a single index is green, preventing us to wait forever on a yellow cluster and generally fixes indexing for us.

### DIFF
--- a/engine/Shopware/Bundle/ESIndexingBundle/ShopIndexer.php
+++ b/engine/Shopware/Bundle/ESIndexingBundle/ShopIndexer.php
@@ -137,7 +137,7 @@ class ShopIndexer implements ShopIndexerInterface
      */
     private function updateSettings(ShopIndex $index)
     {
-        $this->client->cluster()->health(['wait_for_status' => 'yellow']);
+        $this->client->cluster()->health(['index' => $index->getName(), 'wait_for_status' => 'green']);
         $this->client->indices()->close(['index' => $index->getName()]);
 
         foreach ($this->settings as $setting) {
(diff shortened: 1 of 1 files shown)


bcremer:
@OliverSkroblin can you take care of this please.

LarsFronius:
Just to give you a test case. For us the `sw:es:index:populate` task would break when:
- we run a 3 node cluster
- we run a default template for all indexes, setting replica to 2 (a replica per node)

the task would now close and open the index in a yellow state, which is an unsafe operations and would for us actually break the index for us, turning it red.
The fix now does not look at the overall cluster state anymore being yellow as done before, we only look at the index we operate on being green, so that our operations can happen in a safe environment.

This might bring problems with default setting (replica set to 1) on a 1 node cluster, since in this case, you will never get a green state. The fix should be documented (setting replica to 0) and also it should be documented that this is far from a best practice and an unsafe environment for elasticsearch to run in.

bcremer:
Hi @LarsFronius, thanks for describing the test case. I was able to reproduce this issue.

I also think it makes sense to wait for the `green` status.
But we have to find a solution to keep 1 Node (developer) setups easy to test.

One idea was to make `number_of_replicas`  configurable directly in the `config.php`. So for development or single node setups `number_of_replicas` could be set to `0`.

Does this make sense to you?

LarsFronius:
Makes all the sense in the world! :) Thanks!

bcremer:
Related DevDocs PR: https://github.com/shopware/devdocs/pull/322/files

bcremer:
This was merged with commit https://github.com/shopware/shopware/commit/da7b5b24a3f0f578892559295b81dd700108ee80.

Thank you @LarsFronius :+1:

PR #616, opened 2016-07-04
Add support for integer and array values to 'sw:plugin:config:set' cli command

svenmuennich:
Previously it was not possible to set an integer or array value when using the cli command `sw:plugin:config:set`. However, since it is possible to create plugin config elements e.g. for selecting one or multiple dispatch methods, which are saved as their ID (integer) or an array of their IDs (integers) respectively, it is desirable to be able to set the same config value using the cli tools. This PR adds an extra check on the config value passed to `Shopware\Commands\PluginConfigSetCommand::execute()`, which matches integers and arrays of e.g. the form `[val1,val2,val3]` and converts them to PHP arrays. Furthermore, if a value contained in the matched array contains only digits, it is converted to an integer.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,9 @@
 # Shopware Upgrade Information
 In this document you will find a changelog of the important changes related to the code base of Shopware.
 
+## 5.2.1 DEV
+* Add support for integer and array values to the `sw:plugin:config:set` cli command
+
 ## 5.2.0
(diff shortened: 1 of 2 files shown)


bcremer:
Merged with commit https://github.com/shopware/shopware/commit/1c24445d895862d4636c37c9cc8ec1edb83d024e.

Thanks @svenmuennich :+1:

PR #617, opened 2016-07-07
Ensure composer autoloader optimization to prevent incomplete classmap.

kleinmann:
With composer 1.1.2 and 1.1.3 an additional `composer install` on the current release results in an incomplete classmap in `autoload_static.php`. One possible fix is `composer install -o`, another one is to force the optimization in `composer.json`, which is what I've done in this pull request.

This fixes errors like `Doctrine\Common\Proxy\Exception\OutOfBoundsException: Missing value for primary key id on Shopware\Models\Customer\Group`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,12 @@
         "classmap": [
             "engine/Shopware/Core/"
         ],
-        "files": ["engine/Shopware/Application.php"]
+        "files": ["engine/Shopware/Application.php"],
+        "exclude-from-classmap": [
+            "engine/Shopware/Plugins/Community/",
+            "engine/Shopware/Plugins/Local/",
+            "custom/plugins/"
+        ]
     },
     "autoload-dev": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -76,6 +76,7 @@
     },
     "config": {
         "autoloader-suffix": "Shopware",
+        "optimize-autoloader": true,
         "platform": {
             "php": "5.4"
         }


kleinmann:
This PR was created into 5.2, however the fix is also needed for 5.1 (per @bcremer).

svenmuennich:
We ran into the same issue during development, because we usually don't use `ant build-composer-install` but just `composer install` (see discussion in 9138d7ef7d00ec7ee6b61e80f904de2cc653c30d). Therefore I think it's a good idea to enforce the autoloader optimisation in `composer.json`. In any case, the plugins are included in the resulting class map though, as I pointed out in [this comment](https://github.com/shopware/shopware/commit/9138d7ef7d00ec7ee6b61e80f904de2cc653c30d#commitcomment-17945414). This can be prevented by adding the following to `composer.json`:

``` json
"autoload": {
    "exclude-from-classmap": [
        "engine/Shopware/Plugins/",
        "custom/plugins/"
    ]
}
```

@kleinmann Maybe you can extend you PR to exclude the plugin directories from the autoloader optimisation.

kleinmann:
Good point @svenmuennich.
I just tested your proposed change. Judging from a comparison with the 5.2.1 release's `composer.json` we are still _way_ off from the official classmap. Maybe someone from the core team can shed some light on the actual `composer` build so we can have reproducible installations in our development and production environments.

I suggest not merging this PR until we have cleared this up.

bcremer:
We are using the following composer options for the deployed release packages:

``` xml
<exec executable="php" failonerror="true">
    <arg value="${basedir}/output/composer.phar" />
    <arg value="install" />
    <arg value="--prefer-dist" />
    <arg value="--no-interaction" />
    <arg value="--no-dev" />
    <arg value="--optimize-autoloader" />
    <arg value="--quiet" />
    <arg value="--working-dir=@{workingDir}" />
</exec>
```

As for `exclude-from-classmap` I am open to include that too. But please keep in  `engine/Shopware/Plugins/Default` as some core plugins rely on the composer classmap to function properly.

kleinmann:
I have added the exclusion for both Local and Community plugins as well as the new 5.2 flavoured plugins.
With the commandline from Benjamin I have verified a matching classmap. Looking good from my end! :-)

bcremer:
Thanks @kleinmann and @svenmuennich. Merged into 5.1 and into 5.2.

oemmes:
Hello,

we are using composer to add our custom plugins to the Shopware installation:
```
composer require our/package
```

We have some third-party classes which we need to ship but which do not match PSR-0 or PSR-4 autoloading scheme and we have to add those in the plugin's composer.json via classmap like this:
```
{
	"name": "our/pacakge",
	"type": "shopware-plugin",
	"autoload": {
		"psr-4": {
			"Our\\Package\\": ""
		},
		"classmap": [
			"src/other_classes/"
		]
	}
}
```

The problem is that the folder "custom/plugins" is excluded from classmap generation.

This breaks the autoload/classmap settings of our plugin as the map is incomplete.

I think it is reasonable that the old plugin folders (engine/Shopware/Plugins/...) are excluded because they might be included due to the other core autoload settings.

Though, custom/plugins should not be excluded so that the plugin's autoloader information can also be part of the generated classmap.

I also filed this as an issue in the issue tracker: [SW-17339](https://issues.shopware.com/issues/SW-17339)

svenmuennich:
Well, since you also added more dependencies (your plugins) to the `composer.json` of your shop, you can just remove `"custom/plugins/"` from your `"exclude-from-classmap"` entry if you wish. However, I think in this case it would be better to just manually `include`/`require` the respective classes in your plugin's `Bootstrap.php` or wherever you need them to be available.

PR #618, opened 2016-07-08
new command for executing migrations

SvenHerrmann:
build/ApplyDeltas.php has been removed in 5.2


bcremer:
Good catch! Thanks @SvenHerrmann.

PR #619, opened 2016-07-10
Remove redundant isset in FormSynchronizer for description

Haehnchen:


bcremer:
Interesting, I was sure this is needed to prevent a PHP notice in the case `description` is not set in the array.

You prooved me wrong:
https://3v4l.org/Bd8Gc

Haehnchen:
you are not alone, my default code review issue. :) also note `isset($arr['a'], $arr['b'])`;

PR #620, opened 2016-07-12
CheapestPriceService getHighestQuantityDiscount() throws Exception if PriceGroup is null

jemmel:
After updating a shopware 4 version to shopware 5.1.5 we get an error while loading the product detail page (see attached Screenshot). 

While debugging this error we noticed that the method call $product->getPriceGroup() returns null. We fixed this problem by adding the following check:

```
$id = $product->getPriceGroup()->getId();
 if (!isset($priceGroups[$id])) {
     return null;
 } 
```

error_detail_page

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/CheapestPriceService.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/CheapestPriceService.php
@@ -197,6 +197,11 @@ class CheapestPriceService implements Service\CheapestPriceServiceInterface
             return null;
         }
 
+        $checkPriceGroup = $product->getPriceGroup();
+        if (empty($checkPriceGroup)) {
+            return null;
+        }
+
         $id = $product->getPriceGroup()->getId();
         if (!isset($priceGroups[$id])) {
             return null;


bcremer:
@jemmel It's seems like you have some inconsistencies in your product database.
Some articles seem to have `pricegroupActive` no either no `pricegroupID` or a `pricegroupID` that  point to a no longer existing pricegroup.

Can you please run the following (read-onyl) query to verify my assumptions:

``` sql
SELECT * FROM s_articles 
LEFT JOIN s_core_pricegroups ON s_core_pricegroups.id = s_articles.pricegroupID
WHERE s_articles.pricegroupActive = 1 
AND 
(
    s_articles.pricegroupID IS NULL -- pricegroupID is set to null
    OR 
    s_core_pricegroups.id IS NULL -- pricegroupID point to not existing pricegroup
)
```

bcremer:
There are already some saveguards in the codebase that check if the priceGroup property is set. Like here: https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/GraduatedPricesService.php#L115

A more general solution to the problem would be to "fix" `\Shopware\Bundle\StoreFrontBundle\Struct\ListProduct::isPriceGroupActive()`:

```
    public function isPriceGroupActive()
    {
        return $this->isPriceGroupActive && $this->priceGroup;
    }
```

Can you please verify if that also would fix your problem.

jemmel:
@bcremer Thanks fot that query. We found a couple articles with invalid relationsships.

But the article which showed up the error is missing in the result set. Therefore we still need to go with our hotfix. :)

mitelg:
Hi @jemmel 
thanks for your PR :+1: 
this issue was already fixed with fe4c2152670 . I will close this for now. If you still have problems with that, feel free to open another PR oder a ticket on our issue tracker.
best regards

PR #629, opened 2016-07-19
Added AccountMode field to customer list

shyim:
This PR adds a Column to Customer List which displays the Account Mode, some customers has more than 3-4 Accounts and thats annoying to open all details to find the right account which is not a guest.


themes/Backend/ExtJs/backend/customer/view/list/list.js
```diff
@@ -334,6 +340,20 @@ Ext.define('Shopware.apps.Customer.view.list.List', {
...
     /**
+     * Formats the accountMode column
+     *
+     * @param [string] - accountMode
+     * @returns [string] - description
+     */
+    accountModeRenderer:function (value) {
+        if (value) {
+            return '{s name="accountModeGuest"}Guest Account{/s}';
+        } {
```
janbuecker: unnecessary `else` if it was supposed to be an if/else

PR #636, opened 2016-07-27
fix error in unifiedToSQL()

hrombach:
If the `unifiedToSQL()` function doesn't recognize a given type, it returns `'string'`. I assume this is done to make string the default type in case a type is not recognized. Simply returning `'string'` however leads to a confusing SQL error message. Instead it should return the SQL value that is mapped to `'string'`. The function is also used in `CrudService::getList` to set the field `sqlType` but it doesn't make much sense here either because `'string'` is not an actual SQL type.

I don't think this should break anything, because it doesn't work right now.


bcremer:
Good catch, merged with commit https://github.com/shopware/shopware/commit/0439654a10bef8d323f88888ba7f04b5743a5706.

Thank you @hrombach :+1:

PR #638, opened 2016-07-28
Update sRewiteTable.php

netformic:
Vgl. https://issues.shopware.com/#/issues/SW-15786


janbuecker:
Thanks for your contribution, @netformic.

We are already aware of this issue and it will be fixed with SW-15829.

PR #640, opened 2016-08-03
Catch ServiceCircularReferenceException in DI Container

FiveDigital:
@bcremer As discussed in IRC

We had the following subscriber in one of our projects to replace the Product Number Search service and decorate the Product Search service:

``` php
    {
        return [
            'Enlight_Bootstrap_InitResource_shopware_search_es.product_number_search' => 'onInitProductNumberSearch',
            'Enlight_Bootstrap_AfterInitResource_shopware_search.product_search' => 'onAfterInitProductSearch',
        ];
    }

    public function onInitProductNumberSearch()
    {
        // This caused an infinite loop
        \Shopware()->Container()->get('shopware_search.product_search');
    }

    public function onAfterInitProductSearch()
    {
        Shopware()->Container()->set(
            'shopware_search.product_search',
            new \ShopwarePlugins\SwagPluginSystem\StoreFrontBundle\ProductSearch(
                Shopware()->Container()->get('shopware_search.product_search')
            )
        );
    }
```

After debugging we found out that this code produces a circular reference which should throw a `ServiceCircularReferenceException` in Symfonys DI container. The `doLoad()` method in Shopware`s DI container does not catch this exception but triggers the`Enlight_Bootstrap_AfterInitResource_`event in the`finally` block. So even if the exception is thrown the event will be dispatched. In our case until PHP is out of memory.

This PR is an approach to not dispatch the `Enlight_Bootstrap_AfterInitResource_` event if an `ServiceCircularReferenceException` exception is thrown. We have included a unit test to verify the exception is properly thrown.

We think this approach is not perfect, yet but a good starting point for discussion. You also need to know that this behaviour occured after the 5.2 update and was working (some how) in 5.1

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/Container.php
+++ b/engine/Shopware/Components/DependencyInjection/Container.php
@@ -25,6 +25,7 @@
 namespace Shopware\Components\DependencyInjection;
 
 use Symfony\Component\DependencyInjection\Container as BaseContainer;
+use Symfony\Component\DependencyInjection\Exception\ServiceCircularReferenceException;
 
 /**
  * @category  Shopware
@@ -160,6 +161,7 @@ class Container extends BaseContainer
      * @param string $id already normalized
      * @param int $invalidBehavior
      * @return mixed
+     * @throws ServiceCircularReferenceException
      */
     private function doLoad($id, $invalidBehavior = self::NULL_ON_INVALID_REFERENCE)
     {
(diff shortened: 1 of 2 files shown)


bcremer:
This bug was introduced in commit https://github.com/shopware/shopware/commit/f179b80b344.

In my opinion your fix is OK, `ServiceCircularReferenceException` has to be handled special so the state variable in `doLoad` seems fine.

PR #642, opened 2016-08-03
Hide big menu if parent ist hidden (SW-15795)

s-bessing:
If a menu element is selected for hiding in main menu, then the big menu is still created. When hovering a main menu element the wrong big menu gets displayed. The template variable for hiding in main navigation is hideTop not hidetop.


janbuecker:
Thanks for your PR. This is a regression of SW-13117.

PR #644, opened 2016-08-05
unnecessary permission

taltholtmann:
config.php does not exist at that moment.
It will be created later.


bcremer:
Thank you @taltholtmann :+1:

PR #645, opened 2016-08-08
change smarty-block for paymentframe-js

jppeter:
Actually this JS is appended to the "frontend_index_header_javascript"-block.
But it requires jQuery, which is initialised later. So this must be moved to another block.


Phil23:
Hey @jppeter,

you made a nice catch. This file was not refactored for Shopware 5. _hoopsy_ :sweat_smile: 

Thank you for your PR. We will close this one and refactor the whole thing. thx!

PR #648, opened 2016-08-09
Product Modifiers for ProductProvider

ghost:
**Why:**
At some point you want to modify a product before indexing to elasticsearch inside a plugin.
**What:**
Tagged product modifier services via compiler pass for the elasticsearch product provider.

**This is not a breaking change.**

**Usage:**
Create your modifier by implementing the ProductModifierInterface.
Define it as a tagged services with the tag: shopware_elastic_search.product_provider.modifier,
also located as a constant in ProductModifierInterface::SERVICE_TAG_ID


OliverSkroblin:
Thanks for your pull request,
but you can decorate the ProductProvider service and modify the returned data to extend or modify the indexed product data

PR #649, opened 2016-08-09
normalized the return value of Album/Settings getThumbnailSize

jochenmanz:
**Why:**

When no settings are present, the current behaviour will return a array with an empty value instead of an empty array. This makes it difficult to check if there are sizes as the array needs to be analysed.

if (empty($sizes) vs if (empty($sizes[0]))

**What:**

Changed the return value to return a empty array if its empty.

**Breaking something?**

Anything that relies on the old format might break

### DIFF
--- a/tests/Functional/Models/Album/SettingsTest.php
+++ b/tests/Functional/Models/Album/SettingsTest.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+/**
+ * @category  Shopware
+ * @package   Shopware\Tests
+ * @copyright Copyright (c) 2014, shopware AG (http://www.shopware.de)
+ */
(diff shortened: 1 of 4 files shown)


tests/Functional/Models/Album/SettingsTest.php
```diff
@@ -0,0 +1,42 @@
+<?php
```
bcremer: This test should be moved into the `Unit` directory as it has no external dependencies.
You can run the Unit tests separatly using the `composer test-unit` command (`./vendor/bin/phpunit -c tests/phpunit_unit.xml.dist`)

tests/Functional/Models/Album/SettingsTest.php
```diff
@@ -0,0 +1,42 @@
...
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+/**
+ * @category  Shopware
+ * @package   Shopware\Tests
+ * @copyright Copyright (c) 2014, shopware AG (http://www.shopware.de)
```
bcremer: Old header.

``` diff
- * @copyright Copyright (c) 2014, shopware AG (http://www.shopware.de)
```

tests/Functional/Models/Album/SettingsTest.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    public function testGetThumbnailSizeReturnValue()
+    {
+        $settings = new \Shopware\Models\Media\Settings();
+        $settings->setThumbnailSize([]);
+
+        $size = $settings->getThumbnailSize();
+
+        $this->assertEquals([], $size);
+        $this->assertNotEquals([0 => ''], $size);
```
bcremer: Please remove the `$this->assertNotEquals` line. The result cannot equal `[]` and not-not equal `[0 => '']`. 
I suppose this line was there to show the old behavior.

jochenmanz:
Hi,

thanks for the feedback, i just added the requested changes.

Best regards,
Jochen

tests/Unit/Model/Album/SettingsTest.php
```diff
@@ -0,0 +1,41 @@
...
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+/**
+ * @category  Shopware
+ * @package   Shopware\Tests
+ * @copyright Copyright (c) shopware AG (http://www.shopware.de)
+ */
+class Shopware_Tests_Models_Album_SettingsTest extends Enlight_Components_Test_TestCase
```
bcremer: This does not work when running the unit tests separatly with `composer test-unit`

``` diff
- Enlight_Components_Test_TestCase
+ PHPUnit_Framework_TestCase
```

PR #650, opened 2016-08-10
Implementing __clone() method in ModelEntity

barbieswimcrew:
To allow cloning a model entity it's necessary to implement the abstract ModelEntity's __clone() interceptor method in which the clone's id property will be set to null

### DIFF
--- a/engine/Shopware/Components/Model/ModelEntity.php
+++ b/engine/Shopware/Components/Model/ModelEntity.php
@@ -33,6 +33,17 @@ namespace Shopware\Components\Model;
  */
 abstract class ModelEntity
 {
+    
+    /**
+     * Implementing the __clone interceptor method to set the entities id property 
+     * to null on using the clone method
+     */ 
+    public function __clone(){
+        if($this->id){
+            $this->id = null;
+        }
+    }
+    
     /**
      * Example:
      *


bcremer:
A clone is not that easy. You also have to take care of related entities.

See for example: https://github.com/shopware/shopware/blob/v5.2.3/engine/Shopware/Models/Form/Form.php#L175

``` php
    public function getClone()
    {
        $clonedForm = clone $this;
        /* @var $field \Shopware\Models\Form\Field */
        foreach ($this->getFields() as $field) {
            $clonedField = clone $field;
            $clonedForm->fields->add($clonedField);
            // update owning side
            $clonedField->setForm($clonedForm);
        }
        return $clonedForm;
    }
```

Sometimes related entities should be reused, for some entities the relations have to be cloned too.
So it's not sensible to write a generalized clone method.

See also: 
- http://stackoverflow.com/questions/14158111/deep-clone-doctrine-entity-with-related-entities
- http://stackoverflow.com/questions/9088603/symfony2-doctrine-how-to-re-save-an-entity-with-a-onetomany-as-a-cascading-new

You still can clone an object without having it to implement `__clone()`.
This involves reflection or alternatively some closure magic.

``` php
$article = new \Shopware\Models\Article\Article();
$article->setName('wusel');

$copyObject = function($object) {
    $instantiator = new \Doctrine\Instantiator\Instantiator();
    $instance = $instantiator->instantiate(get_class($object));
    foreach ($object as $property => $value) {
        if ($property === 'id') {
            continue;
        }

        $instance->{$property} = $value;
    }
    return $instance;
};

$copyObject = Closure::bind($copyObject, null, $article);
var_dump($copyObject($article));
```

You can find more about this in ocramius's blog: https://ocramius.github.io/blog/accessing-private-php-class-members-without-reflection/

barbieswimcrew:
You are totally right about that deep cloning issues. But honestly, cloning flat objects that don't hold any relations would be much easier when clearing the id property in the way I did suggest rather than using reflection. But thank you for your advice and sorry for the inconvenience.

bcremer:
Not a inconvenience at all. We are always open for new ideas and appreciate your participation  :+1:

PR #652, opened 2016-08-11
[Papercut][SW-15353] Add missing block to sidebar

screeny05:
This adds a block around the content of index/sidebar.tpl. without this block it's not possible to add content to the bottom sidebar.

This change breaks nothing.

The new block is called `frontend_index_left_inner`

In Shopware < 5.2 the block `frontend_index_left_campaigns_bottom` has been used for this, but it was removed


FiveDigital:
Is this going to be merged? We need this in out projects, too.

PR #653, opened 2016-08-11
Disable csrf-header for cors-ajax

screeny05:
The automatic CSRF-Header for AJAX-Requests has to be disabled for CORS-Requests, as those might not allow custom headers to be set / don't allow custom headers in preflight.

Without this fix all CORS-Requests **will fail**.

This change breaks nothing, as all regular AJAX-Requests still get an X-CSRF-Token Header added.

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.csrf-protection.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.csrf-protection.js
@@ -131,13 +131,17 @@
         },
 
         /**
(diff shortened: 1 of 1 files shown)


screeny05:
Has this been closed by https://github.com/shopware/shopware/commit/bdee6921d8c19ead3403cb4149b2b1ab7243196c?

@klarstil

mitelg:
hey @screeny05,
you are right :+1: the regarding ticket is closed with the commit you mentioned

PR #654, opened 2016-08-11
Register controllers automatically

jeboehm:
Hi,
i'm really in love with the new plugin implementation. It works like a charm!
While nearly everything is configured by convention, the controllers have to be registered in the old way.

I tried to improve this by scanning the plugin directory for controller files. The next step will be to add a CompilerPass to the `ContainerBuilder`. It creates a new listener definition and adds the ControllerPath event (Enlight_Controller_Dispatcher_ControllerPath_X_Y) to a new `shopware.event_listener` tag for each found controller. In an additional array, the listener will store all ControllerPath events and their matching controller paths.

While running at build time, there should be no performance issues. The great advantage is the registration of controllers without any additional configuration, just by adding them to the filesystem.

I look forward to your feedback!

### DIFF
--- a/engine/Shopware/Bundle/ControllerBundle/Finder/ControllerFinder.php
+++ b/engine/Shopware/Bundle/ControllerBundle/Finder/ControllerFinder.php
@@ -3,10 +3,12 @@
 namespace Shopware\Bundle\ControllerBundle\Finder;
 
 use Shopware\Bundle\ControllerBundle\Struct\ControllerStruct;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
 
 class ControllerFinder
 {
-    const MODULES = ['Backend', 'Frontend', 'Widgets'];
+    const MODULES = ['Backend', 'Frontend', 'Widgets', 'Api'];
 
     /**
      * @param string $path
@@ -16,17 +18,23 @@ class ControllerFinder
     public function getControllers($path)
     {
         $controllers = [];
+        $finder = new Finder();
+        $finder
+            ->in($path)
+            ->files()
+            ->name('*.php');
 
         foreach (self::MODULES as $module) {
-            $files = glob(sprintf('%s/%s/*.php', rtrim($path, '/'), $module));
+            $finder->path($module);
+        }
 
-            if ($files === false) {
-                continue;
-            }
-
-            foreach ($files as $file) {
-                $controllers[] = new ControllerStruct($module, rtrim(basename($file), '.php'), $file);
-            }
+        foreach ($finder as $file) {
+            /** @var $file SplFileInfo */
+            $controllers[] = new ControllerStruct(
+                $file->getPathInfo()->getBasename(),
+                $file->getBasename('.php'),
+                $file->getPathname()
+            );
         }
 
         return $controllers;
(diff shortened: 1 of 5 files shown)


engine/Shopware/Bundle/ControllerBundle/Finder/ControllerFinder.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Shopware\Bundle\ControllerBundle\Finder;
+
+use Shopware\Bundle\ControllerBundle\Struct\ControllerStruct;
+
+class ControllerFinder
+{
+    const MODULES = ['Backend', 'Frontend', 'Widgets'];
```
mitelg: you forgot Api here :wink:

engine/Shopware/Bundle/ControllerBundle/Finder/ControllerFinder.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * @param string $path
+     *
+     * @return ControllerStruct[]
+     */
+    public function getControllers($path)
+    {
+        $controllers = [];
+
+        foreach (self::MODULES as $module) {
```
FiveDigital, line 27: You could use Symfony's Finder component here to make this part shorter and skip the nested foreach loops.

jeboehm, line 27: Good point, I'll change that later!

engine/Shopware/Bundle/ControllerBundle/Finder/ControllerFinder.php
jeboehm: Oops! I'll change that later.

jeboehm:
The finder component is now in use and it searches also in the Api folder.

Have a nice weekend!

janbuecker:
Maybe it would be better to declare them in the services.xml and use a tag to register them. Less magic, more control.

What do you think? @bcremer @dnoegel

dnoegel:
For our current routing system, convention over configuration is perfect IMHO. If we had a router such as the symfony router with the possibilities of a routing.xml with placeholders etc, configuration would be a better approach. But for our current system I see no benefit in explicitly registering controllers via services. 

So basically Jeffrey's approach is what I would go for :+1:

bcremer:
@jeboehm Thank you for this Pull Request.

I was a bit warned by the complexity of it. In special the usage of the `CompilerPass` seemed unnecessary to me.

But as I dived deeper into it the beauty of it was getting clear.

The IO-intense filesystem iteration is only done during container build time and not for every request. Also `\Shopware\Bundle\ControllerBundle\Listener\ControllerPathListener` is lazily constructed if a matching Controller Event is triggered.

:+1: 

We have to clarify some problems upfront, though.
For example what about BC with existing Plugins that register Controller Paths conditionally.
One idea was to have a opt-in config setting in `plugin.xml` like `autoregisterControllers=on`.

Also the view Directory should be added if a custom controller is in use.

I will put this into discussion with the team next week and report back as soon as we have some ideas.

Thanks again for your contribution to shopware!

jeboehm:
Thx for including the feature in the latest 5.2.8 release and [irony]mentioning me in the changelogs[/irony].

I like your approach of registering only one listener and adding the tags to it.

bcremer:
This was merged with commit https://github.com/shopware/shopware/commit/d648dc50b1914a849851b673363015d29b46bc71 as part of the 5.2.7 release (https://github.com/shopware/shopware/compare/v5.2.6...v5.2.7).
You are mentioned in the changelog of this release, please find it here: http://community.shopware.com/Downloads_cat_448.html#5.2.7.

I missed to close the related pull request here on github, sorry for this and thank you @jeboehm for this contribution :fireworks:

jeboehm:
Oh, my fault. Thank you! 👍

PR #655, opened 2016-08-12
Add missing order detail attributes upon adding order position in backend

svenmuennich:
When adding a new order position via the shopware backend, there are currently no attributes written to the database for that new position. This might lead to problems, if a plugin relies on order detail attributes (table `s_order_details_attributes`) being present. This PR adds the missing creation of a `Shopware\Models\Attribute\OrderDetail` instance, when adding a new order position.

**Edit:** We've always been wondering (for all shopware versions >= 4.0), why in some cases order details were missing their attributes, but now we are finally able to pin it on adding new positions via the backend. Throughout that time, we've also seen quite a few plugins using workarounds to ensure the existence of the attribute rows when needed.

### DIFF
--- a/engine/Shopware/Controllers/Backend/Order.php
+++ b/engine/Shopware/Controllers/Backend/Order.php
@@ -710,6 +710,8 @@ class Shopware_Controllers_Backend_Order extends Shopware_Controllers_Backend_Ex
         //check if the passed position data is a new position or an existing position.
         if (empty($id)) {
             $position = new Detail();
+            $attribute = new Shopware\Models\Attribute\OrderDetail();
+            $position->setAttribute($attribute);
             Shopware()->Models()->persist($position);
         } else {
             $detailRepository = Shopware()->Models()->getRepository('Shopware\Models\Order\Detail');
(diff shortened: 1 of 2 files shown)


engine/Shopware/Controllers/Backend/Order.php
```diff
@@ -710,6 +710,8 @@ public function savePositionAction()
         //check if the passed position data is a new position or an existing position.
         if (empty($id)) {
             $position = new Detail();
+            $attribute = new Shopware\Models\Attribute\OrderDetail();
```
giginos, line 713: Maybe you want to convert this into "use Shopware\Models\Attribute\OrderDetail;" ?

svenmuennich, line 713: I usually don't like this when working with `Attribute` models, since one cannot tell what exactly is instantiated:

``` php
use Shopware\Models\Attribute\OrderDetail;

// ...

$attribute = new OrderDetail();
```

Now it looks like `$attribute` is an order detail and not the order detail attributes (although the variable is called `$attribute`). Alternatively we could use `use Shopware\Models\Attribute\OrderDetail as OrderDetailAttribute;`. Not sure though what shopware's convention is here.

PR #656, opened 2016-08-12
Add missing order attributes and order detail attributes upon creating a temporary order

svenmuennich:
Currently `sOrder::sCreateTemporaryOrder()` only creates rows in `s_order` and `s_order_details`, but not their attribute rows in `s_order_attributes` and `s_order_details_attributes` respectively. As a result, when later converting a cancelled order to a _normal_ order, both the order and their details are still missing their attributes. This might lead to problems, if a plugin relies on order attributes and/or order detail attributes being present. This PR adds the creation of both order attributes and order detail attributes to `sOrder::sCreateTemporaryOrder()`, which leads to a consistent database even before converting a cancelled order to a _normal_ order.


samuelvogel:
This fixes a similar issue as #655.

PR #657, opened 2016-08-12
Set order number in order details upon converting a cancelled order

svenmuennich:
When converting a cancelled order to a _normal_ order, the newly determined order number is currently not set in the order details. This behaviour diverges from the order creation during the checkout process. This PR fixes this behaviour by setting the new order number in all order details too.


janbuecker:
Merged with 1a24aef2fd0b3d3bfdf7063581238c9bc8884c33. Thanks!

PR #658, opened 2016-08-12
[TASK] Use container to create FormType classes

Tombro:
Hey folks,
as discussed in IRC here is my PR which allows decorating/replacing of the formtypes.

This brings a lot of flexibility. In my case I wanted to use symfony validation_groups which could not be obtained without decorating the forms.

This change does not break anything.

Regards,
Thomas

### DIFF
--- a/engine/Shopware/Controllers/Backend/Address.php
+++ b/engine/Shopware/Controllers/Backend/Address.php
@@ -139,7 +139,10 @@ class Shopware_Controllers_Backend_Address extends Shopware_Controllers_Backend_
         $data['state'] = $data['state_id'];
 
         /** @var \Symfony\Component\Form\FormInterface $form */
-        $form = $this->get('shopware.form.factory')->create(AddressFormType::class, $model);
+        $form = $this->get('shopware.form.factory')->create(
+            get_class($this->container->get('shopware_account.form.addressform')),
+            $model
+        );
         $form->submit($data);
 
         if (!$form->isValid()) {
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -26,6 +26,7 @@ This changelog references changes done in Shopware 5.2 patch versions.
 * Added `Shopware\Components\Emotion\ComponentInstaller` class to install emotion components in plugins.
 * Added `\Shopware\Components\Emotion\EmotionComponentViewSubscriber` to register emotion widget templates 
 * Added `plugin_dir` and `plugin_name` container parameter for each plugin. Parameters are prefixed by `\Shopware\Components\Plugin::getContainerPrefix`
+* Creating Forms in Registration or AddressController use classes from container now
 
 
 ## 5.2.9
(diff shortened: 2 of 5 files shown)


Tombro:
Just updated this PR. Now container is used in Backend Controller as well. Any feedback/questions on this one? Its been almost three months :(

engine/Shopware/Controllers/Backend/Address.php
```diff
@@ -139,7 +139,10 @@ public function save($data)
         $data['state'] = $data['state_id'];
 
         /** @var \Symfony\Component\Form\FormInterface $form */
-        $form = $this->get('shopware.form.factory')->create(AddressFormType::class, $model);
+        $form = $this->get('shopware.form.factory')->create(
+            get_class($this->container->get('shopware_account.form.addressform')),
```
xabbuh, line 143: I am not familiar with the Shopware internals, but this looks like a hack to me that shouldn't be necessary.

In Symfony, we first add form types that are registered as services in the container to the registry so that they will be returned when they are requested by their FQCN before the form type would be instanciated just by their FQCN using the default constructor. Is this different in Shopware? If so, I suggest to rather update the Shopware internals.

Tombro: Thanks for your comment. Basically you are right. In my case I am decorating the core form type like this:

Core form:

```
<service id="shopware_account.form.personalform" class="Shopware\Bundle\AccountBundle\Form\Account\PersonalFormType">
    <argument type="service" id="snippets" />
    <argument type="service" id="config" />
    <argument type="service" id="shopware_storefront.context_service" />
    <tag name="form.type" />
</service>
```

My decoration:

```
<service id="emgroup_registration.form.personalform" class="EMGroupRegistration\Components\Form\Account\PersonalFormType" decorates="shopware_account.form.personalform">
    <argument type="service" id="emgroup_registration.form.personalform.inner"/>
    <argument type="service" id="config" />
    <argument type="service" id="snippets" />
    <tag name="form.type" />
</service>
```

Without the change from this PR I am getting the decorated class, but i get this error though:

```
Fatal error: Uncaught Symfony\Component\Form\Exception\InvalidArgumentException: The type name specified for the service "shopware_account.form.personalform" does not match the actual name. Expected "Shopware\Bundle\AccountBundle\Form\Account\PersonalFormType", given "EMGroupRegistration\Components\Form\Account\PersonalFormType"
```

Any tips? Thanks!

xabbuh: You will probably be better off with creating a form type extension instead: http://symfony.com/doc/current/form/create_form_type_extension.html

Tombro: Thank you very much @xabbuh, works like a charm! :)

PR can be closed.

xabbuh: Cool. :) By the way, you can close the PR yourself. ;)

Tombro:
Oops, Button was too small :)

PR #661, opened 2016-08-15
added an initialize method which is called whenever a plugin is initi…

jochenmanz:
**Why:**

currently its not possible to autoload dependencies during container building except the dependencies which can be autoloaded with psr4 from the plugin root directory. In my case, i wanted to include the viison/address-splitter and use it for my services definied in my Resources/services.xml. When the Container is built, i am able to include my own autoloaded. If the Cntainer is already built, this is not possible.

What:

Added a small initialize function

Breaking something?

No.


jochenmanz:
It's only needded for commands definied inside the services.xml. A different solution might be needed so i am closing this PR and i am working on a better fix.

PR #663, opened 2016-08-15
Add missing method setCookie() to enable testing cookies

giginos:
The method is actually only available in Enlight_Controller_Response_ResponseHttp but it should exist in Enlight_Controller_Request_RequestTestCase, too.

### DIFF
--- a/engine/Library/Enlight/Controller/Response/ResponseTestCase.php
+++ b/engine/Library/Enlight/Controller/Response/ResponseTestCase.php
@@ -88,17 +88,6 @@ class Enlight_Controller_Response_ResponseTestCase
     {
         if (!empty($this->_cookies)) {
             $this->canSendHeaders(true);
-            foreach ($this->_cookies as $name => $cookie) {
-                setcookie(
-                    $name,
-                    $cookie['value'],
-                    $cookie['expire'],
-                    $cookie['path'],
-                    $cookie['domain'],
-                    $cookie['secure'],
-                    $cookie['httpOnly']
-                );
-            }
         }
         return $this;
     }
--- a/engine/Library/Enlight/Controller/Response/ResponseTestCase.php
+++ b/engine/Library/Enlight/Controller/Response/ResponseTestCase.php
@@ -80,6 +80,30 @@ class Enlight_Controller_Response_ResponseTestCase
     }
 
     /**
+     * Sends all cookies
+     *
+     * @return Enlight_Controller_Response_ResponseTestCase
+     */
+    public function sendCookies()
+    {
+        if (!empty($this->_cookies)) {
+            $this->canSendHeaders(true);
+            foreach ($this->_cookies as $name => $cookie) {
+                setcookie(
+                    $name,
+                    $cookie['value'],
+                    $cookie['expire'],
+                    $cookie['path'],
+                    $cookie['domain'],
+                    $cookie['secure'],
+                    $cookie['httpOnly']
+                );
+            }
+        }
+        return $this;
+    }
+
+    /**
      * Gets a cookie value
      *
      * @param string $name


bcremer:
The `ResponseTestCase` should not modify the superglobals, or produce side effects. 

Look for example at `\Zend_Controller_Response_HttpTestCase::sendHeaders()`. It does not send out real headers like the "real" class `\Zend_Controller_Response_Abstract::sendHeaders()`:
https://github.com/shopware/shopware/blob/5.2/engine/Library/Zend/Controller/Response/HttpTestCase.php#L41

To test the cookies that would be send you can access all cookies from `ResponseTestCase` with 
`\Enlight_Controller_Response_ResponseTestCase::getCookies()`.

giginos:
Hi @bcremer!
My problem is to unit-test a service that uses the $response->setCookie() method. So i need it in my test-response-object, too.

EDIT: I just call the "dispatch" method - if the "setCookie()" method does not exist in Enlight_Controller_Response_ResponseTestCase all tests via "dispatch"-method will fail, if the "setCookie()" method is used. Do you have a workaround for this?

giginos:
Please also see file: /engine/Library/Zend/Controller/Request/HttpTestCase.php

```
/**
 * Set a cookie
 *
 * @param  string $key
 * @param  mixed $value
 * @return Zend_Controller_Request_HttpTestCase
 */
public function setCookie($key, $value)
{
    $_COOKIE[(string) $key] = $value;
    return $this;
}
```

bcremer:
I am still not sure what you are trying to solve here.
To test code that modifies the response object you can call `$response->getCookies()` to run a assert on it.

Please provide an example of your testing usecase.

giginos:
I already use "getCookies()" to test them. But i also use the "setCookie()" to create them.

```
/**
     * @param Enlight_Controller_Request_Request $request
     * @param Enlight_Controller_Response_Response $response
     */
    public function setCookieVariablesByParameter(
        Enlight_Controller_Request_Request $request,
        Enlight_Controller_Response_Response $response)
    {
        $urlParams = $request->getQuery();
        $knownGetParameter = ParameterConfig::getKnownParameter();
        $cookieParameter = ParameterConfig::getCookieParameter();

        foreach ($knownGetParameter as $currentParameter) {
            if (array_key_exists($currentParameter, $urlParams)) {
                // Handle cookies
                if (array_key_exists($currentParameter, $cookieParameter)) {
                    $expire = $this->getExpire(
                        $cookieParameter[$currentParameter]['minutes'],
                        $cookieParameter[$currentParameter]['hours'],
                        $cookieParameter[$currentParameter]['days']
                    );

                    $response->setCookie(
                        $currentParameter,
                        $urlParams[$currentParameter],
                        $expire);

                    $response->sendCookies();
                }
            }
        }
    }
```

I've subscribed 'Enlight_Controller_Action_PreDispatch_Frontend_Index' and call the service via:

```
/**
     * @param Enlight_Event_EventArgs $args Passed arguments to event
     *
     * @return void
     * @author Rafael Kutscha <rafael.kutscha@twt.de>
     */
    public function saveFrontendParameters(Enlight_Event_EventArgs $args)
    {
        /** @var \Enlight_Controller_Action $controller */
        $controller = $args->getSubject();
        $request = $controller->Request();
        $response = $controller->Response();
        $this->parameterService->setCookieVariablesByParameter($request, $response);
    }
```

Do you now understand the problem, if $response->setCookie() is missing?

As i wrote before, the ZF1 does also implement it so i see no real problem to add this method in the shopware test response object.

bcremer:
OK, I think I now get what the problem is.

But please note that you were constantly mixing up `$response->setCookie()`, which already exists in `\Enlight_Controller_Response_ResponseTestCase` and `$response->sendCookies()` which is indeed missing in `Enlight_Controller_Response_ResponseTestCase`.

You are right, the method `Enlight_Controller_Response_ResponseTestCase::sendCookies()` should be added.

But it should not modify superglobals or delegate to PHP´s `sendCookies()`.
It should be good enough to provide an empty method so the API ist consistent with 
`Enlight_Controller_Response_ResponseHttp`.

giginos:
Japp. We do not really need the full business logic for testing. I've updated the PR for you.

bcremer:
Thank you @giginos, merged :+1:

PR #668, opened 2016-08-18
SW-12898 Fix sites-navigation target attribute is not set

DerKaito:
If the link-target is set for the sites-navigation in the frontend the target attribute is empty.
Its because a wrong Smarty-variable is used in the template.


fabianlupa (approved):

soebbing:
Hi @DerKaito! Thank you for your PR, good catch! Your fix will be included in the next release.

PR #669, opened 2016-08-19
Add missing article translations in listing

FiveDigital:
We need this fix to translate for example the pack unit.


OliverSkroblin:
Thanks for your pull request.
Article translations are already loaded over the ListProductService/*Gateway and shouldn't be reloaded in the sArticles.
I tried to reproduce the behavior but the unit and the article data (pack unit included) translated completely and in the right language. 
Can you write an test which reproduce the failure?

FiveDigital:
Perhaps this isn't needed anymore in 5.2. We had this issue in 5.1 and needed the translation for additional data. We'll review it without this change.

FiveDigital:
This is not needed anymore in 5.2

PR #671, opened 2016-08-22
Remove articleId from s_articles_attributes schema

samuelvogel:
The field `articleId` in  `s_articles_attributes` is obsolete since 5b1c93462b0ae1c4892124a2b1f0f2f51c613fe4 as it is `NULL` for new variants created with Shopware 5.2.

Exact line which changed the behavior: https://github.com/shopware/shopware/commit/5b1c93462b0ae1c4892124a2b1f0f2f51c613fe4#diff-20e7f1f163560b86045546cca64d4321L2297

Screenshot of `s_articles_attributes`:
screenshot 2016-08-20 14 47 11

Removing it `articleId` makes handling consistent wir all other attribute tables (which only have one primary key) and also resolves confusion on how to get the main variants attributes. This is sometimes necessary and could formerly be done with `Shopware\Models\Article\Article::getAttribute()` which was never reliable (could return the attributes of another non-main variant) and does not work at all anymore for variants created with Shopware 5.2.

This is a breaking change, so not really suitable for Shopware 5.2. On the other hand, the current state of Shopware 5.2 is broken as well. Another solution would be reverting the change to set `articleId` again, fixing broken attributes with a migration and marking `articleId` as deprecated. This way the current regression could be resolved and the PR be merged cleanly with the Shopware 5.2 release.


samuelvogel:
The following change contained in this PR is actually a bugfix without any backwards incompatible changes: https://github.com/shopware/shopware/pull/671/commits/653152544916ace5d7eaddb9cb79357b9be35769

Please reconsider merging at least this part!

soebbing:
Hi @samuelvogel! Thank you for your persistency and patience! 💙  We just merged your fix & test, it will be part of the Shopware release 5.3.4. 🎉

samuelvogel:
Great, thank you! Maybe you can re-evaluate removing the `articleId` column from `s_articles_attributes` schema for Shopware 5.4 again (the rest of this PR). Everybody who relies on it writes "unstable" code, which works for some articles and doesn't for others, it's a pitfall which regularly trips developers.

soebbing:
I agree with you, we should drop that column on 5.4 at least. I'll see what I can do 😄

FiveDigital:
And please drop the whole article details concept in 6.0

mitelg:
@FiveDigital there will be no **article** details in 6.0 :stuck_out_tongue_winking_eye: 
maybe product details :smile:

fixpunkt:
But a more descriptive name than "details" would be a nice touch 😉

PR #677, opened 2016-08-23
Added Category Join to Blog getListQueryBuilder function

phoerig:
In some cases it is necessary to get category information from blog article

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -26,6 +26,7 @@ This changelog references changes done in Shopware 5.2 patch versions.
     * `frontend_detail_image_default_image_slider_item`
     * `frontend_detail_images_image_slider_item`
 * Update `ongr/elasticsearch-dsl` to version 2.0.2
+* Added Category Join to Blog getListQueryBuilder function
 
 ## 5.2.5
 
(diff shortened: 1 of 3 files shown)


janbuecker:
Thanks @phoerig 

phoerig:
is there any way to add only the hook to the emotion widget?

the query extension i need, i can add otherwise

bcremer:
@phoerig All repositories are hookable so you can modify the query builder via the plugin system like this:

```php
<?php
namespace SwagHookExample;

use Doctrine\ORM\QueryBuilder;

class SwagHookExample extends Plugin
{
    {
        return [
            'Shopware\Models\Shop\Repository::getActiveQueryBuilder::after' => 'afterActiveQuery',
        ];
    }
    public function afterListQueryBuilder(\Enlight_Hook_HookArgs $args)
    {
        /** @var QueryBuilder $return */
        $return = $args->getReturn();

        $return->addSelect(['comments', 'category']);
        $return->leftJoin('blog.category', 'category');

        $args->setReturn($return);
    }
}
```

phoerig:
@bcremer I know, but i think it was goot for all Users to have the category data

PR #678, opened 2016-08-24
Changes to make db-connection more reliable and restartable in case it fails/break

fuckmaz:
Handling the DB-Connection this way makes way more sense and is more adaptive for the API-Resources.

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Resource.php
+++ b/engine/Shopware/Components/Api/Resource/Resource.php
@@ -563,7 +563,10 @@ abstract class Resource
                                   ->reset('db_connection')
                                   ->load('models');
 
-        $this->getContainer()->load('db_connection');
+        $kernel = $this->getContainer()->get('kernel');
+        $dbConn = $kernel->getConfig()['db'];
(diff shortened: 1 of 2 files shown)


bcremer:
Can you please elaborate in which cases this is more reliable. Can you please provide some testcases.

fuckmaz:
We experienced that when the DB-Connection breaks after a long time (in our case it was a Importer-Plugin that has runtimes of about 5 hours) it won't reinitialize itself, which results in Errors with every operation. With these two changes we were able to reinitialize the DB-Connection from our Plugin and continue Insert- and Update-Operations to the DB.

So it's more reliable when the DB-Connection is needed for a long timespan.

NiklasTeich:
@postcatz Have you already tried to manually set set_time_limit, default_socket_timeout and mysql.connect_timeout before executing these long-time plugin scripts? Maybe a possible workaround to override the system defaults in your script.

fuckmaz:
Yes we tried all that. It's not a server-side problem as we found. Our limits were already set to 24hours each.

janbuecker:
Thanks @postcatz 

PR #679, opened 2016-08-24
Disable loading of widgets of deactivated plugins

hanneswernery:
Widgets of deactivated plugins should neither be loaded nor be able to be added to the widget list, since they cause a JS error and crash the widget panel.
This fix ensures that only Shopware widgets or widgets of active plugins are loaded. This also ensures they are not shown in the widget list (therefore cannot be added to the widget panel)


bcremer:
Thanks for your Pull Request :+1: 

One thing, can you please revert the chmod changes of 
 `engine/Shopware/Controllers/Backend/Widgets.php`.  The permissions in your PR are changed from to `644` to `755`.

hanneswernery:
I changed the permissions and updated the branch.

PR #681, opened 2016-08-24
Fix category attributes frontend format regression

FiveDigital:
In Shopware < 5.2 the keys of category attribute had the lowerCamelCase format since they where given to the frontend throught `sCategories::sGetCategoryContent()` which didn't use the `LegacyStructConverter` but a doctrine query builder (https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Core/sCategories.php#L529 and https://github.com/shopware/shopware/blob/5.1/engine/Shopware/Core/sCategories.php#L657).

After upgrading our customer's shop to 5.2 we had to change all listing templates because of this regression. This PR offers and approach to remain backwards compatible.


FiveDigital:
This issue is still bugging us in every shop we update from 5.1 to 5.2 and is very annoying.

janbuecker:
Thanks @FiveDigital 

Since 5.2 we are using names generated from the doctrine models. This is a breaking change and we do not intend to change this behaviour. If this change is still important to you, feel free to create a new pull request or create an issue on https://issues.shopware.com.

PR #683, opened 2016-08-24
Fix snippet content when name is given without quotes

andrewisplinghoff:
A snippet definition {s name=foo/bar} without quotes around 'foo/bar' previously led to a PHP warning on runtime, because a division by zero occurs. This was due to the following faulty code being generated:

```
<?php $_smarty_tpl->smarty->_tag_stack[] = array('snippet', array('name'=>'foo'/'bar',
[...]
```

After this change, the following valid code is produced:

```
<?php $_smarty_tpl->smarty->_tag_stack[] = array('snippet', array('name'=>'foo/bar',
[...]
```


janbuecker:
I don't think that this is necessary since it works with quotes. If your trying to use special characters in your bash, you will likely run into issues, too.

/cc @bcremer

janbuecker:
Thanks @andrewisplinghoff 

PR #684, opened 2016-08-25
Changed stripTags to htmlEncode in the backend Snippet-Translation form

oktupol:
In the recent commit [5644302](https://github.com/shopware/shopware/commit/56443027b469aa25b5888042ca53650edabb61b2), the stripTags-call was added for the placeholder in the snippet-translation form to avoid breaking the textareas by having HTML-tags in the snippet. However, this isn't sufficient, because something like `"> This breaks the textarea` still causes an error. For that reason, I changed the stripTags-call with an htmlEncode-call.


htkassner:
Thank your for your contribution. We have slightly modified your PR to still strip the tags and then do an htmlEncode. This will be released with the 5.3 version.

PR #685, opened 2016-08-25
Fix PluginBase._off removing all events

domu1de:
The method `PluginBase._off` falsely removes all events from an element.

This issue is caused by the wrong usage of `$.each` in _jquery.plugin-base.js:214_.
`$.each` expects a callback of type `Function(Integer indexInArray, Object value)`, in this case `Function(Object value)` is given.


htkassner:
Thank you for your contribution. This was already fixed with commit 2383f843. I will close this one now.

PR #686, opened 2016-08-25
Avoid potential conflicts in parseQueryString

domu1de:
The `for...in` loop shouldn't be used to iterate over arrays because `Array.prototype` could have been extended by 3rd-party libraries. This would break this function.

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.infinite-scrolling.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.infinite-scrolling.js
@@ -9,25 +9,25 @@
      * @returns {{}} Object with all extracted parameters
      */
     var parseQueryString = function(url) {
-	    var qparams = {},
-		    parts = (url || '').split('?'),
-		    qparts, qpart;
+        var qparams = {},
+            parts = (url || '').split('?'),
+            qparts, qpart;
 
-	    if (parts.length <= 1) {
-		    return qparams;
-	    }
+        if (parts.length <= 1) {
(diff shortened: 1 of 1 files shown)


klarstil:
Hey @domu1de,

thank you for your pull request & your contribution. I'm not sharing your opinion on this topic though. You're right, 3rd-party libraries can be extending the vanilla javascript objects with additional functionality but this doesn't mean these libraries are aiming to change the complete behavior of these functions nor are they're limited to `Array.prototype`. They're working more like a polyfills (browser support) & adding additional syntax sugar like https://github.com/andrewplummer/Sugar. I'm getting why you want to change it but there's no need for that though.

domu1de:
Thanks for the fast reaction @klarstil!

Maybe I wasn't clear about what's the issue.
I'm totally with you, it's absolutely fine and sometimes necessary to extend `Array.prototype`, but the current implementation of `parseQueryString` doesn't take this into consideration.

Check this fiddle: https://jsfiddle.net/kb9n7tew/

-> `Uncaught TypeError: qparts[i].split is not a function(…)`

Actually we wanna iterate over the array content but instead we're iterating over all array members, including polyfills and extensions.

So I think there's indeed a need to change that. ;)

bcremer:
@domu1de I am not a Javascript guy but I do not understand why you want to replace the foreach.

If a bogus library or userland code modifies "core" Javascript methods in their behavior we have a much bigger problem. 
Why stop with foreach then? What about a library that re-implements `Array.prototype.split = function () {};`?

PR #688, opened 2016-08-26
[FEATURE] Use DI container to load API resources

PortaltechGithub:
Currently the API resources are initialized using `new`. With this change, they are loaded via the DI container, which allows to extend or replace them later on in own plugins. For clarity, I have used the new namespace `shopware.api` in the service definition.

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/services.xml
+++ b/engine/Shopware/Components/DependencyInjection/services.xml
@@ -475,5 +475,21 @@
             <argument>%kernel.root_dir%/engine/Shopware/Components/Check/Data/System.xml</argument>
             <argument type="service" id="db_connection"/>
         </service>
+
+        <service id="shopware.api.resource" class="Shopware\Components\Api\Resource\Resource" shared="false" />
+        <service id="shopware.api.address" class="Shopware\Components\Api\Resource\Address" shared="false" />
+        <service id="shopware.api.article" class="Shopware\Components\Api\Resource\Article" shared="false" />
+        <service id="shopware.api.cache" class="Shopware\Components\Api\Resource\Cache" shared="false" />
+        <service id="shopware.api.category" class="Shopware\Components\Api\Resource\Category" shared="false" />
+        <service id="shopware.api.country" class="Shopware\Components\Api\Resource\Country" shared="false" />
+        <service id="shopware.api.customer" class="Shopware\Components\Api\Resource\Customer" shared="false" />
+        <service id="shopware.api.customergroup" class="Shopware\Components\Api\Resource\CustomerGroup" shared="false" />
+        <service id="shopware.api.manufacturer" class="Shopware\Components\Api\Resource\Manufacturer" shared="false" />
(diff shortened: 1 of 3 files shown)


bcremer:
I am not sure if that is a good idea as pulling an api resource directly from the DI-Container will leave it "unconfigured". So no entity-manager, nor the container get's injected.
If that is left to the consumer I see no benefit against using `new` directly.

The complete solution could be to change the setter injection of the resources to proper constructor parameters and inject them directly in the DI-Container. 

That way these resources can pulled from the DI-Container without any further configuration.

PortaltechGithub:
Right now, the resources are also "unconfigured" using the new command, thus it would be no change. But I like your idea, will you implement it?

mohrekopp:
Wouldn't it make sense, if the configuration is based on tags instead of naming conventions? That would be easier for plugin developers I think...
The consequence is, that the resource-manager have to be instantiated via DI-container and can't be used statically.

jeboehm:
We use the following service definition. Because the resources could be changed later (set\* methods), shared should always set to false.

`
        <service id="shopware.api.order" class="Shopware\Components\Api\Resource\Order" shared="false">
            <factory class="Shopware\Components\Api\Manager" method="getResource"/>
            <argument type="string">order</argument>
        </service>
`

engine/Shopware/Components/DependencyInjection/services.xml
```diff
@@ -475,5 +475,82 @@
             <argument>%kernel.root_dir%/engine/Shopware/Components/Check/Data/System.xml</argument>
             <argument type="service" id="db_connection"/>
         </service>
+
+        <service id="shopware.api.resource_factory" class="Shopware\Components\Api\Manager"/>
+
+        <service id="shopware.api.resource" class="Shopware\Components\Api\Resource\Resource" shared="false">
+            <factory service="shopware.api.resource_factory" method="getResource"/>
```
bcremer: Won't that create a loop? The DI-Container is using `Shopware\Components\Api\Manager` as a factory which in case asks the DI-Container for the Service.

PortaltechGithub:
If we use the API-Manager as factory, we have the problem that the resources must still be instantiated with "new". Trying to load a resource from DI within the getResource method would generate a recursion. To fix the problem, we would have to replace all the parts in code where a resource is instantiated by a static call to the API-Manager. And that's a bigger task, as I don't know, which plugins are using this method. Thus, I keep the first approach to get resources into the DI container and add only the shared=false to the service.xml.

PortaltechGithub:
@bcremer Do you already know when you want to merge the PR?

bcremer:
@PortaltechGithub unfortunately I cannot answer this question. We currently have a evergrowing backlogs of tickets in the issue tracker and PRs here on github. I'll do my best to get this PRs into the next sprints.

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/5da9aadb2944f9b8addd31b82a2d980387b388db. Thanks for your patience :+1:

PR #690, opened 2016-08-26
Convert attributes to array in all convert functions

FiveDigital:
The convert function for products and categories already convert the attributes to an array calles `attributes`. This PR adds the conversion for all extendable structs to that you dont't have to use e.g. `$group.attributes.core->get('my_cool_attribute')` in the frontend templates


FiveDigital:
Time constraints? This one is easy to merge ...

PR #694, opened 2016-08-29
SW-15988 - Fixed backend order PDF download

MyIgel:
Fix: Without the exit the server resets the connection for some reasons

In addition to #666


dnoegel:
Not sure, but could this be a regression from 86e746db2f957870622f102736f338ac202461a2 ?
Before that revision, we prevented plugins from registering to that action for good reason - not sure, why this was removed in the first place.

MyIgel:
I dont't think that it is related..
The Problem here was: The server send an RST-Package to  the client somewhere in the middle of the file download (and that only on the live server and not in my dev environment) -> The client stopped the doenload because of an "Network Error" (interesting: curl has no problem downloading it)

MyIgel:
@dnoegel: I have tested it with an old version (without 86e746d), so it should not be reelated

PR #696, opened 2016-09-01
shortcut for configuring cache clear on Plugin operations

hrombach:
| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | successfully tested manually |
| Related tickets? | none found |
| How to test? | Override the field in a plugin controller class and activate/deactivate. |
## Description

With the current way of clearing the cache, if you have a plugin that extends the theme you would need to override at least the activate, deactivate, and uninstall methods in order to correctly clear the cache and recompile the theme, which felt like a lot of redundant code. 

We figured that, if we add a field to the Plugin Abstract class and get the cache clear flag from there, we would only have to override the field in order to clear the desired caches.

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -2,6 +2,12 @@
 
 This changelog references changes done in Shopware 5.2 patch versions.
 
+## 5.2.7
+
+[View all changes from v5.2.6...v5.2.7](https://github.com/shopware/shopware/compare/v5.2.6...v5.2.7)
+
+* added protected field `Shopware\Components\Plugin::$cachesToClear`, override this in your Plugin main class (YourPluginName.php) to more easily control Cache behavior across update, activate, deactivate, uninstall
+
 ## 5.2.6
 
 [View all changes from v5.2.5...v5.2.6](https://github.com/shopware/shopware/compare/v5.2.5...v5.2.6)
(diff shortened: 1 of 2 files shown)


janbuecker:
You are able to achieve the same behaviour by implementing the methods yourself and manually call `$context->scheduleClearCache(..)` afterwards, but I see your intention here.

I don't like using an open variable, a setter for a private variable would fit here better.

What do you think @bcremer?

bcremer:
I would like to prevent as much magic and state in the plugin class as possible. I think it is fine to implement those  four methods in your own plugin with your preferred cache configuration.
A bit of code duplication is ok in my opinion.

hrombach:
Well our thinking was, why would a plug in that needs to clear the theme cache only clear it in some of these? If a plugin has template components it would make sense to always clear the cache if the plugin is Uninstalled/installed/deactivated/activated. So we/I don't really see why you would need to override 4 methods individually, since in any case I can think of it would make sense to always have the same cache behavior for each of these four actions.

dnoegel:
Couldn't we control the caches to be cleared via plugin.xml? I don't want to bloat the base class as well - but if we could just configure it, everyone'd be happy :)

hrombach:
Sounds like a good idea :)

janbuecker:
Thanks @hrombach 

PR #697, opened 2016-09-01
Remove customergroup blocked articles in basket

shyim:
| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | There was a issue. i cant find it lel |
| How to test? | Add article to cart which is blocked for merchants, login to a merchant account in checkout and the article should be removed |
## Description

We had some troubles with customers, because some products are blocked for merchants and merchants has ordered the products with that bug :)

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -1073,6 +1073,8 @@ class sBasket
      */
     public function sGetBasket()
     {
+        $this->removeCustomerGroupBlockedArticles();
+
         // Refresh basket prices
         $basketData = $this->db->fetchAll(
             'SELECT id, modus, quantity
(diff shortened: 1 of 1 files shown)


janbuecker:
We should prevent adding it too. So please consider this case in `sAddArticle()`.

shyim:
@janbuecker added

dnoegel:
This will currently iterate all items in the basket and perform those queries. This should be changed in a way, that only one query is needed to determine all items, which need to be removed. The deletion of items then could happen in a transaction, for example.

shyim:
@dnoegel updated :)

engine/Shopware/Core/sBasket.php
```diff
@@ -2774,4 +2780,27 @@ private function getBasketQuantity($quantity, $basketProduct, $article)
...
         return $newQuantity;
     }
+
+    /**
+     * @param $articleId
+     * @return string
+     */
+    private function isProductBlockedByCustomerGroup($articleId)
+    {
+        return $this->db->fetchOne('SELECT 1 FROM s_articles_avoid_customergroups WHERE customergroupId = ? AND articleID = ?', [
```
mitelg: I think it is more readable, if you use something like :customerGroupId instead of ?

mitelg (changes requested):

engine/Shopware/Core/sBasket.php
shyim: Aye Aye, Captain!

mitelg (approved):

janbuecker:
We've changed the code entirely so that it resulted in a new commit 474703992c4e2b91e5a099798737b6b5a4804bb4.

PR #704, opened 2016-09-05
[Papercut] Fix search drop down position

klarstil:
## Description

This PR fixes the behavior of the search drop down. The drop down was set with a magic number in the code (which had no effect) and in the SCSS using the !important flag. This PR removes the magic number and updates the position when the customer fires a search request, so the search drop down will be positioned correctly even on resize of the backend.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-16256 |
| How to test? | Open up the backend with the new "Connect" menu entry in your main menu & type something in the search field (at least 3 chars). Now make sure the search drop down is not under the "Connect" menu entry. It should be under the search field again. There was a Safari specific change, so please make sure to check it in Safari (Mac OS X) |


klarstil:
Internal QA test was successful. @bcremer You can merge it if you want.

PR #707, opened 2016-09-05
[Papercut][SW-15997] Global Bcc Mail Option

shyim:
## Description

It adds a global bcc mail fieldin the mail settings

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | [Issue](https://issues.shopware.com/#/issues/SW-15997) |
| How to test? | Go to Grundeinstellungen => Shopeinstellungen=> Stammdaten and insert a bcc mail and order a product :) |

### DIFF
--- a/_sql/migrations/786-add-bcc-mail-field.php
+++ b/_sql/migrations/786-add-bcc-mail-field.php
@@ -1,0 +1,12 @@
+<?php
+
+class Migrations_Migration786 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $this->addSql('SET @menu_id = (SELECT id FROM s_core_config_forms WHERE label=\'Stammdaten\');');
+        $this->addSql('INSERT INTO s_core_config_elements (form_id, name, value, label, description, type, required, scope) VALUES(@menu_id, "mailBcc", "s:0:\"\";", "Shopbetreiber BCC E-Mail", "Setzt bei jeder Mail einen BCC (Mehrere E-Mail Adressen können mit Komma getrennt werden)", "text", 0, 1);');
+        $this->addSql('SET @element_id = (SELECT id FROM s_core_config_elements WHERE name = "mailBcc");');
+        $this->addSql('INSERT INTO s_core_config_element_translations (element_id, locale_id, label, description) VALUES(@element_id, 2, "Show owner bcc mail", "Allows to set on any mail a bcc (multiple email addresses can be separated by commas)");');
+    }
+}
--- a/engine/Shopware/Components/BccSubscriber.php
+++ b/engine/Shopware/Components/BccSubscriber.php
@@ -1,4 +1,27 @@
 <?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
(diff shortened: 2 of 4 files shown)


janbuecker:
Is there any reason why you don't use `$this->addBcc()` here? Your implementation isn't really a bcc.

shyim:
@janbuecker i tried addBcc and dont received messages. I can change it back to addBcc

bcremer:
Instead of modifiing `Enlight_Components_Mail_Send` you should add a separate subscriber class that listens to the `Enlight_Components_Mail_Send` event and adds the global BCC options.

_sql/migrations/786-add-bcc-mail-field.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+class Migrations_Migration786 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $sql = <<<'EOD'
+            SET @menu_id = (SELECT id FROM s_core_config_forms WHERE label='Stammdaten');
```
bcremer: Every Statement has to be a separate call to `addSQL()`. This is due to a long standing bug in PHP's PDO. See: https://bugs.php.net/bug.php?id=61613

bcremer:
@shyim Are you going to change this to an event subscriber as noted in the review? We can also close this PR so the issue is free to pickup for the next person if you prefer that.

engine/Library/Enlight/Components/Mail.php
```diff
@@ -329,6 +335,19 @@ public function send($transport = null)
             )
         );
 
+        $bccs = Shopware()->Config()->get('mailBcc');
```
hlohaus: pls, don't use "Shopware()" in a Service.
 https://developers.shopware.com/blog/2015/11/11/best-practices-of-shopware-plugin-development/#shopware

shyim:
@hlohaus @bcremer its changed :)

engine/Shopware/Components/BccSubscriber.php
```diff
@@ -0,0 +1,41 @@
...
+            return;
+        }
+
+        $bccs = $this->container->get('config')->get('mailBcc');
+
+        /** @var \Enlight_Components_Mail $mail */
+        $mail = $args->get('mail');
+
+        if (!empty($bccs)) {
+            $mail->addBcc(explode(',', $bccs));
```
FiveDigital: Perhaps you should also `trim` the bccs to remove empty spaces.

shyim: fixed

engine/Shopware/Components/BccSubscriber.php
```diff
@@ -0,0 +1,45 @@
...
+            return;
+        }
+
+        $bccs = $this->container->get('config')->get('mailBcc');
+
+        /** @var \Enlight_Components_Mail $mail */
+        $mail = $args->get('mail');
+
+        if (!empty($bccs)) {
+            $bccs = array_map(function ($value) {
```
janbuecker: There is no need for an anonymous function here. You can use `array_map('trim', explode(',', $bccs))`

PR #709, opened 2016-09-06
[Papercut][Sw-15946] Fix-Surchargstring-change-after-remove-country-in-paymentmeans

TobiasWilming:
This fixes a bug where the surcharge for countries isn't correctly change when you deselect a country with a surcharge and then save.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-15946 |
| How to test? | Go to backend->configuration->payment method. Then one of the avaiable paymentmethods. Chose a country and then set a surcharge. After that deselect the country for that surcharge and save. With this Fix there souldn't be the entry for that surcharge in the table s_core_paymentmeans or the surchage in the basket if you chose an article and that paymentmethods. |

### DIFF
--- a/engine/Shopware/Controllers/Backend/Payment.php
+++ b/engine/Shopware/Controllers/Backend/Payment.php
@@ -245,7 +245,7 @@ class Shopware_Controllers_Backend_Payment extends Shopware_Controllers_Backend_
                 }
                 $data['shops'] = $shops;
             }
-            $data = $this->filterSurchargeString($data['surcharge'], $data['countries']);
+            $data['surchargeString'] = $this->filterSurchargeString($data['surchargeString'], $data['countries']);
 
             $payment->fromArray($data);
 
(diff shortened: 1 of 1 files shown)


bcremer:
Please have a look at our [Coding Standards](https://developers.shopware.com/developers-guide/coding-standards/#php-coding-standard) and reformat you code accordingly.
To fix the coding style you can use tools like [php-cs-fixer](http://cs.sensiolabs.org/) or [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer). Also PhpStorm can be configured to use PSR-1 and PSR-2 to reformat your code automatically.

engine/Shopware/Controllers/Backend/Payment.php
```diff
@@ -273,12 +274,49 @@ public function updatePaymentsAction()
...
 
-            $this->View()->assign(array('success'=>true, 'data'=>$data));
+            $this->View()->assign(array('success' => true, 'data' => $data));
         } catch (\Doctrine\ORM\ORMException $e) {
-            $this->View()->assign(array('success'=>false, 'errorMsg'=>$e->getMessage()));
+            $this->View()->assign(array('success' => false, 'errorMsg' => $e->getMessage()));
         }
     }
 
+    public function ChecksurchargeString($data)
```
OliverSkroblin: function names should starts with lower case

PR #712, opened 2016-09-06
[Papercut][SW-15897] make possible to use html mails in status mails

shyim:
## Description

It makes possible to use html mails in status mails

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | [SW-15897](https://issues.shopware.com/#/issues/SW-15897) |
| How to test? | Write to a status mail a html content and change a order to that status |

### DIFF
--- a/engine/Shopware/Controllers/Backend/Order.php
+++ b/engine/Shopware/Controllers/Backend/Order.php
@@ -1065,7 +1065,13 @@ class Shopware_Controllers_Backend_Order extends Shopware_Controllers_Backend_Ex
         $mail = clone Shopware()->Container()->get('mail');
         $mail->clearRecipients();
         $mail->setSubject($this->Request()->getParam('subject', ''));
-        $mail->setBodyText($this->Request()->getParam('content', ''));
+
+        if ($this->Request()->getParam('isHtml')) {
+            $mail->setBodyHtml($this->Request()->getParam('contentHtml', ''));
+        } else {
+            $mail->setBodyText($this->Request()->getParam('content', ''));
+        }
+
         $mail->setFrom($this->Request()->getParam('fromMail', ''), $this->Request()->getParam('fromName', ''));
         $mail->addTo($this->Request()->getParam('to', ''));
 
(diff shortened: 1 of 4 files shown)


engine/Shopware/Controllers/Backend/Order.php
```diff
@@ -1434,11 +1440,15 @@ private function getMailForOrder($orderId, $statusId)
                 'data' => array(
                     'error' => false,
                     'content' => $mail->getPlainBodyText(),
+                    'contentHtml' => $mail->getPlainBody(),
                     'subject' => $mail->getPlainSubject(),
                     'to' => implode(', ', $mail->getTo()),
                     'fromMail' => $mail->getFrom(),
                     'fromName' => $mail->getFromName(),
                     'sent' => false,
+                    'isHtml' => (bool) Shopware()->Db()->fetchOne('SELECT `ishtml` FROM s_core_config_mails WHERE `name` = ?', [
```
janbuecker: I think you can replace the SQL statement with something like `!empty($mail->getBodyHtml())` since the `bodyHtml` property of the mail object only gets filled, if the template's `isHtml` is `true`.

engine/Shopware/Controllers/Backend/Order.php
```diff
@@ -1434,11 +1440,13 @@ private function getMailForOrder($orderId, $statusId)
                 'data' => array(
                     'error' => false,
                     'content' => $mail->getPlainBodyText(),
+                    'contentHtml' => $mail->getPlainBody(),
                     'subject' => $mail->getPlainSubject(),
                     'to' => implode(', ', $mail->getTo()),
                     'fromMail' => $mail->getFrom(),
                     'fromName' => $mail->getFromName(),
                     'sent' => false,
+                    'isHtml' => !empty($mail->getPlainBody()),
```
janbuecker, line 1449: Why do you use `getPlainBody()` instead of `getBodyHtml()` here? Don't you want to check against html mails? Have you tested your code afterwards?

janbuecker (changes requested):

engine/Shopware/Controllers/Backend/Order.php
shyim: Done

shyim: getPlainBody returns the html as string, getBodyHtml returns a Zend_Mime_Part :) Yes i tested the code
setter: https://github.com/shopware/shopware/blob/5.2/engine/Library/Enlight/Components/Mail.php#L198
getter:
https://github.com/shopware/shopware/blob/5.2/engine/Library/Enlight/Components/Mail.php#L209

I know the method name is stupid :D

janbuecker (approved):

PR #720, opened 2016-09-07
[Papercut][Sw-14787] fix voucher select in cancel orders

TobiasWilming:
In the cancelled order the voucher selection now shows % if percental and a symbol if absolute. The voucher Description is also added to tell which one is which.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-14787 |
| How to test? | Reload the snippet database. Create a percental and an absolute voucher. Open the cancelled order window. Select an Element where no voucher is already send. Open the Customer feedback and select a voucher. (Must be tested for eng and ger because of the snippets) |

### DIFF
--- a/themes/Backend/ExtJs/backend/canceled_order/model/voucher.js
+++ b/themes/Backend/ExtJs/backend/canceled_order/model/voucher.js
@@ -60,10 +60,10 @@ Ext.define('Shopware.apps.CanceledOrder.model.Voucher', {
                     return value;
                 }
                 if ( record.get('percental') == 1 ) {
-                    return Ext.String.format("{s name=sendVoucherWorthPercent}Send voucher worth {literal}{0}{/literal}% vouchername: {literal}{1}{/literal}{/s}", value, record.get('description'));
+                    return Ext.String.format("{s name=sendVoucherWorthPercent}Send voucher worth [0]% vouchername: [1]{/s}", value, record.get('description'));
                 }
                 else {
-                    return Ext.String.format("{s name=sendVoucherWorth}Send voucher worth {literal}{0}{/literal} vouchername: {literal}{1}{/literal}{/s}", value, record.get('description'));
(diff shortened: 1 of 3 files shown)


klarstil:
Hey @TobiasWilming,
thanks for your contribution. I checked your PR and having a tip for you. You're using the default syntax of `Ext.String.format`. Due to the fact that our Javascript files in the backend are parsed by Smarty and Smarty's delimiters are `{` & `}`. We created an override fowhich enables you to use `[0]` instead of `{0}` for variable placeholders, so you don't need the `{literal}` tags in your snippets.

Take a look at the override:
https://github.com/shopware/shopware/blob/5.2/engine/Library/ExtJs/overrides/Ext.Template.js

Here's a working example:
https://github.com/shopware/shopware/blob/5.2/themes/Backend/ExtJs/backend/customer/view/detail/window.js#L202

Your branch is out-of date with the base branch. Would you be so kind and rebase it please?

themes/Backend/ExtJs/backend/canceled_order/model/voucher.js
```diff
@@ -58,8 +59,12 @@ Ext.define('Shopware.apps.CanceledOrder.model.Voucher', {
                 if(record && record.get('id') == -1) {
                     return value;
                 }
-                return Ext.String.format("{s name=sendVoucherWorth}Send voucher worth {literal}{0}{/literal}{/s}", value);
-
+                if ( record.get('percental') == 1 ) {
+                    return Ext.String.format("{s name=sendVoucherWorthPercent}Send voucher worth [0]% vouchername: [1]{/s}", value, record.get('description'));
+                }
```
OliverSkroblin: Please move the else in same row as the "}" of the first IF clause

PR #724, opened 2016-09-07
[Papercut][SW-14287] cron rating

teiling88:
## Description

With this PR, no rating e-mail is send if all articles of the order are inactive.

| Questions | Answers |
| --- | --- |
| BC breaks? | yes (no modus != 0 entries are in the email template available) |
| Tests pass? | yes |
| Related tickets? | Ticket: SW-14287 |

### DIFF
--- a/engine/Shopware/Plugins/Default/Core/CronRating/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Core/CronRating/Bootstrap.php
@@ -43,7 +43,7 @@ class Shopware_Plugins_Core_CronRating_Bootstrap extends Shopware_Components_Plu
     /**
      * @param Enlight_Components_Cron_EventArgs $job
      *
-     * @return string|void
+     * @return void|string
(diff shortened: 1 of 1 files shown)


engine/Shopware/Plugins/Default/Core/CronRating/Bootstrap.php
```diff
@@ -40,10 +40,16 @@ public function install()
...
+     * @param Enlight_Components_Cron_EventArgs $job
+     *
+     * @return string
+     * @throws \Exception
+     */
     public function onRun(Enlight_Components_Cron_EventArgs $job)
     {
         if (empty(Shopware()->Config()->voteSendCalling)) {
-            return;
+            return '';
```
Haehnchen: think of this '' != null ;)

teiling88: i fixed it :)

PR #725, opened 2016-09-07
[Papercut][SW14753] increase password hash size

teiling88:
## Description

increase password hash size in s_user table 

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-14753 |

### DIFF
--- a/_sql/migrations/786-increase-password-hash-size.php
+++ b/_sql/migrations/786-increase-password-hash-size.php
@@ -1,0 +1,9 @@
+<?php
+
+class Migrations_Migration786 extends Shopware\Components\Migrations\AbstractMigration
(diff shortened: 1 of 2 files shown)


bcremer:
Has `355` a special meaning or reason to choose? Maybe we can increase it to something like `1024` directly.

teiling88:
no special meaning - I changed it to 1024.

bcremer:
Interesting read on the implications of varchar lenghts for alter tables: https://blog.gabriela.io/2016/09/08/varchar-fields-on-mysql-5-7/

PR #727, opened 2016-09-07
[Papercut][SW-15035] Backend orders list: use company/name instead of customerId for sorting by customer

dneustadt:
## Description

Sorting by customer in backend orders list will order by customer id. The user
expects the list to be ordered alphabetically by company name or last name/ first name.
This pull request changes the order queue to company, firstName, lastName when the
sorting property is customerId.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-15035 |
| How to test? | Order by customer in backend order list |

### DIFF
--- a/themes/Backend/ExtJs/backend/order/view/list/list.js
+++ b/themes/Backend/ExtJs/backend/order/view/list/list.js
@@ -319,7 +319,10 @@ Ext.define('Shopware.apps.Order.view.list.List', {
                 header: me.snippets.columns.customer,
                 dataIndex: 'customerId',
                 flex:2,
-                renderer: me.customerColumn
+                renderer: me.customerColumn,
+                getSortParam: function() {
+                    return 'customerName';
+                }
             },
             {
                 header: me.snippets.columns.customerEmail,
--- a/engine/Shopware/Controllers/Backend/Order.php
+++ b/engine/Shopware/Controllers/Backend/Order.php
@@ -369,11 +369,11 @@ class Shopware_Controllers_Backend_Order extends Shopware_Controllers_Backend_Ex
                     array_splice($sort, 1, 0, [
                         [
                             'property' => 'billing.lastName',
-                            'direction' => $orderBy[0]['direction']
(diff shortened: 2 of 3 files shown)


OliverSkroblin:
Hi saberdude,
thanks for your contribution. I would prefer if you implement this switch in extjs and not in the php repository.
With this solution it is no more possible to sort over the customer id.

dneustadt:
Hi Oliver,
yes, i understand that a strict condition like that doesn't belong in a repository method. I moved the condition to the order backend controller as an additional switch case. considering the getList method will only be used in the context of the backend app. unfortunately i can't invest any more time right now to utilize extjs for the same matter.

dneustadt:
I managed to pass a custom sort parameter when sorting by customer so there is no need replacing customerId property permanently. kept customerId as dataIndex for BC.

themes/Backend/ExtJs/backend/order/view/list/list.js
```diff
@@ -319,7 +319,10 @@ Ext.define('Shopware.apps.Order.view.list.List', {
                 header: me.snippets.columns.customer,
                 dataIndex: 'customerId',
                 flex:2,
-                renderer: me.customerColumn
+                renderer: me.customerColumn,
```
htkassner, line 322: You could also make use of the doSort function of Ext.grid.column.Column. Doing that would not require any changes to the backend code. 

```
doSort: function(state) {
        var ds = this.up('tablepanel').store;
        ds.sort({
            property: 'customerName',
            direction: state
        });
    }
```

As far as I know sort also accepts array, so you could also add billing sorters. What do you think @OliverSkroblin ?

dneustadt: The problem with this solution is that the 'orders' alias is prepended to every property in the backend controller, so you would need a switch case anyway. That is unless you make use of doSort on every column and always add the alias there.

PR #728, opened 2016-09-08
[Papercut][SW-13559] Fix tax display

htkassner:
## Description

- Why is it necessary? Fixes display of tax for numbers without decimals and for german locale
- What does it improve? look & feel

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-13559 |
| How to test? | Just checkout a product and watch taxes display. Until now in german language a floating point number with a point character was shown. Now it shows no decimals for non decimals and a comma in german tax display. |


themes/Frontend/Bare/frontend/checkout/cart_footer.tpl
```diff
@@ -139,7 +139,13 @@
                 {block name='frontend_checkout_cart_footer_field_labels_taxes'}
                     {if $sUserData.additional.charge_vat}
                         {foreach $sBasket.sTaxRates as $rate => $value}
-
+                            {if ($rate * 100) % 100 == 0}
```
bcremer: `($rate * 100) % 100 == 0` is this supposed to be a `isFloat` kind of check?

themes/Frontend/Bare/frontend/checkout/cart_footer.tpl
```diff
@@ -139,7 +139,13 @@
...
+                                {$rate = $rate|number_format}
+                            {else}
+                                {if $Locale == "de_DE"}
+                                    {$rate = $rate|number_format:2:",":"."}
```
bcremer: Should't this be solved generally with a locale aware number formatter?
There is https://github.com/shopware/shopware/blob/5.2/engine/Library/Enlight/Template/Plugins/modifier.number.php#L33 but I am not sure if that solves your problem.

htkassner:
Gonna recap this. Didn´t know there was a custom modifier!

htkassner:
Last commit is more general, uses existing functions and also unifies tax presentation for pdf documents.

snippets/documents/index.ini
```diff
@@ -18,7 +18,7 @@ DocumentIndexOrderID = "Order No.:"
 DocumentIndexPageCounter = "Page {$page+1} of {$Pages|@count}"
 DocumentIndexSelectedDispatch = "Selected shipping type"
 DocumentIndexSelectedPayment = "Selected payment method"
-DocumentIndexTax = "Plus {$key}% VAT:"
+DocumentIndexTax = "Plus {if !$key|is_float}{$key = $key|number_format}{else}{$key = $key|number}{/if}{$key} % VAT:"
```
janbuecker: why don't you use this snippet? or isn't it working?

```
{if !$key|is_float}{$key|number_format}{else}{$key|number}{/if}
```

htkassner: Stupidness.

bcremer:
I have the feeling the following line should be refactored into something re-usable:
`if (!$key|is_float}{$key|number_format}{else}{$key|number}{/if}`

What about extracting this logic into a separate smarty modifier so it's just `{$key|readable_number}` or `{$key|number:strip_superfluous_fraction}`. Please also add an Unit test for that.

The if float then check ist re-occurring logic that should not be duplicated in my opinion.

htkassner:
Then we should do something like we did with currency and call it tax because this is espacially for taxes. So maybe implementing a helper and use it like $tax|tax.

bcremer:
Please write an unit test for `smarty_modifier_tax`

snippets/documents/index.ini
```diff
@@ -18,7 +18,7 @@ DocumentIndexOrderID = "Order No.:"
 DocumentIndexPageCounter = "Page {$page+1} of {$Pages|@count}"
 DocumentIndexSelectedDispatch = "Selected shipping type"
 DocumentIndexSelectedPayment = "Selected payment method"
-DocumentIndexTax = "Plus {$key}% VAT:"
+DocumentIndexTax = "Plus {$key|tax} % VAT:"
```
bcremer: We should add an non breaking space between the tax and the percent sign `&nbsp;`

PR #731, opened 2016-09-08
refactored notification plugin

uehler:
| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-14051 |
| How to test? | Turn on notification for an article which is out of stock and is set as last stock |

I optimized the code for readability, removed some redundant code and moved method calls, so they won't get called if they are not needed.

The sARTICLEAVAILABLE email now contains the right translation and the $sArticleLink is the seo url of the article


shyim:
Hey @uehler,
is this Pull Request still necessary? I checked the link is generated and the mail is also translated.

PR #735, opened 2016-09-09
[Papercut][SW-15978] Add editable pseudoprice column for variants.

kleinmann:
Implementation is based on the price column in the same view. Changing a pseudoprice results in deletion of all existing scale prices for that variant so only one price remains (same semantics as in the price column).

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-15978 |
| How to test? | Create a variant article with at least two variants, change pseudoprice in variant list. |


themes/Backend/ExtJs/backend/article/view/variant/list.js
```diff
@@ -425,6 +461,24 @@ Ext.define('Shopware.apps.Article.view.variant.List', {
...
+     * @param metaData
+     * @param record
+     */
+    pseudoPriceColumnRenderer: function(value, metaData, record) {
+        var me = this,
+            prices = record.getPrice();
+
+        if (prices.getCount() === 0 || prices.first().get('pseudoPrice') === 0) {
+            return me.snippets.columns.pseudoPrice.undefined;
+        } else {
```
janbuecker: you can leave out the `else` since you return within the if condition

themes/Backend/ExtJs/backend/article/view/variant/list.js
```diff
@@ -146,6 +150,21 @@ Ext.define('Shopware.apps.Article.view.variant.List', {
                     me.fireEvent('editVariantPrice', e.record, newPrice);
                 }
 
+            } else if (e.value && e.field === 'pseudoPrice') {
+                var newPseudoPrice = Ext.Number.from(e.value);
+                newPseudoPrice = Ext.Number.toFixed(newPseudoPrice, 2);
+
+                var oldPseudoPrice = Ext.Number.from(e.originalValue);
+                if(undefined === e.originalValue) {
```
janbuecker: - please add a whitespace between the `if` and the `parentheses`.
- please use `Ext.isDefined(e.originalValue)` since `undefined` can be overwritten

themes/Backend/ExtJs/backend/article/controller/variant.js
```diff
@@ -1516,6 +1517,39 @@ Ext.define('Shopware.apps.Article.controller.Variant', {
...
+     */
+    onEditVariantPseudoPrice: function(record, pseudoPrice) {
+        var me = this;
+
+        if (!record) {
+            return false;
+        }
+
+        if (!record.getPrice()) {
+            record.getPriceStore = Ext.create('Ext.data.Store', { model: 'Shopware.apps.Article.model.Price' })
```
janbuecker: `;` is missing at the line ending

themes/Backend/ExtJs/backend/article/controller/variant.js
```diff
@@ -1516,6 +1517,39 @@ Ext.define('Shopware.apps.Article.controller.Variant', {
...
+        var newPrice = Ext.create('Shopware.apps.Article.model.Price', {
+            pseudoPrice: 0,
+            percent: 0,
+            customerGroupKey: me.subApplication.firstCustomerGroup.get('key')
+        });
+
+        if (record.getPrice().getCount() > 0) {
+            newPrice = record.getPrice().first();
+        }
+        newPrice.set('pseudoPrice', pseudoPrice);
```
janbuecker, line 1542: please get rid of the setter and move them into the object creation of `newPrice`

themes/Backend/ExtJs/backend/article/view/variant/list.js
```diff
@@ -146,6 +150,21 @@ Ext.define('Shopware.apps.Article.view.variant.List', {
...
+                    oldPseudoPrice = 0;
+                }
+
+                oldPseudoPrice = Ext.Number.toFixed(oldPseudoPrice, 2);
+
+                if (newPseudoPrice != oldPseudoPrice) {
```
janbuecker: please include type checking `!==`

themes/Backend/ExtJs/backend/article/view/variant/list.js
kleinmann: I am a proponent of explicit `else` cases in all conditions, even if the control flow would not require any. Missing `else` cases tend to stand out if you are used to seeing them. If this is a hard requirement from your side I will remove it, though :-)

themes/Backend/ExtJs/backend/article/controller/variant.js
kleinmann: The setters are needed here because `newPrice` may be an actual new instance from line 1534 as well as an existing instance from the record fetched from line 1541.
I can however remove the pseudoPrice from the constructor in line 1535 to keep only one instance of setting that value.

janbuecker: i see, my bad :) 👍

themes/Backend/ExtJs/backend/article/view/variant/list.js
janbuecker: I understand your point here but... please remove it :)

kleinmann:
I have incorporated all discussed comments, thanks for your feedback! :-)

janbuecker:
I've tested your changes but i can't get back to the state where it says "Undefined". The number `0` or an empty field results in no change at all. Can you fix it to match human behaviour? Like clearing the field would result in "Undefined".

kleinmann:
Good catch, I didn't check for that. Probably just assumed no one would ever reset their pseudoprices 😅 

I changed the PR to allow for empty values and values of `0`. Both will reset the variant pseudoprice to the `Undefined` value.

janbuecker:
Thanks you, good work! 👍

PR #737, opened 2016-09-09
[Papercut][SW-10346] - Add cronjob configuration field to keep cronjobs enabled after errors during execution

kleinmann:
This adds a new configuration option for cronjobs to tell whether the job should be disabled if an error occurs during its execution. The option is enabled by default to keep backwards compatibility with the existing behaviour.

The motivation behind this is that cronjobs communicating with external services may crash due to intermittent connection issues or errors in the external service. These jobs should stay enabled so the next run is able to "fix" these issues.

I have also added a new event based on the cronjob name to allow shop owners to handle errors of default or third-party cronjobs within their own plugins.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-10346 |
| How to test? | Create a new cronjob that throws an exception during execution. Uncheck the new checkbox, execute the job. The job should be disabled. Re-enable the job, set the checkbox and re-run the job. The job should now stay enabled even after the error. |


engine/Shopware/Components/Plugin/Bootstrap.php
```diff
@@ -444,14 +444,15 @@ public function createCronJob($name, $action, $interval = 86400, $active = 1)
...
-                'active'     => $active,
-                'end'        => new \DateTime(),
-                'pluginID'   => $this->getId(),
+                'name'             => $name,
+                'action'           => $action,
+                'next'             => new \DateTime(),
+                'start'            => null,
+                '`interval`'       => $interval,
+                'active'           => $active,
+                'disable_on_error' => 1,
```
shyim: New function parameter were cool, to configure "disable_on_error" while creating new cronjob :)

kleinmann: Good catch! Updated the PR accordingly.

_sql/migrations/786-add-crontab-disableonerror.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    public function up($modus)
+    {
+        $sql = <<<'EOD'
+    ALTER TABLE `s_crontab` ADD COLUMN `disable_on_error` TINYINT(1) NOT NULL DEFAULT 1 AFTER `active`;
+EOD;
+        $this->addSql($sql);
+
+        $sql = <<<'EOD'
+    INSERT IGNORE INTO `s_core_snippets` (`namespace`, `shopID`, `localeID`, `name`, `value`, `created`, `updated`)
```
janbuecker: please don't manage snippets yourself using migrations. you can use the snippet files in the `/snippets` folder.

janbuecker (changes requested):

_sql/migrations/786-add-crontab-disableonerror.php
kleinmann: Thanks for the feedback, must have overlooked that folder. Some differences between the Shopware distribution and the Shopware repository can be missed :-)
I added a new commit accordingly.

janbuecker:
Thanks for your work! 👍

PR #706, opened 2016-09-05
[Papercut][SW-15999] Don´t show inactive articles in crosselling

shyim:
| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | no |
| Related tickets? | [SW-15999](https://issues.shopware.com/#/issues/SW-15999) |
| How to test? | Show in Crosseling for inactive articles |

PR #739, opened 2016-09-12
Revert "[Papercut][SW-15999] Don´t show inactive articles in crosselling"

bcremer:
Reverts shopware/shopware#706


bcremer:
@shyim We has to revert #706. Because of usability and consistency issues as there might be already inactive articles assigned. 
A better solution would be to mark inactive articles via a background color.

PR #741, opened 2016-09-12
[PAPERCUT] SW-16063 - Move media files to manufacturer media album when uploaded…

OliverSkroblin:
## Description

- Why is it necessary?
  - Customers would expect manufacturer media files in the corresponding album
- What does it improve? 
  - Media files are now stored in the correct album
- Does it have side effects? 
  - Existing API behavior changed

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-16063 |
| How to test? | Upload an media over the manufacturer rest api resource (example see implemented tests) |

### DIFF
--- a/tests/Functional/Components/Api/ManufacturerTest.php
+++ b/tests/Functional/Components/Api/ManufacturerTest.php
@@ -22,9 +22,14 @@
  * our trademarks remain entirely with us.
  */
 
-namespace Shopware\Tests\Functional\Components\Api;
+namespace Shopware\tests\Functional\Components\Api;
 
+use Doctrine\ORM\AbstractQuery;
 use Shopware\Components\Api\Resource\Manufacturer;
+use Shopware\Components\Api\Resource\Resource;
+use Shopware\Components\Model\ModelRepository;
+use Shopware\Models\Media\Album;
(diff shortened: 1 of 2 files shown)


tests/Functional/Components/Api/ManufacturerTest.php
```diff
@@ -22,9 +22,14 @@
  * our trademarks remain entirely with us.
  */
 
-namespace Shopware\Tests\Functional\Components\Api;
+namespace Shopware\tests\Functional\Components\Api;
 
+use Doctrine\ORM\AbstractQuery;
```
janbuecker: This import is unused.

tests/Functional/Components/Api/ManufacturerTest.php
```diff
@@ -22,9 +22,14 @@
  * our trademarks remain entirely with us.
  */
 
-namespace Shopware\Tests\Functional\Components\Api;
+namespace Shopware\tests\Functional\Components\Api;
```
janbuecker: Please do not change the namespace unless we're changing all test namespaces. I know it should be lowercase :)

PR #742, opened 2016-09-12
[PAPERCUT] SW-14306 - Increase label width of base components to 155px

OliverSkroblin:
## Description

- Why is it necessary?
  - Unify label width of backend components with backend modules
- What does it improve?
  - Unified label width for all modules with this components 
- Does it have side effects?
  - Existing fields without own defined label width are increased by 25px 

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-14306 |
| How to test? | Download one of the backend tutorial plugins and check label width of detail view |

### DIFF
--- a/themes/Backend/ExtJs/backend/base/application/Shopware.model.Container.js
+++ b/themes/Backend/ExtJs/backend/base/application/Shopware.model.Container.js
@@ -887,7 +887,7 @@ Ext.define('Shopware.model.Container', {
         formField.xtype = 'displayfield';
         formField.anchor = '100%';
         formField.margin = '0 3 7 0';
-        formField.labelWidth = 130;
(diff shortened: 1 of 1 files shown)


dneustadt:
So this is a default value unless set otherwise? I could potentially see a few third party plugins being negatively affected by this. Depending on how complex the layout of the view is it might push form elements beyond reach (checkboxes especially), make them overlap or add scrollbars. Anyway, it should be noted as best practice to set labelWidth when depending on it.

janbuecker:
I agree with @saberdude and after talking with @MarcelSchmaeing, we'll remove this ticket from our papercuts.

PR #750, opened 2016-09-13
[Papercut][SW-14270] fix risk management rule ORDERPOSITIONSMORE

mitelg:
## Description
- Why is it necessary?
  - risk management ORDERPOSITIONSMORE does not behave as expected
- What does it improve?
  - correct behaviour of ORDERPOSITIONSMORE
- Does it have side effects?
  - maybe some settings will break, which rely on this false behaviour

| Questions | Answers |
| --- | --- |
| BC breaks? | maybe |
| Tests pass? | no -> test had to be fixed |
| Related tickets? | [Issue](https://issues.shopware.com/#/issues/SW-14270) |
| How to test? | create new risk management for "Nachnahme" with "Bestellpositionen >=" 2. put one product to cart. "Nachnahme" should not be available. put another product to cart -> "Nachnahme" is available |

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -708,14 +708,14 @@ class sAdmin
 
         $addScopeSql = "";
         if ($this->scopedRegistration == true) {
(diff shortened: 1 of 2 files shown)


mitelg:
my fault. the behaviour is not wrong

PR #751, opened 2016-09-13
[Papercut][SW-14526] add numeric prices to product export

mitelg:
## Description
- Why is it necessary?
  - -
- What does it improve?
  - add numeric prices without rounding to product export
- Does it have side effects?
  - no

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | [Issue](https://issues.shopware.com/#/issues/SW-14526) |
| How to test? | edit some product export template and add {$sArticle.netprice_numeric} and / or {$sArticle.price_numeric} . start product export. the new numbers should be available |


janbuecker:
Thanks @mitelg 👍

PR #758, opened 2016-09-15
[Papercut][SW-13998] - Fix image position attribute

taltholtmann:
## Description

- Why is it necessary?  
  The image position attribute can now be used by plugins / extensions. Before all images got "position = 1" and not the value from the database  
- What does it improve?  
  it now sets the right position from the database  
- Does it have side effects?  
  no

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-13998 |
| How to test? | outprint $sArticel.images in template /bare/frontend/detail/images.tpl and see the right position attribute |

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/Hydrator/MediaHydrator.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/Hydrator/MediaHydrator.php
@@ -94,6 +94,10 @@ class MediaHydrator extends Hydrator
             $media->setDescription($data['__media_description']);
         }
 
+        if (isset($data['__image_position'])) {
+            $media->setPosition($data['__image_position']);
+        }
+
         if (isset($data['__media_type'])) {
             $media->setType($data['__media_type']);
         }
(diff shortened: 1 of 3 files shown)


engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/Hydrator/MediaHydrator.php
```diff
@@ -93,6 +93,9 @@ public function hydrate(array $data)
         if (isset($data['__media_description'])) {
             $media->setDescription($data['__media_description']);
         }
+        if (isset($data['__image_position'])) {
```
janbuecker, line 97: please keep the coding style and add a newline between :)

janbuecker (changes requested):

janbuecker:
Please do not add the position to the media struct since the media struct itself does not care about a position. The position is article related. You may have to refactor the image selection on the detail page to fix this issue.

mitelg:
@janbuecker the position of the images on the detail page is correct. but the position value in the struct is always "1"
if we don't need this value, we should delete it or fill it correctly, so other can use this

janbuecker:
yeah, I told that @taltholtmann face2face :)

taltholtmann:
removed the attribute
[https://github.com/shopware/shopware/pull/761](url)

PR #759, opened 2016-09-15
[Papercut][SW-12795] Fix search with uri related chars

mitelg:
## Description
- Why is it necessary?
  - if searching with URI related characters the filter panel will break. you also can't go to page 2
- What does it improve?
  - improve encoding and decoding of the search string, so the filter panel and pagination won't break anymore. also I cleaned up the search controllers. This fixes also an issue if you search with two terms and you go to page 2 or use the filter. example: initial search "all art" -> go to page 2, search term is "all+art"
- Does it have side effects?
  - nope

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | [Issue](https://issues.shopware.com/#/issues/SW-12795) |
| How to test? | Search for "20% alle" or "20+ alle" switch pages or use the filters. result must be correct now |

### DIFF
--- a/engine/Shopware/Controllers/Frontend/AjaxSearch.php
+++ b/engine/Shopware/Controllers/Frontend/AjaxSearch.php
@@ -23,7 +23,8 @@
  */
 
 use Shopware\Bundle\SearchBundle\ProductSearchResult;
-use Shopware\Bundle\StoreFrontBundle\Struct\ProductContextInterface;
+use Shopware\Bundle\SearchBundle\SearchTermPreProcessorInterface;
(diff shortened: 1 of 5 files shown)


mitelg:
yes, I will write a test :laughing:

janbuecker:
If you work on the 99 papercuts project, please read our contribution guide on how to participate :p

mitelg:
@janbuecker thank you so much for your hint. fixed the title

janbuecker:
I'm glad I could help you out.

PR #762, opened 2016-09-15
new filter event for registration form validation

sbrueggenolte:
## Description

- Why is it necessary? Because there is currently no other possibility to validate custom registration form fields.
- What does it improve? The validation of the registration form.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | no ticket yet |
| How to test? | register an event listener and modify the $errors array |


sbrueggenolte:
We have found a better solution for this problem, so I will close this pull request.

PR #763, opened 2016-09-15
[Papercut][SW-12745] Fix voucher wrong enddate

TobiasWilming:
This PR now compares the voucher validdate with the Date instad of the datetime from today so that the last day is now included to the days that you can use the voucher.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/#/issues/SW-12745 https://issues.shopware.com/#/issues/SW-15604 |
| How to test? | Set the valid_from date to something and the valit_to time of a voucher to the current date and check if you can use the voucher in the frontend. You should also make a check if its available on the next date at 0:00 . |

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -458,12 +458,12 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
 
         // Check if voucher is available
         $sqlVoucher = "SELECT s_emarketing_voucher_codes.id AS id, code, description, value, percental FROM s_emarketing_vouchers, s_emarketing_voucher_codes
-         WHERE  modus = 1 AND (valid_to >= now() OR valid_to IS NULL)
+         WHERE  modus = 1 AND (valid_to >= CURDATE() OR valid_to IS NULL)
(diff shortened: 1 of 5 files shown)


htkassner:
I think this is even more confusing the user, because in the backend you configure a till date. As voucher user I would suggest my voucher is still usable on the day it ends and not the day before.

TobiasWilming:
I think I described it wrong. With this PR if you configure a till date it is now possible to use that voucher until the day ends. Without the PR it isnt possible to use it at the last (configured) day

janbuecker (changes requested):
There are more places where the query needs to be updated. f.e.
- engine/Shopware/Plugins/Default/Core/CronBirthday/Bootstrap.php
- engine/Shopware/Controllers/Backend/Newsletter.php
- engine/Shopware/Controllers/Backend/CanceledOrder.php
- engine/Shopware/Components/Document.php

Please search the whole codebase to find all places where it needs to be fixed.

TobiasWilming:
Updated it.

PR #769, opened 2016-09-15
[Papercut][SW-14342] Use helper methode to get the IP of customer

mitelg:
## Description

- Why is it necessary?
  - get correct IP if the shop uses proxies like varnish
- What does it improve?
  - gathering of IPs
- Does it have side effects?
  - no

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Related tickets? | [Issue](https://issues.shopware.com/#/issues/SW-14342) |
| How to test? | have a look in s_order if `remote_addr` is still set correctly. if you want, set up varnish, etc and test with proxy |

### DIFF
--- a/engine/Shopware/Core/sOrder.php
+++ b/engine/Shopware/Core/sOrder.php
@@ -604,7 +604,7 @@ class sOrder
             'currency'             => $this->sSYSTEM->sCurrency["currency"],
             'currencyFactor'       => $this->sSYSTEM->sCurrency["factor"],
             'subshopID'            => $mainShop->getId(),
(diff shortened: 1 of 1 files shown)


bcremer:
This cannot be merged.

Please check if the problem is resolved when the trusted proxies are configured correctly: 

Using `Shopware()->Front()->Request()->getClientIp(true)` (with `true`) opens a big security issue because it blindly trusts http headers that can be spoofed by the client. See: https://github.com/shopware/shopware/blob/v5.2.6/engine/Library/Zend/Controller/Request/Http.php#L1060

PR #771, opened 2016-09-15
Add interfaces to additional core services

hrombach:
Our plugins use a base plugin to do certain things that many of our plugins have to do to try and reduce redundant code in our plugins. Now we are refactoring our plugins to fit the new 5.2 plugin structure and tried to decorate certain core services like the plugin installer to do these things more easily. We discovered that when using PHP 7 this causes issues as any service that depends on these decorated services would TypeHint the default service. 

We approached developer support about this and were advised to open a pull request to introduce interfaces for these core services. 

The only significant side effect is that plugins that depend on these core services should TypeHint the interface instead of the service class.

| Questions | Answers |
| --- | --- |
| BC breaks? | yes - sort of |
| Tests pass? | yes |
| Related tickets? | n/a |
| How to test? | try to decorate one of the core services, implement the interface |

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -10,6 +10,7 @@ This changelog references changes done in Shopware 5.2 patch versions.
 * getOne function of customer api resource contains now the country and state data for billing and shipping address
 * Changed `jquery.search::onKeyboardNavigation()` method to provide more extension possibilities.
 * Deprecated the execution shopware.php via CLI. Please use the command line tool in `bin/console` instead.
+* Declared interfaces for`AttributeBundle` and `PluginInstallerBundle` services to allow for service decoration by plugins
 
 ## 5.2.6
 
--- a/engine/Shopware/Bundle/AccountBundle/Form/Account/AttributeFormType.php
+++ b/engine/Shopware/Bundle/AccountBundle/Form/Account/AttributeFormType.php
@@ -24,7 +24,7 @@
 
 namespace Shopware\Bundle\AccountBundle\Form\Account;
 
-use Shopware\Bundle\AttributeBundle\Service\CrudService;
+use Shopware\Bundle\AttributeBundle\Service\CrudServiceInterface;
 use Shopware\Components\Model\ModelManager;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormBuilderInterface;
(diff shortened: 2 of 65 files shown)


hlohaus:
Weitere Informationen zu den Änderungen vom Kunden:

>   vom Kunden / 22.09.2016 12:37 Uhr
>   Hallo Herr Lohaus,
> 
>   wir brauchen primär die Interfaces des PluginInstaller-, LegacyPluginInstaller- und AttributeCrud-Service.
> 
>   Die ersten beiden deswegen, weil wir mittlerweile über 40 Stores im Shopware haben und darin wiederkehrende Aktionen vorkommen, die wir in unser Basis-Plugin "Foundation" ausgelagert haben.
> Damit Foundation bei Installation/Update/Aktivierung/Deaktivierung/Deinstallation unserer Plugins eingreifen kann, nehmen wir entsprechende Service-Dekorationen vor.
> 
>   Den AttributeCrud-Service dekorieren wir, damit es uns möglich ist, Attribute mit mehrsprachigen Labels im Backend anzeigen zu lassen.
>   Hier hatte ich in einem Support-Ticket schon einmal nachgefragt, wie wir mehrsprachige Attribut-Labels mit Shopware-Bordmitteln realisieren können.
>   Da die Antwort negativ war, haben wir über eine Service-Dekoration einen Weg gefunden.

FiveDigital:
Imho the part with translatable attribute labels has to become part of Shopware itself.

OliverSkroblin:
@FiveDigital +1 for me, better a translatable label than 50x interfaces more which we have to support (not to break) =)

bcremer:
@hrombach After internal discussion we decided to close this PR.

The introduced interfaces are affecting classes that we consider internal and should not be extended because they are viable for the pluginsystem to work properly.

However, if you must decorate one of these services you still can, even when using PHP 7.
For that you have to extend from the class that is getting decorated to keep the object identity. 

```php
<?php

namespace SomePlugin;
use Shopware\Bundle\PluginInstallerBundle\Service\InstallerService;

class MyInstallerService extends InstallerService
{
    private $inner;

    public function __construct(InstallerService $inner) 
    {
        $this->inner = $inner;
    }

    public function getPluginPath($pluginName)
    {
        return $this->inner->getPluginPath($pluginName);
    }

    [...]
}
```

For the translatable attribute labels this should be implemented with a separate ticket in the core as @FiveDigital  already mentioned.

We appreciate your work on this PR but for the mentioned reasons we will not merge it.

sbrueggenolte:
@bcremer we (Net Inventors GmbH) respect the decision, but we cannot fully understand your point.
If you did not want these services to be decorated, why didn't you implement any precautions to prevent this? We have spent days to develop decorations for our base plugin "NetiFoundation", which handles special installation and updating routines for all our plugins. 
The mentioned solution of extending the core service's class does not serve the purpose of dependency injection. But if there is no other solution, we will have to do it this way.

PR #778, opened 2016-09-16
[PAPERCUT] Sw 16379/5.2/save record after row edit

OliverSkroblin:
## Description

- Why is it necessary?
- What does it improve?
- Does it have side effects?

| Questions | Answers |
| --- | --- |
| BC breaks? | yes/no |
| Tests pass? | yes/no |
| How to test? | Please describe how to best verify that this PR is correct. |

### DIFF
--- a/themes/Backend/ExtJs/backend/base/application/Shopware.grid.Panel.js
+++ b/themes/Backend/ExtJs/backend/base/application/Shopware.grid.Panel.js
@@ -813,7 +813,7 @@ Ext.define('Shopware.grid.Panel', {
             /**
              * Event fired after the grid columns created. This event can be used
              * to add additional column over the Ext JS event system.
-             * 
+             *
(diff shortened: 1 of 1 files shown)


janbuecker:
Just for the record, this is a replacement for #772.

themes/Backend/ExtJs/backend/base/application/Shopware.grid.Panel.js
```diff
@@ -1400,7 +1400,7 @@ Ext.define('Shopware.grid.Panel', {
                 clicksToEdit: 2,
                 listeners: {
                     edit: function (editor, e) {
-                        e.record.save();
+                        me.saveRecord(e.record);
```
htkassner, line 1403: Maybe it would be good to validate model before saving it. That would make it possible to add / use model validations. If no validations are used valid is true.

PR #779, opened 2016-09-16
[Papercut][SW-16376] remove unused image handling in last seen articl…

mitelg:
## Description

- What does it improve?
  - the last seen products widgets shows now the correct preview image. the same as in the listing. also the plugin was cleaned up and unnecessary code was removed. the image info was not saved correctly, so I removed it. Also I changed the scope of two config options, so you can select different amounts for subshops
- Does it have side effects?
  - no

| Questions | Answers |
| --- | --- |
| BC breaks? | maybe if someone relies on the `img` column of `s_emarketing_lastarticles` but it was never filled in 5.2.6 |
| Related tickets? | [Issue](https://issues.shopware.com/#/issues/SW-16376) |
| How to test? | browse through the shop to different products and have a look if the image of the products is correct in the last seen widget |

### DIFF
--- a/_sql/migrations/791-change-last-articles-plugin.php
+++ b/_sql/migrations/791-change-last-articles-plugin.php
@@ -1,0 +1,34 @@
+<?php
+
+use Shopware\Components\Migrations\AbstractMigration;
+
+class Migrations_Migration791 extends AbstractMigration
+{
+    /**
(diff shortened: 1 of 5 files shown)


janbuecker:
This issue has been closed in favour of SW-14197.

PR #780, opened 2016-09-16
[Papercut][SW-16211] resolve double negatives by offering config type boolean-reverse

dneustadt:
## Description

- Why is it necessary? Labels of config fields use double negatives to describe the effect. This is confusing. The consequence of changing the wording without negatives would be the necessity to  refactor every condition where the boolean value is queried.
- What does it improve? By using the config element typ boolean-reverse the wording can be changed so no negatives are used without the need to refactor every condition elsewhere. It just changes the label for false values to "Yes" and respectively the value of true values to "No".
- Does it have side effects? None so far.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/issues/SW-16211#/issues/SW-16211 |
| How to test? | Set Config Form Element  with type boolean-reverse "Yes" should be false "No" should be true |

### DIFF
--- a/themes/Backend/ExtJs/backend/base/component/element/boolean_select.js
+++ b/themes/Backend/ExtJs/backend/base/component/element/boolean_select.js
@@ -39,11 +39,23 @@ Ext.define('Shopware.apps.Base.view.element.BooleanSelect', {
     queryMode: 'local',
     forceSelection: true,
     editable: false,
+    reverse: false,
 
     store: [
         ["", 'Inherited'],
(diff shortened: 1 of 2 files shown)


janbuecker:
Thanks for your PR!

Unfortunately, we won't implement this new `xtype` in near future, because it is not absolutely necessary at the moment.

This ticket will be solved by changing the misleading labels.

PR #781, opened 2016-09-19
[Papercut][SW-15814] Fix sorting of filters in categories if using el…

mitelg:
in addition to https://github.com/shopware/shopware/pull/749


janbuecker:
@mitelg is this still relevant?

PR #784, opened 2016-09-19
[Papercut][SW-16407] collect meta data per variant order number

mitelg:
## Description
- Why is it necessary?
  - collecting of the meta data on creating a document is wrong
- What does it improve?
  - collects the data of a variant, not of the main variant
- Does it have side effects?
  - maybe templates which rely on this information could break

| Questions | Answers |
| --- | --- |
| BC breaks? | maybe, see above |
| Tests pass? | yes |
| Related tickets? | [Issue](https://issues.shopware.com/#/issues/SW-16407) |
| How to test? | extend the document template to show the whole data |


janbuecker:
Merged with cb1a228b14ab5e185158852e8d29c1fb9a371854.

PR #788, opened 2016-09-20
Revert "[Papercut][SW-15997] Global Bcc Mail Option"

janbuecker:
Reverts shopware/shopware#707 (https://issues.shopware.com/#/issues/SW-15997)

After some in-depth testing, it seems that BCC cannot be used with SMTP transport since Zend Framework removes this in their implementation. Therefore we'll reopen this issue and put it in "Not scheduled" because it's simply not possible currently. Maybe in the future, if we change our mailing library.

/cc @bcremer


bcremer:
Line in the ZF implementation: https://github.com/shopware/shopware/blob/v5.2.6/engine/Library/Zend/Mail/Transport/Smtp.php#L238

shyim:
We can clone the Mail and use addTo. That were possible

janbuecker:
Yes but that is not the purpose of a BCC. For the time being, you'll have to use a plugin. Sorry.

PR #795, opened 2016-09-21
enable session usage in the backend / prevent fatal error

sbrueggenolte:
## Description

- Why is it necessary? When a symfony service/subscriber injects the session service, it will lead to a fatal error in the Backend. This pull request prevents this error.
- What does it improve? The session handling

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| How to test? | Inject the session service into a service or subscriber and load the Backend. |

this prevents a fatal error in the backend, when a symfony service/subscriber injects the session service

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/Bridge/Session.php
+++ b/engine/Shopware/Components/DependencyInjection/Bridge/Session.php
@@ -25,6 +25,7 @@
 namespace Shopware\Components\DependencyInjection\Bridge;
 
 use Shopware\Components\DependencyInjection\Container;
+use Shopware\Models\Shop\Shop;
 
 /**
  * Session Dependency Injection Bridge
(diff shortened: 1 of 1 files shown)


hlohaus:
i don't like it. It's start a secound session wich are not really needed.

dnoegel:
@hlohaus What's the alternative? Why isn't it "really needed"? Do you just want to leave it like that?

hlohaus:
the problem are is: it's starts the frontend session and close the backend session.
so now is the backend session broken. 

Shopware()->BackendSession();
Shopware()->Session();
Shopware()->BackendSession()->offsetSet('test' ,true); // Throws a exception

The problem was only moved.

sbrueggenolte:
@hlohaus What would you suggest instead?
The main issue why I created this Pull Request, is that it comes to a fatal error when you use the "session" service as an injected dependency in the backend.
If you have a different solution to avoid this crash, I'd be totally happy with it.

hlohaus:
I would just add an exception (no shop selected) and leave it otherwise.
You can still inject the container and then read out the session in the service.

Or we can do it like symfony. They have a save() method:

http://stackoverflow.com/questions/7192916/how-do-you-access-a-users-session-from-a-service-in-symfony2

hrombach:
@hlohaus injecting the container into a service is really bad practice, though. If Shopware is gonna go DI, I think it should be done in a clean way. The session is needed a lot in my experience, so injecting the container all the time would be a dirty solution.

hlohaus:
I did not say that is perfect.
But to create a frontend session in the backend,
is much much dirtier.

Without a rebuild to another session handler,
The backend session is killed and creates unnecessary entries in the database.
The database can run out of space and even crash.

Please consider this here.

janbuecker:
Thanks @sbrueggenolte 

PR #804, opened 2016-09-29
[SW-16191] Fix Productstream Price Filter when maxvalue is null

TobiasWilming:
The PR don't switch the critera for the Product stream if the maxvalue is 0 (null in the database), so the maxvalue from all articles in the Productstream is taken for the Pricefilter.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | https://issues.shopware.com/issues/SW-16191 |
| How to test? | Create a product Stream for a category and then check if the maxvalue for the Price filter in the Frontend is not null. |

### DIFF
--- a/engine/Shopware/Components/ProductStream/FacetFilter.php
+++ b/engine/Shopware/Components/ProductStream/FacetFilter.php
@@ -179,7 +179,9 @@ class FacetFilter implements FacetFilterInterface
             $condition = $criteria->getBaseCondition('price');
 
             $facet->setMin($condition->getMinPrice());
-            $facet->setMax($condition->getMaxPrice());
+            if ($condition->getMaxPrice() != 0) {
+                $facet->setMax($condition->getMaxPrice());
+            }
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/ProductStream/FacetFilter.php
```diff
@@ -179,7 +179,9 @@ private function switchPriceFilterValues(array $facets, Criteria $criteria)
             $condition = $criteria->getBaseCondition('price');
 
             $facet->setMin($condition->getMinPrice());
```
OliverSkroblin, line 181: same behavior for minPrice.
Min and max price are optional.

TobiasWilming: if i set minPrice to null it is converted into 0 and the Pricerange sould be fine, if I havn't got any articles with a negative price.

PR #806, opened 2016-09-29
[FEATURE] Autoload resource files

PortaltechGithub:
With this change, CSS, JS and LESS files of a plugin are automatically loaded if they are stored in one of the resource directories (Resources/css, Resources/js or Resources/less) in plugin directory.

Example:
- custom/plugins/ReplyTest/Resources/css/test.css
- custom/plugins/ReplyTest/Resources/js/test.js
- custom/plugins/ReplyTest/Resources/less/test.less

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | no |
| How to test? | Create a plugin in custom/plugins and add some resources |

### DIFF
--- a/engine/Shopware/Components/Theme/Compiler.php
+++ b/engine/Shopware/Components/Theme/Compiler.php
@@ -24,9 +24,13 @@
 
 namespace Shopware\Components\Theme;
 
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo as File;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\ORM\AbstractQuery;
 use Shopware\Components\Theme\Compressor\Js;
+use Shopware\Components\Plugin;
+use Shopware\Components\DependencyInjection\Container;
 use Shopware\Models\Shop as Shop;
 
 /**
@@ -81,6 +85,11 @@ class Compiler
     private $timestampPersistor;
 
     /**
+     * @var Container
+     */
+    private $container;
+
+    /**
      * @param $rootDir
      * @param LessCompiler $compiler
      * @param PathResolver $pathResolver
@@ -89,6 +98,7 @@ class Compiler
      * @param Js $jsCompressor
      * @param \Enlight_Event_EventManager $eventManager
      * @param TimestampPersistor $timestampPersistor
+     * @param Container $container
      */
     public function __construct(
         $rootDir,
@@ -98,7 +108,8 @@ class Compiler
         Service $service,
         Js $jsCompressor,
         \Enlight_Event_EventManager $eventManager,
-        TimestampPersistor $timestampPersistor
+        TimestampPersistor $timestampPersistor,
+        Container $container
     ) {
         $this->rootDir = $rootDir;
         $this->compiler = $compiler;
@@ -108,6 +119,7 @@ class Compiler
         $this->pathResolver = $pathResolver;
         $this->jsCompressor = $jsCompressor;
         $this->timestampPersistor = $timestampPersistor;
+        $this->container = $container;
     }
 
     /**
@@ -255,6 +267,37 @@ class Compiler
 
         $file->fwrite($content);
         $file->flock(LOCK_UN);   // release the lock
+    }
+
+    /**
+     * @param string $directory
+     * @param string $type
+     * @return ArrayCollection
+     */
+    private function collectResourceFiles($directory, $type)
+    {
+        $files = new ArrayCollection();
+
+        if (empty($directory)|| empty($type)) {
+            return $files;
+        }
+
+        $type = strtolower(trim($type));
+        $directory = implode(DIRECTORY_SEPARATOR, [rtrim($directory, '\\/'), 'Resources', 'frontend', $type]);
+
+        if (!is_dir($directory)) {
+            return $files;
+        }
+
+        $finder = new Finder();
+        $finder->files()->name('*.' . $type)->in($directory);
+
+        /** @var File $file */
+        foreach ($finder as $file) {
+            $files->add($file->getRealPath());
+        }
+
+        return $files;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


engine/Shopware/Components/Theme/Compiler.php
```diff
@@ -517,6 +589,21 @@ private function collectPluginLess(Shop\Template $template, Shop\Shop $shop)
     private function collectPluginCss(Shop\Template $template, Shop\Shop $shop)
     {
         $collection = new ArrayCollection();
+
+        /** @var Kernel $kernel */
+        $kernel = Shopware()->Container()->get('kernel');
+
+        foreach ($kernel->getPlugins() as $plugin) {
+            if ($plugin instanceof Plugin) {
```
FiveDigital: You should add `&& $plugin->isActive()` here.

engine/Shopware/Components/Theme/Compiler.php
```diff
@@ -422,6 +462,21 @@ private function collectInheritanceJavascript($inheritance)
     private function collectPluginJavascript(Shop\Shop $shop, Shop\Template $template)
     {
         $collection = new ArrayCollection();
+
+        /** @var Kernel $kernel */
+        $kernel = Shopware()->Container()->get('kernel');
```
hlohaus: Please inject it in the constructor and don't use the Shopware() function.

PortaltechGithub: This is not possible because the theme.xml is interpreted before placing the kernel as a service. Therefore a constructor injection of this service would result in a ServiceNotFoundException.

```
// \Shopware\Kernel

protected function initializeContainer()
{
    // ...
    $container = $this->buildContainer();
    $container->compile();
    // ...
    $this->container->set('kernel', $this);
    // ...
}

protected function buildContainer()
{
    // ...
    $container = $this->getContainerBuilder();
    $container->addObjectResource($this);
    $this->prepareContainer($container);
    // ...
}

protected function prepareContainer(ContainerBuilder $container)
{
    $loader = new XmlFileLoader($container, ...);
    $loader->load('services.xml');
    $loader->load('theme.xml');
    $loader->load('logger.xml');
    // ...
}
```

engine/Shopware/Components/Theme/Compiler.php
PortaltechGithub: Done

engine/Shopware/Components/Theme/Compiler.php
PortaltechGithub: I've now solved the problem with an injection of the DI container instead of the kernel.

dnoegel:
Hi,

thanks for the PR :+1: . As discussed recently it makes perfect sense for me - but I don't want to bloat the Plugin.php with three methods to overwrite (@bcremer?). If we have the convention that files in Resources{js/css/less} are included automatically, people should be able to live with it. If they need conditional includes, they still can move those files to another folder as for example to the views folder where most people have those files anyway.

Another topic is the conventional path itself: I can live with the current suggestion - but just want to point out, that in themes we have this frontend/_public/{js,css,less} convention. On the other hand, we must make sure, that we don't introduce any kind of collisions with themes or existing templates - so probably frontend/_public/{js,css,less} is exactly the one option we should not choose :)

FiveDigital:
👍  for removing the methods. I think we should use the path `Resources/frontend/less/test.less` so that Backend JS can be placed in `Resources/backend/whatever/ext/.js`.

We should also verify that the console command `sw:theme:dump:configuration` includes the JS and LESS files of the new path.

PortaltechGithub:
All right, the resources are no longer loaded via methods in the plugin, but as part of the collection process in the theme compiler. In addition, they are now fetched from the `frontend` directory in `Resources`. And the `sw:theme:dump:configuration` command also provides the automatically loaded resources from the plugins. The change doesn't load any backend resources automatically, this must be implemented in an additional PR.

engine/Shopware/Components/Theme/Compiler.php
```diff
@@ -108,6 +119,15 @@ public function __construct(
         $this->pathResolver = $pathResolver;
         $this->jsCompressor = $jsCompressor;
         $this->timestampPersistor = $timestampPersistor;
+        $this->container = $container;
+    }
+
+    /**
+     * @param Plugin[] $plugins
+     */
+    public function setPlugins(array $plugins)
```
FiveDigital: I think this is not needed anymore and can be removed.

PortaltechGithub: Done.

PortaltechGithub:
@dnoegel Do you already know when you want to merge the PR?

engine/Shopware/Components/Theme/Compiler.php
```diff
@@ -258,6 +270,42 @@ public function compileJavascript($timestamp, Shop\Template $template, Shop\Shop
     }
 
     /**
+     * @param string $directory
+     * @param string $type
+     * @param boolean $returnRelative
+     * @return ArrayCollection
+     */
+    private function collectResourceFiles($directory, $type, $returnRelative = false)
```
bcremer: The `$returnRelative` argument does not to be in use.

bcremer:
@PortaltechGithub unfortunately I cannot answer this question. We currently have a evergrowing backlogs of tickets in the issue tracker and PRs here on github. I'll do my best to get this PRs into the next sprints.

engine/Shopware/Components/Theme/Compiler.php
PortaltechGithub: Fixed. ;-)

bcremer:
@PortaltechGithub @FiveDigital @kairichter  I am currently working on your Pull Request.
The introduced coupling between the ThemeCompiler and the Plugins bugged me a bit so I tried a different approach using a automatically registered Subscriber.
Please see latest commit on https://github.com/shopware/shopware/tree/sw-16671/5.2/autoload-resource-files-via-subscriber.

What do thing about that? For plugin developers the result should be the same.

PortaltechGithub:
Fine by me.

bcremer:
Merged with commit https://github.com/shopware/shopware/commit/5e04b54b028cd309d95d8fab4ac9c1fb7fcafa34

Thanks to all who were involved in this :+1:

PortaltechGithub:
Cool thx @bcremer !

PR #809, opened 2016-10-05
Add hideTop check in Category Side Menue

Gugiman:
## Description

Categories which are marked as "Hide in Top Navigation" are still appearing in the catagory side Menu on mobile view. 

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | no |
| How to test? | Set the "Hide in Top Navigation" mark on a category and reload. It disappears in the Desktop Menue but is still visible on mobile. |


janbuecker:
Thanks @Gugiman 

The config option works as intended and we won't change this behaviour.

PR #811, opened 2016-10-12
Fixed table name case in addtribute administration migration

FiveDigital:
## Description

- Why is it necessary?
- What does it improve?
- Does it have side effects?

| Questions | Answers |
| --- | --- |
| BC breaks? | yes/no |
| Tests pass? | yes/no |
| How to test? | Please describe how to best verify that this PR is correct. |
## Description

When upgrading a 5.1 shop to 5.2 this migration creates entries in `s_attribute_configuration` based on the entries in the `s_core_engine_elements` table. The `name` column in `s_core_engine_elements` has a lowerCamelCase value but the `column_name` column in the `s_attributes_configuration` table needs a snake_case value. This PR fixes the attribute migration. 

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | No |
| How to test? | Upgrade a 5.1 shop to 5.2 and have a look into the `s_attribute_configuration` table where the values in the column `column_name` for existing attributes should be in snake_case. |

### DIFF
--- a/_sql/migrations/793-fix-attribute-administration.php
+++ b/_sql/migrations/793-fix-attribute-administration.php
@@ -1,0 +1,31 @@
+<?php
+
+use Shopware\Components\Migrations\AbstractMigration;
+use Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter;
+
+class Migrations_Migration793 extends AbstractMigration
+{
+    /** @var CamelCaseToSnakeCaseNameConverter */
+    protected $converter;
+
+    /**
+     * @param \PDO $connection
+     */
+    public function __construct(\PDO $connection)
+    {
+        parent::__construct($connection);
+        $this->converter = new CamelCaseToSnakeCaseNameConverter();
+    }
+
+    public function up($modus)
+    {
+        $statement = $this->connection->query("SELECT * FROM `s_core_engine_elements`");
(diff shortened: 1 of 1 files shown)


dnoegel:
Thanks for that - but usually we do not modify existing migrations, as old shopware installations which already have that migration applied will not profit from it. Later you will never be able to easily tell, if a given shopware installation X.Y.Z is one with that patch or without, as it depends on *when* that installation was set up.

So if possible, please create a new migration, that fixes that issue for all versions.

dnoegel:
Thanks for the PR. 

Currently you are renaming the migration but the migration filename is still the old one. Also we usually try to not touch existing migrations in the first place, once they are released. So I guess in this case, you should add a new migration, "fixing" what the old migration messed up. There is no other way to keep all shopware installations consistent - and it shouldn't be a problem as no data is lossed in this case.

FiveDigital:
The Renaming was a bug :( I will add a new migration to fix up. Since the `CamelCaseToSnakeCaseNameConverter` won't change values that are already in snake_case it's safe to run this migration.

FiveDigital:
I have added the new migration. This might lead to a duplicate entry if someone has added the attribute with the correct `column_name` value after the migration 708. Perhaps we should only convert those attributes that do not already have an entry with the correct `column_name` in `s_attribute_configuration` and delete the wrong and unused entries instead?

PR #814, opened 2016-10-13
remove scrollbar styling on filter-panels

screeny05:
## Description

This PR removes the hacked-together styling of Scrollbars in the filter Panel.
### Why is this necessary?
- Scrollbar stylings for IE are almost impossible to reset
- They only look "kinda" nice in the default theme
- It's in the _hacks.less because, yes, it is a hack

If you insist on doing this, please consider using a flag for this and spare us developers from working around this.

We want default scrollbars.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| How to test? | Test filter panels |

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/less/hacks.less
+++ b/themes/Frontend/Responsive/frontend/_public/src/less/hacks.less
@@ -54,32 +54,6 @@ More precise designations are commented inside the document.
     }
 }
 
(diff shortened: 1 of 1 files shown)


janbuecker:
Thanks @screeny05 

We'll not merge this PR into 5.2 but we're looking forward to merge this into 5.3.

janbuecker:
Merged into 5.3.

PR #818, opened 2016-10-17
Fix error when manufacturer is null

romanWienicke:
## Description

- Why is it necessary?
  At the moment it is possible to add articles without manufacturers with the REST API. Articles without manufacturers cannot be displayed on the "details page" because there is a call to a function of an object which is null.
- What does it improve?
  The function manufacturer object is only called when the object is not null.
- Does it have side effects?
  No known side effects.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | manually tested. |
| Related tickets? | none |
| How to test? | add an article (via REST) and view the article details page in shopware. |

### DIFF
--- a/engine/Shopware/Components/Compatibility/LegacyStructConverter.php
+++ b/engine/Shopware/Components/Compatibility/LegacyStructConverter.php
@@ -545,12 +545,25 @@ class LegacyStructConverter
             $data["sLinks"][] = $temp;
         }
 
-        $data["sLinks"][] = array(
+        /*$data["sLinks"][] = array(
             'supplierSearch' => true,
             'description' => $product->getManufacturer()->getName(),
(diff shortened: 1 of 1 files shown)


romanWienicke:
Fixed style ci...

engine/Shopware/Components/Compatibility/LegacyStructConverter.php
```diff
@@ -545,12 +545,25 @@ public function convertProductStruct(StoreFrontBundle\Struct\Product $product)
             $data["sLinks"][] = $temp;
         }
 
-        $data["sLinks"][] = array(
+        /*$data["sLinks"][] = array(
```
hlohaus: You should remove it. :)

PR #821, opened 2016-10-21
Join s_order_basket_attributes in sGetDispatchBasket

GerDner:
## Description

- Why is it necessary? If you want to calculate shipping costs based on bundle informations you need the s_order_basket_attributes table joined.
- What does it improve? More fields for the Field "Calculation" in the dispatch modul. (e.g For Bundle or AboCommerce)
- Does it have side effects?
  no

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | none |
| How to test? | Use fields in your own calculation. For example count bundles with "COUNT(DISTINCT  ba.bundle_id)" |

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -23,6 +23,7 @@ This changelog references changes done in Shopware 5.2 patch versions.
 * Replaced the default filter on `dispatches.active` added in `Shopware_Controllers_Backend_Base::getDispatchesAction()` by a default filter on `active` added in `Shopware.apps.Base.store.Dispatch`
 * Refactored jQuery product slider plugin for sliding infinitely
 * Added `initOnEvent` option to cross selling tabs on detail page for the combination of tabs with product sliders
+* Left join s_order_basket_attributes to use fields in own dispatch calculation
 
 ## 5.2.9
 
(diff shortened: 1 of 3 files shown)


hlohaus:
Hey,

you shoud extends the dispatch query in the product export too.
https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Core/sExport.php#L1330
Or the select on the attr table generated there an error.

Alternatively, you can also use a subselect . Then the JOIN is not need.

Heiner

GerDner:
Thx for the feedback. I changed sExports as well.

danielwolf1:
The change is only available in sExport as the build process failed for sAdmin - does anyone know something about this case?

mitelg:
hey @danielwolf1 
we reverted the changes in 1a9d4660695 and in e8151e27073

danielwolf1:
OK, because it crashed during Export_ We store additional information about a basket in s_order_basket_attributes which are important to calculate shipping costs. What about adding a notify-Filter to extend the query if needed?

mitelg:
@danielwolf1 yes, that could be done. If you like, just create an own PR oder an issue ticket on https://issues.shopware.com :wink:

PR #822, opened 2016-10-23
Set partner id to optional in order entity

synonymous1984:
## Description

- Why is it necessary? The order entity requires adding a marketing partner when creating an order which is not correct IMO cause an order isn´t related to a partner in the most cases.
- What does it improve? Sets the partnerId to be nullable.
- Does it have side effects? None known

| Questions | Answers |
| --- | --- |
| BC breaks? | yes/no |
| Tests pass? | yes/no |
| Related tickets? | no |
| How to test? | Create a new order via the entity and don´t specify a partner. |

### DIFF
--- a/engine/Shopware/Models/Order/Order.php
+++ b/engine/Shopware/Models/Order/Order.php
@@ -124,7 +124,7 @@ class Order extends ModelEntity
     /**
      * @var string $partnerId
      *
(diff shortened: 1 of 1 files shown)


samuelvogel:
This PR seems to be the smaller brother of #581.

janbuecker:
I agree with @samuelvogel. Currently, we do not intend to change this behaviour.

Closing this as a duplicate of #581.

PR #825, opened 2016-10-27
fix modal-sizing on ios safari

screeny05:
This PR fixes works around a [jquery-bug](https://bugs.jquery.com/ticket/6724), returning an incorrect window-height on safari@ios. This bug results in modals not being correctly displayed when the address-bar is visible and a modal is being opened.

See this screenshot:
iphone-ios-safari-modal-bug
_This shows the modal scrolled all the way up._

This is due to the fact, that jquery's `$(window).height()` is more of a guesswork than a precise method to get the window-height.

The modal-plugin uses this wrong size excluding the areas covered by the top-bar to center the modal, resulting in a negative top-value.

This pr fixes this by using the built-in `window.innerHeight` when available. That fixes the issue, because safari then correctly returns the height including the area covered by the topbar.

For a more in-depth look into the issue, see this [jquery.documentsize](https://github.com/hashchange/jquery.documentsize#what-does-it-do-that-jquery-doesnt).
## Description

- Why is it necessary? To improve the usability of the modal in iphone safari

This is the only occurence of `$(window).height()` in all scripts.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | no |
| How to test? | Open a modal on safari ios. I reproduced this with a content-sized modal with overflowing content. Then compare patched and non-patched versions. |

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.modal.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.modal.js
@@ -150,7 +150,7 @@
 
             /**
              * The max height if sizing is set to `content`
-             * 
+             *
              * @type {Number}
              */
             maxHeight: 0,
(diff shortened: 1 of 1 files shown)


Phil23:
Thx @screeny05, nice pick. Will be merged in the next 5.2 patch release.

PR #826, opened 2016-10-28
Avoid fatal error when using the Address API resource

jeboehm:
This avoids the fatal error when using the Address API resource.

`Fatal error: Cannot use Shopware\Models\Customer\Customer as Customer because the name is already in use in /vagrant/movera-app/engine/Shopware/Components/Api/Resource/Address.php on line 31`


bcremer:
Thanks for your PR :+1:

PR #349, opened 2015-09-22
Plugin via Symlink einbinden

buddhaCode:
In machen Testumgebungen macht es Sinn ein Plugin-Verzeichnis über einen Symlink einzubinden. Der {link} Viewhelper hat dabei allerdings ein Problem den richtigen Pfad zu ermitteln. Dieser Patch fixt dieses Verhalten.

PR #829, opened 2016-11-02
Realpath abstraction / Plugin symlinks

hgiesenow:
## Description
With this patch it is possible to access files in symlinked themes or plugins. 
It adds an abstraction layer for the php builtin realpath() function. It is in most parts compatible but does not resolve symlinks. Also there are some tests added for the new realpath abstraction. Another benefit is,  that with the abstraction it would be possible to fully unittest the calling methods by mocking the abstraction.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | https://github.com/shopware/shopware/pull/349 and https://issues.shopware.com/issues/SW-14241
| How to test?     | Create a symlink to a theme inheriting from responsive. See that images are not broken.

This allow symlinks to work

### DIFF
--- a/engine/Library/Enlight/Loader.php
+++ b/engine/Library/Enlight/Loader.php
@@ -152,32 +152,34 @@ class Enlight_Loader
     /**
      * Realpath implementation that is 100% compatible but symlink aware
      *
-     * @param $path
-     *
+     * @param string $path
      * @return string
      */
     public static function realpath($path)
     {
         // No symlink
-        if (realpath($path) == $path) {
+        if (realpath($path) === $path) {
             return $path;
         }
 
+        if (!isset($path['0'])) {
+            return getcwd();
+        }
         // Make path absolute
-        if ($path[0] != DIRECTORY_SEPARATOR) {
+        if ($path[0] !== DIRECTORY_SEPARATOR) {
             $path = getcwd().DIRECTORY_SEPARATOR.$path;
         }
 
         // Remove . and ..
         // See http://php.net/manual/en/function.realpath.php#84012
-        $path = str_replace(array('/', '\\'), DIRECTORY_SEPARATOR, $path);
+        $path = str_replace(['/', '\\'], DIRECTORY_SEPARATOR, $path);
         $parts = array_filter(explode(DIRECTORY_SEPARATOR, $path), 'strlen');
-        $absolutes = array();
+        $absolutes = [];
         foreach ($parts as $part) {
-            if ('.' == $part) {
+            if ('.' === $part) {
                 continue;
             }
-            if ('..' == $part) {
+            if ('..' === $part) {
                 array_pop($absolutes);
             } else {
                 $absolutes[] = $part;
--- a/engine/Library/Enlight/Loader.php
+++ b/engine/Library/Enlight/Loader.php
@@ -150,6 +150,50 @@ class Enlight_Loader
     }
 
     /**
+     * Realpath implementation that is 100% compatible but symlink aware
+     *
+     * @param $path
+     *
+     * @return string
+     */
+    public static function realpath($path)
+    {
+        // No symlink
+        if (realpath($path) == $path) {
+            return $path;
+        }
+
+        // Make path absolute
+        if ($path[0] != DIRECTORY_SEPARATOR) {
+            $path = getcwd().DIRECTORY_SEPARATOR.$path;
+        }
+
+        // Remove . and ..
+        // See http://php.net/manual/en/function.realpath.php#84012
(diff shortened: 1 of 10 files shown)


kuborgh-mspindelhirn:
@bcremer Any Feedback on this one?

bcremer:
Relates to https://github.com/shopware/shopware/pull/304, https://github.com/shopware/shopware/pull/349 and http://issues.shopware.com/issues/SW-17234.

bcremer:
@kuborgh-hgiesenow Thanks for this pull request. It's about time to tackle this issue for real :)

This PR looks very  good overall, I have some minor adjustments, can you please enable push permissions for me as described in https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/#enabling-repository-maintainer-permissions-on-existing-pull-requests

Ping for previously involved people (@SvenHerrmann @arvatis @FiveDigital @buddhaCode @andrewisplinghoff @hikaru-shindo):
You are welcome to review and test this PR in depth. 

Let's get this merged :)

engine/Library/Enlight/Loader.php
```diff
@@ -150,6 +150,50 @@ public static function isReadable($path)
...
     /**
+     * Realpath implementation that is 100% compatible but symlink aware
+     *
+     * @param $path
+     *
+     * @return string
+     */
+    public static function realpath($path)
+    {
+        // No symlink
```
bcremer, line 160: This condition is not a test for a symlink, it tests if the path is already a realpath

engine/Library/Enlight/Loader.php
```diff
@@ -150,6 +150,50 @@ public static function isReadable($path)
...
 
     /**
+     * Realpath implementation that is 100% compatible but symlink aware
+     *
+     * @param $path
+     *
+     * @return string
+     */
+    public static function realpath($path)
+    {
```
bcremer, line 159: You should skip you own implementation it's not a symlink:

```
if (!is_link($path)) {
    return realpath($path);
}
```

engine/Library/Enlight/Loader.php
```diff
@@ -150,6 +150,50 @@ public static function isReadable($path)
...
+        // No symlink
+        if (realpath($path) == $path) {
+            return $path;
+        }
+
+        // Make path absolute
+        if ($path[0] != DIRECTORY_SEPARATOR) {
```
bcremer: That will emit a notice if the string is empty. Check for empty string first.

```
        if (!isset($path['0'])) {
            return getcwd();
        }
        // Make path absolute
        if ($path[0] !== DIRECTORY_SEPARATOR) {
            $path = getcwd().DIRECTORY_SEPARATOR.$path;
        }
```

tests/Unit/EnlightLoaderTest.php
```diff
@@ -83,4 +83,40 @@ public function testAddIncludePath3()
...
+            // Relative paths
+            array('', getcwd()),
+            array('./', getcwd()),
+            array('../', realpath(getcwd().'/../')),
+
+            // Nonexisting paths
+            array('/nonexisting', false),
+            array('../nonexisting', false),
+            array('nonexisting', false),
+        );
```
bcremer: You realpath implementation is not 100% compatible with PHP's realpath.

Please consider the following testcase: 

```
['./nonexisting/..', false],
```

This should return `false` as `nonexisting` does not exists. In Your implementation this is not checked an `getcwd()` will be returned.

tests/Unit/EnlightLoaderTest.php
```diff
@@ -83,4 +83,40 @@ public function testAddIncludePath3()
...
+    }
+
+    /**
+     * Provide test cases
+     *
+     * @return array
+     */
+    public function dataProviderRealpath()
+    {
+        return array(
```
bcremer: This tests do not actually test any symlinks. 
Please create a directory structure with symlinks and add some testcases for that.

engine/Library/Enlight/Loader.php
hgiesenow: This will not work, as it only checks the given filde/folder, not the whole path.

hgiesenow:
> can you please enable push permissions for me as described in https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/#enabling-repository-maintainer-permissions-on-existing-pull-requests

This checkbox was already checked

engine/Library/Enlight/Loader.php
hgiesenow: Yes, that's what the code says. But the intention behind this check is to see if there is no symlink in the path.

PR #831, opened 2016-11-03
SW-16815 - fix empty category in export feed

jkrzefski:
## Description
* Why is it necessary?
  * Because it fixes a bug in the export module. That bug would output an empty string when trying to get a category while having a category without subcategories selected as the parent category.
* What does it improve?
  * It outputs the category name correctly.
* Does it have side effects?
  * None that I know of.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | Somehow not. I don't know why.
| Related tickets? | [SW-16815](http://issues.shopware.com/issues/SW-16815)
| How to test?     | Configure a Google XML export feed in the backend and set a category without subcategories as parent category. If there is a category path in <g:product_type>, it works.

The category field was empty if the selected parent category in the feed settings has no subcategories.

### DIFF
--- a/engine/Shopware/Core/sCategories.php
+++ b/engine/Shopware/Core/sCategories.php
@@ -435,7 +435,7 @@ class sCategories
             $articleId
         ));
 
-        return $id;
+        return ($id) ? $id : $parentId;
     }
 
     /**


vincentfuhler:
Thank you !

mitelg:
Hey @jkrzefski 
thanks for your PR!
as you can see, one test is failing. please be aware, that the `sGetCategoryIdByArticleId` method is also used in some other important places, not only in the `sExport` module. so there might be some side effects with your change. please try to fix your code, so the tests are fine.
best regards

mitelg:
I will close this PR due to inactivity. Feel free to open it again, if you have still issues

PR #832, opened 2016-11-03
Fix shop and locale id for missing snippets

samuelvogel:
## Description
* Why is it necessary? Otherwise snippets referenced in the code but missing form the `.ini` files are imported into the database with `shopID` `111` and `localeID` `0`
* What does it improve? Import such snippets with the correct default `shopID` `1` and `localeID` `1`
* Does it have side effects? Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| How to test?     | Execute the test case while reverting the other commits (or copy the test case to a clean shopware `5.2.9`), the test will fail

The following unwanted mutations of `section` will happen with each subsequent run because from 2. onwards array access on a string is performed:
1. `[1,1]` (array)
2. `'1:1'` (string)
3. `'111'` (string)
...

This then ends up in the database as `shopID` `111` and `localeID` `0`.

### DIFF
--- a/engine/Library/Enlight/Config.php
+++ b/engine/Library/Enlight/Config.php
@@ -107,7 +107,7 @@ class Enlight_Config extends Enlight_Config_BaseConfig implements ArrayAccess
             $this->_adapter = self::$_defaultAdapter;
         }
         if (isset($options['section'])) {
-            $this->_section = $options['section'];
+            $this->setSection($options['section']);
         }
         if (isset($options['extends'])) {
             $this->setExtends($options['extends']);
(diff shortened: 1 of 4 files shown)


mitelg:
hi @samuelvogel 
thanks for your PR :+1: there was already a ticket for this problem. I will reference them to each other

http://issues.shopware.com/issues/SW-15132

janbuecker:
Thanks @samuelvogel. Merged with 11b2da22bbc4a615a9c22bab9d77e6803046e50c.

PR #835, opened 2016-11-08
ArrayCollection Problem setOneToMany Function

ahmad-saad:
## Description
* Why is it necessary? In Models when we have ManyToMany RelationShip variables declared as ArrayCollection but Set function use setOneToMany function to set the values which expected Array not ArrayCollection.
* What does it improve? We have the appilty now to add the variables as Array like in the shopware core or as ArrayCollection as in the declaration 

| Questions        | Answers
| BC breaks?       | yes/no
| Tests pass?      | yes
| Related tickets? | http://issues.shopware.com/issues/SW-16893
| How to test?     | https://forum.shopware.com/discussion/41791/anlegen-eines-configuratorsets-via-models#Comment_177874

**more details hier:**
https://forum.shopware.com/discussion/41791/anlegen-eines-configuratorsets-via-models#Comment_177874

### DIFF
--- a/engine/Shopware/Components/Model/ModelEntity.php
+++ b/engine/Shopware/Components/Model/ModelEntity.php
@@ -166,7 +166,7 @@ abstract class ModelEntity
      * <li>So the parameter expect <b>"customer"</b></li>
      * </ul>
      *
-     * @param array|null $data Model data, example: an array of \Shopware\Models\Order\Order
+     * @param \Doctrine\Common\Collections\ArrayCollection|array|null $data Model data, example: an array of \Shopware\Models\Order\Order
      * @param string $model Full namespace of the association model, example: '\Shopware\Models\Order\Order'
      * @param string $property Name of the association property, example: 'orders'
      * @param string $reference Name of the reference property, example: 'customer'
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Model/ModelEntity.php
```diff
@@ -185,6 +185,10 @@ public function setOneToMany($data, $model, $property, $reference = null)
             $this->$getterFunction()->clear();
             return $this;
         }
+        // if \Doctrine\Common\Collections\ArrayCollection
+		if($data instanceof \Doctrine\Common\Collections\ArrayCollection){
+			$data = $data.toArray();
```
bcremer: This is not valid PHP-Syntax and thus breaks our tests: https://travis-ci.org/shopware/shopware/jobs/174157092#L958

engine/Shopware/Components/Model/ModelEntity.php
```diff
@@ -185,6 +185,10 @@ public function setOneToMany($data, $model, $property, $reference = null)
             $this->$getterFunction()->clear();
             return $this;
         }
+        // if \Doctrine\Common\Collections\ArrayCollection
+		if($data instanceof \Doctrine\Common\Collections\ArrayCollection){
```
bcremer: Please reformat the code according to out coding standards: https://developers.shopware.com/developers-guide/coding-standards/#php-coding-standard

ahmad-saad: I think the problem with tabs imust use spaces . right?

engine/Shopware/Components/Model/ModelEntity.php
ahmad-saad: Ja I think I write this . by mistek

mitelg:
hey @ahmad-saad 
Unfortunately I was not able to import your changes, because the source ist not available anymore. If you still want to fix this issue, please open another PR.
Thanks and best regards

PR #836, opened 2016-11-10
5.2. voucher safety

LotharHoffmann:
## Description

While writing a plugin for our shop I've realized a design flaw within the vouchersystem.
Scenario:
Given you have 2 vouchers where the first one has individual codes, providing f.e. 50% discount on everything and the second one, which has a common code, giving 10% discount.
The vouchers got the IDs 1 (individual) and 2 (common) in the s_emarketing_voucher table.
While the code for the common voucher is held in the same table, the codes for the first vouchers are held within s_emarketing_vouchers_codes.
Now let's assume that a user starts to collect some articles in his basket, which he want's to order later. Now the user adds the common voucher. So the database saves those articles and while the user keeps on collecting, the admin deletes the common voucher. 
As soon as the sGetVoucher-function in the sBasket-file would be called, it would return the second entry of the s_emarketing_vouchers_codes-table, since the first Select would return a blank and the Select on the codes-table doesn't verify that the ordernumbers are correct.
This could lead to the effect that the user gets a 50% vouchervalue with a 10% voucher-code.
Normally this should be caught by some other mechanisms - like that vouchers are removed from the basket if they are deleted. There is still a theoretical chance (deadlock, defect sector at the harddrive, etc.) that the original basket-entry of the voucher remains, which would occur the described effect.
This is more like a design-flaw than a real risk, but it could occur in larger shops where the outcome could be desasterous.
By adding a new number to the modus-column in the basket-table the system is able to tell if a voucher is a common one or has individual codes, selecting only the required table.

There should be only a little side-effect, that themes need to add the new number to their cart-item.tpl-files.

| Questions        | Answers
| Tests pass?      | yes
| How to test?     | Manually add an entry to the table s_order_basket giving it an articleID which has the same value an individual code


engine/Shopware/Core/sBasket.php
```diff
@@ -790,7 +793,7 @@ public function sAddVoucher($voucherCode, $basket = '')
           sessionID, articlename, articleID, ordernumber, shippingfree,
           quantity, price, netprice,tax_rate, datum, modus, currencyFactor
         )
-        VALUES (?,?,?,?,?,1,?,?,?,?,2,?)
+        VALUES (?,?,?,?,?,1,?,?,?,?,".($voucherDetails['modus'] == 1 ? 5 : 2).",?)
```
hlohaus, line 796: A prepared statement should be used here.

janbuecker:
Thanks @LotharHoffmann 

PR #842, opened 2016-11-14
Add article number to customer questions form

samuelvogel:
## Description
Please describe your pull request: Adds the article number to the customer questions form for an article.
* Why is it necessary? Makes answering the question easier, since the article number provides an exact reference to the article.
* What does it improve? Searching for the article the question is about in the shop is easer.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | https://issues.shopware.com/issues/SW-15693
| How to test?     | Use the customer article question form in the frontend

screenshot 2016-11-14 18 51 59

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Forms.php
+++ b/engine/Shopware/Controllers/Frontend/Forms.php
@@ -171,7 +171,9 @@ class Shopware_Controllers_Frontend_Forms extends Enlight_Controller_Action
                             if ($this->Request()->getParam('sOrdernumber', null) !== null) {
                                 $getName = Shopware()->Modules()->Articles()->sGetArticleNameByOrderNumber($this->Request()->getParam('sOrdernumber'));
                                 $text = Shopware()->Snippets()->getNamespace('frontend/detail/comment')->get('InquiryTextArticle');
-                                $text .= ' ' . $getName;
+                                // Add name and number of article
+                                $escapedArticleNumber = $this->get('shopware.escaper')->escapeHtml($this->Request()->getParam('sOrdernumber'));
+                                $text .= ' ' . $getName . ' (' . $escapedArticleNumber . '): ';
                                 $this->_elements[$id]['value'] = $text;
                                 $element['value'] = $text;
                             }


jkrzefski:
If I understand your changes correctly, this only affects the prefilled text in the textarea. While that might be sufficient in most cases, I would suggest to make an entirely new field (hidden) that contains the ordernumber (if one exists) and passes it separately. That would make it harder for the customer to remove it from the inquiry.

samuelvogel:
Yes, but this way it's visible to the customer and conforms to how most of the contact form variations in the Shopware frontend work. Also it's the quickest and easiest solution 😉

engine/Shopware/Controllers/Frontend/Forms.php
```diff
@@ -171,7 +171,8 @@ protected function getContent($formId)
                             if ($this->Request()->getParam('sOrdernumber', null) !== null) {
                                 $getName = Shopware()->Modules()->Articles()->sGetArticleNameByOrderNumber($this->Request()->getParam('sOrdernumber'));
                                 $text = Shopware()->Snippets()->getNamespace('frontend/detail/comment')->get('InquiryTextArticle');
-                                $text .= ' ' . $getName;
+                                // Add name and number of article
+                                $text .= ' ' . $getName . ' (' . $this->Request()->getParam('sOrdernumber') . '): ';
```
hlohaus: Hey, this is a XSS hole. Please escape the request parameter, before you but it in the html-text-result.

 https://developers.shopware.com/developers-guide/shopware-5-upgrade-guide-for-developers/#other-changes-0
 https://framework.zend.com/manual/2.4/en/modules/zend.escaper.introduction.html

samuelvogel: Sorry, thanks for finding this! I've update the PR with escaping of the article order number content.

hlohaus:
Hi Samuel,

I remembered in a support ticket, that it already works:
 http://i.imgur.com/t5RdXYf.png
But this also means that we now have a smarty hole. Your request parameter will be also evaluated. Example:
    http://www.shopwaredemo.de/anfrage-formular?sInquiry=detail&sOrdernumber={eval}
The fix would be, you would also do a "smarty"-escape:
https://github.com/VIISON/shopware/blob/4aeeb1a52100b0722d35888bd1dfa955bfa34273/engine/Shopware/Controllers/Frontend/Forms.php#L386
(a little helper would be nice)

Greeting Heiner

samuelvogel:
As this feature is not super important to us (just nice-to-have) and a slightly different approach can be taken (using a new form field like you suggested), we will close this here.

PR #843, opened 2016-11-15
Disabled strip_tags on password parameters.

jkrzefski:
## Description
* Why is it necessary?
  * It allows more secure passwords.
* What does it improve?
  * It allows "<" and ">" in customer passwords.
* Does it have side effects?
  * No.

| Questions        | Answers
| BC breaks?       | no
| Tests pass? | yes
| Related tickets? | [#SW-16954](http://issues.shopware.com/issues/SW-16954)
| How to test?     | Make a frontend user with password "abcd<3defg"

These changes allow "<" and ">" in passwords of customers.

### DIFF
--- a/_sql/migrations/795-prepare-password-types-for-reencoding.php
+++ b/_sql/migrations/795-prepare-password-types-for-reencoding.php
@@ -1,0 +1,9 @@
+<?php
+
+class Migrations_Migration795 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $this->addSql("UPDATE s_user SET encoder = 'StripTags';");
+    }
+}
--- a/_sql/migrations/795-prepare-password-types-for-reencoding.php
+++ b/_sql/migrations/795-prepare-password-types-for-reencoding.php
@@ -4,6 +4,6 @@ class Migrations_Migration795 extends Shopware\Components\Migrations\AbstractMig
 {
     public function up($modus)
     {
-        $this->addSql("UPDATE s_user SET encoder = 'StripTags';");
+        $this->addSql("UPDATE s_user SET encoder = 'StripTags' WHERE encoder IN ('bcrypt', 'sha256', 'md5');");
     }
 }
--- a/engine/Shopware/Components/Password/Encoder/StripTags.php
+++ b/engine/Shopware/Components/Password/Encoder/StripTags.php
@@ -25,7 +25,15 @@ class StripTags implements PasswordEncoderInterface
      */
     public function isPasswordValid($password, $hash)
     {
-        return password_verify(strip_tags($password), $hash);
+        $passwordManager = new \Shopware\Components\Password\Manager(
+            Shopware()->Config()
+        );
+
+        $password = strip_tags($password);
+
+        return $passwordManager->isPasswordValid($password, $hash, 'bcrypt') ||
+            $passwordManager->isPasswordValid($password, $hash, 'sha256') ||
+            $passwordManager->isPasswordValid($password, $hash, 'md5');
     }
 
     /**
--- a/engine/Shopware/Components/Password/Encoder/StripTags.php
+++ b/engine/Shopware/Components/Password/Encoder/StripTags.php
@@ -2,47 +2,47 @@
 
 class StripTags implements PasswordEncoderInterface
 {
-	/**
-	 * @return string
-	 */
-	public function getName()
-	{
-		return 'StripTags';
-	}
+    /**
+     * @return string
+     */
+    public function getName()
+    {
+        return 'StripTags';
+    }
 
-	/**
-	 * @return boolean
-	 */
-	public function isCompatible()
-	{
-		return version_compare(PHP_VERSION, '5.3.7', '>=');
-	}
+    /**
+     * @return boolean
+     */
+    public function isCompatible()
+    {
+        return version_compare(PHP_VERSION, '5.3.7', '>=');
+    }
 
-	/**
-	 * @param  string $password
-	 * @param  string $hash
-	 * @return bool
-	 */
-	public function isPasswordValid($password, $hash)
-	{
-		return password_verify(strip_tags($password), $hash);
-	}
+    /**
+     * @param  string $password
+     * @param  string $hash
+     * @return bool
+     */
+    public function isPasswordValid($password, $hash)
+    {
+        return password_verify(strip_tags($password), $hash);
+    }
 
-	/**
-	 * @param  string $password
-	 * @return string
-	 */
-	public function encodePassword($password)
-	{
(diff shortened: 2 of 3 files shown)


engine/Shopware/Plugins/Default/Frontend/InputFilter/Bootstrap.php
```diff
@@ -115,7 +115,8 @@ public function onRouteShutdown(Enlight_Controller_EventArgs $args)
                     $process[$key][self::filterValue($k, $regex)] = $v;
                     $process[] = &$process[$key][self::filterValue($k, $regex)];
                 } else {
-                    $process[$key][self::filterValue($k, $regex)] = self::filterValue($v, $regex);
+                    $stripTags = !in_array($k, ['password', 'passwordConfirmation', 'currentPassword']);
```
hlohaus, line 118: I would just make an
`if(!in_array($k, ['password', 'passwordConfirmation', 'currentPassword'])) {`
`//filterValue()`
`}`
here.
👍

jkrzefski: So your solution would not use filterValue() at all? I tried to focus on keeping it and only making the strip_tags() optional. Otherwise this might expose the database to possible injections I think. But I'm not so sure about that part.

screeny05:
What about backwards-compatibility?

If a user registered in an older version without this fix and his password was, say `foo<bar>` that `<bar>` part got stripped out. It isn't the correct password which gets saved to the database (`foo`) and it isn't the correct password which gets checked. And it is less secure. But it works.

With this fix the user wouldn't be able to login to his account with the same password, as the database contained a hash of `foo`, not `foo<bar>`.

This fix is needed, don't get me wrong. But there needs to be a mechanism to catch this case, so the user won't think he's got hacked.

jkrzefski:
@screeny05 You have a good point there. I haven't thought about it like this yet. I'm going to look into this and hopefully come up with a solution to this soon.

engine/Shopware/Plugins/Default/Frontend/InputFilter/Bootstrap.php
hlohaus: It would at least be consistent. 
And the loss of security would be minimal.
(The input filter is only the last wall of defence)

hlohaus:
Yes, you could implement a "password encoder" for that.
Example from:
 engine/Shopware/Components/Password/Encoder/Bcrypt.php
...
    public function isPasswordValid($password, $hash)
    {
        return password_verify(strip_tags($password), $hash);
    }

jkrzefski:
@hlohaus Actually I think I won't reencrypt old passwords automatically. I see that as a possible security issue for the customer. Let's have the following scenario (ignoring minimum password-length):

The customer has the password "foo<bar". He can now log into his account with "foo<bar" but also with "foo<foo", "foo<barfoobar" and so on. If I were to reencrypt his password the next time he logs in, there is a chance to encrypt a wrong password. Let's have the customer accidentally put "foo<bat" this time. The login would work and the password would be successfully reencrypted but the customer is now unable to log into his account again.

I'm grateful for possible solution suggestions on this.

hlohaus:
@jkrzefski I think rather the opposite. It would solve the security problem.
And if the customers enter the wrong password, they can still use the password reset.

screeny05:
How about we just check if the password the user has given us is valid with the old encryption and only then re-encrypt it

hlohaus:
That's exactly what a (deprecated)-"password encoder" will do in shopware.

jkrzefski:
Ok, that sounds possible. Can you give me a quick hint on how I could accomplish that?

hlohaus:
We need only this "password encoder":
 https://gist.github.com/hlohaus/482bb27227cca8081ab2a968172a875f
And then we can change the "state" of the customer password:
`
 UPDATE `s_user` SET `encoder` = 'StripTags' WHERE `encoder` = 'bcrypt';
`

jkrzefski:
Would that be good practice? I mean that encoder would basically just be a copy of the bcrypt-encoder with the same capabilities (except for the name). It just feels kind of "unclean" to do it like that.
If I were to do it like this, I think I might just extend the bcrypt-encoder and change its name.

Also I need to know, how I get database updates into a pull request.

hlohaus:
It isn't just a "copy". All methods are changed. So is everything fine.

And here you can add your SQL delta:
  https://github.com/shopware/shopware/tree/5.2/_sql/migrations

engine/Shopware/Components/Password/Encoder/StripTags.php
```diff
@@ -0,0 +1,48 @@
...
+        return password_verify(strip_tags($password), $hash);
+    }
+
+    /**
+     * @param  string $password
+     * @return string
+     */
+    public function encodePassword($password)
+    {
+	    throw new Exception('Shoud never be called.');
```
hlohaus: throw new RuntimeException('Should never be called.');

_sql/migrations/795-prepare-password-types-for-reencoding.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+class Migrations_Migration795 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $this->addSql("UPDATE s_user SET encoder = 'StripTags';");
```
hlohaus: Add "WHERE encoder = 'bcrypt'"
because the shop owner can have old password hashes from "oxid" for example.

engine/Shopware/Components/Password/Encoder/StripTags.php
jkrzefski: What is the difference?

hlohaus: the "l" in "Should" is missing. And it is a "RuntimeException" :)

jkrzefski: Ok, thanks.

_sql/migrations/795-prepare-password-types-for-reencoding.php
jkrzefski: But what if SHA256 or MD5 are set? Would it be better to make it `WHERE encoder IN ('bcrypt', 'sha256', 'md5')`?

hlohaus: It would not work, because we test only with "password_verify" (bcrypt).
So we have to extends "isPasswordValid" first.

jkrzefski: I edited the isPasswordValid()-method to take those three encodings into account. Would it now be a good idea to change this to `WHERE encoder IN ('bcrypt', 'sha256', 'md5')`?

PR #847, opened 2016-11-18
Fix single position rounding in basket and order

thomasnordquist:
## Description

* **Why is it necessary?**
In some cases (for example when an order is net) and one tries to modify the order, the method **Order/Order::calculateInvoiceAmount** calculates the prices different from sBasket this is due to tax-calculation based rounding issues. (I'll explain in more details further down)
* **What does it improve?**
    * Tax calculation in **Order/Order** and **sBasket**
    * Overall test coverage (for calculating the basket/order) in both modules
* **Does it have side effects?**
    no

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | not to my knowledge
| How to test?     | run tests without fixes (screenshots attached)

**Why does the calculation has to change**
Prices for articles have to be seen as atomic values, meaning that changing the quantity of articles should have no effect on the price per article.
Example: If we have an article that costs 1.0049 € and we would order 2 of these, the invoice would read 2.01 € while buying them separately one would have 2 invoices with 1.00 € each.
When calculating the net / gross values this principle has to be respected as well.

These „errors“ are caused by multiplying a quantity>1 with an price that is more precise than one could actually pay.

In order to address this issue completely we first added tests for `Models/Order/Order::calculateInvoiceAmount()` and `sBasket::sGetBasketData()` / `sBasket::getBasketArticles()` and fixed the issues accordingly.

**Before**
Note: The tests run with quanity=100 so the issue becomes more obvious.
 => Net price for the first test should be 0.67 €; without the fix it will be calculated as 0.67109999 €
bildschirmfoto 2016-11-18 um 17 02 26

**After**
bildschirmfoto 2016-11-18 um 17 03 23

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2245,7 +2245,7 @@ class sBasket
                 || (!$this->sSYSTEM->sUSERGROUPDATA["tax"] && $this->sSYSTEM->sUSERGROUPDATA["id"])
             ) {
                 if (empty($getArticles[$key]["modus"])) {
-                    $priceWithTax = round($netprice, 2) / 100 * (100 + $tax);
+                    $priceWithTax = round(round($netprice, 2) / 100 * (100 + $tax), 2);
 
                     $getArticles[$key]["amountWithTax"] = $quantity * $priceWithTax;
                     // If basket comprised any discount, calculate brutto-value for the discount
@@ -2298,7 +2298,7 @@ class sBasket
                 if (!$this->sSYSTEM->sUSERGROUPDATA["tax"] && $this->sSYSTEM->sUSERGROUPDATA["id"]) {
                     $getArticles[$key]["amountnet"] = $quantity * round($netprice, 2);
                 } else {
-                    $getArticles[$key]["amountnet"] = $quantity * $netprice;
+                    $getArticles[$key]["amountnet"] = $quantity * round($netprice, 2);
                 }
             }
 
--- a/engine/Shopware/Models/Order/Order.php
+++ b/engine/Shopware/Models/Order/Order.php
@@ -1039,7 +1039,7 @@ class Order extends ModelEntity
         //iterate order details to recalculate the amount.
         /**@var $detail Detail*/
         foreach ($this->getDetails() as $detail) {
-            $invoiceAmount += $detail->getPrice() * $detail->getQuantity();
+            $invoiceAmount += round($detail->getPrice(), 2) * $detail->getQuantity();
 
             $tax = $detail->getTax();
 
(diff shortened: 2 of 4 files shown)


hlohaus:
Kann es sein, dass man dieses Problem nur bei eigenen Importen hat?
Normalerweise vertraut Shopware darauf, dass der Einzelpreis in der Datenbank stimmt.
Und bei Netto-Bestellungen sollte nicht der Bruttopreis gerundet werden,
da dort die Kunden den Nettopreis kaufen und sehen.

Wie genau kann man das Problem mit dem Shopware-Backend nachstellen?
Zusätzlich gibt es bald eh ein neuen Warenkorb. ;)

Gruß Heiner

samuelvogel:
Yes, it happens with regular articles created in the backend as well.

hlohaus:
@samuelvogel Can you give us an example?

And that with the "net orders" is still true.
Why we should round the gross unit price for net orders?
Currently this pull request would destroy the net orders in shopware.

Example:
 http://imgur.com/a/tlBBy

The amount should be "1,00", because:
  round(0.84*1.19, 2) == 1.00

thomasnordquist:
To clarify what we want to achieve:
```
Current equation for net orders:
	(1.08 * 100) * 1.19			// netprice * quantity * tax
 +	(1.67 * 1.19)				// + shipping * tax
---------------------------
	130.5073 => 130.51
```

```
Changes by us: “Zeilenweises runden”

	round(1.08 * 1.19,  2)  * 100		// taxed price on article basis
 +	(1,67 * 1,19)				// + shipping * tax
----------------------------
	130,9873 => 130,99
```

When we now see that there is a significant difference of 0,48 €
Even though both methods are **legally acceptable** and both methods, one should stick to one method.

_Examples where shopware rounds "Zeilenweise":_
- https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Core/sBasket.php#L1193
- https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Core/sBasket.php#L2272
- https://github.com/shopware/shopware/blob/c91e3381f269c4e1e8f5add05265c7fb2765d333/engine/Shopware/Models/Order/Order.php#L1056 (at this point the price should already be rounded)

In the end we didn’t introduce a new rounding system (on position basis) but just ensured it is done consistently throughout the basket/order and tested the both with great care.

The calculations above were taken from an actual order:

## Gross order (same with and without fix)
gross_order_without_fix

## Net order (without fix) - the price is off by 48 cents
net_order_without_fix

## After applying the fix (gross price is now congruent with the gross order from above)
net_order_with_fix


## To explain what happened in your given example

rounding-issue

## Why we should round the gross unit price for net orders?
If you are referring to these changes: 
- https://github.com/shopware/shopware/pull/847/commits/2ac50f2a3aa7f40a1853c9bd1a3cb68651539fc9#diff-b23d50242b56bd7accf6f9320717d4a0R1042

- https://github.com/shopware/shopware/pull/847/commits/2ac50f2a3aa7f40a1853c9bd1a3cb68651539fc9#diff-b23d50242b56bd7accf6f9320717d4a0R1042

We noticed that some of our customers have prices in their baskets which are not rounded to two decimal places. (Mostly if there are coupons in the basket)

We also wrote a test that addresses this issue:
https://github.com/VIISON/shopware/blob/6c286d44e20c2ffe48980a92eb657fb54d72fb8d/tests/Functional/Models/Order/TaxCalculationTest.php#L123-L144.

janbuecker:
Thanks @thomasnordquist 

We are currently refactoring the basket which progress you can check here https://developers.shopware.com/blog/2016/12/09/new-cart-bundle-concept/.

As for now, we won't change the current behaviour.

samuelvogel:
Ok. Are you taking the consistent "single position" rounding into account for the update?

PR #853, opened 2016-11-22
Ignore empty elements from (broken) manufacturer filter URLs

marcnoller:
Broken filter URLs for manufacturer filter that look like this

/?s=|1234

lead to an Exception not displaying the listing page. (Oops an error occured)

When using explode() on this parameter, an array with two elements is created,
one of which empty. This leads to the following Exception, when Shopware
processes the Criteria:

core.ERROR: exception 'Assert\InvalidArgumentException' with message 'Value "" is not an integer or a number castable to integer.' in vendor/beberlei/assert/lib/Assert/Assertion.php:212

Using preg_split with PREG_SPLIT_NO_EMPTY prevents the creation of empty
array elements and fixes this problem.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? | n/a |
| How to test? | Test filter parameter in listing with invalid manufacturer parameter. ?s=\|123. With PR no Exception is thrown |

### DIFF
--- a/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
+++ b/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
@@ -176,9 +176,12 @@ class CoreCriteriaRequestHandler implements CriteriaRequestHandlerInterface
             return;
         }
 
-        $manufacturers = explode(
-            '|',
-            $request->getParam('sSupplier')
+        /* Split filter parameter. Do not return empty elements from broken urls */
+        $manufacturers = preg_split(
+            '/\|/',
+            $request->getParam('sSupplier'),
+            -1,
+            PREG_SPLIT_NO_EMPTY
         );
 
         if (!empty($manufacturers)) {


hlohaus:
Hi,

Thank you for your pull request.
But I have a question:
Why exactly should it not throw an exception?
The URL is in this case wrong or broken.
And there are still more cases which are not intercepted:
  http://www.shopwaredemo.de/hoehenluft-abenteuer/?p=1&n=12&s=1%7C10|as
  
Greeting Heiner

marcnoller:
Hi Heiner,

well, that is a valid question.
I saw it in the field, that there are plugins who may generate URL parameters with empty elements. 
In my opinion an empty element should be handled silently, as it happens with non existing manufacturer IDs.
In case of syntax errors, I'm fine with an exception. (yes, strictly speakting =|123 is a syntax error, but this is a too easy one to let it run into a cast error, isn't it?).

greetings, Marc

hlohaus:
Okay, but you should it do in the "supplier" filter too:
 http://www.shopwaredemo.de/deutsch/hoehenluft-abenteuer/?p=1&n=12&s=2%7C10|

:)

marcnoller:
@hlohaus You removed your last comment?

hlohaus:
Yes, i was wrong. For me it is ready. 
Now the product team has to plan and implement it. :)

marcnoller:
Agree. Had a look at the searchBundle also, but seems to be fine.

janbuecker:
Thanks @marcnoller 

PR #854, opened 2016-11-22
Pass pass-by-reference parameters as references in hook proxy classes.

kleinmann:
## Description
* Why is it necessary?

A hook on a method with pass-by-reference parameters currently breaks the hooked method. This is due to an incorrect parameter array being created in `\Enlight_Hook_ProxyFactory:: generateMethods($class)`.

As stated in http://php.net/manual/en/function.call-user-func-array.php#example-5925, the parameter array for `call_user_func_array` has to pass in the values as references, if the called method expects pass-by-reference arguments.
This is currently not the case. The reference sign `&` is only used in the signature of the hooked method in the proxy class, not in the parameter array for the actual method call.

* What does it improve?

This PR adds a reference sign for both `$proxy_params` and `$array_params`. The `$params` are being prefixed correctly already. This allows for hooks on methods that expect a parameter passed as a reference.

* Does it have side effects?

The test suite passes and it fixes the error case for a plugin of ours. There are currently no explicit test cases for hook methods, but I'm assuming this does not affect any use cases, as it merely fixes a broken implementation for a specific use case.

The placeholder `<proxyMethodParameters>` is not used in the class template, but I included the `$proxy_params` in my change. This may or may not be correct, depending on how (if?) these are used. I could not find any actual usage of `<proxyMethodParameters>`in any case.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | no
| How to test?     | Install [SwagImportExport](https://github.com/shopwareLabs/SwagImportExport) in a shopware installation. Create, install and enable a new plugin with a hook for `Shopware\Components\SwagImportExport\DbAdapters\CustomerDbAdapter::prepareCustomer::before`. The hook method doesn't need to do anything. A customer import through Import/Export Advanced will then fail. Apply the PR, try again, the customer import will work correctly.

### DIFF
--- a/engine/Library/Enlight/Hook/ProxyFactory.php
+++ b/engine/Library/Enlight/Hook/ProxyFactory.php
@@ -275,12 +275,17 @@ class <namespace>_<proxyClassName> extends <className> implements Enlight_Hook_P
                     $proxy_params .= ', ';
                     $array_params .= ', ';
                 }
+
+                $array_param = '';
                 if ($rp->isPassedByReference()) {
                     $params .= '&';
+                    $proxy_params .= '$';
+                    $array_param .= '&';
                 }
                 $params .= '$' . $rp->getName();
                 $proxy_params .= '$' . $rp->getName();
-                $array_params .= '\'' . $rp->getName() . '\'=>$' . $rp->getName();
+                $array_param .= '$' . $rp->getName();
+                $array_params .= '\'' . $rp->getName() . '\'=>' . $array_param;
                 if ($rp->isOptional()) {
                     $params .= ' = ' . str_replace("\n", '', var_export($rp->getDefaultValue(), true));
                 }
--- a/tests/Unit/Components/Hook/EnlightHookProxyFactoryTest.php
+++ b/tests/Unit/Components/Hook/EnlightHookProxyFactoryTest.php
@@ -1,0 +1,134 @@
+<?php
+namespace Shopware\Tests\Unit\Components;
+
+use PHPUnit\Framework\TestCase;
+
+interface MyInterface
+{
+    public function myPublic($bar, $foo = 'bar');
+}
+
+interface MyReferenceInterface
+{
+    public function myPublic(&$bar, $foo);
+}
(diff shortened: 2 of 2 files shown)


dnoegel:
Nice debugging, thanks! Will need to have a closer look on that - but seems to make sense

bcremer:
@kleinmann Very nice find indeed.

I would like to add some unit tests for the ProxyManager to test the generated class code output.
Due to the fact that the class is not very test friendly I had to do some reflectionmagic to test the actual relevant method.

You can find the Test-Boilerplate here:
https://gist.github.com/bcremer/25173d16e5eb657026d3bf658adbec3a

Could you please add this to your PR with a testcase for the added behavior.

kleinmann:
@bcremer Thanks for the basic test. This saved me a lot of work :-)

I have added both the basic test case and a test case with pass-by-reference parameters.

hlohaus:
Hi,

@kleinmann I would not use the hooks, although I am there in the file.
It is better to make your services decorable. See also:
   https://developers.shopware.com/blog/2015/06/09/understanding-the-shopware-hook-system/

And actually we need the hooks only for the core classes,
Since they are too large for this and have no interfaces yet.
The core classes have no references.

Greeting Heiner

kleinmann:
@hlohaus I'm 100% on your page on that. 😃 
We always prefer decorating services and designing our plugins around interfaces and exchangeable services. I actually stumbled upon the bug in one of Shopware's own plugins, [SwagImportExport](https://github.com/shopwareLabs/SwagImportExport), where I tried to solve a customer request (which I actually solved in a cleaner way afterwards). 😉 

The hook system has its reason to exist imho, mostly for working with legacy core classes or third-party plugins that don't allow for a cleaner form of extension/interfacing.

hlohaus:
Hey,

we will solve it within the import / export module.
The coming major version will not contain any hooks anymore.

Greetings from Schöppingen
Heiner

bcremer:
:+1:

PR #857, opened 2016-11-24
[FEATURE] Allow prices with amount 0 for free articles

m-knabe-reply:
## Description
Although it is already possible to have free articles by adding articles with some price and create a price group with a 100% discount I see no reason why prices should not be 0 in the first place. There are just a few error checks that need to be removed/adjusted.
Just so we don't change the current behaviour I added a setting that disallows storing prices with 0. It can be changed in System>Backend

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | none
| How to test?     | Execute migration 796, set option in Settings > System > Backend, create Article with Price 0 (optionally with variants), buy article in FE

### DIFF
--- a/_sql/migrations/796-add-free-articles-config-element.php
+++ b/_sql/migrations/796-add-free-articles-config-element.php
@@ -1,0 +1,13 @@
+<?php
+
+class Migrations_Migration796 extends Shopware\Components\Migrations\AbstractMigration
+{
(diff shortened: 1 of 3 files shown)


janbuecker:
Thanks @m-knabe-reply 

The current workflow is to define a price group with a discount of 100%.

PR #866, opened 2016-12-01
Support lazy services

jeboehm:
## Description
This change allows the use of lazy services. The Symfony Kernel does this by default (\Symfony\Component\HttpKernel\Kernel::dumpContainer).
When you install the composer requirement symfony/proxy-manager-bridge, the class is injected and lazy services are generated.

See
https://symfony.com/doc/current/service_container/lazy_services.html

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | -
| How to test?     | Use lazy="true" in your service definition, install symfony/proxy-manager-bridge, rebuild the container. Inject the lazy service somewhere. You should see, instead of the real service, a proxy class should be injected.

### DIFF
--- a/engine/Shopware/Kernel.php
+++ b/engine/Shopware/Kernel.php
@@ -526,6 +526,10 @@ class Kernel implements HttpKernelInterface
     {
         // cache the container
         $dumper = new PhpDumper($container);
+        
+        if (class_exists('ProxyManager\Configuration') && class_exists('Symfony\Bridge\ProxyManager\LazyProxy\PhpDumper\ProxyDumper')) {
+            $dumper->setProxyDumper(new \Symfony\Bridge\ProxyManager\LazyProxy\PhpDumper\ProxyDumper(md5($cache->getPath())));
+        }
 
         $content = $dumper->dump(array('class' => $class, 'base_class' => $baseClass));
 


engine/Shopware/Kernel.php
```diff
@@ -526,6 +526,10 @@ protected function dumpContainer(ConfigCache $cache, ContainerBuilder $container
     {
         // cache the container
         $dumper = new PhpDumper($container);
+        
+        if (class_exists('ProxyManager\Configuration') && class_exists('Symfony\Bridge\ProxyManager\LazyProxy\PhpDumper\ProxyDumper')) {
```
bcremer, line 530: We use neither ProxyManager nor the proxy manager bridge in our composer.json. So this line will always evaluate to false.

If lazy services bring a huge benefit we might consider adding these dependencies.
But without adding these dependencies to our composer.json this pull request makes no sense in my opinion.

jeboehm: I just copied the Symfony implementation. The Proxy* requirements are still optional. Developers can include them in their plugins, if they want to use them. Maybe you should add them to the Composer suggestions?

bcremer: At this point in time I would not like to add "optional" dependencies since there is no supported way in shopware to add those via plugins.

If lazy services deliver a benefit to the core and to the plugin ecosystem we can add this dependency but please keep in mind that this introduces yet another layer of generated files and thus a layer of indirection.

jeboehm: The feature is considered stable for a few years.
It is very useful when you inject a bigger service that does expensive things in the constructor. We just implemented a lazy RabbitMQ connector that now creates the connection only, when it's really needed. It's also useful for Web-APIs.

bcremer:
As already discussed in the code comments this Pull Request is incomplete as it won't work without the ProxyManager dependency.

Doing slow/expensive work in the constructor is a bad practice anyway and there are ways to circumvent that manually. 

Please feel free to reopen a complete PR if there are measurable benefits in terms of performance.
Else this will introduce more dependencies and indirections that make debugging even harder.

Thank you @jeboehm for your input on this.

jeboehm:
This is a very conservative opinion, since this pattern is supported by the most major frameworks.

jeboehm:
To clarify; a service has to be declared as lazy explicitly. So a plugin developer CAN use it, but he doesn't have to.
You're right that a constructor shouldn't do too much expensive work. But at the other hand, it should assign all resources, that are needed later in the class. IMHO this also includes connections.

dnoegel:
> IMHO this also includes connections

I doubt that: The dbal connection for example does not do that in the constructor (\Doctrine\DBAL\Connection::__construct) same for e.g. ZendDB in ZF2 (https://github.com/zendframework/zend-db/blob/master/src/Adapter/Driver/Pdo/Connection.php). Furthermore this has massive downsides regarding performance (as this PR shows) and you need to introduce code generation in order to fix an issue that arises from an non-necessary architectural decision. 

So you want to introduce code into shopware, that a) is not used by shopware itself (which might be ok, if relevant for external devs), b) will only be used by few (if at all) developers and c) can easily be replaced by other architectural decisions. 

At the end of the day it is pretty easy for you to fix it in your code - but you want us to maintain additional code and (perhaps) a new dependency. I don't think this is about being "conservative" - it is just reasonable to think about why and if it is needed. And (just as Benjamin) I am not convinced, yet.

PR #867, opened 2016-12-02
Fire event after saving address to have a chance to save address attr…

dlobato:
## Description
* Why is it necessary?
When creating a new customer through the backend address attributes are not saved nor there's a chance for extensions to catch the success/fail callbacks after saving the address model. This makes difficult saving custom address attributes on customer creation.

* What does it improve?
This PR just fires an event after save success to give extensions a chance to intercept it and save custom address attributes if needed.

* Does it have side effects?
This change has no side effects.

| Questions        | Answers
| BC breaks?       | yes
| Tests pass?      | yes
| Related tickets? | 


nevoeiro:
We need the same logic in other places:

-> Fire event after saving customer, site and blog to have a chance to save attributes on creation

- If we need to move attribute fields into another form we don't have a chance to save attributes when we are creating new customer, site or blog;
- There's no options argument in onSave methods, like we have in onSaveArticle, e.g. Where we can add logic to callback function;
- Adding these fire events after saving, give us the possibility to extend the functionality more dinamically;

/cc @bcremer @dlobato @dnoegel

PR #699, opened 2016-09-02
Allow applying custom filters on base combobox element

svenmuennich:
### Why is it necessary?

Currently it is not possible to apply a filter on a `select` element, when adding one to the plugin config using the short store syntax (e.g. `'store' => 'Shopware.apps.Base.store.Dispatch'`). However, since e.g. the dispatch base store applies a filter on the `active` field by default, it is desirable to be able to override those default filters in a plugin config.
### What does it improve?

This PR changes the `Shopware.apps.Base.view.element.Select` element to check for an optional `filter` option and, if set, clears all filters from the store and applies the passed filters. Furthermore the `remoteFilter` property is set to `true` in all ExtJS base stores, whose corresponding controller actions allow passing a `filter` parameter. Finally, the  default filter on `dispatches.active` added in `Shopware_Controllers_Backend_Base::getDispatchesAction()` is removed and instead a filter on `active` is added to `Shopware.apps.Base.store.Dispatch`. This allows to override the filter and keeps backwards compatibility of the `Shopware.apps.Base.store.Dispatch` at the same time. However, it changes the default behaviour `Shopware_Controllers_Backend_Base::getDispatchesAction()` to respond all dispatch methods instead of only active ones.
### Does it have side effects?

The following base stores will use remote filtering by default:
- `Shopware.apps.Base.store.Country`
- `Shopware.apps.Base.store.CountryArea`
- `Shopware.apps.Base.store.CountryState`
- `Shopware.apps.Base.store.Currency`
- `Shopware.apps.Base.store.CustomerGroup`
- `Shopware.apps.Base.store.Dispatch`
- `Shopware.apps.Base.store.Locale`
- `Shopware.apps.Base.store.OrderStatus`
- `Shopware.apps.Base.store.Payment`
- `Shopware.apps.Base.store.PaymentStatus`
- `Shopware.apps.Base.store.PositionStatus`
- `Shopware.apps.Base.store.Tax`

Furthermore `Shopware_Controllers_Backend_Base::getDispatchesAction()` will now respond all dispatch methods, instead of only active ones.

| Questions | Answers |
| --- | --- |
| BC breaks? | no |
| Tests pass? | yes |
| Related tickets? |  |
| How to test? | Create a plugin containing a config form `select` element with e.g. the `store` `Shopware.apps.Base.store.Language` and a  `filter` of `[['property' => 'name', 'value' => 'English']]`. The select element will only show the store named `English`. |

PR #874, opened 2016-12-06
SW-17149: Combo/Select Elements autocomplete process in plugin config…

ahmad-saad:
### **Why is it necessary?**
after applying the change from this [Pull Request](https://github.com/shopware/shopware/pull/699) we have a big problem with the Stores

> - Shopware.apps.Base.store.Country
> - Shopware.apps.Base.store.OrderStatus
> - Shopware.apps.Base.store.PaymentStatus

we have the same exception as [heir ](https://github.com/shopware/shopware/pull/699#discussion_r87038134) for all the previous stores.
### **What does it improve?**
I use the same way [heir](https://github.com/shopware/shopware/commit/46149885906d31187723008c78e3559733845384) to fix the other stores.
But in this way we maket just the exception disappear but we still have another problem when we try to write something in the combo element Extjs remove it directly. ( In other word the autocomplete not work anymore).
I change the code in select.js file to add '%' around the value to make the php code search for records contain the insert value and not just the exact value.
**Note:**  this code give us the ability to autocomplete multi values in Multiselection Mode.
In my system until now there is no side effects.
but you must check the code in your side, _because this is my first time with Extjs_.
in select.js there is two 'Todo' . I hope someone have more experience can solve it.

> //@todo Find a way to add the custom filter if there is one
> //@todo Find a way to keep the previous values selected after store filtering in multiSelect Mode

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | [SW-17149](http://issues.shopware.com/issues/SW-17149)
| How to test?     | I create a simple plugin to test this issue [TestAutoCompleteBaseStoreIssue](https://github.com/ahmad-saad/TestAutoCompleteBaseStoreIssue)

### DIFF
--- a/engine/Shopware/Plugins/Default/Backend/Auth/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Backend/Auth/Bootstrap.php
@@ -468,7 +468,7 @@ class Shopware_Plugins_Backend_Auth_Bootstrap extends Shopware_Components_Plugin
         }
         if (empty($options['gc_maxlifetime'])) {
             $backendTimeout = $this->Config()->get('backendTimeout', 60 * 90);
-            $options['gc_maxlifetime'] = (int) $backendTimeout;
+            $options['gc_maxlifetime'] = (int) $backendTimeout ?: PHP_INT_MAX;
         }
         unset($options['locking']);
 


ahmad-saad:
any news about this pull request as I see the issues Team had mein Tieckt Scheduled.

so I think this is a global problem not just in my side I hope this Problem will be solve as soon as possibale.

Regard's

PR #875, opened 2016-12-06
Added onInit event to the StateManager.

jkrzefski:
## Description
* Why is it necessary?
** It created a standard way to register custom jQuery plugins.
* What does it improve?
** You don't have to use $(document).ready() anymore.
* Does it have side effects?
** No.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | None
| How to test?     | Register a jQuery plugin by using `$.subscribe('StateManager/onInit', addMyPlugin);`

This event could be useful to register custom jQuery plugins to the StateManager. Instead of using $(document).ready() a plugin could then be registered by $.subscribe('StateManager', addMyPlugin);


bcremer:
@shopware/shopware-webdev Any observations on this PR?

PR #879, opened 2016-12-08
Update config.php.dist

mitelg:
## Description
* What does it improve?
code style of config.php.dist
* Does it have side effects?
no


mitelg:
made PR in stash

PR #880, opened 2016-12-08
Added option to clear httpcache before sw:warm:http:cache

shyim:
## Description
* Why is it necessary?
The cache dir of httpcache should be cleaned before httpcache warmup
* What does it improve?
Cleares the httpcache before warmup
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | Run command

### DIFF
--- a/engine/Shopware/Commands/WarmUpHttpCacheCommand.php
+++ b/engine/Shopware/Commands/WarmUpHttpCacheCommand.php
@@ -26,6 +26,7 @@ namespace Shopware\Commands;
 
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


engine/Shopware/Commands/WarmUpHttpCacheCommand.php
```diff
@@ -41,6 +40,7 @@ protected function configure()
             ->setName('sw:warm:http:cache')
             ->setDescription('warm up http cache')
             ->addArgument('shopId', InputArgument::OPTIONAL, 'The Id of the shop')
+            ->addOption('clearcache', 'c', InputArgument::OPTIONAL, 'Clear complete httpcache before warmup', false)
```
janbuecker: You should use `InputOption::VALUE_NONE` since it's an option.

shyim: its updated

PR #882, opened 2016-12-12
Fixes undefined index error in sw:snippets:to:db --include-plugins

kuborgh-mspindelhirn:
## Description
This fixes some `PHP Notice` errors that will be raised because of an unchecked array access when working with plugins in new structure. This pull request adds a check before accessing the array.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | no
| How to test?     | Create a plugin with new plugin structure. Add snippets. Run import command


bcremer:
Merged with commit https://github.com/shopware/shopware/commit/8c172c310a3b55e045ccf54c415c081f3228c338.
Thanks @kuborgh-mspindelhirn :+1:

PR #884, opened 2016-12-15
Allow the serialization of Enlight exceptions

sbeining:
## Description
* Why is it necessary?
  If for example the advertised ShyimProfiler plugin is used everytime an exception is thrown there will be an error, that a PDO object can not be serialized
* What does it improve?
  It improves compatability with plugins that try to serialize exeptions
* Does it have side effects?
  The backtrace of unserialized exceptions will be missing.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | none
| How to test?     | Example: Install Shyim Profiler plugin and provoke a CSRF token exception

The backtrace often contains objects that required PDO objects as
arguments.
To allow the serialization of exceptions the backtrace is simply
omitted.

### DIFF
--- a/engine/Library/Enlight/Exception.php
+++ b/engine/Library/Enlight/Exception.php
@@ -32,7 +32,7 @@
  * @copyright  Copyright (c) 2011, shopware AG (http://www.shopware.de)
  * @license    http://enlight.de/license     New BSD License
  */
-class Enlight_Exception extends Exception
+class Enlight_Exception extends Exception implements Serializable
 {
     /**
      * Constant that a class could not be found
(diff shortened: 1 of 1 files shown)


bcremer:
@sbeining Thanks for your pull request.

The exception should not be responsible for it's serialization. This will also just fix one of many cases in which a PDO object, or another object that can't be serialized, might be in the collected information for the profiler.

So it's the responsibility of the profiler to filter, prepare and normalize all collected data before it gets serialized. 

Please see the following pull request in the profiler project for an example implementation: https://github.com/shyim/shopware-profiler/pull/23.

That given I'll close this PR.

PR #885, opened 2016-12-15
Fix jquery.plugin-base event listener removal

robert-wettstaedt:
## Description
* Why is it necessary?
The jquery.plugin-base _off function used wrong parameters in one of the $.each function callbacks, resulting in wrong behaviour. Under certain cirumstances this even resulted in the loss of all event listeners attached to the specified element.
* What does it improve?
The PR fixes the parameters in the $.each function callback so that only the passed event is removed as a listener from the element.
* Does it have side effects?
No.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | http://issues.shopware.com/issues/SW-17258
| How to test?     | Add some event listeners to an element using plugin._on and remove one using plugin._off. Check event listeners on the element using $._data(element)

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js
@@ -153,7 +153,9 @@
                 name = me.getName();
 
             $.each(me._events, function (i, obj) {
-                obj.el.off(obj.event);
+                if (typeof obj !== 'undefined') {
+                    obj.el.off(obj.event);
+                }
             });
 
             // remove all references of extern plugins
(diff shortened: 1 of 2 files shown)


themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js
```diff
@@ -153,7 +153,9 @@
                 name = me.getName();
 
             $.each(me._events, function (i, obj) {
-                obj.el.off(obj.event);
+                if (typeof obj !== 'undefined') {
```
screeny05, line 156: probably better to check if `typeof obj === 'object'`

klarstil:
Thanks for the PR. I agree with @screeny05 on his comment.

mitelg:
@r-wettstaedt thanks for your contribution :+1:

PR #890, opened 2016-12-20
Configurable plugin configuration directory

jochenmanz:
## Description
* Why is it necessary?
Added a way to change the default locations of the menu.xml, config.xml and the snippet directory to allow Plugins with a different folder structure without handling these things on their own.
* What does it improve?
Needs more work / needs discussion
* Does it have side effects?
None

| Questions        | Answers
| BC breaks?       | yes
| Tests pass?      | yes
| Related tickets? | none
| How to test?     | Create a plugin an override public methods


janbuecker:
Thanks @jochenmanz 

We want to keep a consistent schema of where you have to put your files to let the magic happen. Recently, we also closed a PR regarding to set the default caches to be cleared after install/update. So basically, that decision applies here too.

PR #893, opened 2016-12-22
Add State Info to order billing and shipping address information

ahmad-saad:
### **Why is it necessary?**
in Reset Api document man can see that by the [address ](https://developers.shopware.com/developers-guide/rest-api/models/#address) there is Country and State info.
but when you check the return data you find just the country info there is no State info.
### **What does it improve?**
This change will add the State info to the return data.
no

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | just try to call an order by id by Reset Api.


mitelg:
thanks for your PR @ahmad-saad :+1:

PR #894, opened 2016-12-27
Fix overwriting ssl for newsletter mailer 53/5.3

wildi1:
## Description

* Why is it necessary?
When the default mailer in shopware is ssl/tls, then it is not possible to set an non ssl/tls mailer for the newsletter, because the setting ssl => null is getting always overwritten. 0 or an emtpy string is also not possible because of the Zend_Mail_Protocol_Smtp

* What does it improve?
You can set a non ssl/tls newsletter mailer, even the default mailer is ssl/tls

* Does it have side effects?
No

| Questions        | Answers
| BC breaks?       | yes
| Tests pass?      | yes
| Related tickets? | None
| How to test?     | Try to setup a non ssl/tls newsletter mailer when the default mailer is ssl/tls

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/Bridge/MailTransport.php
+++ b/engine/Shopware/Components/DependencyInjection/Bridge/MailTransport.php
@@ -56,9 +56,6 @@ class MailTransport
                 $options['username'] = $config->MailerUsername;
                 $options['password'] = $config->MailerPassword;
             }
-            if (!isset($options['ssl']) && !empty($config->MailerSMTPSecure)) {
-                $options['ssl'] = $config->MailerSMTPSecure;
-            }
             if (!isset($options['port']) && !empty($config->MailerPort)) {
                 $options['port'] = $config->MailerPort;
             }


OliverSkroblin:
Hi @wildi1 

I'll check out your PR in a few and let you know what we'll do with it. I'm concerted about the backward compatibility and would like to check your PR beforehand.

Best regards

bcremer:
@wildi1 This change breaks/removes support for TLS/SSL connections that are configured in the shopware backend completely. 

The variable is a bit hard to find because of some name magic in the config reader. The actual config variable in the database is called `mailer_smtpsecure` and can be configured in the shopware backend under general settings in the mailer settings "Connection Prefix /Verbindungs Präfix" (See http://en.community.shopware.com/Mailer_detail_944.html#Verbindungs_Pr.E4fix).

This issue has to be fixed in the Newsletter Plugin. Please open a ticket for that.

wildi1:
What about a solution that we can write "none" to SSL:

if (!isset($options['ssl']) && !empty($config->MailerSMTPSecure)) {
    $options['ssl']=$config->MailerSMTPSecure;
}
elseif ($options['ssl']=='none') {
$options['ssl']=null;
}

Because when you only fix it in the Newsletter Plugin, and in the future perhaps, another Plugin needs to set a different Mailer, then it is fixed for that too.

bcremer:
This has to be fixed and tested properly in the newsletter module. A plugin should not use the MailTransport Factory from the core directly but create a new mail transport for themselves. 
There is already ticket http://issues.shopware.com/issues/PT-5463 that you can comment and vote on.

PR #897, opened 2017-01-04
[TASK] Added console command sw:rebuild:seo:index

FiveDigital:
* Why is it necessary? Because you cannot rebuild the SEO index on CLI.
* What does it improve? Enables you to rebuild the SEO index on CLI on demand.

Inspired by `Shopware_Plugins_Core_RebuildIndex_Bootstrap::onRefreshSeoIndex()`. I have also added the possibility to add a `shopId` argument to rebuild URLs for this shop only.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | no
| How to test?     | Truncate table `s_core_rewrite_urls` and run the new command `bin/console sw:rebuild:seo:index`. After the command is finished the generated SEO urls should be present in the database.

### DIFF
--- a/engine/Shopware/Commands/RebuildSeoIndexCommand.php
+++ b/engine/Shopware/Commands/RebuildSeoIndexCommand.php
@@ -1,0 +1,135 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


FiveDigital:
I have additionally fixed some PHP notices that appeared on CLI when running this command. Please let me know if you want to use this and if I should squash the commits or drop e0b8e26.

engine/Shopware/Commands/RebuildSeoIndexCommand.php
```diff
@@ -0,0 +1,131 @@
...
+            ->setHelp('The <info>%command.name%</info> rebuilds the SEO index')
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->database = $this->container->get('db');
```
janbuecker: Please use `dbal_connection` instead of `db`

FiveDigital: Done. Please use this in core, too ;)

mitelg:
thanks @FiveDigital :+1:

PR #898, opened 2017-01-10
Add articleId check to Variant Api Resource

Gugiman:
## Description
Allow the mainDetail->number to be changed within the pool of variant ordernumbers of that specific article that already exists over the API

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | none
| How to test?     | Change mainDetail->number to another ordernumber of a variant already exists within this article over API.

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Variant.php
+++ b/engine/Shopware/Components/Api/Resource/Variant.php
@@ -467,7 +467,7 @@ class Variant extends Resource implements BatchInterface
             $connection = Shopware()->Container()->get('dbal_connection');
 
             // Number changed, hence make sure it does not already exist in another variant
-            $exists = $connection->fetchColumn('SELECT id FROM s_articles_details WHERE ordernumber = ? AND articleID != ?', [$data['number'],$article->getId()]);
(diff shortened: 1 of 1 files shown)


janbuecker:
Hi @Gugiman & thanks for your PR.

Can you describe a little bit further, what your change should fix? Can you provide some example code or request and tell what's wrong with the current implementation?

In addition, the continuous build ist broken.

Gugiman:
I think it is better to close this PR for now because maybe I was a bit too fast opening it ;)

giginos:
@janbuecker : Maybe i can describe:

Imagine you like to update an variant-article and like to change the main-variant to a existing variant.

You would pass the new data (only the new main-variant is set) to the update-method in engine/Shopware/Components/Api/Resource/Article.php.

This method does load the old "mainDetail" via DQL-Command. Later the prepareMainVariantData in  engine/Shopware/Components/Api/Resource/Variant.php is called - This method is actually only a wrapper for prepareData(). 

You will end up in the "A variant with the given order number "XYZ" already exists."-Exception, because the method does get the id from old mainVariant and the new main variant is already an "normal" variant of the article.

I'm not sure, if it must be fixed in prepareData() or if it is required to implement a different logic in prepareMainVariantData() instead.

Can you follow my description?

Regards,
Rafael

PR #902, opened 2017-01-11
Added user specific info to sCONFIRMPASSWORDCHANGE mail

shyim:
## Description
* Why is it necessary?
It is currently not possible to user firstname or other fields in mail template
* What does it improve?
Adds the fields
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     |

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Account.php
+++ b/engine/Shopware/Controllers/Frontend/Account.php
@@ -474,6 +474,34 @@ class Shopware_Controllers_Frontend_Account extends Enlight_Controller_Action
             'sKey'      => $hash
         );
 
+        $sql = 'SELECT 
+          s_user.accountmode,
+          s_user.active,
+          s_user.affiliate,
+          s_user.birthday,
+          s_user.confirmationkey,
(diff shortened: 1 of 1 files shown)


OliverSkroblin:
Hi @shyim,

thanks for the pull request. 
Could you please change the following things:
- Don't use the user_addresses
- Customer names are stored (address refactoring) in s_user
- Only select a whitelist of s_user fields (not password, encoder, ...)
- If you want to select attributes too, please consider ambiguous column selection. s_user.id and s_user_attributes.id is duplicated. 
- Please store the user_attributes inside a sub array `attribtues`

shyim:
What would you think if I get the customer model convert it as array and remove some fields like password. So we would have anything about the user in mail context

janbuecker:
I'd prefer a whitelist instead of a blacklist.

shyim:
Aye Aye Captain its updated :)

mitelg:
Thanks again @shyim :+1:

PR #903, opened 2017-01-12
fix common attribute translation migration helper

resslinger:
escape the sql insert statements

## Description
* Why is it necessary? On the Showpare update, it caused an mysql error, because the sql query is not escaped.
* What does it improve? The migriation prozess.
* Does it have side effects? No.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | http://issues.shopware.com/issues/SW-17440.
| How to test?     | Migrate translations with ' in column objectdata.

### DIFF
--- a/_sql/migrations/common/AttributeTranslationMigrationHelper.php
+++ b/_sql/migrations/common/AttributeTranslationMigrationHelper.php
@@ -81,9 +81,12 @@ EOL
             $ids = array_keys($values);
 
             $this->connection->exec('DELETE FROM s_core_translations WHERE id IN (' . implode(',', $ids) . ')');
-            $this->connection->exec(
-                'INSERT INTO s_core_translations (objecttype, objectdata, objectkey, objectlanguage) VALUES ' . implode(',', $values)
-            );
+			
+			foreach($values as $id => $value)
+			{
+				$query = $this->connection->prepare('INSERT INTO s_core_translations (objecttype, objectdata, objectkey, objectlanguage) VALUES(:objecttype, :objectdata, :objectkey, :objectlanguage)');
+				$query->execute([':objecttype' => $value['objecttype'], ':objectdata' => $value['objectdata'], ':objectkey' => $value['objectkey'], ':objectlanguage' => $value['objectlanguage']]);
+			}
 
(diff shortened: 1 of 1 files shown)


bcremer:
Hi @k-30, thanks for your PR and for submitting the issue. We will escape the values and are not going to use prepared statements in this migration for performance reasons. The multi-insert is much faster then separate prepared statements.

```php
            $values = array_map(function($value) {
               return $this->connection->quote($value);
            }, $values);
```

Can you confirm that the above lines fix the issue for you?

resslinger:
Hi @bcremer ,

it doesn't work well:
Could not apply migration (Migrations_Migration799). Error: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''(\'article\', \'a:6:{s:10:\"txtArtikel\";s:50:\"Blaklader work pants trousers c' at line 1

soebbing:
Please try this new function (replaces the old one):

  ```
  private function getUpdatedTranslations($rows, $columns)
    {
        $values = [];

        foreach ($rows as $row) {
            try {
                $updated = $this->filter(unserialize($row['objectdata']), $columns);
            } catch (Exception $e) {
                //serialize error - continue with next translation
                continue;
            }

            if ($updated === null) {
                continue;
            }

            $row = array_map(function ($value) {
                return $this->connection->quote($value);
            }, $row);

            $updated = $this->connection->quote(serialize($updated));

            $values[$row['id']] = "({$row['objecttype']}, {$updated}, {$row['objectkey']}, {$row['objectlanguage']})";
        }

        return $values;
    }
```

resslinger:
The same problem...

soebbing:
That's interesting, that function works for our tests. Could you please make a dump of your s_core_translations-table and link it here so we can analyse the problem?

resslinger:
Hi @soebbing ,

the problem was solved, something was wrong in the dev environment. after a backup, your fix works now.

bcremer:
@k-30  Thanks again for helping us fixing this issue :+1:

PR #904, opened 2017-01-12
fixed the entity creation inside the acl component

jochenmanz:
## Description
* Why is it necessary?
I recently tried to remove and add resources. Unfortunatly after the second createResource call the whole thing fired an exception:

  [Doctrine\ORM\ORMInvalidArgumentException]
  A new entity was found through the relationship 'Shopware\Models\User\Privilege#resource' that was not configured to cascade persist operations for entity: Shopware\Models\User\Resource@00000000366
  c554a000000006ab7bc45. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOn
  e(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'Shopware\Models\User\Resource#__toString()' to get a clue.

* What does it improve?
reworked the resource creation and deletion to create proper relationships between the two models. 
* Does it have side effects?
ACL Resource functions as it is. 

TBD: The createPrivilege function can be removed and the PR needs review if thats the best way to solve the issue

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | none
| How to test?     | remove my changes to the acl component and rerun attached test, it should break without

### DIFF
--- a/engine/Shopware/Components/Acl.php
+++ b/engine/Shopware/Components/Acl.php
@@ -164,6 +164,8 @@ class Shopware_Components_Acl extends Zend_Acl
         $resource->setPluginId($pluginID);
 
         if (!empty($privileges)) {
+            $privilegeObjects = [];
+
             foreach ($privileges as $name) {
                 $privilege = new \Shopware\Models\User\Privilege();
                 $privilege->setName($name);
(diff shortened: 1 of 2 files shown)


mitelg:
thanks @jochenmanz :+1:

larsbo:
maybe a related issue https://issues.shopware.com/issues/SW-16480?

bcremer:
@larsbo good catch, thanks.

PR #907, opened 2017-01-16
Fix IDE warning for wrong return type in InstallContext

Haehnchen:
| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes


bcremer:
I think the method type is correct, the properties type has to be fixed to meixed as the array values may be a string (`scheduleMessage()`) or an array (`scheduleClearCache()`).

bcremer:
Fixed with https://github.com/shopware/shopware/commit/629db3229ea4d08e6a29fe5367417a4ab940972f.

PR #910, opened 2017-01-17
Accept urls with and without trailing slash

ahmad-saad:
### **Why is it necessary?**
if the url saved in s_core_rewrite_urls table with trailing slash for example 'womens/' then if I call it with 'womens' without slash it not find the page and redirect me to the homepage.

and the problem that the code generate the urls sometimes with trailing slash (like the categories) and sometimes without trailing slash (like the products or static pages).

so I can't do anything with .htacess to add or remove the trailing slash from the urls.

### **What does it improve?**
after this change the code will search for the page name und not the exact value with trailing slash. and that will be more senses.

No

| Questions        | Answers
| BC breaks?       |  no
| Tests pass?      |  yes
| Related tickets? | [SW-17514](https://issues.shopware.com/issues/SW-17514)
| How to test?     | simply call the Shopware demo page and then remove or add the trailing slash for example : http://www.shopwaredemo.de/beach-relax/fashion/women/ and http://www.shopwaredemo.de/beach-relax/fashion/women

### DIFF
--- a/engine/Shopware/Components/Routing/Matchers/RewriteMatcher.php
+++ b/engine/Shopware/Components/Routing/Matchers/RewriteMatcher.php
@@ -75,6 +75,7 @@ class RewriteMatcher implements MatcherInterface
           SELECT subshopID as shopId, path, org_path as orgPath, main
           FROM s_core_rewrite_urls
           WHERE path LIKE :pathInfo
+             OR Path LIKE :pathInfoWithoutslach
           ORDER BY main DESC, subshopID = :shopId DESC
           LIMIT 1
         ';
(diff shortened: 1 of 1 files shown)


ahmad-saad:
Hallo @shopwareBot ,
thank's for accepting this pull request.

but I have Q, why the check ist failed. my pull request always failed with php7.1 check.

can you tell me why this happened.

thank's,
Ahamd

bcremer:
@ahmad-saad thanks for your contribution. The failing test is not related to your PR. Restarted the build and now it's green again :)

janbuecker:
Sorry, but this PR has been reverted because of potential duplicate content.

dnoegel:
@janbuecker Shouldn't it actually do a proper redirect to the canonical URL?

janbuecker:
Unfortunately, is doesn't. In addition, there are missing canonical URLs on some pages e.g. forms.

So, the proposed change would introduce more problems than it solves.

ahmad-saad:
so , I think the best solution is use the same way to create the rewrite urls ( with or without trailing slash).
then use .htaccess to redirect to the correct url format ( with or without).

PR #912, opened 2017-01-17
[BUGFIX] check client IP against REMOTE_ADDR rather than SERVER_ADDR …

philippmahlow:
…on BAN Request

BAN requests for cache invalidation fail because of a comparison between client IP and SERVER_ADDR. In almost all average hosting environments SERVER_ADDR will be the internal server IP. But the client IP will be the external IP (like REMOTE_ADDR). Shopware will answer that request with 504, because the internal IP is in most cases not equal to the external IP. This causes exceptions on cache invalidation BAN requests "core.ERROR: Reverse proxy returned invalid status code". 

We have modified this check to match average hosting environments by checking REMOTE_ADDR and not SERVER_ADDR.

See also: https://forum.shopware.com/discussion/40903/reverse-proxy-error

### DIFF
--- a/engine/Shopware/Components/HttpCache/AppCache.php
+++ b/engine/Shopware/Components/HttpCache/AppCache.php
@@ -287,8 +287,8 @@ class AppCache extends HttpCache
      */
     protected function isPurgeRequestAllowed(Request $request)
     {
-        if ($request->server->has('SERVER_ADDR')) {
-            if ($request->server->get('SERVER_ADDR') == $request->getClientIp()) {
+        if ($request->server->has('REMOTE_ADDR')) {
+            if ($request->server->get('REMOTE_ADDR') == $request->getClientIp()) {
                 return true;
             }
         }


dnoegel:
Hi,

are you looking for the trusted proxy configuration: https://developers.shopware.com/developers-guide/http-cache/#trusted-proxies? IIRC this should take care of that issue. Or did I miss something?

philippmahlow:
Hi Daniel,

thank you for your reply. You are right, we could also use the trustedproxies-config. But don't you think it would be easier for users to use REMOTE_ADDR?

This is relevant for normal shared hosting environments. 

As described in your docs trustedproxies should be used for bigger environments (with Varnish, Loadbalancers and stuff).
"Another topic that is quite relevant for you, if you are working with bigger environments, is the trustedproxies configuration."
"It does not affect the cache directly, but as soon as you have e.g. a Varnish and a load balancer in play, you will need to deal with the fact that your IP address is replaced with e.g. the proxy's IP address."

We have seen this issue in various "normal" setups at Mittwald and found some other examples in the community forum.

bcremer:
I am not sure if I understand your problem. If the PURGE request ist issued from the same appserver as the inbuilt http cache runs on, it will be allowed here: https://github.com/shopware/shopware/blob/v5.2.14/engine/Shopware/Components/HttpCache/AppCache.php#L296

Using your modified condition **everybody** would be allowed to send a BAN requests to the shop and empty the cache. This would be a huge attack vector for DOSing the shop.

If you absolutely insist in this behavior you can add the following to your `config.php`:

```
    'httpcache' => [
        'purge_allowed_ips' => [
            '127.0.0.1',
            '::1',
            $_SERVER['REMOTE_ADDR'],
        ],
    ],
```

Can you confirm this results in the desired (insecure) behavior for you.

philippmahlow:
Hi Benjamin,
we have figured out that this issue is solved by simply update shopware to 5.2.14. I don't know exactly why, but its not a problem anymore. 

Thanks for your time - this is closed.

dnoegel:
Oh, I mixed up purge_allowed_ips and trusted_proxies. Thanks @bcremer

PR #914, opened 2017-01-19
Media converter fix

MFCBS:
## Description
* Why is it necessary?
Under PHP 7 if NULL is passed to convertMediaStruct, e.g. if the image is missing, an unrecoverable fatal error is thrown. If this happens inside digital publishing, it can cause whole shopping worlds to fail to load. The first line of the function checks that $media is an instance of StoreFrontBundle\Struct\Media meaning the type declaration isn't strictly necessary.

* What does it improve?
Fixes a bug.

* Does it have side effects?
IDE auto-complete.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | no
| Related tickets? | n/a
| How to test?     | Set up a digital publishing banner with a broken image, migrate from local image storage to s3 then try and load the digital publishing banner

### DIFF
--- a/engine/Shopware/Components/Compatibility/LegacyStructConverter.php
+++ b/engine/Shopware/Components/Compatibility/LegacyStructConverter.php
@@ -659,10 +659,10 @@ class LegacyStructConverter
     }
 
     /**
-     * @param StoreFrontBundle\Struct\Media $media
+     * @param $media
      * @return array
      */
-    public function convertMediaStruct(StoreFrontBundle\Struct\Media $media)
+    public function convertMediaStruct($media)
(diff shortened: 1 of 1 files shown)


janbuecker:
Hey @MFCBS and thanks for your contribution.

I think we should declare the parameter as nullable so we can get rid of the duck typing and stay consistent with other methods.

Could you change and test it, please?

mitelg:
Hey @MFCBS,
we fixed the problem in our Digital Publishing plugin, as we don't want to change die typehint in the media converter. 
Also fixed our storytelling plugin. 

Best regards

PR #915, opened 2017-01-19
Don’t round voucher net values in basket

samuelvogel:
## Description
Before this PR the net prices of vouchers were rounded to three decimal places in the basket. Articles in contrast have an exact net price, where the precisions is just limited by PHP itself.

This might seem like a minor thing, as only prices rounded to two decimal places are displayed to the customer. However this breaks the [PAYONE plugin](http://store.shopware.com/payone00261/payone-payment-fuer-shopware.html) for certain baskets which calculates the gross value itself. The calculation itself is correct but is off by one cent with some constellations because of the missing precisions of the vouchers net value and PHPs imperfect internal float handling.

Example basket:

| Position        | Quantity           | Value  |
| ------------- |:-------------:| -----:|
| Sample Article      | 1 | 112,50 € |
| Voucher 5 %      |      |   -5,63 €|
|      |     **SUM** |   106,87 €|

But the PAYONE plugin takes the rounded net value of the voucher `-5.257` and recomputes it to `-5.62499` which is then then rounded to the incorrect value of `-5.62` (off by 0,01 €). The code doing this can be found [here](https://github.com/PAYONE-GmbH/shopware-5/blob/v3.4.7/Frontend/MoptPaymentPayone/Components/Classes/PayoneParamBuilder.php#L1046) and is correct, it's just not fed with enough precision in case of vouchers.

Orders like this are rejected by the PAYONE API as the summed basket value displayed to the user is transmitted as well and cross checked against the sum of the individual positions. The error returned is `1610: Article list faulty or incomplete` in this case.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | -
| How to test?     | Test case is included in the PR.

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2017,14 +2017,14 @@ class sBasket
             }
         } else {
             if ($voucherDetails["taxconfig"] == "default" || empty($voucherDetails["taxconfig"])) {
-                $tax = round($voucherDetails["value"] / (100 + $this->config->get('sVOUCHERTAX')) * 100, 3) * -1;
+                $tax = $voucherDetails["value"] / (100 + $this->config->get('sVOUCHERTAX')) * 100 * -1;
                 $taxRate = $this->config->get('sVOUCHERTAX');
                 // Pre 3.5.4 behaviour
             } elseif ($voucherDetails["taxconfig"] == "auto") {
                 // Check max. used tax-rate from basket
                 $tax = $this->getMaxTax();
                 $taxRate = $tax;
-                $tax = round($voucherDetails["value"] / (100 + $tax) * 100, 3) * -1;
+                $tax = $voucherDetails["value"] / (100 + $tax) * 100 * -1;
             } elseif (intval($voucherDetails["taxconfig"])) {
                 // Fix defined tax
                 $temporaryTax = $voucherDetails["taxconfig"];
@@ -2033,7 +2033,7 @@ class sBasket
                     array($temporaryTax)
                 );
                 $taxRate = $getTaxRate;
-                $tax = round($voucherDetails["value"] / (100 + (intval($getTaxRate))) * 100, 3) * -1;
+                $tax = $voucherDetails["value"] / (100 + (intval($getTaxRate))) * 100 * -1;
             } else {
                 // No tax
                 $tax = $voucherDetails["value"] * -1;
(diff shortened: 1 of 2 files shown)


htkassner:
Hi @samuelvogel. I evaluated this one again. For my case I used the above described scenario. If we would save the not rounded value, which would be 4.7268907563025 for this scenario and this value runs through the payone code, I would still get 5.62 which would not be the by the basket calculated value. (round(4.7268907563025 * 1.19, 2)). So maybe your example is incomplete or I miss out a point. But this does not seem to be the solution for the payone calculation. Can you evaluate your example again?

samuelvogel:
But you used 19 % VAT 😉

The example posted above is based on 7 % VAT (which I should have mentioned obviously), where you can see that the problem really does occur (it's taken from a real world example at https://shop.rapunzel.de BTW).

Don't ask me why I did not use the real world example in the test case itself, I think I wanted to keep it simple without changing the voucher tax config.

htkassner:
That is true. I used default 19% calculation. But shouldn´t this be also compatible with 19% voucher calculation then? I guess we really have to be very careful here to not close one issue and open 3 new ones, because now 19% calculation would be failing. But thank you anyways for your contribution!

samuelvogel:
Your corner case with 19 % is caused by a PHP floating point precisions issue, see the following PHP 7 output:
```
php > echo 4.7268907563025 * 1.19;
5.625
php > echo round(5.625, 2);
5.63
php > echo round(4.7268907563025 * 1.19, 2);
5.62
```
This only happens when you copy the `4.7268907563025` value to your clipboard and then again paste it into a file or the command line. Internally PHP works with higher precisions than when outputting, see [here](http://php.net/manual/en/language.types.float.php). So in the basket the issue you mentioned will not occur, which you can see here (screenshot taken from Shopware 5.2 `master` just with the commit from this PR and a 19 % voucher):
screenshot 2017-07-11 16 28 40

The PR adds more precisions for all cases and removes the random rounding to 3 decimal places which accidentally leads to a different value in your manual example.

I've just resent the PR against Shopware 5.3 (just some code style changes): #1204

samuelvogel:
As you merged most of our other basket improvements for Shopware 5.3 already it would be great to include this one as well, especially as the next good opportunity (a big release like 5.3) will probably still be some time off.

htkassner:
That is the reason why we are again looking at this PR. We also want to improve our code. I also could not yet find an issue in shopware´s calculated basket, I recognized that leave out the early 3 digits rounding of the voucher "seems" to not affect the final result, but we are always very carefully in evaluating basket changes. 
I was not aware of the fact that c&p the float value, leads to lose of precision. 

We´re still evaluating and maybe this can be merged too. But let us test all the cases first.

PR #916, opened 2017-01-20
Make all attributes from s_articles_attributes available to productfeed

hanneswernery:
## Description
As for now, only attr1-attr20 are added to the SQL query and therefore are available to a productfeed. This PR adds all non-id fields of s_articles_attributes to the SQL query. That way custom attributes from plugins are available as well.
No other fields of the query are overwritten, attr1-attr20 are accessed the same way as before `{sArticle.attr1} `, plugin attributes are accessed via `{sArticle.["attribute.someplugin_someattribute"]} `

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
|Related tickets?| -
|How to test?| Add a custom attribute of any installed plugin to any template of a ProductFeed `{sArticle.["attribute.someplugin_someattribute"]} `and export it. It now appears in the export document.

### DIFF
--- a/engine/Shopware/Core/sExport.php
+++ b/engine/Shopware/Core/sExport.php
@@ -750,6 +750,7 @@ class sExport
         $sql_add_join   = array();
         $sql_add_select = array();
         $sql_add_where  = array();
+        $sql_select_attributes = array();
 
         $skipBackend = $this->shop->get('skipbackend');
         $isoCode = $this->shop->get('isocode');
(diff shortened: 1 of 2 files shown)


engine/Shopware/Core/sExport.php
```diff
@@ -837,6 +838,18 @@ public function sCreateSql()
...
         }
 
+        $tableColumns = Shopware()->Container()->get('models')->getConnection()->getSchemaManager()->listTableDetails('s_articles_attributes')->getColumns();
+        foreach ($tableColumns as $column) {
+            $columnName = $column->getName();
+            if (
+                $columnName !== 'id'
+                && $columnName !== 'articleID'
+                && $columnName !== 'articledetailsID') {
+                $sql_select_attributes[] = 'at.' . $columnName;
```
OliverSkroblin: please prefix all attribute columns of plugins, to prevent ambiguous column exception

```
if ($name = 'attr*') {
  //no prefix
} else {
  //with prefix
}

```

samuelvogel: Yes, we will update the PR and prefix all non-standard attributes. We initially thought this is not needed, since for the method `ModelManager::addAttribute()`, the parameter `$prefix` is mandatory, so conflicts should be almost impossible.

But we already noticed that arbitrary identifiers without any specific prefix can be chosen through the new attribute administration of Shopware 5.2.

hanneswernery: I prefixed the plugin attributes in 36e410c0.

shyim: We need also "at.attr1, at.attr2, ...." as select without it it would be a bc break :)
Old product exports with "{sArticle.attr1}" would be empty :/

samuelvogel: `at.attr1-20` are still selected in the old format, to avoid a BC break only other attributes are selected with a prefix.

OliverSkroblin:
Hi @hanneswernery,
i scheduled because the attribute translations doesn't work for own properties.

PR #917, opened 2017-01-20
Allow listing limit only on available values

shyim:
## Description
* Why is it necessary?
I can go to very big shops with big categories and call the listing page with 99999 limit and use useless cpu and space(http-caching)
* What does it improve?
Sets the default listing limit, if limit value is not available
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | Use limit 9999 or something else in listing page

### DIFF
--- a/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
+++ b/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
@@ -340,8 +340,14 @@ class CoreCriteriaRequestHandler implements CriteriaRequestHandlerInterface
      */
     private function addLimit(Request $request, Criteria $criteria)
     {
+        $pageSizes = explode("|", $this->config->get('numberArticlesToShow'));
         $limit = (int) $request->getParam('sPerPage', $this->config->get('articlesPerPage'));
         $limit = $limit >= 1 ? $limit: 1;
+
+        if (!in_array($limit, $pageSizes)) {
+            $limit = (int)$this->config->get('articlesPerPage');
+        }
+
         $criteria->limit($limit);
     }
 
--- a/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
+++ b/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
@@ -344,7 +344,7 @@ class CoreCriteriaRequestHandler implements CriteriaRequestHandlerInterface
         $limit = (int) $request->getParam('sPerPage', $this->config->get('articlesPerPage'));
         $limit = $limit >= 1 ? $limit: 1;
 
-        if (!in_array($limit, $pageSizes) && $request->getControllerName() == 'listing' && $request->getModuleName() == 'frontend') {
+        if (!in_array($limit, $pageSizes) && in_array($request->getControllerName(), ['listing', 'search']) && $request->getModuleName() == 'frontend') {
             $limit = (int)$this->config->get('articlesPerPage');
         }
 
--- a/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
+++ b/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
@@ -344,7 +344,7 @@ class CoreCriteriaRequestHandler implements CriteriaRequestHandlerInterface
         $limit = (int) $request->getParam('sPerPage', $this->config->get('articlesPerPage'));
         $limit = $limit >= 1 ? $limit: 1;
 
-        if (!in_array($limit, $pageSizes)) {
+        if (!in_array($limit, $pageSizes) && $request->getControllerName() == 'listing' && $request->getModuleName() == 'frontend') {
             $limit = (int)$this->config->get('articlesPerPage');
         }
 


engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
```diff
@@ -340,8 +340,14 @@ private function addOffset(Request $request, Criteria $criteria)
      */
     private function addLimit(Request $request, Criteria $criteria)
     {
+        $pageSizes = explode("|", $this->config->get('numberArticlesToShow'));
         $limit = (int) $request->getParam('sPerPage', $this->config->get('articlesPerPage'));
         $limit = $limit >= 1 ? $limit: 1;
+
+        if (!in_array($limit, $pageSizes)) {
```
OliverSkroblin: Make sense for the normal listing, but if someone implements a slider or an own listing with custom limit, it would be broken with this change. 
The limit validation could be moved to the listing controller.

shyim: Good argument. I added a comment to prevent that. I couldn't move the logic to listing controller, because my value would be overwritten by `CoreCriteraRequestHandler` `handleRequest`

engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
```diff
@@ -340,8 +340,14 @@ private function addOffset(Request $request, Criteria $criteria)
      */
     private function addLimit(Request $request, Criteria $criteria)
     {
+        $pageSizes = explode("|", $this->config->get('numberArticlesToShow'));
         $limit = (int) $request->getParam('sPerPage', $this->config->get('articlesPerPage'));
         $limit = $limit >= 1 ? $limit: 1;
+
+        if (!in_array($limit, $pageSizes) && $request->getControllerName() == 'listing' && $request->getModuleName() == 'frontend') {
```
OliverSkroblin: looks okay for me, i will test it at monday ;)

dnoegel: Should this also be available for the "search" controller?

OliverSkroblin: Good hint

shyim: I have extended it :)

aragon999:
Maybe this problem has to be addressed in a more general way. Since also problems can occoure when passing the parameter max={581.00} with any possible price value this query: https://github.com/shopware/shopware/blob/27507fe404cede0b75ad561fc081ce4fc9554569/engine/Shopware/Bundle/SearchBundleDBAL/FacetHandler/PriceFacetHandler.php#L134  is executed. 
In our shop installation this is called with the root category id (3 in our case).
In larger shop installations without elastic search this query can take over 200s and the query is executed twice.
I am still debugging the problem, since in the installation this problem occours randomly and we have not figured out where exactly this problem comes from.

aragon999:
I have debugged the problem a bit further and the problem in our case was, that there existed a modified `robots.txt` which did not disallowed `/listing` so we had requests in the following way from the BingBot:
```json
"URI": "/listing/?p=1&s=124",
"Params": {
    "p":"1",
    "s":"124",
    "module":"frontend",
    "controller":"listing",
    "action":"index",
    "sPage":"1",
    "sSupplier":"124",
    "sCategory":3
},
"Referer": false,
"User Agent": "Mozilla/5.0 (compatible; bingbot/2.0; +http://www.bing.com/bingbot.htm)"
```
which then triggered the query mentioned in the previous comment on the whole shop. The query looks like the following: 
```sql
SELECT Min(Round(defaultPrice.price * ( ( 100 - Ifnull(priceGroup.discount, 0) ) 
                                        / 100 
                                             ) * 
                            ( ( ( 
                                  CASE tax.id 
                                  WHEN 1 THEN 19 
                                  WHEN 4 THEN 7 
                                  WHEN 5 THEN 0 
                                  end ) + 100 ) / 100 ) * 1, 2)) AS 
       cheapest_price 
FROM   s_articles product 
       INNER JOIN s_articles_details variant 
               ON variant.id = product.main_detail_id 
                  AND variant.active = 1 
                  AND product.active = 1 
       INNER JOIN s_core_tax tax 
               ON tax.id = product.taxid 
       INNER JOIN s_articles_categories_ro productCategory16 
               ON productCategory16.articleid = product.id 
                  AND productCategory16.categoryid IN ( 3 ) 
       LEFT JOIN s_articles_avoid_customergroups avoidCustomerGroup 
              ON avoidCustomerGroup.articleid = product.id 
                 AND avoidCustomerGroup.customergroupid IN ( 1 ) 
       INNER JOIN s_articles_details availableVariant 
               ON availableVariant.articleid = product.id 
                  AND availableVariant.active = 1 
       INNER JOIN s_articles_prices defaultPrice 
               ON defaultPrice.articledetailsid = availableVariant.id 
                  AND defaultPrice.pricegroup = 'EK' 
                  AND defaultPrice.from = 1 
       LEFT JOIN s_articles_prices customerPrice 
              ON customerPrice.articleid = product.id 
                 AND customerPrice.pricegroup = 'EK' 
                 AND customerPrice.from = 1 
                 AND availableVariant.id = customerPrice.articledetailsid 
       LEFT JOIN s_core_pricegroups_discounts priceGroup 
              ON priceGroup.groupid = product.pricegroupid 
                 AND priceGroup.discountstart = 1 
                 AND priceGroup.customergroupid = '1' 
                 AND product.pricegroupactive = 1 
       INNER JOIN s_articles_attributes productAttribute 
               ON productAttribute.articledetailsid = variant.id 
WHERE  avoidCustomerGroup.articleid IS NULL 
GROUP  BY product.id 
ORDER  BY cheapest_price DESC 
LIMIT  1 offset 0
```
Though there was nothing about the supplier in the query, I am not sure if this is another bug or if this is wanted, I am still debugging this issue further. The shop is running Shopware 5.1.6.

shyim:
Fixed with https://github.com/shopware/shopware/commit/284c657d954d34ea9da65e8399c299d6b651acbb

PR #921, opened 2017-01-24
fix breadcrumb-items without links

screeny05:
This PR prevents breadcrumb-items without link from being rendered with an `<a>`-element.

Instead items without link get rendered as spans. This doesn't break styling or semantics.

## Description
* Why is it necessary? Prevent customer confusion from "broken" links
* What does it improve? UX

| Questions        | Answers
| BC breaks?       | yes/no
| Tests pass?      | yes/no

### DIFF
--- a/themes/Frontend/Bare/frontend/index/breadcrumb.tpl
+++ b/themes/Frontend/Bare/frontend/index/breadcrumb.tpl
@@ -9,10 +9,16 @@
                 <li class="breadcrumb--entry{if $breadcrumb@last} is--active{/if}" itemprop="itemListElement" itemscope itemtype="http://schema.org/ListItem">
                     {if $breadcrumb.name}
                         {block name="frontend_index_breadcrumb_entry_inner"}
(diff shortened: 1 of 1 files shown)


janbuecker:
@screeny05 can you please provide a test scenario?

screeny05:
@janbuecker afaik there's none in the default shopware-theme. My particular case is, that the article-name should be displayed in the breadcrumb, when the user is on the detail-page. Also it shouldn't be clickable, as it's obvious to the user that this is the active page. So i added the following code to the detail-template:

```smarty
{block name='frontend_index_start'}
    {$smarty.block.parent}
    {$sBreadcrumb[] = ['name'=> $sArticle.articlename]}
{/block}
```

Notice the lack of the `link`-property.

This would cause shopware to render the breadcrumb as `<a href="#">...</a>`, which is an unnecessary link.

janbuecker:
Oh, this has already been merged with 1d3cb76b7d824a3b793b23e41c3e130f2c966f63.

Closing :)

screeny05:
That's funny - thanks! 👍

PR #925, opened 2017-01-25
Check if javascript files exist before compiling

screeny05:
This PR adds a check to see if all given javascript-files exist before compiling.

This check only runs in development.

The check runs in under 200ms on my machine which is nothing compared to the compilation of the assets.

## Description
* Why is it necessary?
Uglify doesn't check whether or not the given files exist, so if there's a spelling error in someones Theme.php they're left to wonder why e.g. their javascript-plugin is not available.

* What does it improve?
Developer experience

* Does it have side effects?
No

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| How to test?     | run `grunt`

### DIFF
--- a/themes/Gruntfile.js
+++ b/themes/Gruntfile.js
@@ -10,12 +10,12 @@ module.exports = function (grunt) {
 
     lessTargetFile['../' + config.lessTarget] = '../web/cache/all.less';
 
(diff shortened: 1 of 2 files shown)


DennisGarding:
Hello,

thanks for your contribution!

The branch is merged.

PR #929, opened 2017-01-25
Added payment attributes

shyim:
## Description
* Why is it necessary?
Work with attributes in frontend
* What does it improve?
Adds attribute
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | debug vars

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/Hydrator/PaymentHydrator.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/Hydrator/PaymentHydrator.php
@@ -1,0 +1,89 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Bundle\StoreFrontBundle\Gateway\DBAL\Hydrator;
+
+use Shopware\Bundle\StoreFrontBundle\Struct;
+
+/**
+ * @category  Shopware
+ * @package   Shopware\Bundle\StoreFrontBundle\Gateway\DBAL\Hydrator
+ * @copyright Copyright (c) shopware AG (http://www.shopware.de)
+ */
+class PaymentHydrator extends Hydrator
+{
+    /**
+     * @var AttributeHydrator
+     */
+    private $attributeHydrator;
+
+    /**
+     * PaymentHydrator constructor.
+     *
+     * @param AttributeHydrator $attributeHydrator
+     */
+    public function __construct(AttributeHydrator $attributeHydrator)
+    {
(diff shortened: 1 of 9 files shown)


shyim:
@janbuecker again restart :o

engine/Shopware/Core/sAdmin.php
```diff
@@ -426,6 +430,10 @@ public function sGetPaymentMeans()
             array('subject' => $this)
         );
 
+        foreach ($getPaymentMeans as &$getPaymentMean) {
```
OliverSkroblin: Hi @shyim,
store front attributes are a little bit tricky ;).
To implement store front attributes you have to consider the following things:
- Added leftJoin query to initial data query (in this case: https://github.com/shopware/shopware/pull/929/files#diff-0a0564f026f3b27a8946fc6e96862a4bR208)
- Consider translations for all attributes which are stored differently. As own s_core_translation entry with table name as object key or inside the main entity translation like `article` or `variant`.
- Don't execute sql queries in loops to prevent unnecessary server load

My solution here would be to implement a new service and gateway for payments inside the store front bundle and load data there. The FieldHelper and attribute hydrator has different helper function to simplify attribute loading.

If you have fun and time you can implement this to ;)

Service and gateway should only have a "getList" function with $ids and ShopContextInerface parameter, like the other gateways =)

shyim:
@OliverSkroblin I have built now a PaymentStruct, Hydrator and Interface like the other in StoreFront :). I moved also the logic from sAdmin to the Service.

Have you a idea why the last test fails?

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/Hydrator/PaymentHydrator.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * @param array $data
+     * @return Struct\Payment
+     */
+    public function hydrate(array $data)
+    {
+        $payment = new Struct\Payment();
+
+        $translation = $this->getTranslation($data, '__payment');
```
janbuecker: The translations are merged. You can get rid of the next few lines by adding additional parameters.

```php
$translation = $this->getTranslation($data, '__payment', [], $data['__payment_id']);
```

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/PaymentGateway.php
```diff
@@ -0,0 +1,178 @@
...
+        return array_shift($payments);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getList(array $paymentIds, Struct\ShopContextInterface $context, $fallback = false, $user = null)
+    {
+        $isMobile = ($this->front->Request()->getDeviceType() == 'mobile');
+        $sEsd = Shopware()->Modules()->Basket()->sCheckForESD();
```
janbuecker: You don't want the dependency of the sBasket or Front controller here. Please move the whole logic back into the core classes and simply select the data needed for the payment you provide via PaymentService which you need to implement.

engine/Shopware/Bundle/StoreFrontBundle/Gateway/PaymentGatewayInterface.php
```diff
@@ -0,0 +1,52 @@
...
+    public function getList(array $paymentIds, Struct\ShopContextInterface $context, $fallback = false, $user = null);
+
+    /**
+     * @param int $paymentId
+     * @param Struct\ShopContextInterface $context
+     * @param bool $fallback
+     * @param $user
+     * @return Struct\Payment
+     */
+    public function get($paymentId, Struct\ShopContextInterface $context, $fallback = false, $user = null);
```
janbuecker: Please do not implement a `get()` method. Only batch, only getList.

engine/Shopware/Core/sAdmin.php
```diff
@@ -1611,6 +1466,33 @@ private function convertToLegacyAddressArray(Address $address)
         return $output;
     }
 
+    private function convertToLegacyPaymentArray(Payment $payment)
```
janbuecker: Please move this method into the `LegacyStructConverter`

janbuecker (changes requested):

shyim:
@janbuecker changes are done. Travis is also green ❤️

PR #931, opened 2017-01-25
Added config to toggle backend menu hover

shyim:
## Description
* Why is it necessary?
Some customers hate the hovering of the menu in backend
* What does it improve?
Adds a config to disable it
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | Switch the config and clear the cache


janbuecker:
Thanks @shyim 

We've discussed this pull request internally and decided to not implement this feature as we do not want to have this behaviour changed.

shyim:
The default behaviour is hovering and the customers will not see a difference. I don't understand the reason :/

PR #932, opened 2017-01-25
Clear config cache after saving config forms

shyim:
## Description
* Why is it necessary?
I guess everyone know it which has done plugin support. The main question to the customer is have you cleared the cache after saving plugin configuration :)
* What does it improve?
Cleares automaticly after saving form
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | #526 last comment

### DIFF
--- a/engine/Shopware/Controllers/Backend/Config.php
+++ b/engine/Shopware/Controllers/Backend/Config.php
@@ -218,6 +218,8 @@ class Shopware_Controllers_Backend_Config extends Shopware_Controllers_Backend_E
             $this->saveElement($elementData, $defaultShop);
         }
 
+        $this->get('shopware.cache_manager')->clearConfigCache();
+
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Config.php
```diff
@@ -218,6 +218,8 @@ public function saveFormAction()
             $this->saveElement($elementData, $defaultShop);
         }
 
+        $this->get('shopware.cache_manager')->clearConfigCache();
```
OliverSkroblin, line 221: Hi @shyim,
we already implemented this in previous version but decided to remove the automatically cache clear to prevent unnecessary cache clear each time the user saves a form.
In case that the user configures multiple values inside the base settings, the cache will cleared multiple times.
For this reason we closed the pull request.

PR #934, opened 2017-01-30
remove baseprice column

jochenmanz:
## Description
* Why is it necessary?
The baseprice column was removed from the price model in 5.2. This is only a small cleanup PR.
* What does it improve?
Database size
* Does it have side effects?
none

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | none
| How to test?     | reinstall shopware from scratch or update an existing 5.2 installation. No baseprice column should be present in s_articles_prices


_sql/install/latest.sql
```diff
@@ -360,7 +360,6 @@ CREATE TABLE IF NOT EXISTS `s_articles_prices` (
   `articledetailsID` int(11) NOT NULL DEFAULT '0',
   `price` double NOT NULL DEFAULT '0',
   `pseudoprice` double DEFAULT NULL,
-  `baseprice` double DEFAULT NULL,
```
mitelg: don't change the install/latest.sql

mitelg (changes requested):

mitelg:
the builds are failing, because you changed the latest.sql :wink:

mitelg (approved):

_sql/migrations/815-remove-base-price-column.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+class Migrations_Migration815 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $statement = $this->connection->query('SELECT * FROM s_articles_prices');
```
shyim: show columns FROM s_articles_prices would be better here :)

shyim (changes requested):

dnoegel:
If we merge this, I'd suggest 5.3 due to the schema change.

samuelvogel:
Just for reference, why the field is currently still there: https://github.com/shopware/shopware/pull/527#issuecomment-218442872

mitelg:
@dnoegel yeah, we already discussed that. and we will do it on 5.3. IF we do it :wink: :smile:

PR #937, opened 2017-01-30
Add missing classes-block to content-main element

screeny05:
This PR adds a missing block to change the classes of the `.content-main`-element.

This provides an easy way to remove the `.container`-class from the element, which prevents the page from going full-width. Otherwise one would have to use javascript to remove the class or overwrite the whole containing block `frontend_index_content_main`. Both is cumbersome.

## Description
* Why is it necessary? Make creating full-width pages easier

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| How to test?     | In a custom theme, modify the `frontend_index_content_main_classes`-block

### DIFF
--- a/themes/Frontend/Bare/frontend/index/index.tpl
+++ b/themes/Frontend/Bare/frontend/index/index.tpl
@@ -82,7 +82,7 @@
             {/block}
 
             {block name='frontend_index_content_main'}
-                <section class="content-main container block-group">
+                <section class="{block name="frontend_index_content_main_classes"}content-main container block-group{/block}">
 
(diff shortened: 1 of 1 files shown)


klarstil:
Fixed https://github.com/shopware/shopware/commit/1e9668285dfafe1d1695aab710e6f2622c5fa41b

zetno (commented):
The default bare classes content-main and block-group are not working in this setup. You need to give additional space in front and after the classes.
"content-main container block-group" to " content-main container block-group "

screeny05:
@zetno What's the problem without spaces?

PR #940, opened 2017-01-30
Fixed plugin cronjob.xml boolean evaluation

nlubisch:
## Description
* Why is it necessary? if an boolean value equals integer 1 the value got evaluated as false
* What does it improve? boolean evaluation

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes/no
| How to test?     |  install plugin [LubischTest_cronjob_bool.zip](https://github.com/shopware/shopware/files/740228/LubischTest_cronjob_bool.zip) - the configured "disableOnError" with 1 gets evaluated and saved as 0.

Reference: https://www.w3.org/TR/xmlschema11-2/#boolean

### DIFF
--- a/engine/Shopware/Components/Plugin/XmlCronjobReader.php
+++ b/engine/Shopware/Components/Plugin/XmlCronjobReader.php
@@ -77,9 +77,9 @@ class XmlCronjobReader
 
         $cronjobEntry['name'] = $this->getFirstChild($entry, 'name');
         $cronjobEntry['action'] = $this->getFirstChild($entry, 'action');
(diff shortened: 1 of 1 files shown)


Haehnchen:
:thumbsup:

janbuecker:
Thanks @nlubisch, looks good to me. I'll schedule this change and do some testing on it. 👍

nlubisch:
Im working on some changes to the xml reader, like refactor all readers because they share the most functionality and would test it to near 100% code coverage. IMO It looks a bit nasty because each reader does their own thing.

mitelg:
thanks for your contribution @nlubisch :+1:

PR #941, opened 2017-01-30
Fix InvalidArgumentException if plugin services are empty

nlubisch:
## Description
* Why is it necessary? an empty plugin services.xml leads after installation to an InvalidArgumentException and the whole system gets "503 Service Unavailable" as shop owner there is no non technical way to get rid of that
* What does it improve? the shop is not corrupted after plugin installation and the exception gets logged

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes/no
| How to test?     | install plugin [LubischTest_services_empty.zip](https://github.com/shopware/shopware/files/740292/LubischTest_services_empty.zip) and go to frontend or reload backend

### DIFF
--- a/engine/Shopware/Components/Plugin.php
+++ b/engine/Shopware/Components/Plugin.php
@@ -167,7 +167,11 @@ abstract class Plugin implements ContainerAwareInterface, SubscriberInterface
             new FileLocator()
         );
 
-        $loader->load($this->getPath().'/Resources/services.xml');
+        try {
+            $loader->load($this->getPath().'/Resources/services.xml');
+        } catch (\InvalidArgumentException $e) {
+            error_log($e);
+        }
     }
 
     /**


shyim:
I think its the developers fault and should not be catched. When a services.xml is empty it should be removed

bcremer:
I second the comment from @shyim. The new plugin system was built to make it very obvious to the plugin developer when things failed.

The legacy plugins system was known to be very forgiving to errors. But that made it very hard to debug.

janbuecker:
Thanks @nlubisch 

Like @shyim and @bcremer already said, the hard failing is by design and won't be changed.

nlubisch:
@bcremer @janbuecker Any chance in getting a validation in the plugin install process like the other xml files?
So that when a validation error occurs a GrowlNotification appears with "Parse error in file: ... your file"

PR #943, opened 2017-01-31
Make LegacyStructConverter use ContextServiceInterface instead of ContextService

DerRidda:
## Description
* Why is it necessary?

It is necessary to assure that the service "shopware_storefront.context_service" can be cleanly replaced or decorated by third party plugins.
Previously the constructor for the LegacyStructConverter had a type hint for ContextService which naturally causes an exception when Shopware tries to instantiate it with a replaced/decorated service that simply implements ContextServiceInterface.

As LegacyStructConverter doesn't seem to need the concrete implementation of ContextService and only uses the getShoptContext() method which is already guaranteed by the ContextServiceInterface this fix seems logical.

* What does it improve?

Makes it possible to decorate/replace "shopware_storefront.context_service" in third party plugins.

* Does it have side effects?

None that I know of, as LegacyStructConverter does not seem to depend on the specific implementation provided by ContextService.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | (Didn't appear to be covered/ I didn't run the whole test suite)
| Related tickets? | SW-15923
| How to test?     | As basic shop operation is still quite dependent on LegacyStructConverter regressions should appear immediately


DerRidda:
I actually found this ticket mentioning this precise issue: https://issues.shopware.com/issues/SW-15923

I updated the QA in the original message.

PR #946, opened 2017-01-31
Create attributes using attributes.xml

shyim:
## Description
* Why is it necessary?
Added a way to create plugin attributes with xml
* What does it improve?
plugin install/update simplification
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | 

Example File:

```xml
<?xml version="1.0" encoding="utf-8"?>
<tables xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="../../../../engine/Shopware/Components/Plugin/schema/attributes.xsd">

    <table name="s_articles_attributes">
        <attribute name="test" type="combobox">
            <label>I am a checkbox</label>
            <supportText>supportText</supportText>
            <helpText>helpText</helpText>
            <displayInBackend>true</displayInBackend>
            <translatable>true</translatable>
            <custom>true</custom>
            <updateDependingTables>true</updateDependingTables>
            <arrayStore>
                <option key="1">Yes</option>
                <option key="0">No</option>
            </arrayStore>
        </attribute>
    </table>
</tables>
```

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
@@ -27,12 +27,14 @@ namespace Shopware\Bundle\PluginInstallerBundle\Service;
 use Doctrine\DBAL\Connection;
 use Shopware\Components\Model\ModelManager;
 use Shopware\Components\Plugin as PluginBootstrap;
+use Shopware\Components\Plugin\AttributeSynchronizer;
 use Shopware\Components\Plugin\Context\ActivateContext;
 use Shopware\Components\Plugin\Context\DeactivateContext;
 use Shopware\Components\Plugin\Context\InstallContext;
 use Shopware\Components\Plugin\Context\UninstallContext;
 use Shopware\Components\Plugin\Context\UpdateContext;
 use Shopware\Components\Plugin\RequirementValidator;
+use Shopware\Components\Plugin\XmlAttributeReader;
 use Shopware\Components\Snippet\DatabaseHandler;
 use Shopware\Kernel;
 use Shopware\Models\Plugin\Plugin;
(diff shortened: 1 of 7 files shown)


nlubisch:
Would prefer the keyword "arrayStore" to be "store" to unify it with config.xml.

mitelg:
Hey @shyim 
as we have to test this properly, we will schedule the ticket for one of the next sprints

shyim:
@nlubisch Its unified

taltholtmann:
Hi @shyim 
we tested your implementation and found a problem which is not easy to solve. If you change the structure of your attributes.xml and then perform an update the old attributes and their data remains in the database. There is also no possibility for renaming your attribute tables and migrate the data.
In case of the other xml files, all the data is actually in the xml and the database rows are referenced to the plugin ids. So the cleanup there is simple.
According to this we have to close this pull request because we do not find a suitable solution for the problem.

shyim:
@taltholtmann Whats about a table which holds added attributes and from which plugin?

OliverSkroblin:
Hi @shyim,
we discussed the feature and found different problems. The most import problem is that some attributes should not be removed even if the plugin is uninstalled, like customer or order data. For that reason we want to keep the responsibility to the plugin developers.

shyim:
This could be solved by a attribute in config.xml (```<removeOnUninstall>false</removeOnUninstall>``` or with the secureUninstall question in backend

OliverSkroblin:
Would be possible, but we would need to many configuration parameters to solve all possible scenarios ;)
Other scenario would be a shopware version validation ( 5.2.16 => attribute table only exists in this version)

shyim:
Other scenarios could be solved by using crud service in install method. Attribute.xml should not solve all possible scenarios :P

FiveDigital:
I'm not a fan of how attributes are added in the Plugin file right now, too. This is too much overhead and code duplication. In my opinion the best way is to add an remove attributes with the migrations that have been announced quite a long time ago - ping @bcremer @dnoegel The migration should offer handy shortcut methods to work with attributes.

janbuecker:
@FiveDigital that's true, a migration system for plugins would solve many problems, but that is out of scope for this PR.

I've published a ticket for the plugin migration system where you can up-vote this feature.
https://issues.shopware.com/issues/SW-16207

buddhaCode:
would have loved such a feature :/

PR #947, opened 2017-02-01
emotion viewport connection improvment

kuborgh-mspindelhirn:
## Description
Changes the current emotion designer behavior on editing emotions with at least one element.
Not sure if changes like this would be applicable for 5.2 but as it changes a backend behavior I set the PR to 5.3 branch. Hope this is correct.

Old behavior:   
If an emotion contains at least one element do not automaticly connect viewports on load. 

New behavior:
Connect viewports that do not contain any hidden elements

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | Create new emotion; Create element; Unconnect Tablet; Add new element on tablet; Save and reopen emotion. Tablet is not connected but all other viewports are connected.

### DIFF
--- a/themes/Backend/ExtJs/backend/emotion/view/detail/designer.js
+++ b/themes/Backend/ExtJs/backend/emotion/view/detail/designer.js
@@ -267,14 +267,7 @@ Ext.define('Shopware.apps.Emotion.view.detail.Designer', {
 
     createGridView: function() {
         var me = this,
-            stateConnections = [ 'xl' ];
-
-        /**
-         * All viewports are initially connected when the emotion has no elements.
-         */
-        if (me.emotion.getElements().getCount() === 0) {
             stateConnections = [ 'xs', 's', 'm', 'l', 'xl' ];
-        }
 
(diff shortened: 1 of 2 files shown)


kuborgh-mspindelhirn:
Is there any plan when we can expect an update on this one? Would be nice to know if the way I change the behavior goes in the right direction, or if you have other plans.

kuborgh-mspindelhirn:
Any feedback on this one? @Phil23 @bcremer 
The current behavior feels really strange and it would be nice to know what your perspective is

Phil23:
Hey @kuborgh-mspindelhirn,

thank you for your contribution. We decided to close this PR because we think the bahaviour of the viewport connections already works as intended. When there are hidden elements they still are part of the content and the user should always be aware that these elements need to be placed at least at one viewport. If you still think this is important you can create an issue at https://issues.shopware.com/ with detailed information about the feature. If the ticket gets enough votes we will consider an own solution.

PR #948, opened 2017-02-02
fix checkout header generation not uniform

schliflo:
## Description
* Why is it necessary?
Headers are not generated uniformly
* What does it improve?
Uses the header generation from confirm in shipping_payment to have a single base
* Does it have side effects?
none that i can think of

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | --
| How to test?     |

### DIFF
--- a/themes/Frontend/Bare/frontend/checkout/shipping_payment.tpl
+++ b/themes/Frontend/Bare/frontend/checkout/shipping_payment.tpl
@@ -1,55 +1,4 @@
-{extends file="frontend/index/index.tpl"}
-
-{* Shop header *}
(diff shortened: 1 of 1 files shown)


schliflo:
By the way I think your tests are broken. Travis build for PHP 5.6 failed in the first run: https://travis-ci.org/shopware/shopware/builds/197600294
I had to amend and force push without changing code to re-trigger travis. Build passed the second time around...

wolv-dev:
Ticket opened!
https://issues.shopware.com/issues/SW-18547
Something like this shouldn't be closed, cause the code quality should be maintained.

PR #949, opened 2017-02-02
SW-16471 | API addresses - Customer Info fehlt

ahmad-saad:
## Description
* Why is it necessary?
[the Address API Documentation](https://developers.shopware.com/developers-guide/rest-api/api-resource-address/) in Get ( list )  said:

> This API call returns an array of elements, one for each address. Each of these elements has the same structure like a single element above.

but when you call this API you can't find the Customer info.

* What does it improve?
add the Customer Info to the returned list.
* Does it have side effects?
NO

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | [SW-16471](http://issues.shopware.com/issues/SW-16471).
| How to test?     | just call the API.

### DIFF
--- a/engine/Shopware/Models/Customer/AddressRepository.php
+++ b/engine/Shopware/Models/Customer/AddressRepository.php
@@ -158,6 +158,7 @@ class AddressRepository extends ModelRepository
 
         $builder->select([
             'address',
+            'customer',
             'attribute',
             'country',
             'state'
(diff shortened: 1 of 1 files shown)


htkassner:
Thank you for your contribution. There are failing tests and due to performance issues probably caused by this join, we will close this pull request.

Thanks anyways for working on shopware!

PR #955, opened 2017-02-07
change type from documentId in document model

chrisko5:
## Description
* Why is it necessary?
alphanumeric document number (column "docID" in database has type varchar)
* What does it improve?
flexible document number
* Does it have side effects?
nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     |

### DIFF
--- a/engine/Shopware/Models/Order/Document/Document.php
+++ b/engine/Shopware/Models/Order/Document/Document.php
@@ -89,9 +89,9 @@ class Document extends ModelEntity
     private $amount;
 
     /**
-     * @var integer $documentId
+     * @var string $documentId
      *
-     * @ORM\Column(name="docID", type="integer", nullable=false)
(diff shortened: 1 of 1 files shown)


mitelg:
Hi @chrisko5,

thanks for your PR :+1: 
As the document ID is based on the values of the number ranges, these IDs have to be integers.
If you want to change this behavior, you will have to do much more, then just changing the PHPdoc in the doctrine models. For example a migration SQL is missing, for changing the column in the database. 
Beside that, you should know, that the Shopware DB is **not** created on base of the doctrine models. 

So we decided to close this PR because the change is incomplete and does not fix the underlying issue. If this change is still important to you, feel free to create a new pull request or create an issue on https://issues.shopware.com.

Best Regards

PR #957, opened 2017-02-07
fixed typo for documents template variable

nexxome:
## Description
* Why is it necessary?
because its typo
* What does it improve?
typo :)
* Does it have side effects?
no

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     |

### DIFF
--- a/themes/Frontend/Bare/documents/index.tpl
+++ b/themes/Frontend/Bare/documents/index.tpl
@@ -74,7 +74,7 @@ td.head  {
 </head>
 
 <body>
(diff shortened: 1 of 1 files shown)


shyim:
It could be a bc break if someone using in {block name="document_index_address"} as example the old postions variable :)

nexxome:
The "document_index_address" block was just added 13 days ago and was added after the last release.
But I think you are right, someone could use the variable in a block between.

PR #861, opened 2016-11-28
When clearing proxy cache, clear APCu cache too

fixpunkt:
## Description
Please describe your pull request:
* Why is it necessary? Hard-to-debug Doctrine errors may occur for users of plugins that change attribute configuration when APCu is enabled and used by Doctrine (see below for a detailed example).
* What does it improve? Clears the APCu cache when clearing the proxy cache, which should prevent instances of the issue described above.
* Does it have side effects? Clears the APCu cache, so may cause some additional cache misses after updating plugins, but this should not be a relevant drawback.

| Questions        | Answers
| ---------------- | -------------------------------------------------------
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | no
| How to test?     | since this is a once-in-a-blue-moon bug, it's not really testable

The log below shows an instance of the bug. This was observed on a customer's shop installation after the customer updated one of our plugins. During the upgrade, the `s_order_details_attributes` entry `viison_coupon_code_ids` was removed by the plugin. As the log output shows, Doctrine was still trying to access the attribute in question, although it was already removed from the model and from the database. Attributes were re-generated and the proxy cache was manually cleared several times, but the problem persisted. Calling `apcu_clear_cache()` once resolved the problem permanently.

```
[2016-11-18 16:57:27] plugin.ERROR: ViisonPickwarePOS: Failed to create order via REST API {"exception":"[object] (ReflectionException(code: 0): Property Shopware\\Models\\Attribute\\OrderDetail::$viisonCouponCodeIds does not exist at /var/www/web1/htdocs/shopware4/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/RuntimeReflectionService.php:80)"} {"uid":"11cc323"}
[2016-11-18 16:57:27] plugin.ERROR: Stack trace:
 #0 /var/www/web1/htdocs/shopware4/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/RuntimeReflectionService.php(80): ReflectionProperty->__construct('Shopware\\Models...', 'viisonCouponCod...')
 #1 /var/www/web1/htdocs/shopware4/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php(959): Doctrine\Common\Persistence\Mapping\RuntimeReflectionService->getAccessibleProperty('Shopware\\Models...', 'viisonCouponCod...')
[...]

PR #959, opened 2017-02-08
Added OPCache clearing capabilities to Backend and API

oktupol:
## Description
This pull requests adds the capability to clear the Zend OPCache through the Shopware backend.

* Why is it necessary?
  - There is no way to easily clear the Zend OPCache yet
* What does it improve?
  - A checkbox in the backend performance-module, where you can select to clear the OPCache.
  - If you clear all caches through the api, you also clear the OPcache now.
* Does it have side effects?
  - No

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| How to test?     | Set up shopware on an environment with an OPCache, then clear the cache through the backend performance-module.


fixpunkt:
In which cases is it necessary to clear the op cache?

If this is also related to Doctrine hiccups, I think automatically clearing the op cache when clearing the proxy cache is preferable (which is what I did for https://github.com/shopware/shopware/pull/861).

PR #960, opened 2017-02-08
use DI parameters for class names

kuborgh-mspindelhirn:
## Description
* Why is it necessary?

For some services a decorator pattern is not viable for changing/extending service implementations.
* What does it improve?

Using parameters to propagate class names for all services allow 3rd parties to change the class to use for a service. New implementation can extend the original class.
* Does it have side effects?

No

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     |

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/services.xml
+++ b/engine/Shopware/Components/DependencyInjection/services.xml
@@ -9,6 +9,112 @@
             <parameter key="regexp">/([^A-Za-z0-9\.]|-)+/</parameter>
             <parameter key="lowercase">false</parameter>
         </parameter>
+
+        <parameter key="bootstrap.class">Shopware_Bootstrap</parameter>
+        <parameter key="locale_factory.class">Shopware\Components\DependencyInjection\Bridge\Locale</parameter>
+        <parameter key="currency_factory.class">Shopware\Components\DependencyInjection\Bridge\Currency</parameter>
+        <parameter key="currency.class">Zend_Currency</parameter>
+        <parameter key="snippet_resource.class">Enlight_Components_Snippet_Resource</parameter>
+        <parameter key="template_factory.class">Shopware\Components\DependencyInjection\Bridge\Template</parameter>
+        <parameter key="template.class">Enlight_Template_Manager</parameter>
+        <parameter key="session_factory.class">Shopware\Components\DependencyInjection\Bridge\Session</parameter>
+        <parameter key="session.class">Enlight_Components_Session_Namespace</parameter>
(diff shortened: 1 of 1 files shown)


bcremer:
You can use the `decorates` attribute to replace any service in the DI-Container. You are not forced to inject the `.inner` service back into the constructor. The `decorates` name from the Symfony Container might be a little confusing but it can act as a replace.

```xml
<?xml version="1.0" ?>

<container xmlns="http://symfony.com/schema/dic/services"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">

    <services>
        <service id="swag_example.list_product_service"
                 class="SwagExample\Bundle\StoreFrontBundle\ListProductService"
                 decorates="shopware_storefront.list_product_service">
        </service>
    </services>
</container>
```

The class `SwagExample\Bundle\StoreFrontBundle\ListProductServic` may not be a decorator but an own implementation that extends from the core class.

kuborgh-mspindelhirn:
This is true but you cannot see the `list_product_service` as a black box in your example as you need to do the dependency injection for your decorator. So the above example is unlikely to work as you now need to inject all arguments from the original service
```
<service id="shopware_storefront.list_product_service" class="Shopware\Bundle\StoreFrontBundle\Service\Core\ListProductService">
            <argument type="service" id="shopware_storefront.list_product_gateway"/>
            <argument type="service" id="shopware_storefront.graduated_prices_service"/>
            <argument type="service" id="shopware_storefront.cheapest_price_service"/>
            <argument type="service" id="shopware_storefront.price_calculation_service"/>
            <argument type="service" id="shopware_storefront.media_service"/>
            <argument type="service" id="shopware_storefront.marketing_service"/>
            <argument type="service" id="shopware_storefront.vote_service"/>
            <argument type="service" id="shopware_storefront.category_service" />
            <argument type="service" id="config" />
        </service>
```

Replacing the implementation class via parameter allows you to just change one method and if something changes in the rest of the implementation including changes to dependencies you dont have to worry.

OliverSkroblin:
Hi, 
we discussed this issue internally an decided to decline your changes. 
It would increase the complexity to maintain the product and the way how developers can extend service definitions.

hgiesenow:
See http://symfony.com/doc/current/best_practices/business-logic.html#service-no-class-parameter - the class parameters are indeed often used but not intended ;-)

xabbuh:
To add some more information: In Symfony, we dropped the class parameters when moving to 3.0 as they proved to be a hard to maintain extension point and you are seldom able to customise a service just by replacing it class (and if that's the case you are still able to do that in a compiler pass).

PR #961, opened 2017-02-09
Fix typo in block name in ArticleList suggest controller

tyurderi:
## Description
This pull requests fixes a typo in a block name of a backend controller of the module ArticleList.

* Why is it necessary?
It duplicates an overriding controller by an plugin in the code when you append the block.
* What does it improve?
Less duplicated code when extending the current (non-fixed) block.
* Does it have side effects?
No

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | no test needed
| How to test?     | no test needed

### DIFF
--- a/themes/Backend/ExtJs/backend/article_list/controller/suggest.js
+++ b/themes/Backend/ExtJs/backend/article_list/controller/suggest.js
@@ -34,7 +34,7 @@
 /**
  *
  */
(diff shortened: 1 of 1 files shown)


mitelg:
hey @tyurderi 
thanks for your PR, good catch! :+1: 
we will merge this as soon, we released SW 5.2.17

PR #962, opened 2017-02-09
fix combobox set null value on suggest

kuborgh-mspindelhirn:
## Description
* Why is it necessary?
There is a bug in comboxes in emotion widgets if the store has a remote data source.
If you start typing one of the choices the store loads matching entries from the backend. That triggers the "load" event on the store. The event handler changed in this commit is used to select the saved value on initial load. But it is triggered every time and so the event tries to find a record with the typed search query as the ID witch fails and then "null" is set as selected value 
* What does it improve?
Fixes bug
* Does it have side effects?
No. The combobox will still relay the load event so there is no BC break.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | Create emotion widget. Add combobox with store that loads data from the backend. Create widget and try to select an option by typing.


DennisGarding:
Hello,

thanks for your contribution!

The branch is merged.

PR #968, opened 2017-02-13
Changing the PriceFacetHandler to save filterloading time

TobiasWilming:
## Description

This pull request changes the PriceFacet, so that only one query is send to the Database.

* Why is it necessary? The two querys can be executed in one.
* What does it improve? It saves about 30-50% of the time the PriceFilter needs.
* Does it have side effects? No it doesn't.

| Questions        | Answers
| BC breaks?       |no
| Tests pass?      | yes
| Related tickets? | -
| How to test?     | The functional tests should test if the right values where show. For testing the time saving you need to disable the Http-Cache, fill the category with about 1000-3000 articles and measure the category loading time with and without the changes.

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/FacetHandler/PriceFacetHandler.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/FacetHandler/PriceFacetHandler.php
@@ -126,22 +126,18 @@ class PriceFacetHandler implements FacetHandlerInterface
         $selection = $this->priceHelper->getSelection($context);
         $this->priceHelper->joinPrices($query, $context);
 
-        $query->select('MIN('. $selection .') as cheapest_price');
-
-        /**@var $statement \Doctrine\DBAL\Driver\ResultStatement */
-        $statement = $query->execute();
-
-        $min = $statement->fetch(\PDO::FETCH_COLUMN);
+        $query->select('MIN(' . $selection . ') as cheapest_price');
 
         $query->groupBy('product.id')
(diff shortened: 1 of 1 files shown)


engine/Shopware/Bundle/SearchBundleDBAL/FacetHandler/PriceFacetHandler.php
```diff
@@ -126,22 +126,18 @@ public function generateFacet(
...
-            ->orderBy('cheapest_price', 'DESC')
-            ->setFirstResult(0)
-            ->setMaxResults(1);
+            ->orderBy('cheapest_price', 'DESC');
 
         /**@var $statement \Doctrine\DBAL\Driver\ResultStatement */
         $statement = $query->execute();
 
-        $max = $statement->fetch(\PDO::FETCH_COLUMN);
+        $result = $statement->fetchAll(\PDO::FETCH_COLUMN);
```
bcremer, line 137: You could use `$statement->fetch()` to fetch the first row and `$statement->fetch(PDO::FETCH_ORI_LAST )`  to fetch the last row.

bcremer: The synxtax is a bit different:

```
$stmt = $pdo->prepare('SELECT id FROM table', array(PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL));
$stmt->execute();

$first = $pdo->fetch(PDO::FETCH_ASSOC, PDO::FETCH_ORI_FIRST);
$first_row = $first['id'];
```

PR #973, opened 2017-02-15
Remove old syntax to enable support shop-struct

giginos:
The call $this->getShop()->get('isocode') does nothing else than $this->getShop()->getId() - but the old syntax does not support the shop-struct, which is returned using the ShopContextService. So this line should be changed.

## Description
* Why is it necessary? Support shop-struct
* What does it improve? Performance (because you can use the struct)

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes

### DIFF
--- a/engine/Shopware/Components/TemplateMail.php
+++ b/engine/Shopware/Components/TemplateMail.php
@@ -164,7 +164,22 @@ class Shopware_Components_TemplateMail
             );
             $isoCode = $this->getShop()->getId();
             $translationReader = $this->getTranslationReader();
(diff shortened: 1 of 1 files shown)


giginos:
Also added fallback for translations. So there is no need to create translation for each subshop.

soebbing:
Hi @giginos, we merged your PR in 4aafa2aa7b509334257f8fb56bf69f1b42af2f75, sorry it took us so long. :( 
Also we only used the reading of fallback-translations. We think there was a misunderstanding of what Shop instance should be given to the TemplateMail-class, so we also added a proper typehint to the setter.

PR #974, opened 2017-02-15
False guess on snippet export

Gugiman:
On snippet export, shopware querys a $baseLocale for creating an alias.
A few lines further, shopware guesses that the baseLocale is a record with localeID = 1 and shopId = 1

But in some cases, this record isn't the first result on the query before. 
Like this: 
image

This causes, that the german translation is put doubled either in german column and in englisch (or some other language that was the first line in BaseLocale query):
image

There is either this workaround or the sql query must be sorted after localeID = 1 and shopId = 1
I think this one is more dynamic

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| How to test?     | I don't know exactly how to reproduce that the german language isn't on first place. But that causes the error. Simply export snippets in csv or csvexcel.


mitelg:
thanks for your contribution @Gugiman :+1: :tada:

PR #975, opened 2017-02-15
Update main.ini

Gugiman:
In consideration of https://github.com/shopware/shopware/pull/809

On mobile view the top navigation is replaced by the canvas side navigation.
Since the option for hiding a category in top navigation won't effect the side navigation and the PR was closed, I think this label is misleading the user by thinking the whole category would be hidden.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related PR? | https://github.com/shopware/shopware/pull/809


Phil23:
Hi,

the text is misleading, because the user could think, that the whole category is only shown on mobile devices. But this option is only for the top navigation.

If you have other ideas for an alternate text for this snippet, feel free to open a new pull request.

Thx

PR #977, opened 2017-02-15
Fix typo "nocache" in AppCache.php

pascalheidmann-bedarf:
Fix default CacheControlDirective to "no-cache"  instead of "nocache" (see https://developer.mozilla.org/en/docs/Web/HTTP/Headers/Cache-Control#Cacheability )

## Description
* Why is it necessary?
In its current form browsers won't recognise the no-cache directive
* What does it improve?
Shopware sends the correct no-cache directive (if not otherwise changed)
* Does it have side effects?
None

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | not tested
| Related tickets? | none
| How to test?     | Check any page with default Cache-Control headers set by Shopware. In combination with a high max-age set this would result in not checking for changes by the browser until max-age is reached.

### DIFF
--- a/engine/Shopware/Components/HttpCache/AppCache.php
+++ b/engine/Shopware/Components/HttpCache/AppCache.php
@@ -114,7 +114,7 @@ class AppCache extends HttpCache
         $response = parent::handle($request, $type, $catch);
 
         $response->headers->remove('cache-control');
-        $response->headers->addCacheControlDirective('nocache');
+        $response->headers->addCacheControlDirective('no-cache');
 
         return $response;
     }


FiveDigital:
@pascalheidman-bedarf I think your last commit should be on a separate branch and a separate PR.

pascalheidmann-bedarf:
@FiveDigital yeah, forgot to select create another branch in github ui :/

FiveDigital:
@Phil23 Would you merge a backport of this in the 5.2 branch if I create a PR?

PR #978, opened 2017-02-16
Fixed sShopUrl Url on alwaysSecure

shyim:
## Description
* Why is it necessary?
When always ssl enabled in shop, it should be also in mails
* What does it improve?
Uses https url in mails
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | Enable always secure and show the sShopUrl link in mail template

### DIFF
--- a/engine/Shopware/Controllers/Backend/Mail.php
+++ b/engine/Shopware/Controllers/Backend/Mail.php
@@ -179,7 +179,7 @@ class Shopware_Controllers_Backend_Mail extends Shopware_Controllers_Backend_Ext
 
         $defaultContext = array(
             'sShop'    => Shopware()->Config()->get('ShopName'),
-            'sShopURL' => 'http://'. $shop->getHost() . $shop->getBasePath(),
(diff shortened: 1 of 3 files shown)


engine/Shopware/Controllers/Backend/Mail.php
```diff
@@ -179,7 +179,7 @@ protected function getSingleMail($id)
 
         $defaultContext = array(
             'sShop'    => Shopware()->Config()->get('ShopName'),
-            'sShopURL' => 'http://'. $shop->getHost() . $shop->getBasePath(),
+            'sShopURL' => 'http' . ($shop->getAlwaysSecure() ? 's' : '') . '://'. $shop->getHost() . $shop->getBasePath(),
             'sConfig'  => Shopware()->Config(),
         );
```
nlubisch: looks like here is some redundant code.
A method like "getMailContext" with parameter $shop could simplify this much more

shyim: its replaced :)

PR #984, opened 2017-02-16
Save grid user preferences also in customers, orders

shyim:
## Description
* Why is it necessary?
User can customize columns of orders, customers
* What does it improve?
User grid preferences in orders and customers will be also saved like article list 
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     |


klarstil:
Hey @shyim, thanks for your contribution. We decided to decline your PR due to the fact we experienced a lot of trouble and unwanted behavior with the ExtJS StateManager. We had planned to enable it in every administration module we're having but sadly found out it has major faults in the ExtJS version we're using.

PR #986, opened 2017-02-18
Fixed opcache check when opcache is disabled with php.ini

shyim:
## Description
* Why is it necessary?
Backend Module shows opcache is everything ok, but it is not enabled..
* What does it improve?
Checks opcache is also enabled in php.ini
* Does it have side effects?
Nope

PS: I have done also cosmetic changes

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | [SW-17721](https://issues.shopware.com/issues/SW-17721)
| How to test?     |

### DIFF
--- a/engine/Shopware/Controllers/Backend/Performance.php
+++ b/engine/Shopware/Controllers/Backend/Performance.php
@@ -505,7 +505,7 @@ class Shopware_Controllers_Backend_Performance extends Shopware_Controllers_Back
                 'id' => 1,
                 'name' => Shopware()->Snippets()->getNamespace('backend/performance/main')->get('cache/apc'),
                 'value' => extension_loaded('apcu'),
-                'valid' => extension_loaded('apcu') === true ? self::PERFORMANCE_VALID : self::PERFORMANCE_INVALID
+                'valid' => extension_loaded('apcu') === true && ini_get('opcache.enable') === '1' ? self::PERFORMANCE_VALID : self::PERFORMANCE_INVALID
             ),
             array(
                 'id' => 3,


engine/Shopware/Controllers/Backend/Performance.php
```diff
@@ -500,27 +494,27 @@ protected function getPerformanceCheckData()
...
-            array(
+            ],
+            [
                 'id' => 3,
                 'name' => Shopware()->Snippets()->getNamespace('backend/performance/main')->get('cache/zend'),
                 'value' => extension_loaded('Zend OPcache'),
-                'valid' => extension_loaded('Zend OPcache') === true ? self::PERFORMANCE_VALID : self::PERFORMANCE_INVALID
-            ),
-            array(
+                'valid' => extension_loaded('Zend OPcache') === true && ini_get('opcache.enable') == 1 ? self::PERFORMANCE_VALID : self::PERFORMANCE_INVALID
```
nlubisch: you can better check the type with "===" too.
ini_get returns always a string if the key is there - only when its not there it returns false

nlubisch (changes requested):

engine/Shopware/Controllers/Backend/Performance.php
```diff
@@ -45,11 +45,7 @@ class Shopware_Controllers_Backend_Performance extends Shopware_Controllers_Back
      * Stores a list of all needed config data
      * @var array
      */
-    protected $configData = array();
-
-    protected function initAcl()
-    {
-    }
+    protected $configData = [];
```
samuelvogel: This PR is really hard to digest as it also contains a lot of (probably unwanted) PHP syntax / code style changes.

nlubisch (approved):

soebbing:
PR was merged with https://github.com/shopware/shopware/commit/9a2f16b12c828a477cd208a6f3dc453872c67e93

Thanks again for your help! @shyim 👍

PR #990, opened 2017-02-20
Added paymentID as event property in "Shopware_Modules_Admin_Execute_…

shyim:
## Description
* Why is it necessary?
Do custom things with rule and payment
* What does it improve?
Adds paymentId to event
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | .
| How to test?     |

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -1690,14 +1690,14 @@ SQL;
         foreach ($queryRules as $rule) {
             if ($rule["rule1"] && !$rule["rule2"]) {
                 $rule["rule1"] = "sRisk".$rule["rule1"];
-                if ($this->executeRiskRule($rule["rule1"], $user, $basket, $rule["value1"])) {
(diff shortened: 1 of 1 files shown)


engine/Shopware/Core/sAdmin.php
```diff
@@ -1712,17 +1712,19 @@ public function sManageRisks($paymentID, $basket, $user)
      * @param array $user
      * @param array $basket
      * @param string $value
+     * @param integer $paymentID
      * @return bool
      */
-    public function executeRiskRule($rule, $user, $basket, $value)
+    public function executeRiskRule($rule, $user, $basket, $value, $paymentID)
```
OliverSkroblin: paymentID should defined with default value null, otherwise it is a api break

shyim: Done

mitelg:
thanks @shyim :+1:

PR #1001, opened 2017-02-22
Url rewrite

egoist-dd:
## Description
* Why is it necessary? Suppliers with & in his name was saved as &amp; in database. So Supplier-Url would be test-amp-test not test-test

* What does it improve? better seo-urls

| Questions        | Answers
| BC breaks?       | yes
| Tests pass?      | yes
| Related tickets? | no
| How to test?     | create supplier test & test and create seo urls


janbuecker:
Thanks for your contribution @egoist-dd 

We've implemented slugify for exactly this use-case: customization.

There is an interface `Shopware\Components\Slug\SlugInterface` which can be used to decorate the `shopware.slug` service in the DI container. Alternatively, you can configure the DI container parameters `shopware.slug.config` for your needs.

egoist-dd:
Hi @janbuecker,

thanks for your reply.

Can you tell me how to set shopware.slug.config?

A small documentation about this would be great

janbuecker:
It's a simple service decoration or parameter overwriting. I've attached two plugins with 2 different approaches:

* SwagCustomSlugConfig, adds the `&` symbol to allowed values. This will result in: `/test-&-test/`
* SwagCustomSlugService, runs custom functions before. This will result in: `/test-test/`

You can find these plugins here:

* https://developers.shopware.com/exampleplugins/SwagCustomSlugConfig.zip
* https://developers.shopware.com/exampleplugins/SwagCustomSlugService.zip

janbuecker:
I've documented the configuration in shopware/devdocs#516. 

https://developers.shopware.com/developers-guide/plugin-slugify/

Thanks for the hint!

PR #1003, opened 2017-02-23
Fix empty form field clear data in record

kuborgh-mspindelhirn:
## Description
* Why is it necessary?
Unable to clear form fields like comboboxes. Empty combobox does not return any value thus the data in the record is never removed
* What does it improve?
If a form field exists and is empty the data in the record will be cleared as well
* Does it have side effects?
no

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | Create new backend CRUD; Add e.g. combobox as form element. Select a value from the store. Remove value from combobox. Save.

### DIFF
--- a/engine/Library/ExtJs/overrides/Ext.form.Base.js
+++ b/engine/Library/ExtJs/overrides/Ext.form.Base.js
@@ -55,6 +55,7 @@ Ext.override(Ext.form.Basic, {
      */
     updateRecord: function(record) {
         record = record || this._record;
+        var me = this;
 
         var values = this.getValues(),
             fields = record.fields,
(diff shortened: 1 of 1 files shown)


kuborgh-mspindelhirn:
Fixed a bug where an empty value was send for disabled fields.

PaddyS:
Hello @kuborgh-mspindelhirn,

sorry for letting you wait for so long.

Unfortunately your changes might change the behavior of pretty much any custom field and thus might break some plugins, as well as some core functionality.

Yet I understand your usecase and came up with a custom solution for you:
You could create your very own ExtJS ComboBox component, which extends from the default Ext.form.field.ComboBox.
Your custom combobox then has to implement a method called "getSubmitData", which could internally call the "getModelData", which is also part of the default combobox.
This should already fix the issue you've mentioned for your specific combobox.
Here's a quick boilerplate example to show you what I mean:
https://gist.github.com/PaddyS/70c561c94a3740a7cd489a97ce299880

PR #1004, opened 2017-02-23
fixed NaN-bug in custom date field when value is 0000-00-00

Stapelchips:
## Description
* When article has a custom field of date-type and is saved without an input in said field, the value '0000-00-00' is saved. When article is opened again in backend, it shows 'NaN.NaN.0NaN'. This 'value' however can not be saved in the date-type-field since this is not valide input.
* This fix checks if the value creates a NaN-Output. If it does, null is returned. If not, the valid date is returned.
* No known side effects.

| Questions        | Answers
| BC breaks?       | unknown
| Tests pass?      | unknown
| Related tickets? | unknown
| How to test?     | Create custom field for article of type "date". Create new article and leave said field empty. Reopen the new article: the custom field displays NaN. Now article can't be saved until NaN is deleted from the field. With this fix however, the NaN-output is catched and null is returned instead.

### DIFF
--- a/themes/Backend/ExtJs/backend/base/component/element/date.js
+++ b/themes/Backend/ExtJs/backend/base/component/element/date.js
@@ -41,7 +41,12 @@ Ext.define('Shopware.apps.Base.view.element.Date', {
         if(!value) {
             return null;
         } else if (typeof(value) == 'string') {
-            return new Date(value);
+            var date = new Date(value);
+            if (!isNaN(date)) {
+                return date;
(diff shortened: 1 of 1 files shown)


mitelg:
hey @Stapelchips  
this was already fixed with https://github.com/shopware/shopware/commit/6641138b26f3720a4b98725430901c363a3e493a#diff-e686ca4a7f6ea4cf4e87d178a7b99634

thanks :blush: :+1:

PR #1009, opened 2017-02-26
Update FPDF and include it via composer

samuelvogel:
## Description
* Why is it necessary?
The shipped FPDF version is not fully PHP 7 compatible and emits warnings.
* What does it improve?
Includes PHP 7 compatible versions and cleans up the source code tree by using composer.
* Does it have side effects?
Plugins which might have included the `.php` files directly before will emit a PHP warning and if the used `require` it will even be a fatal error.

| Questions        | Answers
| BC breaks?       | Besides plugins which `require()`ed the files directly before, no, the [FPDF changelog](http://www.fpdf.org/en/changelog.php) lists no breaking changes, same goes for the [FPDI release notes](https://www.setasign.com/products/fpdi/downloads/#releaseNotes)
| Tests pass?      | yes
| Related tickets? | none
| How to test?     | Use the batch processing in the backend order list to create multiple documents and activate the option to get one merged PDF including all of them. It will still work just like before returning the single, merged PDF.

With PHP 7 and `error_reporting(E_DEPRECATED)` and following "How to test?" from above this is emitted instead of the PDF:
```
Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; FPDF has a deprecated constructor in /var/www/engine/Library/Fpdf/fpdf.php on line 12

Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; fpdi_pdf_parser has a deprecated constructor in /var/www/engine/Library/Fpdf/fpdi_pdf_parser.php on line 22

Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; pdf_parser has a deprecated constructor in /var/www/engine/Library/Fpdf/pdf_parser.php on line 51

Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; pdf_context has a deprecated constructor in /var/www/engine/Library/Fpdf/pdf_context.php on line 22
```
When diffing the FPDF and FPDI and FPDF_TPL versions included in the Shopware core with the original sources of the same versions the only change was the following CSS rule, which is only relevant for the FPDF docs, which where not even included anyways.

```diff
diff --git shopware/engine/Library/Fpdf/ original/engine/Library/Fpdf/
index c4acb86..8cfa33d
--- shopware/engine/Library/Fpdf/fpdf.css
+++ original/engine/Library/Fpdf/fpdf.css
@@ -1,6 +1,6 @@
 body {font-family:"Times New Roman",serif}
 h1 {font:bold 135% Arial,sans-serif; color:#4000A0; margin-bottom:0.9em}
-h2 {font:bold 100% Arial,sans-serif; color:#900000; margin-top:1.5em}
+h2 {font:bold 95% Arial,sans-serif; color:#900000; margin-top:1.5em; margin-bottom:1em}
 dl.param dt {text-decoration:underline}
 dl.param dd {margin-top:1em; margin-bottom:1em}
 dl.param ul {margin-top:1em; margin-bottom:1em}
```


shyim:
I would change the target branch to 5.3. Some plugins use fpdf can require 'engine/Library/Fpdf/fpdf.php'; and that would break it

samuelvogel:
Yes, you are right, I just updated the PR! I mentioned this as a possible problem already. It could be mitigated with placing empty files there as well, as the classes themselves will be autoloaded.

PR #1010, opened 2017-02-27
jquery.search - abort previous ajax-requests

screeny05:
This PR modifies the jquery.search plugin to abort any previous (still-running) ajax-requests to the search controller. This prevents multiple ajax-request coming in at the same time. Also it prevents an older request from superseeding a new request, in case it takes longer (for whatever reason).

## Description
* Why is it necessary? Performance improvement, preventing possible bugs
* What does it improve? Ajax-Search

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| How to test?     | Open up the browsers dev-tools, throttle the network-speed and start typing into the search-box. See screenshot below for what it should look like.

image

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.search.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.search.js
@@ -384,7 +384,11 @@
 
             $.publish('plugin/swSearch/onSearchRequest', [ me, searchTerm ]);
 
-            $.ajax({
+            if(me.lastSearchAjax){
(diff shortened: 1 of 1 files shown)


FiveDigital:
@screeny05 Was this PR intended to be 5.3 only? This would be useful in 5.2, too.

screeny05:
i don't see why it can't be backported to 5.2

screeny05:
backported https://github.com/shopware/shopware/pull/1057

PR #1014, opened 2017-02-28
add guzzle config array optional param in factory

slax0rr:
Makes initializing a guzzle client much simpler, and eliminates the
requirement to make consecutive calls to set default options, and
provides a convenient way to set other options in the Guzzle HTTP
client.

## Description
* Why is it necessary? Convenience
* What does it improve? Initialization of the Guzzle HTTP Client instance through the factory class

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | Not that I am aware of.
| How to test?     | Pass a config array to the *->createClient* method as you would to the *Client* class of the Guzzle Library

### DIFF
--- a/engine/Shopware/Components/HttpClient/GuzzleFactory.php
+++ b/engine/Shopware/Components/HttpClient/GuzzleFactory.php
@@ -30,11 +30,12 @@ use GuzzleHttp\ClientInterface;
 class GuzzleFactory
 {
     /**
+     * @param array $guzzleConfig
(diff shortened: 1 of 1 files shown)


soebbing:
Hi @slax0rr, thanks for your PR and helping to make Shopware better! 👍  Your change was merged with https://github.com/shopware/shopware/commit/2a244c12bf5fa90fb7207ebb2d4203a95cd9dbdd

PR #1016, opened 2017-02-28
Smarty - add mediaById compiler

screeny05:
This PR adds a new smarty-compiler `mediaById`.

This tag accepts two parameters: an `id` which identifies the media-entry and a `size` which identifies a thumbnail-size if there's a matching one.

This tag is needed due to the fact, that attributes of the type `single_selection`, `Shopware\Models\Media\Media` are returned to the frontend only as media-ids.

For example: you add an attribute to the supplier:
```php
$attrCrudService->update(
    's_articles_supplier_attributes',
    'banner',
    'single_selection',
    [
        'label' => 'Banner',
        'displayInBackend' => true,
        'entity' => 'Shopware\Models\Media\Media'
    ]
);
```

This attribute is accessable in the frontend (detail-page) via `{$sArticle.supplier_attributes.core->get('banner')}` but this only returns the id of the selected media-entity. With this new smarty-tag the developer is able to use this id like `{mediaById id=$sArticle.supplier_attributes.core->get('banner') size=0}`, which returns the path to the selected media with size 0 if available.

## Description
* Why is it necessary? Developer experience with the new attributes
* What does it improve? Getting a path to a media-file via it's id

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| How to test?     | See example given above

### DIFF
--- a/engine/Library/Enlight/Template/Plugins/compiler.mediaById.php
+++ b/engine/Library/Enlight/Template/Plugins/compiler.mediaById.php
@@ -1,0 +1,76 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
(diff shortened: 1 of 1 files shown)


screeny05:
Needs additional work - better to return the full media-object if asked for

PR #1017, opened 2017-02-28
Userfriendly mailer options

shyim:
## Description
* Why is it necessary?
Some users are not so clever and types "SMTP" in Mailer and that breaks the mailer...
* What does it improve?
Replaces text field with combobox, so the user CANT BREAK IT
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     |


EtienneBruines:
See https://issues.shopware.com/issues/SW-17988

soebbing:
Thanks once again, @shyim! 💙

PR #1018, opened 2017-02-28
Added sArticle variable to mail sARTICLEAVAILABLE

shyim:
## Description
* Why is it necessary?
Tell more informations about article in mail
* What does it improve?
Adds sArticle to mail
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | .
| How to test?     | .

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
@@ -21,6 +21,7 @@
  * trademark license. Therefore any rights, title and interest in
  * our trademarks remain entirely with us.
  */
+use Shopware\Models\Shop\Shop;
 
 /**
  * Shopware Notification Plugin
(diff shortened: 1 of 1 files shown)


OliverSkroblin:
@shyim 
you register the active default shop here: https://github.com/shopware/shopware/pull/1018/files#diff-a9f5d7c2c2c84b8859d3d5d251258f87R267

Problem is, if the customer notified in a different language shop, you get the article with the default language.
Additionally the store front list product service validates if the loaded product is available in the current shop. This would be broken and produce a strange behavoir.

Please check if you can find a solution for this case, otherwise we would decline the pull request.

shyim:
@OliverSkroblin
It uses now the shop context with the filled language ID in the s_articles_notification table. I also fixed the rewriting of the article link :D

PS: The mail doesn't have a translation

OliverSkroblin:
Hi @shyim,

i tested your changes and got an exception in case of shops with base path.

My shop is stored in a sub directory "/52", after your changes the mail contains the following links:

**Main shop:**          _http://localhost/genusswelten/koestlichkeiten/272/spachtelmasse_
**Language shop:** _http://localhost/worlds-of-indulgence/brandies/272/spachtel-en_

Please verify this issue and change the router context to support base path, base urls, virtual urls and ssl

PR #1023, opened 2017-03-01
Attributes Values ​​for ESD items in the basket

christopherholst:
If there is an ESD article, also the attributes in the cart should arrive with. This has the advantage that one at the shipping costs explicit or exact shipping costs can take off and one does not get a multiple choice of shipping modes.

## Description
* Why is it necessary? 
for the shipping costs
* What does it improve?
attributes in esd items for the basket
* Does it have side effects?
no

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -2586,7 +2586,7 @@ SQL;
             LEFT JOIN s_articles a
             ON b.articleID = a.id
             AND b.modus = 0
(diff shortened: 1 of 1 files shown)


OliverSkroblin:
This changes would introduce unexpected behavior for shipping cost calculation, shipping method selection and other calculations.
Please add unit tests and different test cases for expected behavior.

PR #1033, opened 2017-03-04
Renamed block for comment count in frontend_blog_detail_comments_count

shyim:
## Description
* Why is it necessary?
The old blockname was duplicate in same file. I can't choose which block i can really override
* What does it improve?
Renames duplicate block to frontend_blog_detail_comments_count
* Does it have side effects?

| Questions        | Answers
| BC breaks?       | yes/no
| Tests pass?      | yes/no


ndzoesch:
Thank you for adding the PR, sadly we have to decline it since your solution falls short when it comes to BC. We implemented our own fix which will be released with 5.2.21.

PR #1035, opened 2017-03-07
Add empty condition

Gugiman:
The getOrderStatusHistoryListQueryBuilder() Method throws an Doctrine exception when $orderBy is not set as parameter although it's optional.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| How to test?     | Use getOrderStatusHistoryListQueryBuilder() Method without second param.


ngoedde:
Thank you for your contribution @Gugiman. Your PR will be merged soon.

PR #1038, opened 2017-03-07
Abort pending ajax requests when the search word is updated

m-knabe-reply:
## Description
When you type in the search field, a result preview is fetched via ajax as soon as you stop typing. When you start typing again, while the first ajax request is still pending, a second ajax request is executed, but the first one is not aborted. That can lead to the situation, (especially when the server is under high load) where the second ajax request is finished before the first one. In this case, the preview shown does not match the current search term, but the one from the first request.
Solution: Abort the pending search request as soon as you start a new one.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | no
| How to test?     | Have a slow server, type in the search field, and see how the search results are not updated properly

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.search.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.search.js
@@ -384,10 +384,16 @@
 
             $.publish('plugin/swSearch/onSearchRequest', [ me, searchTerm ]);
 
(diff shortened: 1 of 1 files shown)


janbuecker:
Hey, thanks for your pull request.

This has been recently implemented in shopware/shopware#1010. :)

PR #1048, opened 2017-03-14
don't echo readfile's return

jeboehm:
Since readfile() echoes the given file directly to the output buffer, it is not necessary to echo its return value. In addition, echoing the number of bytes read _could_ mess up the download in environments with output compression.

See http://php.net/manual/en/function.readfile.php


roman-1983 (approved):

mitelg:
hey @jeboehm 
thanks for your PR :+1: 
it will be merged soon into SW 5.2

PR #1049, opened 2017-03-14
Move 'Shopware_Modules_Export_ExportResult_Filter'

hrombach:
The Event `Shopware_Modules_Export_ExportResult_Filter` currently supplies a `Zend_Db_Statement_Pdo` instead of the actual result. The statement can't actually be filtered, thus the event should be moved down and filter the actual result before it is assigned to Smarty.

## Description
* Why is it necessary?
The current implementation does not allow developers to filter the query result.
* What does it improve?
The result can be filtered.
* Does it have side effects?
If this event was previously used to manipulate the actual Statement object, that will not work anymore.

| Questions        | Answers
| BC breaks?       | possibly
| Tests pass?      | yes
| Related tickets? | could not find any
| How to test?     | subscribe to the event and filter the results

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -4,6 +4,7 @@ This changelog references changes done in Shopware 5.2 patch versions.
 
 ## 5.2.23
 * Added conditional statement in `themes/Frontend/Responsive/frontend/_public/src/js/jquery.product-slider.js` to prevent the jquery plugin from sending ajax requests indefinitely
+* The event `Shopware_Modules_Export_ExportResult_Filter_Fixed` was added and now filters the processed export result. Previously with `Shopware_Modules_Export_ExportResult_Filter`, an instance of `Zend_Db_Statement_Pdo` was supplied, which could not be used to filter the actual result.
 
 ## 5.2.22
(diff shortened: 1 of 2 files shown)


janbuecker:
Hey @hrombach, thanks for your PR.

To make it backward compatible, you should just create a new event with the needed parameters. Please keep the current event as it is.

hrombach:
The current event can't be used for its intended function, and its name would only be confusing if it was left intact. Do you have a suggestion how to make this clear, so that it doesn't create additional problems?

tpierschel:
@janbuecker we realy like to see this commit in the next version :-)
Here https://github.com/NetInventors/sw.ext.neti_discount_promotion_export is a working plugin with our solution.

janbuecker:
@hrombach I see the confusion about the filter naming but this will introduce a BC since the whole behaviour of this event has changed.

Your point is absolutely valid, but we can't merge BCs. So, you must add a new event instead of changing the existing one, even if the naming is confusing. Welcome to product development. :)

hrombach:
@janbuecker I renamed the event and left the old one intact.

I also rebased my changes while I was at it and squashed the commits together.

PR #1059, opened 2017-03-19
Capability to use Namespaces in Controllers

shyim:
## Description
* Why is it necessary?
Default benefits of namespacing
* What does it improve?
Allows "Enlight_Controller_Dispatcher_ControllerPath_X_X" Event to return a Class name, which should be used
* Does it have side effects?
Nope

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     |

Example Plugin:

Plugin:
```php
<?php


namespace Test;

class Test extends Plugin
{
    {
        return [
            'Enlight_Controller_Dispatcher_ControllerPath_Frontend_Test' => 'onTest'
        ];
    }

    public function onTest(\Enlight_Event_EventArgs $args)
    {
        return \Test\Controllers\Test::class;
    }
}
```

Controller:
```php
<?php


namespace Test\Controllers;

class Test extends \Enlight_Controller_Action
{
    public function indexAction()
    {
    }
}

### DIFF
--- a/engine/Library/Enlight/Controller/Dispatcher/Default.php
+++ b/engine/Library/Enlight/Controller/Dispatcher/Default.php
@@ -445,7 +445,7 @@ class Enlight_Controller_Dispatcher_Default extends Enlight_Controller_Dispatche
             return true;
         }
         $path = $this->getControllerPath($request);
-        return Enlight_Loader::isReadable($path);
+        return Enlight_Loader::isReadable($path) || class_exists($path);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jeboehm:
I'm wondering if the class_exists() statement with an optimized composer autoloader is faster, than the is_readable() filesystem call.
If it is faster, it should stand at the first position in the isDispatchable() method.

Maybe it would be useful to integrate these changes into RegisterControllerCompilerPass.

janbuecker:
Thanks @shyim, your PR will be merged soon. Please update the corresponding documentation [here](https://github.com/shopware/devdocs/blob/master/source/developers-guide/plugin-system/index.md) to have it documented within the next release.

PR #1061, opened 2017-03-21
[BUGFIX] Added padding to unordered lists in custom pages

wolv-dev:
## Description
Styling looks bugged on custom pages when unordered lists are used. Cause is, that ul styling receives padding: 0.

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | Add list to static page.


klarstil:
Thanks for your contribution @wolv-dev I added sorted list to your commit, so we're having the same behavior across both list types. The commit should be pushed in a few.

PR #1064, opened 2017-03-23
Fix calculation in Order::calculateInvoiceAmount

thomasnordquist:
## Description

* Why is it necessary?
  * The amount calculated by `Order/Order::calculateInvoiceAmount()` differs from the basket,
     this might lead to inconsistent prices calculations if the order is saved later in the backend.
     (Can be tested by changing shipping costs, saving and reverting to old value) 
* What does it improve?
  * Consistency of price calculation
* Does it have side effects?
  * No

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | This PR is a subset of #847
| How to test?     | Run the tests before applying fix in calculateInvoiceAmount

## Detailed writeup
While we originally tried to test/fix the first issue, the second issue emerged after we finished the tests and needed to be fixed as well, even though the implications weren't as extensive as with the first issue.

**First Issue:**
  * **Line 1069** in Order.php (here it's irrelevant if shop is in net mode or not)
  * Values rounded to 3 decimal places In s_order_details result in deviations, rounding them to two places will match the sBasket behaviour.
  * Solution: Extract `getPrice()` to `$price = round($detail->getPrice(), 2);` in all 3 occurrences.
  * Reproduce: Buy **5 x "SW10239"** in Frontend (shop in net mode), change shipping costs in backend and save and revert shipping costs back to old value. The amount now has changed from 96.48 to 96.47.

### Screenshot basket: amount = 96.48
bildschirmfoto 2017-03-23 um 15 14 05

### After `calculateInvoiceAmount` was triggered by changing shipping costs.
bildschirmfoto 2017-03-23 um 15 16 06

**Second Issue:**
  * **Line 1085** in Order.php
  * Missing rounding after calculation will result in unrounded values in the database.
  * Solution: round result to two decimal places.
  * Reproduce: Buy article in Frontend, change quantity in Backend and save, check s_order table for amount. The amount now stored in the database is not rounded at all. (Shop must not be in net mode to reproduce)

## What has been tested ?
* That `Order/Order::calculateInvoiceAmount()` calculation matches `sBasket`
* Checkout process (via frontend controllers)

pr-01

Edit: extract $detail->getPrice(), update PR message

### DIFF
--- a/engine/Shopware/Models/Order/Order.php
+++ b/engine/Shopware/Models/Order/Order.php
@@ -1066,7 +1066,9 @@ class Order extends ModelEntity
         //iterate order details to recalculate the amount.
         /**@var $detail Detail*/
         foreach ($this->getDetails() as $detail) {
-            $invoiceAmount += $detail->getPrice() * $detail->getQuantity();
+            $price = round($detail->getPrice(), 2);
+
+            $invoiceAmount += $price * $detail->getQuantity();
 
             $tax = $detail->getTax();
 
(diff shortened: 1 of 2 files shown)


mitelg:
merged with ecf9ac2d5eb0dbb156e962d353e395988772bc03

PR #1065, opened 2017-03-24
Don't loose array type hints when hooking methods

Ancillius:
If a method with type hints is hooked (e.g. 'Shopware\Models\Category\Repository::getListQueryBuilder::after' => 'myAwesomeFunc') the ProxyFactory will generate: 
public function getListQueryBuilder( $filterBy,  $orderBy, $limit = NULL, $offset = NULL, $selectOnlyActive = true)

But the original function is declared via
public function getListQueryBuilder(array $filterBy, array $orderBy, $limit = NULL, $offset = NULL, $selectOnlyActive = true)

This will result in
PHP Strict Standards:  Declaration of Shopware_Proxies_ShopwareModelsCategoryRepositoryProxy::getListQueryBuilder() should be compatible with Shopware\Models\Category\Repository::getListQueryBuilder(array $filterBy, array $orderBy, $limit = NULL, $offset = NULL, $selectOnlyActive = true)

## Description
* Why is it necessary?
See above
* What does it improve?
See above
* Does it have side effects?
Did not test deeply

| Questions        | Answers
| BC breaks?       | yes/no
Dont know
| Tests pass?      | yes/no
Dont know
Private Support Ticket
See Description

### DIFF
--- a/engine/Library/Enlight/Hook/ProxyFactory.php
+++ b/engine/Library/Enlight/Hook/ProxyFactory.php
@@ -282,6 +282,7 @@ class <namespace>_<proxyClassName> extends <className> implements Enlight_Hook_P
                     $proxy_params .= '$';
                     $array_param .= '&';
                 }
+                $params .= $rp->isArray() ? 'array ' : '';
                 $params .= '$' . $rp->getName();
                 $proxy_params .= '$' . $rp->getName();
                 $array_param .= '$' . $rp->getName();


aragon999:
Wouldn't it be better not only to check if it should be an array, but to check for all types (via [`ReflectionParameter::getType`](http://php.net/manual/de/reflectionparameter.gettype.php))

Ancillius:
aragon999:
Define "better". GetType is available only in PHP 7. Shopware supports 5.6 as of now.

bcremer:
Just as an idea it should be considered to include https://github.com/Ocramius/ProxyManager to generate the hook proxies. Proxy generation get's harder with every new PHP Version and Ocramius/ProxyManager is widely used in Symfony and Zend Framework.

PR #1071, opened 2017-03-31
[FEATURE] Updated media smarty compiler with id parameter

wolv-dev:
## Description
When using entity media with attribute system you only get id instead of path. To get the media path by id I updated the media smarty compiler with new parameter "id".

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     | Call media smarty function with parameter id: {media id='1'}


janbuecker:
Thanks for your PR.

The media smarty tag will be replaced with actual PHP, so it will introduce a huge performance issue because there is a SQL query for each image you request with the `id`. This is why we only enabled the `path` property of medias as we can simply get the absolute url from the media service without querying the database at all.

PR #1075, opened 2017-04-02
Deprecated Enlight_Controller_ActionEventArgs

shyim:
## Description
* Why is it necessary?
Enlight_Event_EventArgs can handle the same functions with magic method __call and the consts(PRE_EVENT, POST_EVENT...) are not used.

* What does it improve?
Makes typehinting of events callbacks clearer with one single class and removes enlight code :D

* Does it have side effects?
Nope, its marked as deprecated 

| Questions        | Answers
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | 
| How to test?     |

### DIFF
--- a/UPGRADE-5.3.md
+++ b/UPGRADE-5.3.md
@@ -298,6 +298,7 @@ did not work anymore because the smarty rendering is off. The string {$offerPosi
 * Deprecated `\Shopware_Controllers_Widgets_Listing::ajaxListingAction`, use `\Shopware_Controllers_Widgets_Listing::listingCountAction` instead
 * Deprecated method `sArticles::sGetAffectedSuppliers()` without replacement, to be removed with 5.5
 * Deprecated `Shopware\Models\Article\Element`, to be removed with 6.0
+* Deprecated `Enlight_Controller_ActionEventArgs` use `Enlight_Event_EventArgs` instead, to be removed with 5.4
 
 ### Backend Components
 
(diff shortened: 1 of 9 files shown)


bcremer:
I am -1 for this one. `Enlight_Controller_ActionEventArgs` provides better IDE autocompletion and code analysis in event callbacks.

dneustadt:
I'd rather have the helper methods of `Enlight_Controller_ActionEventArgs` in `Enlight_Event_EventArgs`

I see that ActionEventArgs hints explicitly at what kind of subscription it is. But I'd rather have the convenience of just always using `Enlight_Event_EventArgs`

I think the naming is also a little confusing, seeing how we have:
- **Enlight_Controller**_ActionEventArgs
- **Enlight_Hook**_HookArgs
- **Enlight_Event**_EventArgs

oktupol:
@dneustadt I disagree with the first part of your comment since many events don't have a subject, request and response attached to them. Having them suggested by the IDE would be confusing then, and wrong.

That said, I use the constants in Enlight_Controller_ActionEventArgs a lot, it reduces the probability of typos (which I have to admit, is quite high for me :) )

dneustadt:
@oktupol just add `null` (or `void`?) as return type to the annotation. And as ease of use goes, you would have to know the nature of the event beforehand anyway to type hint the obscurely named `Enlight_Controller_ActionEventArgs`. `Enlight_Event_EventArgs ` is just referred to much more often within the docs and examples.

PR #1082, opened 2017-04-05
Make jquery.product-slider infinite sliding opt-out

screeny05:
## Description

| Why?                    | Infinite sliding was a breaking change which should be at the very least an opt-out-able feature. It has the power to destory many existing features, built without cloned elements in mind. |
| Tests exists & pass?    | no tests |
| How to test?            | create a product-slider element with the `infiniteSlide` option set to false |
| Requirements met?       | yas |


mitelg:
thanks for your contribution @screeny05 :+1:

PR #1084, opened 2017-04-05
Fixed saving of empty dates

hgiesenow:
## Description
When clearing a custom attribute date field, the database contains "0000-00-00" instead of null, leading to the value of "Nan.Nan.0Nan" in the edit form. This PR fixes it to be reseted to NULL.

| Why?                    | To remove a date in a custom field |
| How to test?            | Create a Custom field of type Date. Select a date and save. Clear the field an save again. The field now stays empty and does not display "Nan.Nan.0Nan". The database field contains now NULL instead of "0000-00-00" |
| Requirements met?       | Maybe |


King-of-Babylon:
Thank you @hgiesenow for your PR.
This has already been implemented in Shopware 5.3.0 and therefore I will close this PR.

PR #1085, opened 2017-04-07
Process more than 50 manufacturer seo urls

jeboehm:
The SearchCriteria's limit property has to be set to NULL to process more than 50 manufacturer seo urls. :)


jeboehm:
What happens with the 10000st manufacturer? ;-)

PR #1090, opened 2017-04-15
add api endpoint /users

larsbo:
## Description

This is my first larger pull request - I hope I didn't mess it up totally 
There is one part I'm unsure about: It looks like there is no well matching bundle for the new user validation services. I chose the account bundle but perhaps a new one would make more sense?

| Why?                    | This PR adds the REST API endpoint `/users` for manipulating backend users. |
| How to test?            | Running PHPUnit tests. |
| Requirements met?       | I hope so :) |

### DIFF
--- a/engine/Shopware/Bundle/AccountBundle/Constraint/UserEmail.php
+++ b/engine/Shopware/Bundle/AccountBundle/Constraint/UserEmail.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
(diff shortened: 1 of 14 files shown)


jenskueper:
Hi @larsbo,

thank you for your work and the extensive PR! I tested it a bit and one thing that appears to be not working is the automatic password generation. I created a new user and tried the login with the generated password but I had no success. Could you check this?

larsbo:
@jenskueper thanks for reviewing/testing :)

I fixed the password bug and also added a user name validator because there was no duplicate check so far.

mitelg:
thanks for your contribution @larsbo :+1: :tada:

PR #1037, opened 2017-03-07
smarty cleanup

screeny05:
This PR fixes a lot of mess in the smarty template-files:

* missing or wrong indentation
* some missing blocks
* removed all usage of `append` and `prepend` for blocks
* removed `CDATA` in javascript
* replaced `frontend_index_header_javascript` with a more reliable and extendable version of building these variables

## Description
Please describe your pull request:
* Why is it necessary? General Cleaning
* What does it improve? Developer Experience
* Does it have side effects? No

| Questions        | Answers
| ---------------- | -------------------------------------------------------
| BC breaks?       | no
| Tests pass?      | yes

PR #1092, opened 2017-04-19
Prevent Javascript and Smarty Error with IndexNoCookiesNotice if single or doubl…

ahmad-saad:
…e quote is used

## Description

| Why?                    | this [PR](https://github.com/shopware/shopware/pull/899) fixed the single quote error but not fix the all problem now if we used a double quote inside the text we have Smarty error  |
| Related tickets?        | __ |
| How to test?            | just open the snippets manger search for IndexNoCookiesNotice and add a double quote to the text |
| Requirements met?       | ja |

### DIFF
--- a/themes/Frontend/Bare/frontend/index/index.tpl
+++ b/themes/Frontend/Bare/frontend/index/index.tpl
@@ -169,7 +169,8 @@
             {rdelim};
 
             var snippets = snippets || {ldelim}
-                'noCookiesNotice': '{"{s name='IndexNoCookiesNotice'}{/s}"|escape}'
+                {capture name="IndexNoCookiesNotice"}{s name='IndexNoCookiesNotice'}{/s}{/capture}
+                'noCookiesNotice': '{$smarty.capture.IndexNoCookiesNotice|escape:"javascript"}'
             {rdelim};
 
             var themeConfig = themeConfig || {ldelim}


ahmad-saad:
hallo @shopwareBot 
the capture is the solution. the javascript escaping is just something I added to make the escaping better.

the real Problem is that we have this code 
`'{"{s name='IndexNoCookiesNotice'}{/s}"|escape}'`

so let us say that IndexNoCookiesNotice value is 

> Cokies notice for "shopware" . testing double quote Problem.

then the code will become :
`'{"Cokies notice for "shopware" . testing double quote Problem."|escape}'`

this will return a smarty compile error because we have Shopware between tow texts and that make no since for smarty

you can change it to 
`"{'{s name='IndexNoCookiesNotice'}{/s}'|escape}"`

but now if I use a single quote in the text I have the same problem.

so capture is the solution I found to but the snippet value inside a variable without thinking what the user will add in the snippet text field in the backend.

I hope I clear the problem in a better way now.

screeny05:
This could probably be fixed with this PR: https://github.com/shopware/shopware/pull/1037

ahmad-saad:
hallo @screeny05 ,

I think it will not, but in any way try to add the following text  to IndexNoCookiesNotice  and check if it break the code.

> Cokies notice for "shopware" . testing double quote Problem.

thank's

screeny05:
@ahmad-saad the PR changes the output of the snippet, so it gets encoded as a json object. See here: https://github.com/shopware/shopware/pull/1037/files#diff-e53861ea23b76fbba835e8e51d01d35bR151 on line 166

ahmad-saad:
Hallo @screeny05 your changes prevent the javascript error but not the Smarty Problem.

if you use any value with double quote you will have the following error:

> Fatal error: Uncaught SmartyCompilerException: Syntax Error in template &quot;/......./themes/Frontend/Bare/frontend/index/index.tpl&quot; on line 167 &quot;'noCookiesNotice' =&amp;gt; {&amp;quot;{snippet name='IndexNoCookiesNotice' namespace='frontend/index/index'}Cokies notice for &amp;quot;shopware&amp;quot; . testing double quote Problem.{/snippet}&amp;quot;|escape:'javascript'}&quot; - Unexpected &quot;shopware&quot;, expected one of: &quot;}&quot; , &quot; &quot; , ATTR in /.../engine/Library/Smarty/sysplugins/smarty_internal_templatecompilerbase.php:657 
> 
> Stack trace: 
> #0 /...../engine/Library/Smarty/sysplugins/smarty_internal_templateparser.php(3144): Smarty_Internal_TemplateCompilerBase->trigger_template_error() 
> #1 /.../engine/Library/Smarty/sysplugins/smarty_internal_templateparser.php(3209): Smarty_Internal_Templateparser->yy in /.../engine/Library/Smarty/sysplugins/smarty_internal_templatecompilerbase.php on line 657

best Regard's,
Ahmad

PR #1094, opened 2017-04-20
Changed the check for mod rewrite

mamazu:
## Description

| Why?                    | I could not install shopware because according to it I didn't have mod_rewrite on. After some digging I found out that the $_SERVER variable doesn't have a mod_rewrite and the correct way to check for it is the apache modules function. |
| BC breaks?              | no, works with PHP 4.3.2 and greater|
| Tests exists & pass?    | no need. |
| How to test?            | If unset the $_SERVER['mod_rewrite'] and it will still work (as you have the rewrite mod already installed). |

### DIFF
--- a/recovery/install/index.php
+++ b/recovery/install/index.php
@@ -63,7 +63,7 @@ if (PHP_SAPI === 'cli') {
 ini_set('max_execution_time', 120);
 
 // Redirect to no mod rewrite path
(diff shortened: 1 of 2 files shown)


janbuecker:
Thanks @mamazu. I think this will enforce the usage of apache2 and will fail when using a different web server like nginx.

The info about `mod_rewrite` can be ignored when using an alternative web server.

shyim:
I would try to get with ajax the sitemap.xml as example and if it response a 404 , rewrite is then not configured.  It would be server independent

PR #1102, opened 2017-04-28
Added getConnection method to Plugins to handle special things on con…

shyim:
## Description

| Why?                    | Handle special things while container building like depending on plugins |
| Related tickets?        | @jochenmanz  's IRC Request :D |
| How to test?            | Add a compilerpass or load a services.xml when a specific plugin is installed |


shyim:
@janbuecker have a better idea.. :) I will create a new one

bcremer:
Can you please elaborate why this is necessary?
You should be able to get the PDO-Connection from within the plugin from the container like so:

```
// in plugin
$this->container->get('db_connection')
```

janbuecker:
For the use case to check if a depending plugin is installed, we could add a collection parameter into the container.

PR #1108, opened 2017-04-30
Improve input handling for the sw:rebuild:category:tree command and set correct limit

aragon999:
## Description

| Why?                    | The current default limit is set not the same as suggested in the documentation of the command. Also the input for this command is not checked and will raise warnings, thus the behavior of the command is probably not as expected if not properly used. The options could also supplied empty, which makes no sense. |
| How to test?            | Run `php bin/console sw:rebuild:category:tree -o=123` for example and view output with and without the patch |

### DIFF
--- a/engine/Shopware/Commands/RebuildCategoryTreeCommand.php
+++ b/engine/Shopware/Commands/RebuildCategoryTreeCommand.php
@@ -48,13 +48,15 @@ class RebuildCategoryTreeCommand extends ShopwareCommand
                 'offset',
                 'o',
                 InputOption::VALUE_REQUIRED,
-                'Offset to start with. Default: 0'
+                'Offset to start with. Default: 0',
+                0
             )
             ->addOption(
                 'limit',
                 'l',
                 InputOption::VALUE_REQUIRED,
-                'Categories to build per batch. Default: 3000'
+                'Categories to build per batch. Default: 3000',
+                3000
             )
             ->setHelp(<<<EOF
 The <info>%command.name%</info> command will rebuild your category tree.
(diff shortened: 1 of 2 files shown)


engine/Shopware/Commands/RebuildCategoryTreeCommand.php
```diff
@@ -43,21 +44,35 @@ protected function configure()
...
             ->setDescription('Rebuild the category tree')
-             ->addOption('offset', 'o', InputOption::VALUE_OPTIONAL, 'Offset to start with.')
-             ->addOption('limit', 'l', InputOption::VALUE_OPTIONAL, 'Categories to build per batch. Default: 3000')
+            ->addOption(
+                'offset',
+                'o',
+                InputOption::VALUE_REQUIRED,
+                'Offset to start with. Default: 0'
+            )
+            ->addOption(
```
Haehnchen: whats about using default parameter?

```
public function addOption($name, $shortcut = null, $mode = null, $description = '', $default = null) { }
```

aragon999: Thank you. I thought this would only work if the options are not optional. I corrected this. Shall I crunch the changes into one commit?

Haehnchen: depends of project guidlines but i suggest you should ;)

aragon999: Yes I know, but I could not find anything in the contribution guideline, that's why I asked. I will wait for an official feedback and then change the PR, if necessary :)

mitelg:
fixed with https://github.com/shopware/shopware/pull/1487

PR #1110, opened 2017-05-02
Fixed default value assignment for skipInit parameter

oktupol:
## Description

| Why?                    | Right now, if you pass anything that is not `undefined` for the `skipInit`-Parameter, it'll always be true |
| BC breaks?              | yes |
| How to test?            | Try to load a SubApplication into the Shopware backend with the Shopware.app.Application.addSubApplication function and set the `skipInit` parameter to false. The SubApplication will load, but nothing happens afterwards. See example below |

Example: Type this into the developer tools of your browser
```javascript
Shopware.app.Application.addSubApplication({
    name: 'Shopware.apps.Analytics',
    action: 'index'
}, false);
```

This is especially annoying if you want to use the callback parameter:

```javascript
Shopware.app.Application.addSubApplication({
    name: 'Shopware.apps.Analytics',
    action: 'index'
}, false, function () {
    console.log('callback');
});
```

What you'd have to do at the moment to actually set `skipInit` to false:

```javascript
Shopware.app.Application.addSubApplication({
    name: 'Shopware.apps.Analytics',
    action: 'index'
}, undefined);
```

### DIFF
--- a/engine/Library/ExtJs/overrides/Ext.app.Application.js
+++ b/engine/Library/ExtJs/overrides/Ext.app.Application.js
@@ -91,7 +91,9 @@ Ext.override(Ext.app.Application, {
 	 * @param subapp
 	 */
 	addSubApplication: function(subapp, skipInit, fn, showLoadMask) {
-        skipInit = (skipInit === undefined) ? false : true;
+        if (skipInit === undefined) {
+			skipInit = false;
+		}
 		subapp.app = this;
 
         if(subapp.hasOwnProperty('showLoadMask')) {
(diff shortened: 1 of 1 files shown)


engine/Library/ExtJs/overrides/Ext.app.Application.js
```diff
@@ -91,7 +91,9 @@ Ext.override(Ext.app.Application, {
 	 * @param subapp
 	 */
 	addSubApplication: function(subapp, skipInit, fn, showLoadMask) {
-        skipInit = (skipInit === undefined) ? false : true;
+        if (skipInit === undefined) {
```
shopwareBot: You can replace your logic by this simple line.

```
skipInit = skipInit === true;
```

Does this fix your issue?

oktupol: Yes, that'd work, too.

mitelg:
thanks for your PR @oktupol :+1:

PR #1123, opened 2017-05-10
Add Form Attributes to View Data

egoist-dd:
## Description

| Why?                    | Customize Forms with Attributes. |
| Tests exists & pass?    |no |
| Related tickets?        | -- |
| How to test?            | Extend Form with Attributes and use it in template or hooks. |
| Requirements met?       | -- |


engine/Shopware/Controllers/Frontend/Forms.php
```diff
@@ -218,6 +218,7 @@ protected function getContent($formId)
             'metaTitle' => $form->getMetaTitle(),
             'metaDescription' => $form->getMetaDescription(),
             'metaKeywords' => $form->getMetaKeywords(),
+            'attribute' => $form->getAttribute(),
```
shyim: Is $form->getAttribute() not returning a Doctrine model instead a array?

PR #1130, opened 2017-05-16
Share base stores among config selects

fixpunkt:
## Description

The settings of most of our plugins each contain multiple selects which are backed by stores, usually `Shopware.apps.Base.store.Payment`, `Shopware.apps.Base.store.PaymentStatus` and `Shopware.apps.Base.store.Dispatch`.

The situation right now is that for each of these selects, the backing store is loaded twice per subshop. This means that for customers who have more than a few subshops (8, for instance), opening the Pickware Mobile settings causes more than 200 HTTP requests to be made to the server in order to load the same content of three stores. This basically amounts to a small DoS attack on the server and can cause timeout errors to occur.

Here is an example waterfall graph for a standard installation (DE and EN subshops) loading one of our plugins' basic settings page:

image

This PR improves the situation by changing the `getStoreById()` method on `Shopware.apps.Base.view.element.Select` to use a managed store instance which prohibits reloading for all selects that use the same store. When the select requests a filter using the features @svenmuennich has implemented in https://github.com/shopware/shopware/pull/699, the filter gets encoded into the `storeId` to ensure that each combination of filters uses its own store instance.

In addition to implementing store sharing, I've removed `autoload: true` from the code that instantiates the store. This was causing all stores to be loaded twice because `setValue` will force loading of the store anyway and that always gets called when the config form is loaded. 

Here's the waterfall graph for the same plugin's basic settings  again after this patch has been applied:

image
 
| Why?                    | Silly number of requests when loading basic settings pages that use multiple store-backed selects. See explanation above. |
| Tests exists & pass?    | none exist |
| Related tickets?        | None that I know of|
| How to test?            | Use this [test plugin](https://gist.github.com/e469296b8a654cdd3bf75050a2cbe999). |

### DIFF
--- a/themes/Backend/ExtJs/backend/base/component/element/select.js
+++ b/themes/Backend/ExtJs/backend/base/component/element/select.js
@@ -117,12 +117,39 @@ Ext.define('Shopware.apps.Base.view.element.Select', {
      * @return Ext.data.Store|null
      */
     getStoreById: function(storeId) {
+        // Try to find any previously-loaded config select store for the given storeId
+        var configSelectStoreId = 'Shopware.apps.Base.view.element.Select.store.' + storeId;
+        // If filters are defined on this element, we need to use a different store for that filter combination
+        // The filter gets applied in `initComponent()` above.
+        if (this.filter) {
+            configSelectStoreId += '.filter.' + JSON.stringify(this.filter);
+        }
+        var store = Ext.data.StoreManager.lookup(configSelectStoreId);
+
+        if (store) {
+            return store;
+        }
+
         try {
-            return Ext.create(storeId, {
-                pageSize: 1000,
-                autoLoad: true
+            store = Ext.create(storeId, {
+                storeId: configSelectStoreId,
+                pageSize: 1000
             });
+            // Override the load() method so the store is only ever loaded once
+            Ext.override(store, {
+                load: function () {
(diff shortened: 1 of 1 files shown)


mitelg:
thanks for your contribution @fixpunkt

shyim:
Due this change the basic example of https://developers.shopware.com/developers-guide/plugin-configuration/#selectionfield-/-remote-combobox does not work more.

I patched my local shopware copy back and it works normal. I guess the problem is that normal stores like Shopware.apps.Base.store.Category don't have autoLoad, so they would be never loaded.

@fixpunkt i would like create a new pr which adds autoLoad: true to the Base stores to solve the issue or have you a better idea?

fixpunkt:
Hmm, you could try stepping through `setValue` to see why `store.load` is not being called:

https://github.com/VIISON/shopware/blob/bc8c405784bf51abe10d25aca87a76f286fa5df7/themes/Backend/ExtJs/backend/base/component/element/select.js#L85

shyim:
The value null condition in setValue breaks it, because after the plugin installation is the value null 🙄 

I have created a pr: https://github.com/shopware/shopware/pull/1497
Thank you for your idea with the setValue ❤️

fixpunkt:
No worries, thanks for fixing my mess!

PR #1131, opened 2017-05-17
Renamed Shopware.model.AttributeConfig to Shopware.apps.Base.model.AttributeConfig

oktupol:
## Description

| Why?                    | This fixes a consistency flaw. The class now follows the same naming scheme as all other classes in this directory. |
| BC breaks?              | no, old class name is still defined as alternateClassName |
| How to test?            | see below |

### How to test
Test, if new class name works:
```javascript
// In developer console
Ext.create('Shopware.apps.Base.model.AttributeConfig');
```

Test, if old class name still works
```javascript
// In developer console
Ext.create('Shopware.model.AttributeConfig');
```

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -2,6 +2,9 @@
 
 This changelog references changes done in Shopware 5.2 patch versions.
 
+## 5.2.24
+* Renamed `Shopware.model.AttributeConfig` to `Shopware.apps.Base.model.AttributeConfig`
+
 ## 5.2.23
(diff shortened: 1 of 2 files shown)


mitelg:
thanks @oktupol :+1:

htkassner:
Sorry @oktupol, but this commit has to be reverted, because there are some problems opening several modules. It is correct that the model can be instantiated via both names but it seems that the way the attribute plugin is loaded into several modules requires the old model class name. Thank you anyways.

oktupol:
I see. Wouldn't it be better to rename those occurrences as well though?

htkassner:
I guess the benefit of a renaming is too small in this case compared to the issues we had and the unexpected ones it may cause. I hope you understand that.

PR #1136, opened 2017-05-18
Bugfix proposal for SW-18748

hgiesenow:
## Description
Fixes issues, when using a http proxy on the same host without using ssl.

| Why?                    | This fix is needed to diable SSL in a http proxy setup |
| Related tickets?        | https://issues.shopware.com/issues/SW-18748 |
| How to test?            | Setup a nginx proxy and disable SSL. Images should now load |


dnoegel:
Hi,
thanks, we've just closed the related ticket. The main issue from my perspective is the support of old, external proxy SSL services and the check for the X_FORWARDED_HOST header, which we cannot break easily. 

The fix you are suggesting is not sufficient for setups, where the actual request is not secure (e.g. SSL Offloading in the LB). 
As stated in the ticket, we are planning to remove the support for external SSL proxies in the next major release - but we will probably not be able to fix this right now without breaking other systems already using this behaviour. 

Thanks for your effort anyway!

PR #1143, opened 2017-05-23
Removed the default option touchControls of the swImageSlider plugin

aragon999:
Regarding the PR #1118
It is really sad, that you just close the pull requests without a feedback, that the PR can be discussed. It was just closed, no way to discuss the issue. In my opinion, the problem is, that the fix for the underlying issue, is probably very complicated and may even introduce a bc break. This PR fixes one issue which exists at the moment and even removes redundant code. It is not that there is more code added for a workaround etc. 
So please reconsider your communication with the community, since we are doing the same thing when we write the PR messages. Otherwise I will probably no longer create PRs since they are time consuming to create (message) and then will just be closed with a standard message. I am a bit disappointed.

## Description

| Why?                    | The image slider will be destroyed and reinstantiated. In our case it sometimes came to a race condition where `trackThumbnailControls` was already executed but the plugin was "destroyed" and thus the function resulted in an error `Uncaught TypeError: Cannot read property 'position' of null`. Where the line of failure was `pos = $slide.position(),`. This is probably a deeper issue which should also be looked into. |
| BC breaks?              | No |
| Tests exists & pass?    | None exist |
| How to test?            | Create a template with an image slider (instantiated by `data-image-slider="true"`) and then for example load an shopping world. |

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -4,6 +4,7 @@ This changelog references changes done in Shopware 5.2 patch versions.
 
 ## 5.2.23
 * Added conditional statement in `themes/Frontend/Responsive/frontend/_public/src/js/jquery.product-slider.js` to prevent the jquery plugin from sending ajax requests indefinitely
+* Removed the default option `touchControls: true` from the instantiation of the `swImageSlider`, since it is default by the plugin itself and caused reinstantiation of the plugin when updated by the StateManager without this particular option
 
 ## 5.2.22
 * Fixed the picture implementation of the `box-emotion.tpl` to load the correct image sizes
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.ajax-variant.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.ajax-variant.js
@@ -139,7 +139,7 @@
                     $(me.opts.footerJavascriptInlineSelector).replaceWith($response.filter(me.opts.footerJavascriptInlineSelector));
 
                     StateManager.addPlugin('select:not([data-no-fancy-select="true"])', 'swSelectboxReplacement')
-                        .addPlugin('*[data-image-slider="true"]', 'swImageSlider', { touchControls: true })
+                        .addPlugin('*[data-image-slider="true"]', 'swImageSlider')
                         .addPlugin('.product--image-zoom', 'swImageZoom', 'xl')
                         .addPlugin('*[data-image-gallery="true"]', 'swImageGallery')
                         .addPlugin('*[data-add-article="true"]', 'swAddArticle')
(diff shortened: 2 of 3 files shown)


aragon999:
Thank you, if I know you cannot reproduce the problem I can look into it more closely :-) In fact I did look at the problem more closely and the issue was unfortunately not as I said, that the plugin configuration was different, but in the state manager, has `me._previousState = ""`. Therefore `StateManager._getPluginConfig(me._previousState, selector, pluginName)` will return an empty object, which obviously differs from `{touchControls: true}`. That is why the patch fixed the issue. 
I have created an demo plugin which has the issue, see here: [SwkweSliderTouchControllsBug](https://github.com/aragon999/SwkweSliderTouchControlsBug) As soon as the Emotion Shopping World is loaded the plugin `swImageSlider` is destroyed.
Unfortunately I was not able to reproduce the race condition. There is another plugin on the page which manipulates the `swImageSlider` plugin. Maybe there is the issue, but with this patch, the site works without errors. I also created some screenshots that hopefully help.
bildschirmfoto von 2017-05-28 13-26-44
bildschirmfoto von 2017-05-28 13-26-21
I suggest the following fix, change: https://github.com/shopware/shopware/blob/5.2/themes/Frontend/Responsive/frontend/_public/src/js/jquery.state-manager.js#L926 to: `previousState = me._previousState = me._currentState || state,`
I have not tested this or checked for any other problems.

mbdesign:
For me the issue still existing with Version 5.3.5 on the product detail site

`Uncaught TypeError: Cannot read property 'position' of null
    at Plugin.trackThumbnailControls (1515502851_90b031c30e6e579712a116feba1ba0c3.js:286)
    at HTMLDivElement.f (1515502851_90b031c30e6e579712a116feba1ba0c3.js:5)
    at HTMLDivElement.d.complete (1515502851_90b031c30e6e579712a116feba1ba0c3.js:5)
    at i (1515502851_90b031c30e6e579712a116feba1ba0c3.js:5)
    at Object.fireWith [as resolveWith] (1515502851_90b031c30e6e579712a116feba1ba0c3.js:5)
    at i (1515502851_90b031c30e6e579712a116feba1ba0c3.js:5)
    at n.fx.tick (1515502851_90b031c30e6e579712a116feba1ba0c3.js:5)`

mitelg:
thanks for your contribution @aragon999 :tada: : 

@mbdesign please open a ticket :blush:

klarstil:
Hey @aragon999 thanks for your breakdown of the issue. I have assigned the issue to myself and will take a closer look if we're dealing with bigger issue with the plugin management in the state manager. I'll keep you up to take using this issue and will create a new ticket in our issue tracker with detailed information.

aragon999:
@mbdesign you could try to implement the suggested fix, i.e. replace the following line:
https://github.com/shopware/shopware/blob/60d79b05ab4f81c478ff144f13741a74776ab6d0/themes/Frontend/Responsive/frontend/_public/src/js/jquery.state-manager.js#L935
by
```
previousState = me._previousState = me._currentState || state,
```
and see if this fixes your problem.

mbdesign:
Hi @aragon999   yes it fix my problem. 

FYI The issue still exist in shopware 5.4.1 when i use shopware Quickview (Professional Edition plugin)

Thanks

PR #1145, opened 2017-05-24
Use Shopware container as container property class

shyim:
## Description

| Why?                    | ContainerInterface doesn't provide all shopware container methodes like initialized |
| How to test?            | Use method initialized and you got a phpstorm warning |


shyim:
CDN is broken :(

giginos:
If these methods are really "required" - wouldn't it be better to define them in ContainerInterface or create a new ExtendedContainerInterface containing them?

shyim:
It was not possible to modify the Shopware default ContainerAwareInterface to ContainerInterface, because it would introduce many bc-breaks. It could be done with 5.4 :)

shyim:
ok

PR #1151, opened 2017-05-29
Basket float value issue

JonidBendo:
## Description

| Why?                    |There is an unhandled exception for float values, that gets corrected by this |
| Related tickets?        | https://issues.shopware.com/issues/SW-17016 |
| How to test?            | There is a test case created for this specific case |
| Requirements met?       | Does your PR fulfil our [contribution requirements](https://developers.shopware.com/contributing/contribution-guideline/#requirements-for-a-successful-pull-request)? |


fixpunkt:
https://github.com/shopware/shopware/pull/1027 and https://github.com/shopware/shopware/pull/1029 address the same issue.

JonidBendo:
@fixpunkt you are correct, although the approach is slightly different and the test cases do not utilise sGetBasketData. The approach in the other pull requests leaves space for errors, since in my case the error came from the exact test case i created and #1029 would not fix that. #1027 would pass but it would just address the empty cart issue but not actually provide the proper values.

fixpunkt:
Both PRs in combination fix the practical issues that occur when you discount a cart with a 100% voucher without introducing new numerical instabilities. It won't fix your test case, but should not make a difference in a real-world test.

JonidBendo:
@fixpunkt This is actually a case that came in one of the live shops :). So I would consider it a real-world case. That is why I submitted an issue in shopware a long time ago and only now decided to create a pull request since the issue remained. I just want to avoid having to apply core hacks or running my own shopware version.

fixpunkt:
I think you may have misunderstood my last comment. I was not doubting that your scenario does happen in real-world circumstances (we've reproduced this too), only that your unit test does not reflect that real world scenario perfectly. My PRs won't fix your unit test since that tests the return value of `sGetBasketData()`. However, that is irrelevant since my PRs will fix your actual real-world scenario: The basket will no longer become invalid and all displayed amounts will be rounded correctly. Since the results of `sGetBasketData()` get rounded again further down the line, making sure these results are rounded perfectly is not necessary.

JonidBendo:
You are correct, I misunderstood your comment. I agree with you partially, I am not sure where the rounding occurs afterwards and if every value is rounded but nonetheless I still believe it is better to actually do that from the start and not later on. This is my point of view and it might certainly be biased or faulty :)

shyim:
Hey @JonidBendo ,
can you please sign the CLA? otherwise we cannot merge the pull request

JonidBendo:
@shyim signed.

bneumann97:
Hey @JonidBendo,

thanks for your contribution 💙 

Is this issue still relevant?

JonidBendo:
@bneumann97 according to my knowledge yes, this is still an issue. It is partially solved with other PR's, but if someone uses the events present on that function and tries to compare one of the values of `$totalAmount` or `$totalAmountNet` with 0 it might fail depending on the calculation.

As said before this is a really specific issue that should not happen normally but depending on the business needs it could appear. Currently this is not an issue for me but I can see how it can be one for other people.

PR #1152, opened 2017-05-29
Don't load inactive plugins

shyim:
## Description

| Why?                    | The Plugin bootstrap is currently also constructed, if plugin is inactive. If a plugin includes libs example from composer and use the include the autoload.php in the head, they will be loaded everytime. I have local currently 90 Plugins, and i got a issue on registration page, because a library use symfony 3.x components was loaded and broke any thing, also if all plugins are inactive. Thats totally terrible to detect which plugin does it... |
| Related tickets?        | Nopeee |
| How to test?            | Create a new plugin, make composer require symfony/options-resolver in plugin folder, include it on the plugin bootstrap header. And try a registration everything is broken |

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInitializer.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInitializer.php
@@ -76,6 +76,11 @@ class PluginInitializer
                 continue;
             }
 
+            $isActive = in_array($pluginName, $activePlugins, true);
+            if (!$isActive) {
+                continue;
+            }
+
             $namespace = $pluginName;
             $className = '\\' . $namespace . '\\' . $pluginName;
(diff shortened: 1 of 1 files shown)


bcremer:
The (new) plugin system was design so that plugin construction is a cheap operation. I tried my best to prevent people from doing heavy stuff in the plugin's constructor (hence the `final` constructor https://github.com/shopware/shopware/blob/v5.2.24/engine/Shopware/Components/Plugin.php#L65).

I would propose to add a new callback function `registerAutoloader` that is called when a plugin is constructed and is active.

Otherwise this would be a compatibility break.

shyim:
The method registerAutoload would also allow people to do heavy stuff while Kernel is starting. Whats about change the target branch to 5.3, if is possible to merge it before release :)

PR #1156, opened 2017-06-11
option to compile theme from current asset timestamp and skip cache clear

wbob:
## Description

This evolved from a need in release handling on multiple nodes.

| Why?                    | the `theme:cache:generate` command creates a new assetTimestamp and clears the caches by default. Multiserver setups need finer controls in rolling out new releases and nodes. Setting a new timestamp is not always neccessary and if it is, then only once per rollout. The other nodes have to build from the new stamp instead of setting it again.<br/>For building assets: the first frontend request would trigger the process if the assets do not exist on the filesystem yet. Having them ready for the new release/node prior to getting real views gives flexibility and avoids slow requests / seconds of unstyled content.  Cache invalidation/clearing should be handled by existing dedicated commands. |
| Related tickets?        | no |
| How to test?            | execute `bin/console sw:theme:cache:generate --current` |
| Requirements met?       | Upgrade.md change: ✓ <br/>PR to correct shopware version: ✓ <br/>implementation missing important parts: no<br/>provideed necessary tests: ✓ <br/>do duplicate pull request exist on the same issue: no |

To me it was cleaner to create a new function in Components/Theme/Compiler.php with a subset of compile(), since the function is already doing more than it is advertising. If you prefer to supply arguments to compile(), please make a suggestion.

If you agree on approach, please merge for 5.2 as well.

### DIFF
--- a/UPGRADE-5.3.md
+++ b/UPGRADE-5.3.md
@@ -8,6 +8,7 @@ This changelog references changes done in Shopware 5.3 patch versions.
 
 ### Additions
 
+* Added command option `--current` to theme:cache:generate to compile from current asset timestamp and skip cache clear
 * Added config element `displayOnlySubShopVotes` to display only shop assigned article votes
 * Added parameter `displayProgressOnSingleDelete` to `Shopware.grid.Panel` to hide progress window on single delete action
 * Added parameter `expression` in `Shopware.listing.FilterPanel` to allow define own query expressions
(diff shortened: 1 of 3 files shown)


Isengo1989:
@wbob I do not get exactly the purpose of not generating a new file over timestamp since browser caches will not pick up the changes. Can you explain please? I know it's already merged but I would like to understand the benefits.

wbob:
You're benifiting by getting flexibility and regaining control. Imagine you need to build new assets in an environment with multiple nodes and a common httpcache. Without `--current`, every node receiving the command not in the very same second would reset the timestamp and httpcache again and again, effectively pulling the rug from the other nodes until the last node is reached.
I'm sure there are alternative setups/workarounds (nfs, or lsyncd from a master node, see cluster guide) and methods (a request to trigger the build) to make sure the assets are present before getting end-user requests, but why not getting a little help from the app instead?
Too, you want to control the moment of the httpcache invalidation: when every node has the new assets ready. This is why `compile()` is unfit to build a new theme if you orchestrate commands individually per node in a multinode environment.

PR #1158, opened 2017-06-12
Fix null being a string when persisting a date attribute and the value is empty.

bestit-el-bardan:
## Description

Fixes an issue when saving a date attribute with an empty value.
Currently this leads to "0000-00-00" being saved to the Database which then ExtJS cannot render correctly: https://i.imgur.com/Tnhqrqg.png

| Why?                    | Fixes a bug where ExtJS cannot render a field value correctly. |
| BC breaks?              | No |
| Tests exists & pass?    | Cannot find tests. |
| How to test?            | Add a new date attribute, enter an empty value and save. Check the value in the DB, it should be NULL not 0000-00-00. |

### DIFF
--- a/engine/Shopware/Bundle/AttributeBundle/Service/DataPersister.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Service/DataPersister.php
@@ -227,7 +227,7 @@ class DataPersister
             $value = $data[$column->getName()];
 
             if ($this->isDateColumn($column) && empty($value)) {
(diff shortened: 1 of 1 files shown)


janbuecker:
This has recently been an issue for us and we've already fixed it. It has been merged today here 9b66bd52fe29f382831fe5eedbd65eb0d3ea271e.

Thanks @bestit-el-bardan.

PR #1159, opened 2017-06-12
Fix proxy generation for classes implementing type-hinted interfaces

svenmuennich:
## Description

When [`Enlight_Hook_ProxyFactory`](https://github.com/shopware/shopware/blob/74c0df5115eb7d79935372c1f34b3fa508da962a/engine/Library/Enlight/Hook/ProxyFactory.php) generates a proxy class, all hooked methods of the original class are implemented, defining all parameters defined in the original method signatures. However, only the parameter and optionally a default value is read from the original signature and added to the proxy method signature. That is, no parameter type hints are copied over, e.g.:

```php
interface FooInterface
{
    /** 
     * @param array $a
     */
    public function bar(array $a = null);
}

class Foo implements FooInterface
{
    /** 
     * @inheritdoc
     */
    public function bar(array $a = null)
    {
        // ...
    }
}

class FooProxy
{
    public function bar($a = null)
    {
        // ...
    }
}
```

As the example shows, the generated method signature for `FooProxy::bar()` does not match the signature defined in the interface `FooInterface` 100%, because of the missing type hint of parameter `$a`. This results in a fatal error, as also stated in a note in the [PHP docs on interfaces](http://php.net/manual/en/language.oop5.interfaces.php#language.oop5.interfaces.implements):

> The class implementing the interface must use the exact same method signatures as are defined in the interface. Not doing so will result in a fatal error.

This PR fixes `Enlight_Hook_ProxyFactory::generateMethods()` to check all added parameters for a type hint and, if present, adds it to the respective parameter declaration in the proxy method signature:

```php
class FooProxy
{
    public function bar(array $a = null)
    {
        // ...
    }
}
```

| Why?                    | Currently `Enlight_Hook_ProxyFactory::generateMethods()` generates code that results in a fatal error when hooking a method defined in an interface (see above). |
| Tests exists & pass?    | This PR adds unit test for `Enlight_Hook_ProxyFactory::generateMethods()` |
| Related tickets?        | #1065 |
| How to test?            | Run the unit test |

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -5,6 +5,7 @@ This changelog references changes done in Shopware 5.2 patch versions.
 ## 5.2.25
 * Added notify event `Shopware_Modules_Basket_AddArticle_Added` in `engine/Shopware/Core/sBasket.php`
 * The event `Shopware_Modules_Export_ExportResult_Filter_Fixed` was added and now filters the processed export result. Previously with `Shopware_Modules_Export_ExportResult_Filter`, an instance of `Zend_Db_Statement_Pdo` was supplied, which could not be used to filter the actual result.
+* Fixes a fatal error caused by `Enlight_Hook_ProxyFactory::generateMethods()` generating proxy method sigantures that are not compatible with an interface implemented by the hooked class
 
 ## 5.2.23
 * Added conditional statement in `themes/Frontend/Responsive/frontend/_public/src/js/jquery.product-slider.js` to prevent the jquery plugin from sending ajax requests indefinitely
--- a/engine/Library/Enlight/Hook/ProxyFactory.php
+++ b/engine/Library/Enlight/Hook/ProxyFactory.php
@@ -276,6 +276,20 @@ class <namespace>_<proxyClassName> extends <className> implements Enlight_Hook_P
                     $array_params .= ', ';
                 }
 
+                // Add parameter type/class for interface compliance
+                if (method_exists($rp, 'hasType') && $rp->hasType()) {
+                    // PHP >= 7.0 only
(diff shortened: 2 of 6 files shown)


shyim:
Related PR: https://github.com/shopware/shopware/pull/1065#issuecomment-304790460

svenmuennich:
Sorry, I must have overlooked this one. Still, I think this should be fixed now instead of waiting for the day https://github.com/Ocramius/ProxyManager is added to shopware, if it will ever happen.

svenmuennich:
Well, the point is that you **do** need a type hint when hooking a method that implements an interface, in whose methods type hints are used (as also described in my original PR description). The PHP docs are pretty clear about that fact:

> The class implementing the interface must use the exact same method signatures as are defined in the interface. Not doing so will result in a fatal error.

So currently it is impossible to hook a method, if it is defined in an interface. And since many components/services in shopware implement respective interfaces, leaving this unfixed means that it is impossible to hook many of their methods.

bcremer:
Absolutely correct @svenmuennich.
A class implementing or extending an existing interface or class must declare all methods with with the same types.

This behavior in PHP is actually a violation of the "Liskov substitution principle" wich states that the input of a method should be allowed to be widened.

This will be fixed in the upcoming PHP 7.2 release (https://wiki.php.net/rfc/parameter-no-type-variance).

Example:
https://3v4l.org/mi2SJ

The inclusion of https://github.com/Ocramius/ProxyManager should be do-able without major Breaks to the public API.

svenmuennich:
I just sent a new PR #1174 that refactors `Enlight_Hook_ProxyFactory` to use [`ocramius/proxy-manager`](https://github.com/Ocramius/ProxyManager).

PR #1162, opened 2017-06-14
Fixed issue with PHP notice when no id for an option is defined

synonymous1984:
## Why
When creating or upgrading an article via API resources a PHP notice if thrown if the $option['id'] key is not set (which is no required field).

## How to test
Send an array like the following to article resource for update operation:

```
$params = array(
                'images' => array(
                    array(
                        'mediaId' => $media->getId(),
                        'options' => array(
                            array(
                                array(
                                    'name' => '60 Stück'
                                )
                            )
                        )
                    )
                )
            );

$articleResource->update($article->getId(), $params);
```
This operation throws the following notice:
**Notice: Undefined index: id in /engine/Shopware/Components/Api/Resource/Article.php on line 1846**

## Fix
The fix ensures that the key "id" exists in the option array.

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Article.php
+++ b/engine/Shopware/Components/Api/Resource/Article.php
@@ -1843,7 +1843,7 @@ class Article extends Resource implements BatchInterface
             foreach ($mappingData as $option) {
                 $conditions = [];
 
-                if ($option['id']) {
+                if (isset($option['id']) && $option['id']) {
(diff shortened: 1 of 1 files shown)


Phil23:
Thank you @synonymous1984 for your contribution.

Your change was merged with 88bce56994417394f96fc56c6c12574125b018f9.

PR #1164, opened 2017-06-15
Changed orderBy parameter to be optional

Gugiman:
OrderBy parameter can be optional in getListQueryBuilder() of the category repository class because it already has a default order and a condition to check if the "additonal" order array is empty.

With this PR the getListQuery() method can be called without an empty array as second parameter when it is not needed.

**The error thar appeared when second param leaves unsetted just occurs on PHP 7. I'm not quite sure why.**

## Description

| Why?                    | See above
| How to test?            | Call getListQuery of Shopware\Models\Category\Category |
| Requirements met?       | Think so |

### DIFF
--- a/engine/Shopware/Models/Category/Repository.php
+++ b/engine/Shopware/Models/Category/Repository.php
@@ -98,7 +98,7 @@ class Repository extends ModelRepository
      *
      * @return \Doctrine\ORM\Query
      */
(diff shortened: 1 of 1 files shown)


jenskueper:
Hello Gugiman,

thank you for your contribution! I changed the commit to match our code style. It will be merged soon.

PR #1166, opened 2017-06-19
Fix comparison due to which all snippets are being overwritten

m1pl:
## Description

| Why?                    | All snippets are overwritten, regardless whether _dirty_ is set or not. |
| Related tickets?        | none of which I know |
| How to test?            | Mark a snippet as _dirty_ and run the _sw:snippets:to:db_ CLI command. Without this commit, the snippet is being overwritten. |

This bug has been introduced with 97f9cb3ead4b850fd984a37ced007e1597f89f97 and has been released since v5.2.23. Because _fetchAll_ only returns strings, the comparison without type casting (===) between a string and an integer can never be true. This is a major bug, we already had multiple customers complaining about wrong snippets.


mitelg:
thanks for your contribution @mihaiplasoianu :+1:

PR #1167, opened 2017-06-21
take displayDate for documentDate

tinect:
## Description

| Why?                    | It's a wrong way to ignore displayDate while saving a document. For sample: If a customer wanted to change his name in the created Invoice, you have to regenerate the Invoice. in Backend you can set "displayDate" but any interface for the accountancy cannot see this date, so every generated Invoice will get a new date. For accountancy it will be a new invoice. So there would be 2 options. 1. this OR 2. make it impossible to update a invoice in backend :-) |
| BC breaks?              |  |
| How to test?            | Create a invoice, generate a new invoice after a day with the displayDate of yesterday. Now the date of the document will be the date of update, recarding this commit it'll be the displayDate. |
| Requirements met?       | hope so |

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -580,7 +580,7 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
         if (!empty($checkForExistingDocument['ID'])) {
             // Document already exist. Update date and amount!
             $update = '
-            UPDATE `s_order_documents` SET `date` = now(),`amount` = ?
+            UPDATE `s_order_documents` SET `date` = ?,`amount` = ?
             WHERE `type` = ? AND userID = ? AND orderID = ? LIMIT 1
             ';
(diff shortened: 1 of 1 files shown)


htkassner:
Hello @tinect. We don´t want to change this behaviour because we don´t want to allow the user to take influence on the "creation" date of the document. The generated document will correctly contain the display date. We think the document should always have the "correct" creation date due to accountancy rules. We do not see the benefit of changing this globally.
Thank you anyways for your thoughts and your contribution. We will close this pull request.

tinect:
Hello @htkassner 
I cannot understand that.
Why is it correct to change "creation date" in database, but allow manipulating the date in the document? Sounds inconsistent to me. 

Thank you for some information about that.

htkassner:
Hello @tinect,
it is mainly that we dont want do use the display date as new "creation" date because that would be a changed behaviour and could be unexpected for many users. The benefit of this change is not big enough. I hope you understand that.

PR #1168, opened 2017-06-22
New filter-event in PriceCalculationService.

jkrzefski:
Puts the price through a filter-event to let subscribers alter it.

## Description

| Why?                    | With the PriceCalculationService we have a central place for all price-calculations in the system (except the basket). It would make sense to give plugins the ability to alter the prices there. |
| Tests exists & pass?    | not yet |
| Related tickets?        | Not that I know of any. |
| How to test?            | Use the new event in a plugin and alter the price there. |


shyim:
You could also decorate the service, for that reason shopware build for any service a interface

jkrzefski:
That is correct. But I fail to see a way to keep most of the logic intact if you want to alter the price at that point. I built myself a workaround that consists of a decorator, but due it being private I cannot simply override the calculatePrice()-method. If I am missing out on some features a decorator could provide, please let me know.

PR #1174, opened 2017-06-29
Use 'ocramius/proxy-manager' in Enlight_Hook_ProxyFactory

svenmuennich:
## Description

When [`Enlight_Hook_ProxyFactory`](https://github.com/shopware/shopware/blob/74c0df5115eb7d79935372c1f34b3fa508da962a/engine/Library/Enlight/Hook/ProxyFactory.php) generates a proxy class, all hooked methods of the original class are implemented, defining all parameters defined in the original method signatures. However, only the parameter and optionally a default value are currently read from the original signature and added to the proxy method signature. That is, no parameter type hints are copied over:

```php
interface FooInterface
{
    /** 
     * @param array $a
     */
    public function bar(array $a = null);
}

class Foo implements FooInterface
{
    /** 
     * @inheritdoc
     */
    public function bar(array $a = null)
    {
        // ...
    }
}

/**
 * The proxy class as generated by Enlight_Hook_ProxyFactory.
 */
class FooProxy extends Foo implements Enlight_Hook_Proxy
{
    // ...

    public function bar($a = null)
    {
        // ...
    }
}
```

As the example shows, the generated method signature for `FooProxy::bar()` does not match the signature defined in the interface `FooInterface` 100%, because of the missing `array` type hint of parameter `$a`. This results in a fatal error, as also stated in a note in the [PHP docs on interfaces](http://php.net/manual/en/language.oop5.interfaces.php#language.oop5.interfaces.implements):

> The class implementing the interface must use the exact same method signatures as are defined in the interface. Not doing so will result in a fatal error.

This PR changes `Enlight_Hook_ProxyFactory` to use `ocramius/proxy-manager` for generating the proxy class code (as proposed in #1065 and #1159):

```php
/**
 * The proxy class as generated by the revised Enlight_Hook_ProxyFactory.
 */
class FooProxy extends Foo implements \Enlight_Hook_Proxy
{
    // ...

    /**
     * @inheritdoc
     */
    public function bar(array $a = null)
    {
        // ...
    }
}
```

| Why?                    | Currently `Enlight_Hook_ProxyFactory::generateMethods()` generates code that results in a fatal error when hooking a method defined in an interface (see above). |
| Related tickets?        | #1065, #1159 |
| How to test?            | Run the unit test |

### DIFF
--- a/UPGRADE-5.3.md
+++ b/UPGRADE-5.3.md
@@ -194,6 +194,7 @@ This changelog references changes done in Shopware 5.3 patch versions.
 * Changed `Shopware_Controllers_Widgets_Listing::productSliderAction` to `Shopware_Controllers_Widgets_Listing::productsAction`
 * Changed snippet `DetailBuyActionAdd` in `snippets/frontend/detail/buy.ini`, it now contains <span> tags
 * Changed snippet `ListingBuyActionAdd` in `snippets/frontend/listing/box_article.ini`, it now contains another <span> tag
+* Changed `Enlight_Hook_ProxyFactory` to use `ocramius/proxy-manager` for generating proxy classes
  
 ### Removals
 
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,8 @@
         "ramsey/uuid": "3.5.2",
         "setasign/fpdf": "1.8.1",
         "setasign/fpdi": "1.6.1",
-        "roave/security-advisories": "dev-master"
+        "roave/security-advisories": "dev-master",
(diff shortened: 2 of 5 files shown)


OliverSkroblin:
Nice work @svenmuennich, i will take a look

bcremer:
@OliverSkroblin Github failed to close this PR automatically.
@svenmuennich Great feature, I'm glad it was merged so quickly :tada:

svenmuennich:
Awesome! Thank you for this super fast merge 😃

PR #1175, opened 2017-06-29
Fixed field names in Shopware.apps.Base.model.Address

oktupol:
## Description

| Why?                    | The fields `firstname` and `lastname` are written with lowercase n's in the [Doctrine-Model](https://github.com/shopware/shopware/blob/5.2/engine/Shopware/Models/Customer/Address.php) and with uppercase N's in the ExtJs-Model. Since the field names are case sensitive, there is no simple way to access those fields' values from ExtJs. |
| How to test?            | No easy way to test this, since there is no store tied to this model. You could create a new backend module and associate its model with the Address model. |
| Requirements met?       | kinda |


roman-1983 (changes requested):
In `Customer/Address` the property is `$firstname` and in in `Customer/Billing` it is `$firstName`.
I think this should also be fixed in this PR.

PR #1176, opened 2017-06-29
RestApi-Plugin also checks for raw JSON post, on FrontPreDispatch

Diu:
## Description
 This PR allows JSON data to be sent as a POST parameter.
When processing RestApi requests, the plugin now checks for raw JSON in POST, if nothing was found in the body.

| Why?                    | The current media endpoint allows file uploads in the body, but restricts the remaining data request to pure POST parameters. |
| How to test?            | Send JSON as POST parameter with 'application/json' as name. |
| Requirements met?       | mostly |

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -7,6 +7,7 @@ This changelog references changes done in Shopware 5.2 patch versions.
 [View all changes from v5.2.26...v5.2.27](https://github.com/shopware/shopware/compare/v5.2.26...v5.2.27)
 
 * Fixed id collision in `themes/Backend/ExtJs/backend/base/component/Shopware.ModuleManager.js`
+* Added check for JSON data in POST parameter, when processing RestApi requests
 
 # 5.2.26
 
--- a/engine/Shopware/Plugins/Default/Core/RestApi/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Core/RestApi/Bootstrap.php
@@ -146,7 +146,12 @@ class Shopware_Plugins_Core_RestApi_Bootstrap extends Shopware_Components_Plugin
             if ($rawBody != '') {
                 $input = Zend_Json::decode($rawBody);
             } else {
-                $input = null;
+                $rawPost = $request->getPost('application/json');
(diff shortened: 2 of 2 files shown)


Diu:
I want to upload images/media with the native HTTP file upload, but also transfer the data with JSON, as in every other endpoint.
I think it is weird that this is the only endpoint, which doesn't accept JSON as a data source.

Also, to be honest, i didn't want to implement a method, which converts the media object into HTTP array parameters, just for one use case.

Maybe there are other ways, i can't see right now.

svenmuennich:
Do you talk about uploading images via a `POST` to the `/api/media` endpoint? This is already possible since 02b6258d10f712710b0849080072f400d6c2e742. You just have to use `multipart/form-data` and add the file in the `file` section. In that case all other `POST` values will be parsed from the request as well.

Diu:
Yes, that is what i wanted to enhance, because it doesn't parse JSON submitted via POST.

svenmuennich:
I think every _good_ HTTP client should be able to encode more complex objects in `multipart/form-data` bodies. E.g. the JSON object

```json
{
  "foo": {
    "bar": "value"
  }
}
```

should look something like the following as `multipart/form-data` parameters:

```
foo[bar]=value
```

I do understand your intent though. You like to send a `multipart/form-data` that looks something like this:

```
--SomeBoundary
Content-Disposition: form-data; name="json"
Content-Type: application/json

{"foo": {"bar": "value"}}
--SomeBoundary
Content-Disposition: form-data; name="file"
Content-Type: image/jpeg

<THE FILE DATA>
--SomeBoundary--
```

This doesn't look like a bad idea to me, but IMHO the proposed fix does not quite capture it. It would  probably be better to parse the JSON of all _parts_ whose `Content-Type` is `application/json` instead of limiting the scope to a single part `name`.

Diu:
I never expected my PR to get accepted, in the form I submitted it. But to start a discussion and show my intention, it was good enough.
Your last idea looks good. But after a brief look into it, I'd rather transform my media object into POST parameters, as I don't have that much time to work on this issue.
Thanks anyway for your input.

PR #1177, opened 2017-07-01
Remove Warning Messages on command line

s-bessing:
If the building thumbnails on the command line warning messages a printed if the file does not exists.

## Description

| Why?                    | command line not readable |
| BC breaks?              | yes/no |
| Related tickets?        | none |
| How to test?            | execute console sw:thumbnail:generate white missing thumbs |
| Requirements met?       | don't know|

### DIFF
--- a/engine/Shopware/Bundle/MediaBundle/OptimizerService.php
+++ b/engine/Shopware/Bundle/MediaBundle/OptimizerService.php
@@ -47,6 +47,9 @@ class OptimizerService implements OptimizerServiceInterface
      */
     public function optimize($filepath)
     {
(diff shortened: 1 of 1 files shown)


htkassner:
Thank you for your contribution. Can you still confirm this problem in our 5.3 version? We have several checks implemented here which check if a file exists. So maybe you could explain in which scenario you can reproduce this. That would be very helpful.

htkassner:
Hello @s-jakob, due to the fact that we cannot reproduce this problem in out 5.3 version, we will close this PR right now. Feel free to open another one with a more detailed description if you can still reproduce this. Thank you for your contribution to shopware.

PR #1178, opened 2017-07-03
Set s_articles_attributes.articleID null in demo-data

shyim:
## Description

| Why?                    | Shopware fills NULL in s_articles_attributes.articleID since 5.2. This PR corrects all demo articles |
| How to test?            | Open s_articles_attributes table |
| Requirements met?       | Yep |

Please change that also in the plugin store demo data or make it opensource ;)


samuelvogel:
We tried to remove the field with #671 already, which would actually be the correct path forward. Maybe this should be merged for Shopware 5.3?

shyim:
Unnecessary with #1505

PR #1179, opened 2017-07-03
Fix execution model of `replace` hooks

svenmuennich:
## Description

Currently the execution model of `replace` hooks is broken, once more than one `replace` hook exists for the same method. The problem is that all `replace` hooks are executed sequentially and each of these hooks has the opportunity to call `executeParent()`. Hence, if all `replace` hooks do call `executeParent()`, the original (parent) implementation is executed more than once. This basically breaks `replace` hooks on methods that have side effects, because these side effects are applied for every call of `executeParent()`. As a result, it is currently very risky to have more than one `replace` hook on the same method. In fact, we at Pickware refrain from using `replace` hooks altogether because of that behaviour!

Example:

```php
class HookTarget
{
    public function fooBar()
    {
        echo "HookTarget: fooBar!\n";
    }
}

class FooSubscriber implements \Enlight\Event\SubscriberInterface
{
    //...

    /**
     * First hook
     */
    public function onHookTargetFooBar(\Enlight_Hook_HookArgs $args)
    {
        echo "FooSubscriber: before executeParent()\n";
        $args->getSubject()->executeParent($args->getMethod(), $args->getArgs());
        echo "FooSubscriber: after executeParent()\n";
    }
}

class BarSubscriber implements \Enlight\Event\SubscriberInterface
{
    //...

    /**
     * Second hook
     */
    public function onHookTargetFooBar(\Enlight_Hook_HookArgs $args)
    {
        echo "BarSubscriber: before executeParent()\n";
        $args->getSubject()->executeParent($args->getMethod(), $args->getArgs());
        echo "BarSubscriber: after executeParent()\n";
    }
}
```

Output:

```
FooSubscriber: before executeParent()
HookTarget: fooBar!
FooSubscriber: after executeParent()
BarSubscriber: before executeParent()
HookTarget: fooBar!
BarSubscriber: after executeParent()
```

The desired execution model of `replace` hooks prevents multiple calls of the original implementation, if all hooks call `executeParent()` only once. That said, the `replace` hooks should be parents of each other, based on the position they are sorted by in the event manager. This execution model is very similar to the one used by _ExtJS_ and hence should be more familiar and most importantly predictable to shopware developers.

This commit refactors the whole hook manager as well as changes the generated proxy classes (which is easy now thanks to #1174) to add an execution context, which executes the registered `replace` hooks as a chain in which each call to `executeParent()` calls the respective next hook, if available. Only the last (highest position) hook’s call to `executeParent()` calls the original implementation of the hooked method. Finally the respective return values can be passed down the chain again. That is, the `replace` hook with the lowest position is able to ultimately override the return value, before it is passed to any `after` hooks:

```php
class HookTarget
{
    public function fooBar()
    {
        echo "HookTarget: fooBar!\n";
    }
}

class FooSubscriber implements \Enlight\Event\SubscriberInterface
{
    //...

    /**
     * First hook
     */
    public function onHookTargetFooBar(\Enlight_Hook_HookArgs $args)
    {
        echo "FooSubscriber: before executeParent()\n";
        $args->getSubject()->executeParent($args->getMethod(), $args->getArgs());
        echo "FooSubscriber: after executeParent()\n";
    }
}

class BarSubscriber implements \Enlight\Event\SubscriberInterface
{
    //...

    /**
     * Second hook
     */
    public function onHookTargetFooBar(\Enlight_Hook_HookArgs $args)
    {
        echo "BarSubscriber: before executeParent()\n";
        $args->getSubject()->executeParent($args->getMethod(), $args->getArgs());
        echo "BarSubscriber: after executeParent()\n";
    }
}
```

Output:

```
FooSubscriber: before executeParent()
BarSubscriber: before executeParent()
HookTarget: fooBar!
BarSubscriber: after executeParent()
FooSubscriber: after executeParent()
```

Of course it is still possible to execute the original implementation more than once, by calling `executeParent()` in the same hook repeatedly.

On a side note, this fix kind of makes `before` and `after` hooks obsolete, since it is now possible to just subscribe `replace` hooks, perform custom steps _before_ the original implementation, call `executeParent()` and then perform some more steps _after_ that call (just like in _ExtJS_).

**Edit:** Please also consider [my comment down below](https://github.com/shopware/shopware/pull/1179#issuecomment-313393092) regarding recursion.

| Why?                    | Having more than one `replace` hook on the same method that all call `executeParent()` results in the original/parent method to be executed more than once.|
| Tests exists & pass?    | yes (add new tests for the hook manager) |
| How to test?            | Just run the unit tests |

### DIFF
--- a/UPGRADE-5.3.md
+++ b/UPGRADE-5.3.md
@@ -36,6 +36,7 @@ View all changes from v5.3.0-RC1...v5.3.0-RC2](https://github.com/shopware/shopw
 * Merged `account/sidebar.tpl` and `account/sidebar_personal.tpl`
 * Moved snippets from `account/sidebar_personal.ini` to `account/sidebar.ini`
 * Changed `Enlight_Hook_ProxyFactory` to use `ocramius/proxy-manager` for generating proxy classes
+* Changed the execution model of `replace` hooks to prevent multiple calls of the hooked method, if more than one `replace` hook on the same method exists and all of them call `executeParent()` once
 
 ### Removals
 
--- a/engine/Library/Enlight/Hook/HookExecutionContext.php
+++ b/engine/Library/Enlight/Hook/HookExecutionContext.php
@@ -1,0 +1,183 @@
+<?php
+/**
+ * Enlight
+ *
+ * LICENSE
+ *
+ * This source file is subject to the new BSD license that is bundled
+ * with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://enlight.de/license
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@shopware.de so we can send you a copy immediately.
+ *
+ * @category   Enlight
+ * @package    Enlight_Hook
+ * @copyright  Copyright (c) 2017, shopware AG (http://www.shopware.de)
+ * @license    http://enlight.de/license     New BSD License
+ * @version    $Id$
+ * @author     Heiner Lohaus
+ * @author     $Author$
+ */
+
+/**
+ * The Enlight_Hook_HookExecutionContext represents a single execution of a hook.
+ *
+ * In order to execute a proxy method and all its 'before', 'replace' and 'after' hooks, a new context must be created
+ * and executed.
+ *
+ * @category   Enlight
+ * @package    Enlight_Hook
+ * @copyright  Copyright (c) 2017, shopware AG (http://www.shopware.de)
+ * @license    http://enlight.de/license     New BSD License
+ */
+class Enlight_Hook_HookExecutionContext
+{
+    /**
+     * @var Enlight_Hook_HookManager
+     */
+    protected $hookManager;
+
+    /**
+     * @var Enlight_Hook_HookArgs
+     */
+    protected $args;
+
+    /**
+     * @var int
+     */
+    protected $parentExecutionLevel = 0;
+
+    /**
+     * @param string $className
+     * @param string $method
+     * @param string $hookType
+     * @return string
+     */
+    public static function createHookEventName($className, $method, $hookType)
+    {
+        return $className . '::' . $method . '::' . $hookType;
+    }
+
+    /**
+     * @param Enlight_Hook_HookManager $hookManager
+     * @param Enlight_Hook_Proxy $class
+     * @param string $method
+     * @param array $args
+     */
+    public function __construct(
+        Enlight_Hook_HookManager $hookManager,
+        Enlight_Hook_Proxy $class,
+        $method,
+        array $args
+    ) {
+        $this->hookManager = $hookManager;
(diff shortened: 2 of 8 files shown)


UPGRADE-5.3.md
```diff
@@ -36,6 +36,7 @@ View all changes from v5.3.0-RC1...v5.3.0-RC2](https://github.com/shopware/shopw
 * Merged `account/sidebar.tpl` and `account/sidebar_personal.tpl`
 * Moved snippets from `account/sidebar_personal.ini` to `account/sidebar.ini`
 * Changed `Enlight_Hook_ProxyFactory` to use `ocramius/proxy-manager` for generating proxy classes
+* Changed the exeuction model of `replace` hooks to prevent mutliple calls of the hooked method, if more than one `replace` hook on the same method exists and all of them call `executeParent()` once
```
bcremer: Typo: execution

svenmuennich:
Please note that I just pushed some changes that fix some more complicated scenarios with recursive methods. Although my original implementation could handle recursion, it fell apart as soon as one of the replace hooks called `executeParent()` more than once. Of course this is a very special case and I can't imagine a scenario in which someone would want to do that, but nonetheless the behaviour in that case should be well defined. That is, repeated calls to a recursive `executeParent()` should not change the overall result of the original function call. Instead, the recursion should be repeated within the context of the hook, resulting in a tree-like call graph. First, the call graph of a simple recursion with depth `1`, in which `executeParent()` is only called once per hook:

```
  recMethod()
       |
    HOOK_R0
       |
executeParent()
       |
   METHOD_R0
       |
  recMethod()
       |
    HOOK_R1
       |
executeParent()
       |
   METHOD_R1
```

And this is how it looks when calling `executeParent()` twice in the `replace` hook:

```
  recMethod()
       |
    HOOK_R0 ----------------------------------
       |                                     |
executeParent()                       executeParent()
       |                                     |
   METHOD_R0                             METHOD_R0
       |                                     |
  recMethod()                           recMethod()
       |                                     |
    HOOK_R1 ---------------               HOOK_R1 ---------------
       |                  |                  |                  |
executeParent()    executeParent()    executeParent()    executeParent()
       |                  |                  |                  |
   METHOD_R1          METHOD_R1          METHOD_R1          METHOD_R1
```

OliverSkroblin:
Hi @svenmuennich,
thanks for the pull request. We decides to merge this changes with version 5.4, because the changes have a deep effect on the system and all existing plugins with replace hooks.

soebbing:
Hi @svenmuennich!
we had to revert your PR since it has one major drawback: It only works for public methods. Since the ExecutionContext does not extend the hooked class anymore, it is not able to work with protected methods like the current implementation does.

Maybe you can extend your solution to also work in that usecase? We would be happy to merge it again afterwards.

svenmuennich:
I can't believe that I missed this, sorry about that 😕 Anyway, I opened a new PR #1501 which includes support for `protected` methods.

PR #1180, opened 2017-07-03
Removed extends from box-emotion.tpl

shyim:
## Description

| Why?                    | The extends in the file doesn't have a effect, because the file overwrite the entire box-basic.tpl. It reduces the confusion why my box-basic doesnt have a effect :D |
| How to test?            | Remove it and see nothing happens |

### DIFF
--- a/themes/Frontend/Bare/frontend/listing/product-box/box-emotion.tpl
+++ b/themes/Frontend/Bare/frontend/listing/product-box/box-emotion.tpl
@@ -1,5 +1,3 @@
-{extends file="frontend/listing/product-box/box-basic.tpl"}
-
 {namespace name="frontend/listing/box_article"}
 
 {block name="frontend_listing_box_article"}


roman-1983:
When `box-basic.tpl` will get a new block , then `box-emotion.tpl` won't benefit from it.

roman-1983 (commented):
When `box-basic.tpl` will get a new block , then `box-emotion.tpl` won't benefit from it.

klarstil:
@roman-1983 That's not the case 'cause `box-emotion.tpl` overrides the most upper block of the `box-basic.tpl` therefore it was superfluous to extend from `box-basic.tpl`. Even if the `{extends}` call would be in place, new blocks are not reflected in the `box-emotion.tpl`. This PR doesn't changes the behavior of the `box-emotion.tpl` at all.

mitelg:
thanks for the PR @shyim :+1:

PR #1184, opened 2017-07-05
Replace File with UploadedFile

roman-1983:
*Why*
To keep the file-ending and treat every file as an uploaded file.
If you are using `File` for files added via API, Shopware tries to guess the file-extension. This is sometimes wrong.
It's better to treat the files as `UploadedFile` instead of `File` so Shopware will use the given file-extension.

*What*
Use `UploadedFile` instead of `File`

## Description

| Why?                    | To keep the file-ending and treat every file as an uploaded file.
If you are using File for files added via API, Shopware tries to guess the file-extension. This is sometimes wrong.
It's better to treat the files as UploadedFile instead of File so Shopware will use the given file-extension. |
| How to test?            | Upload a file via API.
Before: If you upload a `.stp` file, it will be saved as `.txt`
After: If you upload a `.stp` file, it will be saved as `.stp` |

### DIFF
--- a/themes/Backend/ExtJs/backend/customer/view/address/list.js
+++ b/themes/Backend/ExtJs/backend/customer/view/address/list.js
@@ -159,7 +159,7 @@ Ext.define('Shopware.apps.Customer.view.address.List', {
 
         return {
             columns: columns,
-            searchField: false,
+            searchField: true,
             detailWindow: 'Shopware.apps.Customer.view.address.detail.Window'
         }
     },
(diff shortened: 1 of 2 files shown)


themes/Backend/ExtJs/backend/customer/view/address/list.js
```diff
@@ -159,7 +159,7 @@ Ext.define('Shopware.apps.Customer.view.address.List', {
 
         return {
             columns: columns,
-            searchField: false,
+            searchField: true,
```
mitelg, line 162: what is your intention behind this change? I can't see any connection between uploaded files and the listing of addresses in the backend :wink: 

please provide another PR if you want to change this behaviour

mitelg (changes requested):

roman-1983:
@mitelg Sorry, something went wrong here.
Because i deleted my repository in the meanwhile, i had to create a new pull-request.

This one can be closed. Here is the new PR.
https://github.com/shopware/shopware/pull/1257

PR #1191, opened 2017-07-06
Force late-calls to `asyncReady` to be async

screeny05:
## Description
To stay consistent with the early call to `document.asyncReady`, the late call must also call the callback asynchronously. Otherwise Zalgo will be unleashed.

Issues arising from synchronously calling the callback may be:
```
console.log('1. before ready');

document.asyncReady(() => { console.log('3. is ready'); });

console.log('2. after ready');
```

This code has inconsistent behaviour depending on when it's executed.

If it gets executed before the page is ready, this will happen:
```
1. before ready
2. after ready
3. is ready
```

if the code instead gets called after the page is ready, this will happen:
```
1. before ready
3. is ready
2. after ready
```

See this blog post for more information, regarding this Issue:

_https://blog.ometer.com/2011/07/24/callbacks-synchronous-and-asynchronous/_

| Why?                    | Keep the API consistent |
| How to test?            | Example given above |

### DIFF
--- a/themes/Frontend/Bare/frontend/index/script-async-ready.tpl
+++ b/themes/Frontend/Bare/frontend/index/script-async-ready.tpl
@@ -14,7 +14,7 @@
 
         document.asyncReady = function (callback) {
             if (typeof callback === 'function') {
-                callback.call(document);
+                setTimeout(callback.apply(document), 0);
             }
         }
     });


Phil23:
Thank you @screeny05 for your contribution.

The change was merged with 8817b93bfd9d5967e26aa63e1ceb1902dc9106f3.

Have nice day!

PR #1192, opened 2017-07-06
Make loading attributes consistent

kekster:
Due to inconsistencies within the shopware core, some db items which should have an empty attribute by default, just have no attribute at all. This change makes sure that loading an attribute always returns an array.

| Why?                    | Due to inconsistencies within the shopware core, some db items which should have an empty attribute by default, just have no attribute at all. This change makes sure that loading an attribute always returns an array. |
| Related tickets?        | none |
| How to test?            | load an order attribute which doesnt exist |


mitelg:
thanks @kekster :+1:

PR #1195, opened 2017-07-06
Added the possibility to ignore url parameters for the cache key gene…

aragon999:
## Description

The problem is, that often additional query parameters are added to the url (e.g. from Google Adwords). With this, often unique, parameters a new HTTP Cache version is generated, although these parameters are only needed by some JavaScript tracking scripts. This PR adds the possibility to ignore such parameters. I have taken the parameters from the varnish cache documentation, which we use in several projects and added the parameter `_ga`. See here: https://www.varnish-software.com/wiki/content/tutorials/varnish/sample_vclTemplate.html#removing-google-analytics-added-parameters

| Why?                    | See above |
| BC breaks?              | not sure |
| How to test?            | Activate HTTP Cache and try to load the page with a parameter like gclid=123123, the `x-content-digest` should be the same for each value of the parameter and without the parameter |


soebbing:
Hi @aragon999!

We just merged your commit but changed your implementation a bit. Also we didn't want to define a set of filtered parameters in a patch release, so we added the parameters in the defaults but commented them out for the moment. We'll add them in 5.4.

Thanks for helping to make Shopware better! 💙

PR #1199, opened 2017-07-10
count orders after eventFilter

tinect:
## Description

| Why?                    | Adding orders from foreign source wouldn't have a nice result, cause getPagerStructure just looks for own orders. In our sample, we add orders live from out ERP and the pager is broken :-) |
| BC breaks?              | hope not |
| Tests exists & pass?    |  |
| How to test?            | Manipulate the result in Event Shopware_Modules_Admin_GetOpenOrderData_FilterResult with more orders and see what's happening with the pager |


tinect:
unnessary... sorry

PR #1204, opened 2017-07-11
Do not round voucher net values

samuelvogel:
## Description

Updated version of #915 against Shopware 5.3, please also see the discussion there.

Before this PR the net prices of vouchers were rounded to three decimal places in the basket. Articles in contrast have an exact net price, where the precisions is just limited by PHP itself.

This might seem like a minor thing, as only prices rounded to two decimal places are displayed to the customer. However this breaks the [PAYONE plugin](http://store.shopware.com/payone00261/payone-payment-fuer-shopware.html) for certain baskets which calculates the gross value itself. The calculation itself is correct but is off by one cent with some constellations because of the missing precisions of the vouchers net value and PHPs imperfect internal float handling.

Example basket with 7 % VAT:

| Position        | Quantity           | Value  |
| ------------- |:-------------:| -----:|
| Sample Article      | 1 | 112,50 € |
| Voucher 5 %      |      |   -5,63 €|
|      |     **SUM** |   106,87 €|

But the PAYONE plugin takes the rounded net value of the voucher `-5.257` and recomputes it to `-5.62499` which is then then rounded to the incorrect value of `-5.62` (off by 0,01 €). The code doing this can be found [here](https://github.com/PAYONE-GmbH/shopware-5/blob/v3.4.7/Frontend/MoptPaymentPayone/Components/Classes/PayoneParamBuilder.php#L1046) and is correct, it's just not fed with enough precision in case of vouchers.

Orders like this are rejected by the PAYONE API as the summed basket value displayed to the user is transmitted as well and cross checked against the sum of the individual positions. The error returned is `1610: Article list faulty or incomplete` in this case.

| Why?                    | see above |
| How to test?            | Test case is included in the PR. |
| Requirements met?       | sure |

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2086,14 +2086,14 @@ class sBasket
             }
         } else {
             if ($voucherDetails['taxconfig'] == 'default' || empty($voucherDetails['taxconfig'])) {
-                $tax = round($voucherDetails['value'] / (100 + $this->config->get('sVOUCHERTAX')) * 100, 3) * -1;
+                $tax = $voucherDetails['value'] / (100 + $this->config->get('sVOUCHERTAX')) * 100 * -1;
                 $taxRate = $this->config->get('sVOUCHERTAX');
                 // Pre 3.5.4 behaviour
             } elseif ($voucherDetails['taxconfig'] == 'auto') {
                 // Check max. used tax-rate from basket
                 $tax = $this->getMaxTax();
                 $taxRate = $tax;
-                $tax = round($voucherDetails['value'] / (100 + $tax) * 100, 3) * -1;
+                $tax = $voucherDetails['value'] / (100 + $tax) * 100 * -1;
             } elseif (intval($voucherDetails['taxconfig'])) {
                 // Fix defined tax
                 $temporaryTax = $voucherDetails['taxconfig'];
@@ -2102,7 +2102,7 @@ class sBasket
                     [$temporaryTax]
                 );
                 $taxRate = $getTaxRate;
-                $tax = round($voucherDetails['value'] / (100 + (intval($getTaxRate))) * 100, 3) * -1;
+                $tax = $voucherDetails['value'] / (100 + (intval($getTaxRate))) * 100 * -1;
             } else {
                 // No tax
                 $tax = $voucherDetails['value'] * -1;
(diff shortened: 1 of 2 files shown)


htkassner:
Hey @samuelvogel. I can confirm that this is working for the 7% calculation of your example. But I have another question. The calculated voucher values are directly stored into the s_order_basket. Payone, I guess, will use these values for calculation. When we save to the database we also lose precision. Thats why I get 5.62 in the 19% example when I calculate gross with the in s_order_basket stored value. It seems that the result totally depends on the prices that get calculated and in some cases these are correct and in some not. Can you confirm this error with 19% and your example prices? I calculated gross in calculateVoucherValues where full precision is available (5.63 correct) and in sGetBasketData after loadingBasketArticles from s_order_basket which results in 5.62.

How does Payone retrieve these values?

Edit: Look what we are doing if the shop is in net mode.
https://github.com/shopware/shopware/blob/6b1ae56879ee500d2e0030074589fc61d9d76283/engine/Shopware/Core/sBasket.php#L2322

We´re rounding the net value and then multiply tax rate which results in 5.6287. This value gets rounded another time which results in 5.63 when shopware calculates the net price. So I guess it is the difference in how gross is calculated.

samuelvogel:
Payone works on the "fresh" PHP basket which gets recalculated on every request and not the "stale" database values. At [shop.rapunzel.de](https://shop.rapunzel.de) the fix is deployed since 6 months and successfully prevents the problem from happening.

Can you reproduce the "wrong" value of `5.62` in the storefront actually? I can't in my own testing.

Also, as stated before, I think we can't go wrong with adding _more_ precision during the calculation. Even if it leads to a corner case which accidentally led to a seemingly more correct value before. Anyhow, we can't even reproduce this drawback in the real world right now, can we?

Also this is not an issue about the rounding strategy in general, as only one voucher can be used per order, therefore concerns about when to round (before multiplying by quantity or after) do not apply.

Just for reference, this get's stored in the `s_order_basket` when the PR is applied:
voucher rounding

htkassner:
Hey. No this was in my tests never an issue in the storefront. It is more that we have to decide between benefit, risks of breaking things etc. That Payone uses the "fresh" values makes then indeed a difference. We´ll discuss that.

htkassner:
Hello @samuelvogel, we evaluated this PR with a bigger group and came to the same conclusion. We do not want to change this behaviour of the basket right now and it will be reimplemented in our new version with the new basket. Thank you for your contribution. We always appreciate that.

PR #1207, opened 2017-07-13
SW-19299: Fix persisting of existing article entity if only attribute…

ck-1:
…s of articleDetails etc. are changed, but nothing from the article entity itselft.

## Description

| Why?                    | Saving of ArticleDetails in the backend fails |
| How to test?            | Save dimensions of an article only |


ck-1:
Doctrine will only check the owning side of an association for changes:
http://docs.doctrine-project.org/en/latest/reference/working-with-associations.html
So either persist the articleDetails or flush the article entity!

ck-1:
But this is the current error of the shopware backend. If I change articleDetails only, these changes are not saved. Handing over the article object to the flush method let's Doctrine save the article entity **including** all related entities, in this case the articleDetails.

This is because Doctrine will only check the owning side of an association for changes:
http://docs.doctrine-project.org/en/latest/reference/working-with-associations.html

So this fix would solve the problem. Or what is your solution so our backoffice could use the backend again?

PR #1208, opened 2017-07-14
Reset OPcache after extracting plugin zip

svenmuennich:
When downloading a plugin ZIP file from the community store or manually uploading it in the plugin manager, the contents of the ZIP file are extracted and, if the same plugin is already available in the file system, replace the existing files. However, if the OPcache is enabled and [`opcache.revalidate_freq`](http://php.net/manual/de/opcache.configuration.php#ini.opcache.revalidate-freq) is set to an arbitrary long interval (e.g. 2 minutes), you most likely run into the case that outdated plugin files are executed, until the OPcache is revalidated automatically. This is particularly problematic if the user runs the plugin update directly after downloading it, because some update steps that are only present in the downloaded `Bootstrap.php` might not be executed.
This commit fixes these issues by adding a single `opcache_reset()` call just after the actual extraction of the plugin ZIP file.

## Description

| Why?                    | After extracting a plugin zip file the plugin files loaded in the OPcache do not match the extracted files. |
| Tests exists & pass?    | n/a |
| How to test?            | Enable OPcache, set `opcache.revalidate_freq` to a long interval, change something in a plugin and upload the new version as a ZIP. |

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/DownloadService.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/DownloadService.php
@@ -123,6 +123,11 @@ class DownloadService
         } else {
             throw new \RuntimeException('No Plugin found in archive.');
         }
+
+        // Reset the OPcache to make sure no outdated versions of the unzipped files are loaded from the cache
+        if (extension_loaded('Zend OPcache') && ini_get('opcache.enable')) {
+            opcache_reset();
+        }
     }
 
     /**


engine/Shopware/Bundle/PluginInstallerBundle/Service/DownloadService.php
```diff
@@ -123,6 +123,9 @@ public function extractPluginZip($file, $pluginName)
         } else {
             throw new \RuntimeException('No Plugin found in archive.');
         }
+
+        // Reset the OPcache to make sure no outdated versions of the unzipped files are loaded from the cache
+        opcache_reset();
```
shyim: You should check like in CacheManager is opcache active and loaded :) otherwise can the function be undefined and break the uploading. Like on development environments 
https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Components/CacheManager.php#L228

svenmuennich: Good find! Thanks for pointing that out 👍

dnoegel:
Hi,
IMHO the PluginExtractor and LegacyPluginExtractor are already handling this exact case:

https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Bundle/PluginInstallerBundle/Service/LegacyPluginExtractor.php#L49
https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginExtractor.php#L96

Or did I miss something?

svenmuennich:
Yea, you are right. The problem we had can actually not be solved by clearing the OPCache upon extracting the zip file. We built a workaround into our plugins instead.

PR #1211, opened 2017-07-19
Fix critical random plugins disappearing bug when refreshing plugin list concurrently

audaki:
## Description

| Why?                    | There is a critical bug when the plugin list is refreshed concurrently (e.g. by two users or two tabs using the backend at the same time, another possibility which happened several times is when a timed license is expired, and you opened the backend and then the plugin manager, the plugin list was also loaded twice concurrently. And whenever that happens there is a chance that the plugins have different refresh_date in s_core_plugins, which will cause those to be deleted later on in the cleanupPlugins because it only relied on this column in the database. I'd recommend a rewrite of these functions to pass the expired plugin IDs directly in the code but for now this bug is also 100% fixed by wrapping these writes and reads into a transaction, so that the concurrent requests do not see the refresh_dates of other requests, which I've done in this pull request. |
| Related tickets?        | I have not found a ticket |
| How to test?            | Execute the refreshPluginListAction several times concurrently (e.g try with two people or do it with CURL) until you suddenly have disappearing plugins... Then try again with this pull request |

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/InstallerService.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/InstallerService.php
@@ -317,10 +317,20 @@ class InstallerService
     {
         $refreshDate = new \DateTimeImmutable();
 
+        // Start transaction so that concurrenct requests can't overwrite the refresh_date, which causes plugins
+        // to be deleted in cleanupPlugins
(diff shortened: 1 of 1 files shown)


janbuecker:
Close as duplicate of #1212. See my comment [here](https://github.com/shopware/shopware/pull/1212#issuecomment-316641186).

PR #1212, opened 2017-07-19
Fix critical random plugins disappearing bug when refreshing plugin list concurrently

audaki:
## Description

| Why?                    | There is a critical bug when the plugin list is refreshed concurrently (e.g. by two users or two tabs using the backend at the same time, another possibility which happened several times is when a timed license is expired, and you opened the backend and then the plugin manager, the plugin list was also loaded twice concurrently. And whenever that happens there is a chance that the plugins have different refresh_date in s_core_plugins, which will cause those to be deleted later on in the cleanupPlugins because it only relied on this column in the database. I'd recommend a rewrite of these functions to pass the expired plugin IDs directly in the code but for now this bug is also 100% fixed by wrapping these writes and reads into a transaction, so that the concurrent requests do not see the refresh_dates of other requests, which I've done in this pull request. |
| Related tickets?        | I have not found a ticket |
| How to test?            | Execute the refreshPluginListAction several times concurrently (e.g try with two people or do it with CURL) until you suddenly have disappearing plugins... Then try again with this pull request |

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/InstallerService.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/InstallerService.php
@@ -317,10 +317,20 @@ class InstallerService
     {
         $refreshDate = new \DateTimeImmutable();
 
+        // Start transaction so that concurrenct requests can't overwrite the refresh_date, which causes plugins
+        // to be deleted in cleanupPlugins
+        $this->em->beginTransaction();
+
+        // Ensure consistent state (@see https://dev.mysql.com/doc/refman/5.7/en/innodb-transaction-isolation-levels.html)
+        // Consistent read is only guaranteed after the first read
+        $this->em->getConnection()->fetchAll('SELECT id, refresh_date FROM s_core_plugins');
+
         $this->pluginInstaller->refreshPluginList($refreshDate);
         $this->legacyPluginInstaller->refreshPluginList($refreshDate);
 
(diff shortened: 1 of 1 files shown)


janbuecker:
Thank you @EarthlingKira for taking the time to debug this issue! Your requested change makes sense and seems to fix the issue. We'll test it a little bit further in some scenarios to verify that nothing else will be affected by this change.

I'll close the other pull requests as the `5.3` branch will be our new master shortly. We'll merge the changes into `5.4` with the next up-merge and may backport it to `5.2` if it is necessary.

jenskueper:
@EarthlingKira I just wanted to give you an short update about the progress. First of all, thank you again for finding this critical bug! We were able to fix the issue but we decided to take a different approach. After evaluating the logic behind the cleanupPlugins method, we came to the conclusion that using a SQL transaction would indeed fix the race condition but if a plugin is removed from the file system, the cleanupPlugins routine would still remove the plugin from the s_core_plugins table without being able to run the uninstall logic of the affected plugin.

This can cause an inconsistency and can cause a state in which the user is unable to reinstall the plugin again. We therefore decided to remove the cleanupPlugins logic completely, which has the positive side effect that this race condition is no longer an issue.

We made theses changes based on your commit since you found the bug in the first place. I hope you are fine with this solution.

These changes will be merged soon.

jenskueper:
Merged with commit https://github.com/shopware/shopware/commit/8314fd46880fc32d25f3a8d297037655d979e230

PR #1214, opened 2017-07-19
Load attributes of product exports in backend

ckilb:
This will load attributes of a product export if it's dialog was opened in the backend.
Currently after you save a product export with it's attributes the attributes will be lost after you open the dialog (and save the export) a second time - because all attribute fields will be empty again.


mitelg:
thanks for your contribution @ckdot 
the changes will be part of the next release

PR #1221, opened 2017-07-25
Patch smarty recursive mkdir race condition

fixpunkt:
**This issue is affecting multiple customers, rendering the backend unusable after a cache clear for some of them. Because if this, we would really appreciate this patch being backported to 5.2 soon.**

This is a backport of the code in https://github.com/smarty-php/smarty/pull/379 to Shopware's vendored Smarty 3.1.8.

## Description

| Why?                    | PHP's recursive `mkdir()` implementation is not concurrency-safe, which means smarty compilation on multiple requests is not concurrency safe, which has been causing significant problems for Shopware ERP customers in the last few weeks. See description below for details.|
| BC breaks?              | No. |
| Tests exists & pass?    | not really unit-testable |
| Related tickets?        | dunno |
| How to test?            | Clear the cache and reload the backend. Repeat a couple of times. |
| Requirements met?       | Yes. |

## Gory details

PHP's recursive `mkdir()` implementation may fail to create a nested directory structure when another process tries to create a nested directory structure that shares a prefix with the original process at the same time. This is a known issue (cf. https://bugs.php.net/bug.php?id=35326), but the PHP maintainers have decided that providing concurrency-safe behaviour is out of scope for PHP and should be handled by the application instead.

Here is [a small demonstrator script](https://gist.github.com/fixpunkt/5881bfeba51d64926c99d0339d3b8c02) that triggers the race condition in PHP's recursive `mkdir()` implementation and shows that my PHP reimplementation of recursive `mkdir()` behaves correctly even in the concurrent scenario.

Smarty as used in shopware uses a nested directory structures for its template cache. This directory structure is created on-the-fly here: https://github.com/shopware/shopware/blob/5.3/engine/Library/Smarty/sysplugins/smarty_internal_write_file.php#L40. When multiple requests try to compile the same template after a cache has been cleared, the race condition within `mkdir()`-recursive may cause both requests to fail, resulting in a template compilation error, which is shown as an error in the backend and prevents some Subapplications from loading correctly.

Some changes in ERP have resulted in this bug being triggered much more frequently over the last few weeks. Some of our customers have been hitting this issue with increasing frequency lately, which made the backend completely unusable for them.

Props to Sebastian Grüny (intoCommerce) and @stefanheppenheimer for their help in figuring out this issue!

### DIFF
--- a/engine/Library/Smarty/sysplugins/smarty_internal_write_file.php
+++ b/engine/Library/Smarty/sysplugins/smarty_internal_write_file.php
@@ -36,8 +36,9 @@ class Smarty_Internal_Write_File {
 
         $_dirpath = dirname($_filepath);
         // if subdirs, create dir structure
-        if ($_dirpath !== '.' && !file_exists($_dirpath)) {
-            mkdir($_dirpath, $smarty->_dir_perms === null ? 0777 : $smarty->_dir_perms, true);
+        if ($_dirpath !== '.' && !@self::ensureDirectoryExists($_dirpath, $smarty->_dir_perms)) {
+            error_reporting($_error_reporting);
+            throw new SmartyException("unable to create directory {$_dirpath}");
         }
 
         // write to tmp file, then move to overt file lock race condition
@@ -83,5 +84,27 @@ class Smarty_Internal_Write_File {
 
         error_reporting($_error_reporting);
         return $success;
+    }
+
+    /**
+     * Recursively creates the missing parts of a directory path in a manner that is concurrency-safe.
+     *
+     * @see https://bugs.php.net/bug.php?id=35326
+     *
+     * @param string $pathname a (nested) directory path to create
+     * @param integer $mode the permission to use
+     * @return bool true iff the directory path was successfully created
+     */
+    private static function ensureDirectoryExists($pathname, $mode)
+    {
+        $path_segments = explode(DIRECTORY_SEPARATOR, $pathname);
+
+        $current_pathname = '';
+        foreach ($path_segments as $path_segment) {
+            $current_pathname = $current_pathname . $path_segment . DIRECTORY_SEPARATOR;
+            @mkdir($current_pathname, $mode);
+        }
+
+        return is_dir($pathname);
     }
 }


fixpunkt:
Oh, and here's a Screenshot of how the bug presents in the shopware backend:

bildschirmfoto 2017-07-24 um 14 57 57

Stacktrace looks like this:
```
2017-07-24 15:30:51] core.ERROR: SmartyException: unable to write file /html/shopware/var/cache/production_201706221543/templates/backend_de_DE_cfdd9/ff/b5/f0/wrt5975f68b045564.82124959 in /html/shopware/engine/Library/Smarty/sysplugins/smarty_internal_write_file.php:48 Stack trace:
  #0 /html/shopware/engine/Library/Smarty/sysplugins/smarty_internal_template.php(201): Smarty_Internal_Write_File::writeFile('/html/shopware/...', '<?php /* Smarty...', Object(Enlight_Template_Manager))
  #1 /html/shopware/engine/Library/Smarty/sysplugins/smarty_internal_templatebase.php(155): Smarty_Internal_Template->compileTemplateSource()
  #2 /html/shopware/engine/Library/Smarty/sysplugins/smarty_internal_template.php(286): Smarty_Internal_TemplateBase->fetch(Object(Enlight_Template_Default), NULL, NULL, NULL, false, false, true)
  #3 /html/shopware/var/cache/production_201706221543/templates/backend_de_DE_cfdd9/96/c9/61/96c96106912c0a71deddeb551ea9c878e2124db0.snippet.3567b6264bda188e18092f04cb532fa461d57b3d.php(24): Smarty_Internal_Template->getSubTemplate('/html/shopware/...', NULL, 'backend_de_DE_c...', NULL, NULL, Array, 0)
  #4 /html/shopware/engine/Library/Smarty/sysplugins/smarty_internal_templatebase.php(180): content_5975f68b030781_63382942(Object(Enlight_Template_Default))
  #5 /html/shopware/engine/Library/Enlight/View/Default.php(274): Smarty_Internal_TemplateBase->fetch()
  ...
```

aragon999:
As noted here https://github.com/smarty-php/smarty/pull/379#issuecomment-318875750 maybe only the setting [`cache_locking`](https://www.smarty.net/docs/en/variable.cache.locking.tpl) should be activated to prevent the (often unnecessary) call of `mkdir()` on each subfolder of a file. Maybe also the smarty sysplugin `smarty_internal_runtime_writefile` should be backported, so that the creation of a file is retried if it failed. 
Disclaimer: I have not tested the effects of this option and if it has any negative side effects, nor evaluated the needed effort to backport the writefile plugin, just noticed the comment on the smarty pull request.

fixpunkt:
@aragon999 yup, good point.

I'll verify the `cache_locking` approach as soon as we have another support case, if that works, I'd prefer that approach.

I personally don't think retrying an arbitrary number of times is the best approach to solve race conditions, in particular since this does not guarantee success and may only cause the problem to occur less often. IMHO, this would be awful for customers because they'd still feel the pain, but be faced with an issue that cannot be reproduced reliably.

I'm also not certain whether the cost of the extra in-PHP `mkdir` calls is significant, especially in the case of Shopware since cache misses should be rare anyway.

fixpunkt:
So I just gave both of the suggestions from the smarty maintainer a spin. `cache_locking` had no effect at all, and the retry implementation did not solve the problem either. Increasing the max number of retries introduced a performance issue on top. See https://github.com/smarty-php/smarty/pull/379#issuecomment-319091092 for details.

mitelg:
thanks for your contribution @fixpunkt :+1:

PR #1222, opened 2017-07-26
Set service dbal_connection while container building

shyim:
## Description

| Why?                    | Register services or set parameters to container while container building from informations from database.  |
| How to test?            | See below example |

```php
class ShyimExample extends Plugin
{
    public function build(ContainerBuilder $container)
    {
        parent::build($container);

        $config = new Plugin\DBALConfigReader($container->get('dbal_connection'));
        $pluginConfig = $config->getByPluginName('ShyimExample');
        if ($pluginConfig['fooIntegration']) {
            $container->addCompilerPass(new FooIntegrationCompilerPass());
        }

        $container->setParameter('shyim_example.plugin_config', $pluginConfig);
    }
}
```

### DIFF
--- a/engine/Shopware/Kernel.php
+++ b/engine/Shopware/Kernel.php
@@ -24,6 +24,7 @@
 
 namespace Shopware;
 
+use Doctrine\DBAL\DriverManager;
 use Enlight_Controller_Request_RequestHttp as EnlightRequest;
 use Enlight_Controller_Response_ResponseHttp as EnlightResponse;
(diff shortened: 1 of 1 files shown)


janbuecker:
The container build should be as fast as possible with no more queries than needed. What is your use case for the change?

shyim:
Adding subscribers on some plugin configuration conditions or just easy access to plugin config without reading it on any subscriber again.. (if configs are not per subshop)

janbuecker:
We've discussed this change internally again and have decided to decline this pull request as we do not want to provide the database connection while the container is being build. This process should be as fast as possible.

Thanks @shyim

PR #1230, opened 2017-07-31
Fix SW-19032

DerRidda:
This commit fixes the issue outlined in https://issues.shopware.com/issues/SW-19032

I personally encountered the very same issue in Shopware 5.2.27 the exact same way and came to the same conclusion as the ticket creator.

Either way this is a very, very, insanely easy fix that doesn't break tests (coverage says this condition is covered in 49 tests) and it also allows third-party plugin creators to create entities with circular dependencies that can be used via the API without problems.

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Resource.php
+++ b/engine/Shopware/Components/Api/Resource/Resource.php
@@ -424,7 +424,7 @@ abstract class Resource
                 );
                 continue;
             }
-            if ($entity->$method() == $value) {
+            if ($entity->$method() === $value) {
                 return $entity;
             }
         }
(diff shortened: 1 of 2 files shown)


wildi1:
We encountered the same issue in 5.2.27 too and we please Shopware to merge this very easy fix.

janbuecker:
@DerRidda I think there went something wrong with your branch. There are changes from another open pull request in here.

DerRidda:
Oops, seems like I messed things up when making that other PR.
I did not realize PRs could be mutated after they were purely by operations I did on my fork.

Since I didn't branch this commit it's probably not going to be cleaned up, so I would recommend reopening the dupe and use that instead.

TIL to always use a new branch for every commit.

mitelg:
I messed something up :smile: 
sorry for that. we will use your first commit and drop the other one

will be merged soon, btw

Ticket https://issues.shopware.com/issues/SW-19032

PR #1231, opened 2017-08-02
Whitelist array() built-in in smarty security

fixpunkt:

Smarty erroneously treats the PHP `array()` built-in as a function and applies its PHP function security policy to it (https://github.com/shopware/shopware/blob/9a116f29f78c0005e031e34ffcce919b6a50e3a9/engine/Library/Smarty/sysplugins/smarty_internal_templateparser.php#L2871). This causes template security to prevent `array()` from being used from templates.

Right now, we work around this for Shopware 5.3 by using `array_fill(0, 0, null)` within our templates instead of `array()`.

It adds the `array()` built-in to Shopware's whitelist of functions that are allowed within smarty templates.

Call `array()` from a template (e.g. by doing `{assign var=a value=array()}`) and try to compile that on Shopware 5.3. Smarty security will throw an Exception.

After applying the patch, remember to clear the cache before testing the fix.

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Configs/smarty_functions.php
+++ b/engine/Shopware/Configs/smarty_functions.php
@@ -28,6 +28,7 @@ return [
     'acosh',
     'addcslashes',
     'addslashes',
+    'array',
     'array_change_key_case',
     'array_chunk',
     'array_column',


Phil23:
Thank you @fixpunkt for the hint.

We have scheduled this for the next 5.3.1 release. But you can simply use the shorthand syntax to get around this.

`{assign var=test value=[]}`

fixpunkt:
Thanks for scheduling and for the suggestion. 

Unfortunately, we cannot use the shorthand syntax since this is code that still needs to work on Shopware 4.3. Or is that a smarty built-in?

Phil23:
The syntax depends on the required PHP Version 5.4

fixpunkt:
Ok, thanks for clarifying. In this case we'll have to stick with `array_fill()`.

mitelg:
thanks again @fixpunkt :+1:

PR #1232, opened 2017-08-03
Fix SW-19523

EtienneBruines:
New cronjobs are never 'active', because the 'end' date is never set. 

Setting the 'end' date of *new* cronjobs, so the 'active' flag will actually be useful. As a result, plugins which are set `active` in the `cronjob.xml` file, will actually be executed / activated. (Which is the expected behavior)

Install any plugin with `cronjob.xml`, and look at the configuration of the cronjobs. It will show as 'not active' in configuration, even though you may have included `<active>true</active>` within the XML. 

https://issues.shopware.com/issues/SW-19523

None, this fixes a bug. 

### 6. Checklist

- [x] [N/A] I have written tests and verified that they fail without my change

### DIFF
--- a/engine/Shopware/Components/Plugin/CronjobSynchronizer.php
+++ b/engine/Shopware/Components/Plugin/CronjobSynchronizer.php
@@ -86,6 +86,7 @@ class CronjobSynchronizer
             $this->connection->update('s_crontab', $cronjob, ['id' => $id]);
         } else {
             $cronjob['next'] = new DateTime();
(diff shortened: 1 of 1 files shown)


mitelg:
hey @EtienneBruines 
there is already an open PR regarding this issue https://github.com/shopware/shopware/pull/1217

EtienneBruines:
@mitelg Oops; apologies. Forgot to check the past few weeks. Wasn't there a month ago :sweat_smile:

PR #1233, opened 2017-08-03
Fix getShopArrayByHostAlias when s_core_shops.hosts has only one line

tyurderi:
Fix \Shopware\Models\Shop\Repository::getShopArrayByHostAlias when s_core_shops.hosts has only one line.

Invalid redirect to default shop when the hosts alias field have only one line.

It fixes the query which tries to search for a shop by host aliases.

Create a new subshop with a custom domain and put in hostalias just an alternative host (e.g. example.com without www) without any new line. Just "example.com" - not "example.com\n" or similar.

### 6. Checklist


htkassner:
Thank you for your contribution. This will be released within the next version.

PR #1235, opened 2017-08-07
Don't display warnings related to session in CLI applications

xrogers:
When using logger in CLI scripts warnings are displayed because of the fact that env processor is accessing session and server variables available only in case of HTTP interface.


soebbing:
Hi @xrogers, thank you for your help to make Shopware better! 💙  Your PR will be part of the Shopware v5.3.4 release!

PR #1237, opened 2017-08-07
Bugfix for ProductAttributeFacetHandler

mkreusch:
Bug in the facet handler of free text facets

0 will be equal to any string because of automatic typecasting - therefore it has to be removed or replaced by a string zero 

Try to setup a free text facet with random strings

### 6. Checklist


mitelg:
thanks for your contribution @mkreusch :+1:

PR #1238, opened 2017-08-08
Fix for nesting level too deep

FKalab:
Uploading child articles with more than one image causes a
"Fatal error: nesting too deep - recursive dependency"
error in
"/engine/Shopware/Components/Api/Resource/Resource.php line 427"
and stops the upload.

This ensures a correct and more clean way to check for the value
and avoids returning wrong entity instead of NULL.

This is fixed by changing the return value check of
$entity->$method() == $value to
$entity->$method() === $value
in
"/engine/Shopware/Components/Api/Resource/Resource.php line 427"

Uploading child articles with more than one image causes a
"Fatal error: nesting too deep - recursive dependency"
error in
"/engine/Shopware/Components/Api/Resource/Resource.php line 427"
and stops the upload.

-

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Resource.php
+++ b/engine/Shopware/Components/Api/Resource/Resource.php
@@ -424,7 +424,7 @@ abstract class Resource
                 );
                 continue;
             }
-            if ($entity->$method() == $value) {
+            if ($entity->$method() === $value) {
                 return $entity;
             }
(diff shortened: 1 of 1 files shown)


mitelg:
duplicates https://github.com/shopware/shopware/pull/1230

PR #1239, opened 2017-08-08
Mistake in label

aherking:
Capital F for Filter in german lable

looks better

change the label in migration

open basic settings and look for search settings

/
### 6. Checklist


janbuecker:
We must not change existing migrations as already updated customers won't get the changes. You need to create a new one.

aherking:
Sorry ;-) I think this should be right now @janbuecker

aragon999:
You do not need to change the old migration, only the new one is needed.

aherking:
Finally 👍

PR #1240, opened 2017-08-09
Keep array keys in array_slice

hkcomputer:
Another bug that prevented working product streams for a slightly different scenario then the one before this

Product streams weren't working because the array keys for the products kept vanishing

Preserve array keys for further usage

1. Install Shopware 5.2.25
2. Create some articles (more then 25 with article numbers counting up from 1)
3. Create a product stream to filter the products
4. Add an article slider to an "Erlebnisswelt" with the newly created product stream
5. Watch as the only articles with IDs from 0 - 24 get selected for display.

(could not reproduce this on shopwaredemo.com... dont know why)

No changes needed

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundle/BatchProductNumberSearch.php
+++ b/engine/Shopware/Bundle/SearchBundle/BatchProductNumberSearch.php
@@ -100,7 +100,7 @@ class BatchProductNumberSearch
         }
 
         // use internal pointer to return different products to each request with the same criteria/context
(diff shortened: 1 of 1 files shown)


janbuecker:
Thanks @hkcomputer. Your change does not fix the underlying issue. We are already aware of this problem and fixed it with the upcoming 5.3.1. The change can be reviewed here: https://github.com/shopware/shopware/commit/40d5170296184668e0ba2bbbac7e4203bbb4d1e9

Closing this PR.

PR #1241, opened 2017-08-09
Change PSD File Type to unknown

s-bessing:
Changing PSD File Type to unknown, because media manager tries to generate an thumbnail and fails.

### 6. Checklist


mitelg:
thanks for your contribution @s-jakob :+1:

PR #1242, opened 2017-08-14
Added service "modules" to di-container via services.xml

jkrzefski:
Added a section for an instance of Shopware_Components_Modules to services.xml to make it possible to depend on it. Also added every core class as a service using the new "modules"-service as a factory.

Without this change it is impossible to have a service with a dependency on "modules". Let's say I need to have some information of sAdmin in my service. Until now the only way to achieve this was to use `Shopware()->Modules()->Admin()`. But since the monolith should not be used, this is bad practice. The reason for this behavior is located in `Shopware_Plugins_Core_System_Bootstrap::onInitResourceModules`. This is called once the service "modules" is instantiated. But here's the catch: This only works after plugin events are registered, which happens after the di-container's logic.

This change adds the service "modules" one step earlier than before. This is achieved through an entry in the services.xml file instead of a registration on an event.

- Create a plugin (in the 5.2-system).
- Create a service and enter it in the services.xml file.
- Give the service an argument that is a service with the id "modules".
- See that your entire system dies because at the time the service "modules" is required it neither exists not is the event `Shopware_Plugins_Core_System_Bootstrap::onInitResourceModules` registered yet.

None (yet).

None (I think).

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/services.xml
+++ b/engine/Shopware/Components/DependencyInjection/services.xml
@@ -21,9 +21,7 @@
             <deprecated>The "%service_id%" service is deprecated since 5.2 and will be removed in 6.0.</deprecated>
         </service>
         
-        <service id="modules" class="Shopware_Components_Modules">
-            <factory service="Shopware_Plugins_Core_System_Bootstrap" method="onInitResourceModules" />
-        </service>
+        <service id="modules" class="Shopware_Components_Modules" synthetic="true" />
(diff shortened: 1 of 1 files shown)


jkrzefski:
I understand that. I had such concerns myself when I made this pr. But what if we scrap all but one line and only keep this one:
```<service id="modules" class="Shopware_Components_Modules" synthetic="true" />```
I find this useful because `modules` is already available as a service. It just cannot be used as a dependency.

PR #1243, opened 2017-08-14
Make Unitize behave like CSS properties

lajosthiel:
Adjust .unitize-padding, .unitize-margin, .unitize-multiple to behave like standard CSS properties. I.E. you can just supply a single value like this:
.unitize-padding(10);

Better usability for FE devs

Change signature of .unitize-multiple, .unitize-margin, .unitize-padding mixins to conform to standard CSS properties

Use .unitize-padding, .unitize-margin, .unitize-multiple with only one argument and try to compile themes

Inline documentation included in PR

### 6. Checklist

- [ n ] I have written tests and verified that they fail without my change
- [ y ] I have squashed any insignificant commits
- [ y ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ y ] I have read the contribution requirements and fulfil them.


soebbing:
Hi @lajosthiel, thank you very much for your support, we just merged your commit: 813a5714ecb6966c658f26c8654682807387ff59! 🎉

PR #1244, opened 2017-08-14
Fix action columns not being extendable dynamically

windaishi:
Hello, I am a employee of VIISON GmbH / Pickware. This is my first pull request.

Actions column items cannot be added or removed after the action column has been instanciated.

Change the renderer closure of the action columns to items can now be removed and added dynamically again.

The problem is because the variable `var l = items.length` is defined outside the `me.renderer` closure ...
https://github.com/shopware/shopware/blob/301699adba5f423cb0b0a11795e912fe21478118/engine/Library/ExtJs/overrides/Ext.grid.column.Action.js#L38
and used inside the closure:
https://github.com/shopware/shopware/blob/301699adba5f423cb0b0a11795e912fe21478118/engine/Library/ExtJs/overrides/Ext.grid.column.Action.js#L55
When now the items get modified after the action column object has been initialized, the value for `l` is not updated.
Solution: Get the length of `me.items` dynamically. (See commit of pull request)

The problem is for example reproducable with the order document table. The action column of this table is instanciated in the createActionColumn method (which is called directly by the initComponent() method)

```
    /**
     * Creates the action columns
     *
     * @returns { Ext.grid.column.Action }
     */
    createActionColumn: function() {
        var me = this;

        return Ext.create('Ext.grid.column.Action', {
            width: 60,
            items: [
                me.createDeleteColumn(),
                me.createSendColumn()
            ]
        });
    },
```
If you try to extend the action column now by doing the following:
```
Ext.define('Shopware.apps.Example.view.list.Document', {

    override: 'Shopware.apps.Order.view.list.Document',

    createActionColumn: function() {
        var actionColums = this.callParent(arguments);
        actionColums.items.push({
            iconCls: 'sprite-mail-send',
            action: 'example'
        });
        return actionColums;
    }
}
```
... the action column is not changed as expected.

### DIFF
--- a/engine/Library/ExtJs/overrides/Ext.grid.column.Action.js
+++ b/engine/Library/ExtJs/overrides/Ext.grid.column.Action.js
@@ -35,7 +35,6 @@ Ext.override(Ext.grid.column.Action, {
         var me = this,
             cfg = Ext.apply({}, config),
             items = cfg.items || [me],
-            l = items.length,
             i,
             item;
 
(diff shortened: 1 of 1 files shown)


soebbing:
Hi @windaishi, thank you for your help! Your PR will be part of the Shopware v5.3.4! 🎉

PR #1245, opened 2017-08-15
Changed visibility of createEntitySearchQuery

dneustadt:

The method createEntitySearchQuery should technically be part of a repository. While this might be too big of a change at the very least the method should be public. That's because I need to be able to alter the returned query builder when the method is called by searchAction for enhancing various entity searches on the backend side.

Change visibility of method createEntitySearchQuery from private to public

-

-

none

### 6. Checklist


mitelg:
thanks for your contribution @dneustadt :+1:

PR #1248, opened 2017-08-23
Allow creation of documents for inactive subshops

samuelvogel:
Shopware crashes with `Call to a member function setCurrency() on null` [here](https://github.com/shopware/shopware/blob/v5.3.2/engine/Shopware/Components/Document.php#L538) when one tries to create an order document (e.g. invoice) for an order with an inactive language subshop. It's a real world issue, as even when the subshop is deactivated and nothing is sold anymore, one might have to create cancelation invoices some time later for older orders.
The problem is also super annoying, as the exception (if visible) is does not really tell you about where the problem might be coming from, which leads to painful and long debugging for an actually easy problem. Especially a regular shop owner will not be able to figure it out by himself.

It fixes the exception which occurred before when creating an order document for an inactive subshop and therefore allows to create documents for inactive subshops as well. Before this was only possible for active subshops.

Create a document for an order of which the language subshop is inactive. It will crash and not create the PDF. Then try again after the fix is applied.

n/a

None

### 6. Checklist


mitelg:
thanks for your contribution @samuelvogel :+1: :tada:

PR #1249, opened 2017-08-23
Improves performance of product listing

barbieswimcrew:
Each {include} causes that the page load time increases. Instead of using {include} in every if/else condition it's better to set the template file path in a variable and load the file afterwards. By testing it in a project we got a massive improvement. While (even if force compile === false) the listing did last about 8 seconds, loading the same listing now lasts about 300 milliseconds.

Removing multiple usage of {include} in the box_article.tpl and instead set the file path of the different box templates into a variable to load the dynamically determined template afterwards.

Create a big list of articles in a category and navigate to the category's listing view in your browser. Loading the page is pretty slow. Now replace the original code in themes/Frontend/Bare/frontend/listing/box_article.tpl by the one I committed and you will see how this change improves the page load time.

---

---

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/listing/box_article.tpl
+++ b/themes/Frontend/Bare/frontend/listing/box_article.tpl
@@ -1,22 +1,26 @@
 {block name="frontend_listing_box_article_includes"}
 
     {if $productBoxLayout == 'minimal'}
-        {include file="frontend/listing/product-box/box-minimal.tpl"}
+        {$path = "frontend/listing/product-box/box-minimal.tpl"}
 
     {elseif $productBoxLayout == 'image'}
-        {include file="frontend/listing/product-box/box-big-image.tpl"}
+        {$path = "frontend/listing/product-box/box-big-image.tpl"}
 
     {elseif $productBoxLayout == 'slider'}
-        {include file="frontend/listing/product-box/box-product-slider.tpl"}
+        {$path = "frontend/listing/product-box/box-product-slider.tpl"}
 
     {elseif $productBoxLayout == 'emotion'}
-        {include file="frontend/listing/product-box/box-emotion.tpl"}
-    {elseif $productBoxLayout == 'list'}
-        {include file="frontend/listing/product-box/box-list.tpl"}
+        {$path = "frontend/listing/product-box/box-emotion.tpl"}
 
     {else}
+        {$path = "frontend/listing/product-box/box-emotion.tpl"}
+    {/if}
+    
+    {if $path == ''}
         {block name="frontend_listing_box_article_includes_additional"}
             {include file="frontend/listing/product-box/box-basic.tpl" productBoxLayout="basic"}
         {/block}
+    {else}
+        {include file=$path}
     {/if}
 {/block}
--- a/themes/Frontend/Bare/frontend/listing/box_article.tpl
+++ b/themes/Frontend/Bare/frontend/listing/box_article.tpl
@@ -1,5 +1,6 @@
 {block name="frontend_listing_box_article_includes"}
-
+    {$path = ''}
+    
     {if $productBoxLayout == 'minimal'}
         {$path = "frontend/listing/product-box/box-minimal.tpl"}
 
@@ -12,8 +13,8 @@
     {elseif $productBoxLayout == 'emotion'}
         {$path = "frontend/listing/product-box/box-emotion.tpl"}
 
-    {else}
-        {$path = "frontend/listing/product-box/box-emotion.tpl"}
+    {elseif $productBoxLayout == 'list'}
+        {$path = "frontend/listing/product-box/box-list.tpl"}
     {/if}
     
     {if $path == ''}


themes/Frontend/Bare/frontend/listing/box_article.tpl
```diff
@@ -1,22 +1,26 @@
...
-        {include file="frontend/listing/product-box/box-emotion.tpl"}
-    {elseif $productBoxLayout == 'list'}
-        {include file="frontend/listing/product-box/box-list.tpl"}
+        {$path = "frontend/listing/product-box/box-emotion.tpl"}
 
     {else}
+        {$path = "frontend/listing/product-box/box-emotion.tpl"}
+    {/if}
+    
+    {if $path == ''}
```
shyim: Can $path be empty string? the if has a else :)

shyim:
I tested the pr with the profiler and got very fine results. Thank you ❤️ 

Before:
Before

After:
After

themes/Frontend/Bare/frontend/listing/box_article.tpl
```diff
@@ -1,22 +1,26 @@
...
-        {include file="frontend/listing/product-box/box-product-slider.tpl"}
+        {$path = "frontend/listing/product-box/box-product-slider.tpl"}
 
     {elseif $productBoxLayout == 'emotion'}
-        {include file="frontend/listing/product-box/box-emotion.tpl"}
-    {elseif $productBoxLayout == 'list'}
-        {include file="frontend/listing/product-box/box-list.tpl"}
+        {$path = "frontend/listing/product-box/box-emotion.tpl"}
 
     {else}
```
aragon999: I guess there is some kind of mistake here, in the else should be something like `{$path = "frontend/listing/product-box/box-basic.tpl"}` and `{$productBoxLayout = "basic"}` also the case `$productBoxLayout == 'list'` is missing. 
But nevertheless I like these changes :)

barbieswimcrew: Sorry, I didn't realize that... now I added the missing "list" case and I even added {$path = ''} as default value. Thanks for your review!

Depending on the smarty block "frontend_listing_box_article_includes_additional" that is wrapped around the "box-basic" case I cannot put the "box-basic" stuff into the else condition itself. That's why I had to determine between empty $path and a defined $path in a separate if/else

themes/Frontend/Bare/frontend/listing/box_article.tpl
barbieswimcrew: You were right, sorry for my mistake. I added {$path = ''} as default now. Thanks for your review!

dneustadt:
Such a simple change with a massive result. Great discovery. So nested statements are apparently being handled recursive by smarty?

themes/Frontend/Bare/frontend/listing/box_article.tpl
```diff
@@ -1,22 +1,27 @@
...
-    {elseif $productBoxLayout == 'list'}
-        {include file="frontend/listing/product-box/box-list.tpl"}
+        {$path = "frontend/listing/product-box/box-emotion.tpl"}
 
-    {else}
+    {elseif $productBoxLayout == 'list'}
+        {$path = "frontend/listing/product-box/box-list.tpl"}
+    {/if}
+    
+    {if $path == ''}
```
hrombach, line 20: Wouldn't it be even more efficient to set the `$path` variable here again, and then put a single `{include}` statement at the very end of the block? That way you would save one more `{include}`.

teiling88: I think this creates a breaking change. Cause the block frontend_listing_box_article_includes_additional works in a different way like before.

hrombach: Oh, you're right. It's possible. Forget what I said then :slightly_smiling_face:

hrombach (approved):

FiveDigital:
@shyim How do you get this Smarty Metrics? We think there is a problem with (our) detail page, too and want to analyze this.

shyim:
@FiveDigital https://github.com/shyim/shopware-profiler :)

themes/Frontend/Bare/frontend/listing/box_article.tpl
pthreat: Hi @barbieswimcrew, thanks a lot for your contribution. 

I have analyzed the reason why "it is faster".

screenshot from 2017-09-13 14-50-37.

As you can see, by removing multiple include lines (with your fix) the cache file is reduced by less than half of the lines:

4599 lines  (with the fix ) 
10737 lines (without the fix) 

Of course these lines are not just mere lines, they are rather expensive Smarty instructions.

Thanks a lot for your contribution, I'm reviewing that nothing is broken. And please consider applying this same fix to other templates in the frontend, always taking in account the suggestions that other contributors have told you above.

We would greatly appreciate it

Thank you.

pthreat:
Hi @barbieswimcrew, thanks a lot for your contribution. 

I have analyzed the reason why "it is faster".

screenshot from 2017-09-13 14-50-37.

As you can see, by removing multiple include lines (with your fix) the cache file is reduced by less than half of the lines:

4599 lines  (with the fix ) 
10737 lines (without the fix) 

Of course these lines are not just mere lines, they are rather expensive Smarty instructions.

Thanks a lot for your contribution, I'm reviewing that nothing is broken. And please consider applying this same fix to other templates in the frontend, always taking in account the suggestions that other contributors have told you above.

We would greatly appreciate it

Thank you.

resslinger:
why is this pull request father open? i think, it takes too much time to merge pull requests...

soebbing:
Hi @lacodimizer! I can understand your impatience, I can't wait to get more performance as well! 😆  

We are currently looking for & fixing other places with this pattern of includes, this takes a bit of time. Sorry for the delay 😞

mitelg:
hey @barbieswimcrew 
thanks for your contribution :+1: :tada: 
we merged this with https://github.com/shopware/shopware/commit/772a8a00e52e8ff3798c999380a04411be1d961e and fixed this also in other places

PR #1250, opened 2017-08-24
Show Growl message when creating order document

samuelvogel:
When creating a document for an order in the backend there is neither a success nor a failure message. The document just appears in the list, but especially when recreating/overwriting a document it is not distinguishable when the action is actually finished. This is even more annoying, when an error occurs while creating the document. The problem is not visible for users at all and painful debugging ensues...

Show success and error growl messages when creating an order document. The respective snippets for this where already present in the code BTW but never used.
document-message

Create a document in the backend and a success message should be shown. For some mPDF errors (missing rights on `files/documents` or no disk space for example) there will still be no correct error message unfortunately. The currently shipped mPDF version responds with an error string instead of throwing an exception (leading to a `200 OK` HTTP code, which will not be interpreted as a failure). This was fixed with mPDF `v6.1.0` (see https://github.com/mpdf/mpdf/commit/2d02c208b8d7b15c4208517375b9a10b0ea7cefa), once mPDF gets updated these exceptions will be shown as well.

n/a

None

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/order/controller/detail.js
+++ b/themes/Backend/ExtJs/backend/order/controller/detail.js
@@ -537,7 +537,8 @@ Ext.define('Shopware.apps.Order.controller.Detail', {
      * @param [Ext.container.Container] me
      */
     onCreateDocument: function(order, config, panel) {
-        var store = Ext.create('Shopware.apps.Order.store.Configuration');
+        var me = this,
+            store = Ext.create('Shopware.apps.Order.store.Configuration');
 
         panel.setLoading(true);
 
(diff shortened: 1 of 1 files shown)


soebbing:
Hi @samuelvogel, your commit was merged in 00df971ca1ebf60c0fdf53d07b915db2416c7c70! 🎉  Thank you for your help! 💙

samuelvogel:
Thank you! The commit actually is d5785bcbb57643cf4d732825341803496ce5bb25. As #1252 has been merged for Shopware 5.4 correct error messages will also be shown in the future as well 👍

PR #1252, opened 2017-08-24
Update mPDF and include it via composer

samuelvogel:
mPDF is currently directly bundled into the Shopware repository in version `6.0.0`, which is at least 2,5 years old (judging from the [tag on GitHub](https://github.com/mpdf/mpdf/releases/tag/v6.0.0)). It was last updated with Shopware 5.1.0 (a4ea89a7fd51d152c9869032f0f290ca1383d20e) and is shipped with custom fonts and a minor config change (see 9f921df4db75d3a6b9ee59695dae0fe20c997f4e). Some changes where made afterwards to backport specific fixes for PHP 7 compatibility (see b560f1110ca07bb3de475d1fd7584e1eb31f82e3 and df075d6a020547cf7ab3440065c5a5a245999ca8).

Unfortunately there is still a problem with PHP 7 and SVG usage on documents (a lot of people use a vector graphic to achieve sharply printed logos on their PDFs) when the server is configured to show `E_DEPRECATED` errors: `PHP Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; SVG has a deprecated constructor in 
/engine/Library/Mpdf/classes/svg.php on line 33`

Also since mPDF version `6.1.0` errors are thrown as exceptions which leads to better usability in the backend when there is not enough disk space or the file permissions are wrong (see https://github.com/mpdf/mpdf/commit/2d02c208b8d7b15c4208517375b9a10b0ea7cefa). The improved output for order documents is pending PR #1250.

Update mPDF to version `6.1.4` (the current stable release), which has much better compatibility. Also it includes mPDF via composer to clean up the source tree and ease future updates.

We checked carefully for this to not introduce any backwards incompatibility. We first diffed a4ea89a7fd51d152c9869032f0f290ca1383d20e against a original [mPDF 6.0.0 zip from http://www.mpdf1.com](http://www.mpdfonline.com/repos/MPDF_6_0.zip) to make sure there are no hidden changes. We then reapplied all configuration options in a composer compatible way. The exact same 5 fonts are available (all fonts shipped with the mPDF distribution are disabled, as before) and even the `backupSIPFont` option is still deactivated (even though probably nobody knows what it actually does nowadays).

We also did something very similar for FPDF with #1009.

Just create an order document with a regular Shopware `5.3` instance, save it and then create one with the PR applied. They will both look 100 % the same, pixel perfect. Fortunately mPDF is not under a lot of development, so this was the expected result 

n/a

None

### 6. Checklist


mitelg:
thanks for your contribution @samuelvogel :+1:

soebbing:
Hi @samuelvogel!

FYI: We'll have to move your PR to Shopware 5.4 - a lot of plugins rely on something like `require_once(Shopware()->DocPath() . 'engine/Library/Mpdf/mpdf.php');` (including some Viison plugins) which will break since the mpdf.php-file will have been moved to `vendor/shopware`. We cannot deliver such a breaking change in a patch release, sorry. 😞

shyim:
Just add a new file to engine/Library/Mpdf/mpdf.php which is empty and the problem is fixed :D

soebbing:
This sound like one of those workarounds which tend to stay around forever 😆

samuelvogel:
Looks like this PR is in Shopware 5.4 now, so we are happy 👍

As for our plugins: We have already scoped the FPDF includes to Shopware versions < 5.3 and as soon as a Shopware release includes this change, we will do the same for mPDF.

janbuecker:
@samuelvogel I'm sorry that I have to tell you, that this change has been reverted.

Mpdf comes with a lot of fonts which were not included in Shopware until your pull request. Currently, we are preparing the 5.4 release and noticed a much larger release package - much much larger than expected. After investigation, the problem were the amount of fonts which are delivered by mpdf.

We could have just built a package and accept the new package size but we are speaking of an increase of over 100% (35mb vs. 80mb). That's why we've reverted this pull request.

**BUT**, an improvement is still present: You don't have to require mpdf anymore as it is now part of the composer autoloader.

I hope you're not too unhappy with these changes.

You can review the actual code in 162358dc3c1dea059b9142520c9403ecd2148501.

samuelvogel:
Actually the newer mPDF version got much smaller again:
* `5.7.4`: 9.6MB
* `6.0.0`: 79MB (what Shopware currently uses)
* `6.1.4`: 49.8MB (included in this PR)
* `7.0.3`: 47.6MB (newest version)

It just seems that you stripped a lot of fonts from your `6.0.0` derivate distributed since Shopware `5.1`. While creating this PR we only made sure that the fonts used before are still present so documents look the same by default but the lager installation size is of course unwanted, good find!

We will try to add composer `post-install-cmd` to clean up the fonts which where not present before either and hope to still have this included in Shopware 5.4. Are you otherwise fine with the changes? We will also probably update to the newest mPDF `7.0.3` while we are at it again.

How does your process to bundle a release ZIP work? Any info which helps us with cleaning out the fonts in a way that works for you is appreciated!

samuelvogel:
Quick analysis of the biggest extracted files/folders in mPDF `7.0.3` that could potentially be removed:
* `src/QrCode/data`: 7.2MB of `.dat` files
* `ttffonts`: 91.3MB of font files, some especially big
    * `Sun-ExtA.ttf`: 23MB
    * `Sun-ExtB.ttf`: 17.6MB
    * `UnBatang_0613.ttf`: 6.9MB
    * `Aegyptus.otf`: 5.8MB

All of these sound rather irrelevant, we will check. This should strip 60.5MB of the 104.8MB extracted size or almost 58 %. The size of the zipped mPDF file will shrink from 47.6MB to 19.3MB.

Smaller savings could be achieved by removing also `Jomolhari.ttf` (2.3MB) and `Aegean.otf` (1.7MB).

Size graph:
screenshot 2018-01-30 09 59 01

soebbing:
We thought about that as well, but that would mean we then would have two versions of Shopware: The released package (with some or most fonts) and the git (or soon: composer) version with ALL fonts. That might mean that plugins could behave differently between those two versions, something we really want to avoid.

Maybe we could configure Mpdf 7.* to a degree where it would only use a subset of the available fonts, maybe even in a way that is backward compatible. In that case it might be possible to release a package where some fonts are removed, keeping the size of the release roughly the same as currently.

samuelvogel:
Of course we were not talking about manually removing something only from the release ZIP 😉 Basically we wanted to know if it's save for your release process to do this with a bash script run via composer, which is what we now did with #1495. This will of course then be the same for Git/composer and the ZIP distribution.

We really hope you can include this in Shopware 5.4 still (and therefore really hurried up with the new PR)!

With the scripts in this new PR it is possible to shrink the ZIP size even further if you want to go that route BTW. Would be happy to hear back what final size you see now 😃

soebbing:
I really hate to tell you that it is not likely that this change will make it into the 5.4, it is just to short a time for us to really test this through (though you really did an awesome job with that already)!

But I will make sure this is one of the first things being merged for 5.5, you shouldn't have to wait to long for this to hit a release! 💙

Again, thank you so much for that awesome PR (in record time!), I hope we can include it rather sooner than later!

samuelvogel:
You should remove the paragraph about this in the Shopware 5.4 upgrade docs then 😢: https://developers.shopware.com/developers-guide/shopware-5-upgrade-guide-for-developers/#mpdf

soebbing:
I am not sure what you mean, mPDF still has been updated to 6.1.4 and the class has been registered in the composer autoloader (although it wasn't required using composer). So the paragraph seems to be correct?

samuelvogel:
Ok, nevermind, as mPDF was actually updated with Shopware 5.4, most of the benefits of this PR (no PHP 7 errors with SVGs anymore and improved error message) will be shipped with it. And since the `require()` calls are already gone, the move to a composer dependency (and update to version 7) we did with #1495 will be less breaking.

PR #1254, opened 2017-08-28
Fixed account error when user couldn't found

shyim:
The currently implementation makes no sense and throws a warning invalid usage of sprintf.

Fixes the sprintf usage

Create a password reset, delete the user and open the reset mail

Nope

Nothing

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Account.php
+++ b/engine/Shopware/Controllers/Frontend/Account.php
@@ -770,10 +770,10 @@ class Shopware_Controllers_Frontend_Account extends Enlight_Controller_Action
         /** @var $customer Customer */
         $customer = $this->get('models')->find('Shopware\Models\Customer\Customer', $confirmModel->getData());
         if (!$customer) {
-            throw new Exception($resetPasswordNamespace->get(
-                sprintf('PasswordResetNewMissingId', $confirmModel->getData()),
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Frontend/Account.php
```diff
@@ -770,10 +770,10 @@ private function getCustomerByResetHash($hash)
         /** @var $customer Customer */
         $customer = $this->get('models')->find('Shopware\Models\Customer\Customer', $confirmModel->getData());
         if (!$customer) {
-            throw new Exception($resetPasswordNamespace->get(
-                sprintf('PasswordResetNewMissingId', $confirmModel->getData()),
-                sprintf('Could not find the user with the ID "%s".', $confirmModel->getData())
-            ));
+            throw new Exception(sprintf($resetPasswordNamespace->get(
```
Haehnchen, line 773: properly you want `sprintf` inside `get` and not wrapping it

soebbing: No, @shyim is right. The ->get() just fetches the right translation of the error message, containing a %s that is replaced using the sprintf.

soebbing:
Hi @shyim, thank you once again for your help! 💋

PR #1257, opened 2017-08-29
Use UploadedFile instead of File

roman-1983:
*Why*
To keep the file-ending and treat every file as an uploaded file.
If you are using `File` for files added via API, Shopware tries to guess the file-extension. This is sometimes wrong.
It's better to treat the files as `UploadedFile` instead of `File` so Shopware will use the given file-extension.

*What*
Use `UploadedFile` instead of `File`

## Description

| Why?                    | To keep the file-ending and treat every file as an uploaded file.
If you are using File for files added via API, Shopware tries to guess the file-extension. This is sometimes wrong.
It's better to treat the files as UploadedFile instead of File so Shopware will use the given file-extension. |
| How to test?            | Upload a file via API.
Before: If you upload a `.stp` file, it will be saved as `.txt`
After: If you upload a `.stp` file, it will be saved as `.stp` |

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Media.php
+++ b/engine/Shopware/Components/Api/Resource/Media.php
@@ -244,7 +244,7 @@ class Media extends Resource
         $name = $name . '.' . $ext;
         $path = $this->load($link, $name);
         $name = pathinfo($path, PATHINFO_FILENAME);
-        $file = new File($path);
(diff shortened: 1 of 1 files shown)


janbuecker:
This has already been implemented with decf4daf95f8ab4b6076b8992f4f34ee422dfb23. Therefore i'll close this PR.

PR #1262, opened 2017-09-01
Fix topseller generation

larsbo:
The setting for displaying topseller in the frontend prevents the cron job for generating latest topseller stats from doing his job. This leads to wrong "popular" sorting in listings and product streams because they are based on the generated topseller values.

It removes the config check for `topSellerActive` inside the topseller cron job what allows the job to run even when the topseller frontend display is disabled.

1. Go to `Performance > Einstellungen > Marketing > Topseller` and disable the setting `Topseller im Shop anzeigen`.
2. Enable the cron job `Topseller Refresh` and make sure that the cron job has run through.
3. Look into `s_articles_top_seller_ro` and you will see no updates here.

### 6. Checklist


larsbo:
Dear pull request, all the best for your first anniversary... 🎂 😭

PR #1266, opened 2017-09-06
Remove Hanseatic Bank notice in payment status

pascalheidmann-bedarf:
Right now there is no state to show that you ordered your bank/payment provider to debit charge your customer. It seems like Shopware thought about this some time ago but the current implementation explicitly mentions the Hanseatic Bank which won't be used by the majority and therefore a brand independent solution is to prefer.
Shopware itself doesn't use this state anywhere in its code so there might just be very few plugins (if any).

All mentions of the Hanseatic Bank are removed and replaced by a broader naming.

Payment state for every order allows to switch to this state. This is also shown in the user account at frontend.

### 6. Checklist

- [ ] I have written tests and verified that they fail without my change -> no tests as no usage inside Shopware


samuelvogel:
Great PR, was always wondering what kind of left over that was. Actually "some time ago" was at least Shopware 4.0.0, probably even much earlier in the 3.x days 😉 
https://github.com/shopware/shopware/blob/0cf9253bfe3d181543c9071883927d097d7e3eb9/_sql/install/4.0.0.sql#L5854

engine/Shopware/Models/Order/Status.php
```diff
@@ -80,7 +80,7 @@ class Status extends ModelEntity
     const PAYMENT_STATE_NO_CREDIT_APPROVED = 30;
     const PAYMENT_STATE_THE_CREDIT_HAS_BEEN_PRELIMINARILY_ACCEPTED = 31;
     const PAYMENT_STATE_THE_CREDIT_HAS_BEEN_ACCEPTED = 32;
-    const PAYMENT_STATE_THE_PAYMENT_HAS_BEEN_ORDERED_BY_HANSEATIC_BANK = 33;
+    const PAYMENT_STATE_THE_PAYMENT_HAS_BEEN_ORDERED = 33;
```
fixpunkt, line 83: Strictly speaking, this is a change that breaks backwards compatibility. Instead, maybe you can leave the old constant in as deprecated and alias it to the new name like this:

```php
const PAYMENT_STATE_THE_PAYMENT_HAS_BEEN_ORDERED = 33;
/** @deprecated use PAYMENT_STATE_THE_PAYMENT_HAS_BEEN_ORDERED instead */
const PAYMENT_STATE_THE_PAYMENT_HAS_BEEN_ORDERED_BY_HANSEATIC_BANK = self::PAYMENT_STATE_THE_PAYMENT_HAS_BEEN_ORDERED;
```

_sql/install/latest.sql
```diff
@@ -4031,7 +4031,7 @@ INSERT INTO `s_core_states` (`id`, `description`, `position`, `group`, `mail`) V
 (30, 'Es wurde kein Kredit genehmigt.', 30, 'payment', 1),
 (31, 'Der Kredit wurde vorlaeufig akzeptiert.', 31, 'payment', 1),
 (32, 'Der Kredit wurde genehmigt.', 32, 'payment', 1),
-(33, 'Die Zahlung wurde von der Hanseatic Bank angewiesen.', 33, 'payment', 1),
```
janbuecker: Please do not change the `install/latest.sql` file. Write a migration instead.

pascalheidmann-bedarf:
Updated to include suggested changes by @fixpunkt & @janbuecker

jenskueper:
Hello @pascalheidman-bedarf,

thank you for the PR. I have reviewed the PR and merged it.

PR #1273, opened 2017-09-11
Improved message for local plugin update

samuelvogel:
Currently the update buttons are ambiguous if there is an update available via the store and also already locally uploaded:
local update
The left button tries to download and install the update via the store, the right one just applies the local update.

Clearer wording for the right, local update button.

Manually download an a newer version from the store and upload it via the plugin manager in the backend. Now both buttons will appear and it's unclear which means which.

n/a

n/a

### 6. Checklist


mitelg:
@samuelvogel thanks for the PR :+1:

PR #580, opened 2016-05-30
Added filter events to all convert functions in LegacyStructConverter

FiveDigital:
This is needed to be able to convert the attributes you can add in the decorated services to the struct objects. The new events are consistent to the event `Legacy_Struct_Converter_Convert_Media` that was already present in the code. This also creates the ability to edit the storefront data in one central place.

PR #1274, opened 2017-09-11
Add interface for LegacyStructConverter

kleinmann:
Any plugin decorating the `legacy_struct_converter` service will currently cause a fatal error in the EmotionBundle, because `Shopware\Bundle\EmotionBundle\Service\StructConverter` depends on the actual base implementation.

It introduces an interface for the LegacyStructConverter and uses it in place of the actual implementation.

Decorate the service `legacy_struct_converter` and try to open any page with an emotion/shopping world.

You can use the attached example plugin `MyPlugin` for this:

<details><summary>MyPlugin.php</summary>

```php
<?php

namespace MyPlugin;

use Shopware\Components\Compatibility\LegacyStructConverter as BaseService;

class MyPlugin extends Plugin
{
    {
        return [
            'Enlight_Bootstrap_AfterInitResource_legacy_struct_converter' => 'decorateLegacyStructConverter',
        ];
    }

    public function decorateLegacyStructConverter()
    {
        $this->container->set(
            'legacy_struct_converter',
            new LegacyStructConverter($this->container->get('legacy_struct_converter'))
        );
    }
}

class LegacyStructConverter
{
    /** @var BaseService */
    private $baseService;

    /**
     * LegacyStructConverter constructor.
     *
     * @param BaseService $baseService
     */
    public function __construct(BaseService $baseService)
    {
        $this->baseService = $baseService;
    }

    /**
     * @param string $name
     * @param mixed  $arguments
     *
     * @return mixed
     */
    public function __call($name, $arguments)
    {
        return call_user_func_array([$this->baseService, $name], $arguments);
    }
}
```
</details>

You will then see this error instead of an emotion/shopping world:

```Uncaught TypeError: Argument 1 passed to Shopware\Bundle\EmotionBundle\Service\StructConverter::__construct() must be an instance of Shopware\Components\Compatibility\LegacyStructConverter, instance of MyPlugin\LegacyStructConverter given```

none

Probably none.

### 6. Checklist

- [X] I have written tests and verified that they fail without my change (see example plugin)

### DIFF
--- a/engine/Shopware/Bundle/EmotionBundle/Service/StructConverter.php
+++ b/engine/Shopware/Bundle/EmotionBundle/Service/StructConverter.php
@@ -36,13 +36,13 @@ use Shopware\Bundle\EmotionBundle\Struct\Emotion;
 use Shopware\Bundle\EmotionBundle\Struct\Library\Component;
 use Shopware\Bundle\MediaBundle\MediaServiceInterface;
 use Shopware\Bundle\StoreFrontBundle\Struct\Product\Manufacturer;
-use Shopware\Components\Compatibility\LegacyStructConverter;
+use Shopware\Components\Compatibility\LegacyStructConverterInterface;
 use Shopware\Components\DependencyInjection\Container;
 
 class StructConverter
 {
     /**
-     * @var LegacyStructConverter
+     * @var LegacyStructConverterInterface
      */
     private $converter;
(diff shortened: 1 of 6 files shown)


mitelg:
we had some discussion about that some time ago: https://github.com/shopware/shopware/pull/580

PR #1280, opened 2017-09-18
added two filter events to allow the sql to be amended when additiona…

mark-newman:
…l joins are required in shipping cost terms or calculations

The provided sql is not always enough to allow complex shipping calculations. A common scenario might be for countries to have additional attributes relating to shipping, but the countries attributes table can't be joined without adding these filter events.

Adds two new events to allow the sql to be updated if needed.

If you wish to use data from a table not included in the query you currently have to write a complex list of sub select statements to get the data you require. A simple additional join is much more efficient.

N/A

None

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -2510,7 +2510,7 @@ class sAdmin
         $sql = $this->eventManager->filter(
             'Shopware_Modules_Admin_GetDispatchBasket_FilterSql',
             $sql,
-            ['subject' => $this, 'id' => $userId]
+            ['subject' => $this, 'user_id' => $userId]
         );
 
         $sessionId = $this->session->offsetGet('sessionId');
@@ -2737,7 +2737,7 @@ class sAdmin
         $sql = $this->eventManager->filter(
             'Shopware_Modules_Admin_GetPremiumDispatches_FilterSql',
             $sql,
-            ['subject' => $this, 'id' => $userId]
+            ['subject' => $this, 'user_id' => $userId]
         );
 
         $dispatches = $this->db->fetchAssoc(
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -2508,7 +2508,7 @@ class sAdmin
         $userId = $this->session->offsetGet('sUserId');
 
         $sql = $this->eventManager->filter(
-            'Shopware_Modules_Admin_GetDispatchBasket_FilterResult',
+            'Shopware_Modules_Admin_GetDispatchBasket_FilterSql',
             $sql,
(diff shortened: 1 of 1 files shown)


engine/Shopware/Core/sAdmin.php
```diff
@@ -2506,6 +2506,13 @@ public function sGetDispatchBasket($countryID = null, $paymentID = null, $stateI
         ";
 
         $userId = $this->session->offsetGet('sUserId');
+
+        $sql = $this->eventManager->filter(
+            'Shopware_Modules_Admin_GetDispatchBasket_FilterResult',
```
shyim: Should the event name not end with "_FilterSql"? FilterResult is mostly used for result of queries

mark-newman: Yes, good idea. I've made the change.

engine/Shopware/Core/sAdmin.php
```diff
@@ -2506,6 +2506,13 @@ public function sGetDispatchBasket($countryID = null, $paymentID = null, $stateI
         ";
 
         $userId = $this->session->offsetGet('sUserId');
+
+        $sql = $this->eventManager->filter(
+            'Shopware_Modules_Admin_GetDispatchBasket_FilterSql',
+            $sql,
+            ['subject' => $this, 'id' => $userId]
```
shopwareBot: Please rename the parameter to `user_id`.

engine/Shopware/Core/sAdmin.php
```diff
@@ -2726,6 +2733,13 @@ public function sGetPremiumDispatches($countryID = null, $paymentID = null, $sta
         ";
 
         $userId = $this->session->offsetGet('sUserId');
+
+        $sql = $this->eventManager->filter(
+            'Shopware_Modules_Admin_GetPremiumDispatches_FilterSql',
+            $sql,
+            ['subject' => $this, 'id' => $userId]
```
shopwareBot: Please rename the parameter to `user_id`.

shopwareBot (changes requested):
Please implement the changes we requested in our code comments.

engine/Shopware/Core/sAdmin.php
mark-newman: Done

engine/Shopware/Core/sAdmin.php
mark-newman: Done

shyim:
Hey @mark-newman ,
we have changed these methods to use a QueryBuilder and provided there new events to modifiy the QueryBuilder with commit https://github.com/shopware/shopware/commit/eec8927194a374cd9ac534adee60d9b2aa5bd327.
This should solve your issue, but thanks for your Contribution :blue_heart: :tada:

PR #1283, opened 2017-09-20
Bugfix RewriteMatcher.php

GHNutzer:
Wrong select to find routing rewrite.

Correct the sql to find the url write for a particular shop.

Change category in a shop and browse old category in shop.

### 6. Checklist


OliverSkroblin:
Hey,
i think this isn't a bug inside the sql statement.
Please take a look to at the following line: https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Components/Routing/Matchers/RewriteMatcher.php#L108

At this line we check if the shop id of the found route is not equals with the current shop id. I think this is a condition if the url moved to another shop.

Please verify this behavior and add some tests

PR #1284, opened 2017-09-21
Add support for diacritics to search term handler

Amaridian:
lacking support for diacritics in search term handler
https://issues.shopware.com/issues/SW-19956?_ga=2.38083237.725833049.1505824796-1740009504.1485252782

use the cocur slugifier to slugify the search terms instead of a str_replace 

see 1.

see 1.

none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/SearchTerm/TermHelper.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/SearchTerm/TermHelper.php
@@ -23,6 +23,7 @@
  */
 
 namespace Shopware\Bundle\SearchBundleDBAL\SearchTerm;
+use Cocur\Slugify\Slugify;
 
 /**
  * @category  Shopware
@@ -45,6 +46,11 @@ class TermHelper implements TermHelperInterface
      * @var bool
      */
     private $replaceUmlauts;
+    
+    /**
+     * @var slug
+     */
+    private $slug;
 
     /**
      * @param $config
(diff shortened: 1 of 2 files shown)


engine/Shopware/Bundle/SearchBundleDBAL/SearchTerm/TermHelper.php
```diff
@@ -56,6 +62,7 @@ public function __construct($config, $useBadWords = true, $replaceUmlauts = true
         $this->config = $config;
         $this->useBadWords = $useBadWords;
         $this->replaceUmlauts = $replaceUmlauts;
+        $this->slug = new Slugify();
```
shyim: Should not be the slugify service injected? Otherwise the default slugify configuration wouldn't have a effect (https://github.com/shopware/shopware/blob/b4a17510a79fd60eda4d54b7cfbe8f50de975fa6/engine/Shopware/Components/DependencyInjection/services.xml#L8)

Amaridian: Not so important in this case because default behavior is exactly what is needed... but for consistency reasons a good idea anyway.

Amaridian:
I am not sure if the injection is done right like this... feel free to change.

mitelg:
hey @Amaridian you need to add a new parameter in the constructor

Amaridian:
@mitelg sorry, that injection system is causing me headaches.

PR #1285, opened 2017-09-25
redirect to shopdomain with request uri

egoist-dd:
redirect user to the right url if he request your shop with an url that is not default

if you request http://shopwaredemo.de/beach-relax/fashion/women/ you would be redirected to start page. Some Tools e.g. sistrix call your shop with an invalid url to check the 404 status for non existing pages. In this case you got 301 for the redirect.

call http://shopwaredemo.de/beach-relax/fashion/women/ and you should be redirected to http://www.shopwaredemo.de/beach-relax/fashion/women/

---

---

### 6. Checklist


egoist-dd:
Reason 1: If you change the shopdomain in your backend you have to add or edit the rule in htaccess. 

Reason 2: not everyone has knowledge about editing htacces

PR #1286, opened 2017-09-25
Changed api variant resource to check for existing configurator group…

DerRidda:
…s and options only via the database

These changes are necessary to guarantee that existing configurator options and groups are found according in the case insensitive database tables, which ultimately is the authority on what exists and what doesn't.

This is mostly evident when the api's variant resource checks if a configurator option for any given group already exists via the options name. Because the comparison is done via sting comparison in PHP, any option names that are the same except for capitalization are not recognized as equal.

This means that the PHP code will try to create a new option for a given group with a name that is only different in capitalization and will violate the unique index set on group_id and name name for the s_article_configurator_options database table.

This change converts these strings to lower case for comparison to mimic the database's case insensitivity.

Change the getAvailableGroup and getAvailableOption methods in Shopware\Components\Api\Resource\Variant to convert the names of the respective entities to lower case before performing their checks.

The test method testVariantUpdate in Shopware\Tests\Functional\Components\Api\VariantTest
was altered to include changed configurator options that will not only fail but cause an exception without this fix.

This should serve as a demonstration on how to reproduce.

None I am aware of.

None as far as I can tell. This is all internal implementation detail.

### 6. Checklist

Updated some incorrect/missing PhpDoc hints as well.

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Variant.php
+++ b/engine/Shopware/Components/Api/Resource/Variant.php
@@ -838,7 +838,7 @@ class Variant extends Resource implements BatchInterface
                 throw new ApiException\CustomValidationException(sprintf('Unit by id %s not found', $data['unitId']));
             }
 
-        //new unit data send? create new unit for this variant
+            //new unit data send? create new unit for this variant
         } elseif (!empty($data['unit'])) {
             $data['unit'] = $this->updateUnitReference($data['unit']);
         }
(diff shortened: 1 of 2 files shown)


DerRidda:
The failing test seem to evolve around some helper methods I currently do not know well enough to pin down the actual issue.

Whether this fix is broken/insufficient  or the tests actually need changes or the test inter-dependencies are too big or any other issue.

Any input would be appreciated.

engine/Shopware/Components/Api/Resource/Variant.php
```diff
@@ -26,6 +26,7 @@
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
+use function foo\func;
```
Gugiman: Is this wanted like that? :)

DerRidda: Nope, freaking autcomplete on callbacks always gets that in if I am not careful. Will amend.

DerRidda:
This PR has been changed (and reset to keep the commits in check) to use mb_strtolower on the names instead, this also fixes the unrelated failing unit tests. So I agree it's the simpler and better solution.

FiveDigital:
I think the underlying problem ist bigger because it's not just the case insensitivity. I have reported this issue 1,5 years ago here: https://issues.shopware.com/issues/SW-14308

DerRidda:
I agree, the API is a hot mess.

Especially as far as articles and most notably variants are concerned. I just finished working on a larger plugin (not-public)  where I bumped into all the ugly bits of the API, like a cartoon character falling off a tree and hitting every branch on the way down, to then fall off the cliff on which that tree stands and hit every sharp corner on the way down.

Whenever I found an issue I could not work around I made a PR.

I never bother with issue reporting because issues just usually lay around too long and get abandoned, as your's was.

/rant

DerRidda:
This has been in limbo for ages now. Is it still relevant?

FiveDigital:
Yes, but I don't think that it will be fixed.

PR #1290, opened 2017-09-27
MediaBundle GarbageCollector: Add mediaId index to the helper table

313:
The media bundle's garbage collector suffers from significant performance issues when dealing with large data sets. Especially the query in the final `moveToTrash` method can take a lot of time to process.

Creating an index for the `mediaId` column of the temporary `s_media_used` table greatly increases the GC's performance, without requiring any changes to its logic.

```php bin/console sw:media:cleanup```

\-

\-

### 6. Checklist


mitelg:
closed in favor of #1670

PR #1296, opened 2017-10-05
Update component_youtube.tpl

tinect:
label doesn't match function
modestbranding has to be 0, to disable branding, as the label says.

it will hide the branding

### 6. Checklist


tinect:
missunderstood youtube-docu. Sorry.

PR #1297, opened 2017-10-05
[BUGFIX] Changed all font-weight in less to use available variables

wolv-dev:
Variables are available in responsive theme, but not used entirely troughout the theme.

Replacing all font-weight <= 300 with variable @font-light-weight
Replacing all font-weight 400, 500 and normal with @font-base-weight
Replacing all font-weight >= 600 and bold with @font-bold-weight


mitelg:
merged with 8a692582509
thanks @wolv-dev :+1: :tada: 
great work!

PR #1300, opened 2017-10-09
Fixed emotion default value size to TEXT

shyim:
The current column has a limitation to 256 Characters. A html example is currently not possible with this size.

The migration makes the default value field of elements the same size like the value (s_emotion_element_value.value)

Create a emotion world field with a longer default value

### 6. Checklist


OliverSkroblin:
Hi @shyim,
thanks for the pull request.
The changes are merged and will be available with version 5.3.5.

PR #1304, opened 2017-10-11
Hard-coded user and group id for shopware's docker images

simkli:
*Shopware Labs Branch*

The Docker Team says themselves:
> Eliminate the “it works on my machine” problem once and for all.

That's not true for the shopware docker containers unfortunately.

My local user is in a group with ID 20. Shopware uses this ID for the group inside the docker container, too. The problem:

> Many Linux systems reserve the GID number range 0 to 99 for statically allocated groups[...]

from https://en.wikipedia.org/wiki/Group_identifier#Reserved_ranges

This means:
docker

Now I can't build the container. We should rely on static variables inside our container so we are independent of the host system.

The userID and groupID of the container's user won't be determinated by the host user's ids.

Use a unix system with a user's group's id is 20. Try to install the docker image. It fails.

none

### 6. Checklist

### DIFF
--- a/dev-ops/docker/containers/scriptcreator.sh
+++ b/dev-ops/docker/containers/scriptcreator.sh
@@ -2,8 +2,8 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-USERID=$(id -u)
-GROUP=$(id -g)
+USERID="2020"
+GROUP="2020"
 
 
 cat > $DIR/php7/createuser.sh <<DELIM
--- a/.psh.yaml.dist
+++ b/.psh.yaml.dist
@@ -33,13 +33,13 @@ const:
   PHP_VERSION: "7.1"
   APP_DEBUG: "1"
   APP_ENV: "dev"
+  USER_ID: "2020"
+  GROUP_ID: "2020"
+  USERKEY: "2020:2020"
 
 dynamic:
-  USERKEY: echo "$(id -u):$(id -g)"
   APP_ID: docker-compose ps -q app_server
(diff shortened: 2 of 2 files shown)


dev-ops/docker/containers/scriptcreator.sh
```diff
@@ -2,8 +2,8 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-USERID=$(id -u)
-GROUP=$(id -g)
+USERID="2020"
```
teiling88: please use instead of 2020 the place_holder __USER_ID__

dev-ops/docker/containers/scriptcreator.sh
```diff
@@ -2,8 +2,8 @@
...
+GROUP="2020"
```
teiling88: please use instead of 2020 the place_holder __GROUP_ID__

janbuecker:
We've updated the docker environment to work with macOS as this issue is only present there. Feel free to open another PR if it still does not work for you.

simkli:
@janbuecker I saw that. But to get it „working“ you deleted the group 20 inside the container. But this can only be a temporary solution. If we really want to get the „it works on every machine“ experience we have to cut the dependency to the host system.
There could be another distribution of linux which uses group 21 (or anything in range of 0-99) as default group and it will fail there. 

Another scenario: I want to add my user to the dialout group on my host system... now I can‘t use this image... 

I‘ll reopen this PR when I‘ve also undo the deletion of the dialout group as it would’t be needed anymore.

PR #1305, opened 2017-10-12
[BUGFIX] Fixed error in note uniqueId is newly set and not available …

wolv-dev:
…in sCountNotes (returned 0)

When adding an article to the notes, the first time it will return 0 due to cookie with value sUniqueId is not available on initial call.

Fixes the error that 0 will be returned.

1. Visit the shop
2. Add an article to the note list (on first try it will display 0 above the heart)
3. If not reproduceable clear all your application data of the site and retry


shyim:
Hey @wolv-dev,
this has been already fixed in Ticket https://issues.shopware.com/issues/SW-19657.
Thanks for your contribution :blue_heart:

PR #1307, opened 2017-10-12
Have the plugin manager reload on fail

niklasbuechner:
When a plugin update fails the local plugin listing is not reloaded. The plugin will therefore still be listed under activated even though it gets disabled due to the failed update.
This can leave users confused since the plugin is listed as active but is not.

It reloads the local plugin listing once a plugin update fails. Furthermore, it reloads the menu to remove disabled menu items.

Create a plugin which fails during the update process. Click update in the plugin manager's local plugin listing.

None exists.

None.

### 6. Checklist

By the way, I am a new Pickware team member.


Upgrade.md
```diff
@@ -0,0 +1 @@
+* Adds missing handling of plugin update fails by reloading the plugin listing in the plugin manager window
```
mitelg: please update our existing Upgrade files

mitelg (changes requested):

Upgrade.md
niklasbuechner: The correct file has been updated.

mitelg (approved):

PR #1309, opened 2017-10-13
Allow session namespaces

soebbing:
It allows to store the sessions of multiple Shopware-instances in one redis or memcache-Server by providing them with different namespaces.

It changes how the session-key is being generated by moving the default name to Defaults, allowing it to be overwritten in the config.php.

Install two shopware-instances and point them to one memcache-server.

The documentation regarding how to define memcache or redis as a session backend should be extended to include the namespace-parameter.

### 6. Checklist

- [] This change has comments for package types, values, functions, and non-obvious lines of code


Haehnchen:
:+1:

soebbing:
There doesn't seem to be any use case for this, closing the PR...

PR #1311, opened 2017-10-13
[BUGFIX] JS error caused by empty sidebar filter

mkreusch:
Currently if you have the filter placed in the sidebar and no facets are set up for the current view, the essential information for themes/Frontend/Responsive/frontend/_public/src/js/jquery.listing-actions.js especially the data attributes of the HTML filter form are missing – thus resulting in a JS error.

It takes care of the invisible filter form to be rendered even if there are no facets set up for the current view, as it is if you do not have placed it in the sidebar. Basically it just adapts the (correct) behaviour of the non-sidebar-filter.

1. Place the filter in the sidebar
2. Disable all facets
3. Open a product category in the shop's frontend

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/listing/sidebar.tpl
+++ b/themes/Frontend/Bare/frontend/listing/sidebar.tpl
@@ -8,9 +8,7 @@
 
                 <div class="sidebar-filter">
                     <div class="sidebar-filter--content">
-                        {if $criteria && $facets}
-                            {include file="frontend/listing/actions/action-filter-panel.tpl"}
-                        {/if}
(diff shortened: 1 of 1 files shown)


soebbing:
Hi @mkreusch, I'm closing this PR since we fixed this (more or less at the same time as you) with https://github.com/shopware/shopware/commit/b828d6c21908e547d367353bd9ad6947d592df56 

Sorry that I didn't see your PR before, would have liked to merge it. 😞

mkreusch:
Never mind :-)
Glad it's fixed in the next release :+1:

soebbing:
The fix is part of v5.3.4 already!

PR #1317, opened 2017-10-19
Always use the HttpHost and BasePath from ShopwareBackend

fionera:
When the Request does not have a BasePath because of Rewrites but you have a BasePath in the Backend it will not provide the correct Paths

Removes the fetching of the URL over the Request

Have a Shop with a Basepath and try to display an Image in the Backend

### 6. Checklist


fionera:
@shopwareBot The problem is that if you have set the BasePath it can happen that images etc. get the wrong path, because the BasePath is not necessarily present in the request. This Problem also exists in the Frontend

PR #1318, opened 2017-10-19
[next] storefront twig templates are also inside global namespace

Haehnchen:
support global twig templates in shopware next storefront


janbuecker:
Thanks, as discussed, we'll fix our templates to always include the bundle reference in the template name.

PR #1330, opened 2017-10-23
Implemented fallback into getCurrency in ContextFactory

jkrzefski:
Replaced a generic RuntimeException that is never caught with the only logical fallback. This makes it impossible for the currency-cookie to break the storefront if it is invalid.

Without this change an invalid value in the currency-cookie will break the entire storefront because the RuntimeException for this was never caught.

It replaces the thrown RuntimeException with a fallback. When a currency is being loaded based on a cookie and it is not found in the database, now the result falls back to the default shop currency.

Open the storefront and replace the value of the currency-cookie with a different one that is invalid. This could easily happen when a currency is removed after a shop already had customers that used the currency switch.

None.

None.

### 6. Checklist

### DIFF
--- a/src/Context/Service/ContextFactory.php
+++ b/src/Context/Service/ContextFactory.php
@@ -231,7 +231,7 @@ class ContextFactory implements ContextFactoryInterface
         $currency = $this->currencyRepository->read([$currencyUuid], $context);
 
         if (!$currency->has($currencyUuid)) {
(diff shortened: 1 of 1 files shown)


jkrzefski:
If the question comes up: I purposely did not persist the fallback currency uuid to the cookie because I don't think, this is the right place to do so. In general I think I makes sense to set the currency cookie to the fallback in case it is needed. But I think that should be handled by the Storefront bundle instead of the Context bundle.

PR #1331, opened 2017-10-23
Show a confirmation dialog when deleting an order document

windaishi:
Order documents can be deleted without confirmation. Most order documents, especially invoices, should never be deleted. The user should at least be warned about that.

This functionality is integrated in PickwareERP and does not work in Shopware 5.3.

Adds a confirmation message before deleting an order document.

Delete a document from within an order.


cjuner:
@windaishi, please rebase for Shopware 5.5, or delegate it appropriately.

windaishi:
This has been rebased to 5.4 as it can be merged as a bug fix.

PR #1332, opened 2017-10-23
Expose eventemitter as global

screeny05:
The EventEmitter is a useful pattern already included in shopware. Expose it as a global to open it for reusability for other developers.

Create `window.EventEmitter`

`console.log(EventEmitter)`

-

Possibly add it to the DevDocs

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.state-manager.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.state-manager.js
@@ -286,6 +286,8 @@
         }
     };
 
+    window.EventEmitter = EventEmitter;
+
     /**
      * @public
      * @static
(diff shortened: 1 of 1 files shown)


klarstil:
Hey @screeny05 thanks for your contribution. I really like the idea to have the event emitter exposed globally. I'm just concerned about the expose key e.g. `window.EventEmitter`. It was rumored that a W3C Event Emitter will be coming soon (haven't looked in the docs to verify it though).

Therefore I would like to suggest that you'll expose the event emitter with the state manager which is our own domain. When there's the possibility that we can break standard APIs I would like to avoid it beforehand.

Best regards

screeny05:
Good point. Has been done.

mitelg:
thanks for your contribution  @screeny05 :+1:

PR #1337, opened 2017-10-24
Updated currency cookie after context is created

jkrzefski:
The ContextFactory finds the currency based on the currency uuid in the given currency cookie. If there is none, a fallback will be used. This fallback is never persisted to the cookies.

This change writes the currency uuid to the request attributes where it will later be read from and sent back to the browser as a cookie.

Open the Storefront without a currency cookie.

None.

None.

### 6. Checklist


OliverSkroblin:
Hi,
thanks for the pull request. The fallback shouldn't be written back to the request cookie. 
The currency cookie should only be set if the customer switch the currency. We only set the shop cookie for faster routing.

PR #1340, opened 2017-10-25
Fix document type models

windaishi:
The models `Shopware\Models\Document\Document` and `Shopware\Models\Order\Document\Type` are referencing the same database table. So their definition should match.

Also: The model  `Shopware\Models\Order\Document\Type` cannot be persistet because of missing backticks in the ORM annotation

- Equalizes the definitions of the models `Shopware\Models\Document\Document` and `Shopware\Models\Order\Document\Type` 
- Fixes missing backticks in `Shopware\Models\Order\Document\Type` in the ORM annotation for fields 'left' and 'right' so they can be persisted without a MySQL error.
- Fixes several copy paste errors in the doc blocks for the models `Shopware\Models\Order\Document\Type`, `Shopware\Models\Document\Document` and `Shopware\Models\Document\Element` 

```
<?php
use Shopware\Models\Order\Document\Type;

/** @var ModelManager $modelManager */
$modelManager = Shopware()->Container()->get('models');

$newDocumentType = new Type();
$modelManager->persist($newDocumentType);
$modelManager->flush($newDocumentType);
```

No issue.

No changes required.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Document/Document.php
+++ b/engine/Shopware/Models/Document/Document.php
@@ -24,6 +24,7 @@
 
 namespace   Shopware\Models\Document;
 
+use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\ORM\Mapping as ORM;
 use Shopware\Components\Model\ModelEntity;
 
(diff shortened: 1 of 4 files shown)


windaishi:
I appreciate the solution the remove the `Type` model. I was thinking about that, too.
But removing the `Type` model is a breaking change, isn't it? Won't that break the backwards compatibility to some plugins?

PaddyS:
That's actually true, this change would result in a break and all plugins being subscribed to this class in any way would break.
Therefore we'll only release those changes with an according major release.

PR #1343, opened 2017-10-25
Improved autocomplete of synthetic services

shyim:
Class definition is required to see the service in PhpStorm Symfony Plugin

Adds class definitions to synthetic services

### 6. Checklist


htkassner:
Hello @shyim, thank you for your contribution. Sorry but we have to decline this PR due to the fact, that synthetic services are by definition not intended to have a class definition. We do not want to handle this different in comparison to Symfony standards. Keep up the good work.

http://symfony.com/doc/2.8/service_container/synthetic_services.html

PR #1344, opened 2017-10-25
Show menu and shop sites when a error is occurred

shyim:
The user should have also on error page the categories and shop sites to go to a another page

Adds categories informations to error page

Before:
Before
After:
After

Throw a exception in a controller

### 6. Checklist


janbuecker:
Thanks @shyim. We decided to close this PR because it is possible that in the template base exceptions are thrown so you will see an exception stack which may lead to leaks of your server configuration.

PR #1348, opened 2017-10-26
Fix: Javascript Fehler Im Suchfeld bei drücken der Eingabetaste SW-20346

jhit:
It reduces the amount of Javascript Errors on every Shopware store

It calls the proper Javacript object instead of a missing one.

Just open any Shopware Store, type a search term into the search field and wait for the ajax search to load. Do not use the mouse, just hit the Enter key on your keyboard.

https://issues.shopware.com/issues/SW-20346

None

### 6. Checklist


htkassner:
Thank you for your contribution. This will be released with the next release.

PR #1349, opened 2017-10-26
[FEATURE] Added payment api to get all payments

wolv-dev:
Get a list of all available payments (makes it easier for WaWi apis to get information out of shop)

Adds api endpoint for payments

http://example-shop.de/api/payment

https://developers.shopware.com/developers-guide/rest/
New endpoint

Return of API will look like this:
`data: [
{
id: 2,
name: "debit",
description: "Lastschrift",
position: 4,
active: false
},
{
id: 3,
name: "cash",
description: "Nachnahme",
position: 2,
active: true
},
{
id: 4,
name: "invoice",
description: "Rechnung",
position: 3,
active: true
},
{
id: 5,
name: "prepayment",
description: "Vorkasse",
position: 1,
active: true
},
{
id: 6,
name: "sepa",
description: "SEPA",
position: 5,
active: true
}
],
total: 5,
success: true
}`

### DIFF
--- a/engine/Shopware/Components/Api/Resource/PaymentMethods.php
+++ b/engine/Shopware/Components/Api/Resource/PaymentMethods.php
@@ -1,0 +1,246 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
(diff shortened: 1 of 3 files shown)


wolv-dev:
@shopwareBot Added all CRUD operations. Tested all operations and added unit tests.

svenmuennich:
I think the API endpoint should better be called `paymentMethods` instead of `payment` (also note the pluralisation). After all it handles methods of payment rather than single payment instances of an order.

wolv-dev:
@shopwareBot Done!

mitelg:
@wolv-dev now you have two resource classes and two controllers. :wink: 
please remove the old ones

wolv-dev:
@mitelg How was this possible when I just renamed the file? Anyway: I removed the wrong files.

PR #1350, opened 2017-10-26
Fixed variable existence check in themes/Frontend/Bare/frontend/index…

kfrohwein:
You will get a JS error if for anyreason asyncCallbacks was not defined. In JS there are no simple boolean casts like `if(!asyncCallbacks)` so this will result in an error:

```
Uncaught ReferenceError: asyncCallbacks is not defined
at HTMLScriptElement.
```

I apllied a check like described here: https://stackoverflow.com/a/519157

We activated infinite scroll and disabled the load of the main JS for some reason. So the error occured.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/index/script-async-ready.tpl
+++ b/themes/Frontend/Bare/frontend/index/script-async-ready.tpl
@@ -2,7 +2,7 @@
 <script type="text/javascript">
     document.getElementById('main-script').addEventListener('load', function() {
 
-        if (!asyncCallbacks) {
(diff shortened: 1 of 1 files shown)


Phil23:
Hey @kfrohwein,

this change makes no sense. The variable is defined in the standard template and therefore it is always defined as an array, so the normal check should do just fine. If the variable is not defined you made a template change which brakes this behaviour. Also when you want to do a definite check you are missing the check against a `null` value.

PR #1351, opened 2017-10-28
Catch categoryId notice in seo url

shyim:
Catches the fancy notice in seo urls (http://shopware.local/Notice-Undefined-index-categoryId-in/var/www/html/sw53/var/cache/production-REVISION/templates/frontend-Responsive-en-GB-2/16/c6/27/16c62700d0a9aa95f0d43d859a21dd7d24c9e894.string.php-on-line-26/The-Deli-Garage)

Fills the categoryId with null if not exists, to prevent the notice

Add the following to config.php
```php
'front' => [
    'showException' => true,
    'throwExceptions' => true,
    'noErrorHandler' => false,
],
'phpsettings' => [
    'display_errors' => 1,
],
```
and run ./bin/console sw:rebuild:seo:index and you get the fancy urls in campaign urls
heidisql_2017-10-28_15-47-12

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sRewriteTable.php
+++ b/engine/Shopware/Core/sRewriteTable.php
@@ -637,6 +637,9 @@ class sRewriteTable
         $translation = $translator->readWithFallback($shopId, $fallbackShopId, 'emotion', $campaign['id']);
 
         $campaign = array_merge($campaign, $translation);
+        if (!isset($campaign['categoryId'])) {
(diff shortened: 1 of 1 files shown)


soebbing:
Hi @shyim! We already fixed this issue a bit more thoroughly in https://issues.shopware.com/issues/SW-20439

But thanks for your help anyway and keep 'em coming! 💙

PR #1352, opened 2017-10-29
Add new cronjob event

ThomasOpiolka:
Currently there is no way to get a notification about a finished CronJob. 

Such an event would allow plugins to implement an event based cronjob tracking / logging over selected or all jobs. Cronjobs can be a very crucial part that can use that kind of control.

Introducing a new 'notify' event in the cron manager class after a job has finished. 

Inform developers about the new event

### 6. Checklist


MarcelSchmaeing:
Hi, 

thx, this was merged with b62c20a281ed1a0e83dc012d9c03a2fe4d9aec4b

PR #1354, opened 2017-10-29
Added applyDataAttributes to ajax-product-navigation

jkrzefski:
The jquery plugin `jquery.ajax-product-navigation.js` is now configurable via data-attributes.

To have the ability to configure the `ajax-product-navigation` via markup.

It calls `me.applyDataAttributes()` in the plugins init-function, so that the elements data-attributes are merged into the options-array of the plugin instance.

Add some data-attributes to the element this plugin is called on and see that its options remain unchanged.

None.

None.

### 6. Checklist


MarcelSchmaeing:
Hi,

thx, this was merged with: https://github.com/shopware/shopware/commit/3ac4941e1a0be8124c9f4090418012a95d4833c8

PR #1355, opened 2017-10-29
s_core_states id is not autoincrement but the ORM thinks so

JoshuaBehrens:
The column `id` of `table s_core_states` is not set to autoincrement although the ORM annotations in the model hints that. The most obvious solution to just alter the column does not work as the following command:
```sql
ALTER TABLE s_core_states MODIFY COLUMN id INT auto_increment;
```
cannot be executed due to a conflict with a foreign key constraint `s_core_config_mails_ibfk_1` of table `s_core_config_mails`. In addition: a new attribute to the column may conflict with already used solutions to fill this table by community plugins.

It removes the annotation that the `id` column is automagically generated so the ORM will generate a set clause in the SQL when updating/inserting this column.
It adds a setter so you can set the `id`. Current work-around for setting the `id`:
```php
\Closure::bind(function($instance) { $instance->id = /* insert value here */; }, null, $order)($order);
```

```php
$order = new \Shopware\Models\Order\Status();
$order->setPosition(123456789);
$order->setSendMail(0);
$order->setName('prettyIgnorableNameForThisTest');
$order->setDescription('prettyIgnorableNameForThisTest');
$order->setGroup(\Shopware\Models\Order\Status::GROUP_PAYMENT);
Shopware()->Models()->persist($order);
// all right, BOOM!
Shopware()->Models()->flush($order);
```

None

None as there are no breaking changes.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Order/Status.php
+++ b/engine/Shopware/Models/Order/Status.php
@@ -100,7 +100,6 @@ class Status extends ModelEntity
      *
      * @ORM\Column(name="id", type="integer", nullable=false)
      * @ORM\Id
-     * @ORM\GeneratedValue(strategy="IDENTITY")
      */
     private $id;
 
@@ -151,6 +150,14 @@ class Status extends ModelEntity
         return $this->id;
     }
 
+    /**
+     * @param int $id
+     */
+    public function setId($id)
+    {
+        $this->id = $id;
+    }
+    
     /**
      * @return string
      */
--- a/tests/Functional/Models/Order/StatusTest.php
+++ b/tests/Functional/Models/Order/StatusTest.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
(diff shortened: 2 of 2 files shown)


soebbing:
Hi @JoshuaBehrens, your PR got merged in eb97b739ac7a9afccb07deb9454e08471029d495. Thank you for your contribution 💙 and we're sorry it took so long to be merged. :(

JoshuaBehrens:
One year :astonished:  oh boi. Thanks for merging. At least I did not have to rebase :blush:

element-code:
Could you guys give me a quick answer, how plugins should extend the orderstates?
Our current solution  to resolve a valid id is something like this:
```php
$id = $em->createQueryBuilder()
  ->select('MAX(s.id)')
  ->from(Status::class, 's')
  ->getQuery()
  ->getSingleScalarResult() + 1;
```

I actually dont like the idea that this should be the way to go for plugin developers.
Is there any reason why there should not be a generated value for the Table `s_order_states` and if there is a reason, why doesnt it apply to `s_order_detail_states`?

JoshuaBehrens:
@element-code Read my exposition of my PR. If you can solve this I guess everyone will be fine. You could also add a service to do this. But for now you have to manage them your own.

For discussion you could join on gitter.

PR #1356, opened 2017-10-29
Set \Shopware\Models\Order\Status::$description initially to empty string

JoshuaBehrens:
This is pure developer experience. The annotations hints that one is discouraged to use `\Shopware\Models\Order\Status::setDescription` but one is forced to do so as the default value is not compatible with database constraints. Making the database column nullable is considered as a break change as everyone currently expects it to be non-nullable.

One can write a test for this, but the ORM is not usable for this model, yet. See #1355. This PR is not part of #1355 as this is not part of the core problem in #1355 but a nice-to-have.

It sets the default value of description to empty string. Therefore you do not have to use the setter to use this model properly.

Do not use `\Shopware\Models\Order\Status::setDescription` on a new model instance and then insert it.

#1355 as this ships the fix to use the ORM to work with this model on insertion.

None

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Order/Status.php
+++ b/engine/Shopware/Models/Order/Status.php
@@ -118,7 +118,7 @@ class Status extends ModelEntity
      *
      * @ORM\Column(name="description", type="string", length=255, nullable=false)
      */
-    private $description;
+    private $description = '';
 
     /**
      * @var int


JoshuaBehrens:
EVEN WORSE.
It is not even deprecated:
https://github.com/shopware/shopware/blob/5.3/themes/Backend/ExtJs/backend/order/view/list/position.js#L269
https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Models/Order/Repository.php#L137

Which one is the right file to add accessing the snippet manager?

soebbing:
Hi @JoshuaBehrens! I just merged your PR (https://github.com/shopware/shopware/commit/9e3453d908834340d79cc74be2062d08452ec864), thank you so much for helping us to make Shopware better! 💙

JoshuaBehrens:
@soebbing I am happy to help, but what about my additional comment? https://github.com/shopware/shopware/pull/1356#issuecomment-340577152

soebbing:
@JoshuaBehrens Just because something was deprecated does not mean Shopware cannot or should not use the functionality anymore. Shopware still has to support the values that were set for this field!

Deprecations simply means that at one point in the future this function will be removed and gives you time to plan ahead and change your functionality accordingly. For \Shopware\Models\Order\Status::setDescription we currently are are in this transition period: We plan to remove it in the future but currently still support it.

JoshuaBehrens:
Is there already an alternative I can switch to? Or do I have to wait for a not yet released feature/improvement?

soebbing:
The deprecation message is "Use getName() + snippets instead". You can find an example of how it is supposed to work in the Backend/Mail-controller: https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Controllers/Backend/Mail.php#L100: The name of the status get's translated, using the (deprecated) description as a fallback.

JoshuaBehrens:
I tried it, but I was not successful. After I set the description it worked out. I will work out some details to my issue but it is related to the referenced files from the comment.

PR #1357, opened 2017-10-29
Removed obsolete statement from sAdmin::sCheckUser

JoshuaBehrens:
These lines were obsolete because in the following line the value of `$this->sSYSTEM->sUSERGROUP` is always overwritten.

Although core classes will be obsolete soon, this one was really unpleasant while debugging so I just wanted to fix this.

Nothing, but reducing the lines of code and increasing the scrutinizer score by removing the usage of 3 variables that are marked as deprecated. There are no magic getters or setters involved.

None.

None.

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -853,11 +853,6 @@ class sAdmin
             );
             $this->sSYSTEM->sUSERGROUPDATA = $this->sSYSTEM->sUSERGROUPDATA ?: [];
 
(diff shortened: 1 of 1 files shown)


JoshuaBehrens:
The travis build error for php 7.1 seems not to be reproduceable for me.

janbuecker:
Yeah, that's a false positive. I've restarted the build, should be good after.

MarcelSchmaeing:
Hi,

merged this with: https://github.com/shopware/shopware/commit/d15db54ab09327a2e1cc64895f126299d0b673ce

PR #1360, opened 2017-10-30
add event 'Shopware_Modules_Order_SaveOrder_FilterAttributes'

willemmeert:
add a filter event to easily add/modify the order attributes on saving
an order, similar to SW-17847 which added the event for the attributes
of the order detail(s).

Allow easy modification or addition of the order attributes by a plugin on saving an order.

Adds a new filter event.

Subscribe to the event and you're ready to add/read/modify the order attributes.

Similar to SW-17847 which added the event for the order DETAIL attributes (but not for the attributes at the order level).

None

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sOrder.php
+++ b/engine/Shopware/Core/sOrder.php
@@ -646,7 +646,15 @@ class sOrder
 
         $attributeData = array_merge($attributeData, $this->orderAttributes);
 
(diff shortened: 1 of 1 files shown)


mitelg:
thanks for your contribution @willemmeert :+1: :tada:

PR #1362, opened 2017-11-01
Show configurator option names in article variant detail window

windaishi:
When editing the details of a variant article you cannot see which variant you are currently editing because the assigned configurator options are not shown.

Display the variant options in the article variant detail window.

image

### 6. Checklist


MarcelSchmaeing:
Hi,

thx, merged with 1c4b306351eb31f923d47ac854bdf88d4a295a25

PR #1363, opened 2017-11-01
Fixed theme smarty loading on finish page

shyim:
Currently on the finish page the smarty plugins are not added because, we save the theme settings in the Session Variables and restore it and the ConfigLoader checks for theme variable. If it is empty it fills the variable and adds the smarty folders
See: https://github.com/shopware/shopware/blob/9a116f29f78c0005e031e34ffcce919b6a50e3a9/engine/Shopware/Components/Theme/EventListener/ConfigLoader.php#L79

Removes theme from session variables

Register a custom smarty function and use these in template and open the finish page.

### 6. Checklist


themazim:
+1

shyim:
Fixed with https://github.com/shopware/shopware/commit/056488f8a75790c6470e679ffd461740ae2ef309

PR #1366, opened 2017-11-03
change searchcolumn for article numbers to search for variants

egoist-dd:
If you want to add articles in cross selling tab or stream you have to use always the number of the main detail, if you enter the number from a variant there are no results

allows to search for variants on cross selling tab, filtered streams tab or all filter forms wich use /base/getArticles as data source

open cross selling tab and enter a article number wich is no main detail -> no results
with this PR you will get the article

--

--

### 6. Checklist


egoist-dd:
that function is not only used for cross selling selection, but i checked all usages.

I think there is no problem to add this, because the relation between the elements (articles, blogentries, banners ,...) and assigned articles is made by the article id, so there should be no problem with showing variants in a non-variant supported field

PR #1368, opened 2017-11-06
Fixed captcha validation with double optin

shyim:
The captcha is also in double optin checked and we can't register to newsletter anymore.
bildschirmfoto 2017-11-06 um 12 27 28

Ignores captcha check if newsletter validation is confirmed.

I tested also newsletter registration in account. It works also after this patch 😄 

Enable Double-Optin Newsletter and change captcha to numbers and characters.
Try to register to the newsletter

### 6. Checklist


soebbing:
Hi @shyim! Thanks once again for your help! 💙

PR #1371, opened 2017-11-09
[Bugfix] Consider also currency factors smaller than 1

larsbo:
The condition should consider all currency factors not equal 1.

Replacing the condition inside the document template.

Create a currency in the backend with a factor larger than 1. There will be no currency factor notice in invoice documents.

### 6. Checklist


MarcelSchmaeing:
Hi,

thx, we merged your PR with this commit: https://github.com/shopware/shopware/commit/5096a8c73b31719cae7226e78adcf9039fabe002

PR #1374, opened 2017-11-13
[BUGFIX] Prevent Fatal error when links or download is null

saschanowak:
To prevent the following exception:
Fatal error: Uncaught TypeError: Argument 1 passed to Shopware\Components\Api\Resource\Article::translateAssociation() must be of the type array, null given

Add a condition around the translations

### 6. Checklist


mitelg:
thanks for your contribution @nlx-sascha  :+1: :tada:

PR #1377, opened 2017-11-15
Prevent missing locale from crashing snippet import

hanneswernery:

Right now if you import snippets to the db while installing a plugin and if a specific entry in `s_core_locales` is missing, the whole import will crash with the following exception `Call to a member function getId() on null` in [this line](https://github.com/shopware/shopware/blob/d4c4f2e731d94c9a385791fc5898e9c3c4a18479/engine/Shopware/Components/Snippet/DatabaseHandler.php#L138) and the plugin installation will fail.
We encountered multiple customers who _cleaned up their database_ by removing some locales. Which is bad practise by itself, but it also causes all of our plugin installations to fail.

This PR adds a safeguard that checks if the requested locale does in fact exist before importing a snippet for it. If no locale was found the specific snippets will be skipped. This way all remaining (and working) snippets can still be imported normally, which is better than throwing an exception (which also prevents the plugin from being installed) and probably what mentioned customers want to happen.

If you delete a locale that is needed in the snippets of a plugin from the db and try to import snippets of said plugin, it will fail for all remaining snippets of this plugin.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Snippet/DatabaseHandler.php
+++ b/engine/Shopware/Components/Snippet/DatabaseHandler.php
@@ -135,9 +135,12 @@ class DatabaseHandler
                     $locale = $localeRepository->findOneBy(['locale' => $index]);
                 }
 
-                $databaseWriter->write($values, $namespacePrefix . $namespace, $locale->getId(), 1);
+                // Only write entry if locale was found
+                if ($locale) {
+                    $databaseWriter->write($values, $namespacePrefix . $namespace, $locale->getId(), 1);
 
-                $this->printNotice('<info>Imported ' . count($values) . ' snippets into ' . $locale->getLocale() . '</info>');
+                    $this->printNotice('<info>Imported ' . count($values) . ' snippets into ' . $locale->getLocale() . '</info>');
+                }
             }
 
             $this->printNotice('<info></info>');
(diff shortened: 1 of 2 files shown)


engine/Shopware/Components/Snippet/DatabaseHandler.php
```diff
@@ -135,9 +135,12 @@ public function loadToDatabase($snippetsDir = null, $force = false, $namespacePr
                     $locale = $localeRepository->findOneBy(['locale' => $index]);
                 }
 
-                $databaseWriter->write($values, $namespacePrefix . $namespace, $locale->getId(), 1);
+                // Only write entry if locale was found
+                if (!$locale) {
```
mitelg: shouldn't this be `if ($locale) {` ?

mitelg (changes requested):

engine/Shopware/Components/Snippet/DatabaseHandler.php
hanneswernery: Of course it should. Thanks!

mitelg (approved):

hanneswernery:
I've added a test that fails without the my changes in the `DatabaseHandler.php`.

cjuner:
@hanneswernery, please rebase for Shopware 5.5, or delegate it appropriately.

soebbing:
Hi @hanneswernery, thank you very much for your contribution! 🎉

PR #1378, opened 2017-11-15
Catch Throwable on cronjobs

shyim:
Currently Throwables are not catched by cronjob manager. (http://php.net/manual/de/language.errors.php7.php)

Catches also Throwable's on on php7.x systems

Create a cronjob which sends a mail and have a error in mailtemplate like {$orderTime.date} $orderTime is a DateTime so it throws a unhandled exception currently.

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Components/Cron/Manager.php
+++ b/engine/Library/Enlight/Components/Cron/Manager.php
@@ -286,6 +286,21 @@ class Enlight_Components_Cron_Manager
             ]);
 
             throw $e;
+        } catch (\Throwable $e) {
+            $job->setData(['error' => $e->getMessage()]);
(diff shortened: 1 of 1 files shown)


shyim:
@shopwareBot It is compatible with PHP5.6. I have tested it and many php libraries are using it currently and they also supporting PHP 5.6. Just search in shopware vendor Folder for "Throwable", also symfony using it currently 😛

mitelg:
@shyim I dare you, if this will not work! :laughing:

EtienneBruines:
Regarding compatibility, this may be of interest: 
http://php.net/manual/de/language.errors.php7.php#119652

PR #1031, opened 2017-03-02
Document order patch 1

JonidBendo:
- Add Event after the positions are processed for the invoice generation
- Add set and get methods for the protected variables to allow further modifications for taxes and etc.

## Description
Please describe your pull request:
* Why is it necessary?
There is no other way to actually work/edit the invoice generation system

* What does it improve?
It allows plugin developers to use the event and modify the invoice details

* Does it have side effects?
If not used properly from the plugin developers it might break the invoice generation methods

| Questions        | Answers
| ---------------- | -------------------------------------------------------
| BC breaks?       | no
| Tests pass?      | yes
| Related tickets? | If this PR fixes an existing [Issue](https://issues.shopware.com) ticket, please add its complete issue URL.
| How to test?     | Simply call the event in a plugin...

PR #1380, opened 2017-11-16
Add config flag to allow multiple documents of the same type in an order

hanneswernery:

Currently it is not possibly to create multiple documents of the same type for an order. If another document of type X is created, the the first document is simply overwritten, instead of creating an additional one. For example this prevents the creation of partial invoices or partial delivery notes where you need multiple documents of the same type.

This PR adds an optional config field to the Document component that allows the creation of multiple documents for the same type. This is a purely programmatic change and does not have any immediate effect when just using "plain" Shopware via the frontend or backend. It does have a lot of use cases for plugins though, especially together with the pretty new possibility to filter positions on documents (see https://github.com/shopware/shopware/pull/1031).
Currently super ugly hacks are required to achieve multiple documents of the same type for an order, but there is clearly a need for this, as can be seen with this plugin for example: http://store.shopware.com/mbdus41476730029/teildokumente-erstellen.html

Create a document for a certain type for an order. Create another one. The first document will be overwritten, no additional document will be created.

n/a

None, since it just adds an additional option in the PHP code.

### 6. Checklist

### DIFF
--- a/UPGRADE-5.5.md
+++ b/UPGRADE-5.5.md
@@ -21,6 +21,7 @@ This changelog references changes done in Shopware 5.5 patch versions.
     * `cart_item_rebate.tpl`
     * `cart_item_surcharge_discount.tpl`
     * `cart_item_voucher.tpl`
+* Added option to allow multiple documents of the same type for orders in `engine/Shopware/Components/Document.php`.
 
 ### Changes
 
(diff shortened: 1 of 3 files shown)


cjuner:
@hanneswernery, please rebase to Shopware 5.5, or delegate appropriately.

soebbing:
Hi @hanneswernery, thank you for your contribution! Your change was merged in https://github.com/shopware/shopware/commit/172e2d9278ba8a80139ff1e7b8fd0a9399a2ccdd 🎉

buddhaCode:
Maybe someone needs a plugin to make this available in backend GUI: https://github.com/buddhaCode/BucoAllowMultipleDocuments

PR #1382, opened 2017-11-16
Normalise attribute name, if no data was found

Diu:
So that data attributes can have a consistent style, if they are used to change plugin settings.

It adds a check for a 'normalised' data attribute, when a camel case attributes wasn't found. The attribute "data-tabName" could now be called "data-tab-name", for example.

Change an existing data attribute, which uses camel case, to the described format and verify that the setting still gets applied to the javascript plugin.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js
@@ -46,6 +46,16 @@
     }
 
     /**
+     * Replaces uppercase characters with a dash and its lowercase character.
+     *
+     * @param {String} key
+     * @returns {String}
+     */
+    function normaliseKey(key) {
+        return key.replace(/([A-Z])/g, '-$1').toLowerCase();
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


klarstil:
Hey @Diu 
thanks for your contribution. I like the idea but the change is quite dangerous. For example what happens when a property misspelled (upper & lower case) in HTML and the change applies it to the plugin options? We have to check every jQuery plugin we're having the frontend as well as any third-party plugin. What would be an approach which will be less dangerous and provides the same or almost the same behavior? My idea would be to provide an argument to the `applyDataAttributes()` method which enables the behavior like a feature switch. It would be nice if you can change your PR and provide the argument (default: `false`) and update it on here, so we can review it and schedule it for the next Shopware version.

Diu:
I'm not sure if I can follow your problem.
The setting `tabName` would previously only be overwritten by `data-tabName`. Now it can also be overwritten by `data-tab-name`, but only if the previously mentioned attribute wasn't found. Typos would break the functionality in both cases.

Something like adding a parameter seems like a good idea anyway. I'll look into it, when I have time.

PR #1383, opened 2017-11-16
Enable store autoload in the select element component

rephluX:

Given we have a plugin config.xml containing a combo element:

```
<element type="combo">
    <name>orderStates</name>
    <label>Order status</label>
    <store>Shopware.apps.Base.store.OrderStatus</store>
</element>
```
The appropiate option values for the dropdown list are not fetched from the backend and the list remains empty. Even when the select list is clicked, no options will be lazy loaded. On Chrome DevTools no ajax request is submitted to the backend that fetches any option values.

That behaviour differs to prior versions of Shopware, for example the latest Version for the 5.2 branch.

It enables the `autoload` property for the store on the select element. When the plugin form is opened, the store will be loaded and the options will be fetched from the backend.

Create a plugin config with at least one (remote) combo element and choose a ExtJs store to select options from (e.g `Shopware.apps.Base.store.OrderStatus` or `Shopware.apps.Base.store.PaymentStatus`)

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/base/component/element/select.js
+++ b/themes/Backend/ExtJs/backend/base/component/element/select.js
@@ -133,7 +133,8 @@ Ext.define('Shopware.apps.Base.view.element.Select', {
         try {
             store = Ext.create(storeId, {
                 storeId: configSelectStoreId,
-                pageSize: 1000
+                pageSize: 1000,
+                autoLoad: true
             });
             // Override the load() method so the store is only ever loaded once
             Ext.override(store, {


rephluX:
Actually, this change is not needed, because the `value node` is missing on the elements configuration. I've spend some time researching the the latest code change on the file (see PR #1130). As mentioned in the linked PR, the autoload is not needed, because the form will force load the store when the `setValue` Method is being executed on the element.

mitelg:
hey @rephluX 
thanks for your answer. I will close the PR then :blush: :+1:

EtienneBruines:
@rephluX but isn't the node tag optional? In that case the store never gets loaded.

rephluX:
@EtienneBruines You mean the _value_ node tag in the xml? Yes, that node is indeed optional. In my case i could provide an initial value for the config element to load the store.

But maybe the _setValue_ Method in the select.js file should be refactored slightly to also load the store when the value **is null** (e.g. when no value node is provided in the xml) and only setting the actual value in the dropdown when the value **is not null**.

EtienneBruines:
> But maybe the setValue Method in the select.js file should be refactored slightly to also load the store when the value is null (e.g. when no value node is provided in the xml) and only setting the actual value in the dropdown when the value is not null.

Or loading the store when someone 'opens' the dropdown. That would be a solid alternative as well. 

(Is anyone able to re-open this issue? Doesn't feel like it's quite closed just yet. There's a workaround, but no real fix.)

PR #1385, opened 2017-11-18
fixed generating emotion landingpage seo urls

kekster:

Generating emotion landingpages for every shop available is
1.) unnecessary because the emotion isnt available for every shop
2.) bad because we can overwrite existing, actually valid seo urls for selected shops

The emotion backend controller generates only seo urls for selected shops when saving an emotion. This should be the default behaviour.

Only generate seo urls for those shops which are selected in the emotion.

Create two emotion landingpages with the same name (for example: "Supplier XY")
Emotion A for shop A
Emotion B for shop B
and run the seo refresh cronjob. The emotion B will generate seo urls for both shop A and B and will overwrite the correct seo url from emotion A.

None.

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sRewriteTable.php
+++ b/engine/Shopware/Core/sRewriteTable.php
@@ -619,6 +619,9 @@ class sRewriteTable
             $fallbackId = $fallbackShop->getId();
         }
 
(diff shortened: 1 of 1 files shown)


mitelg:
thanks for your contribution @kekster  :+1: :tada: 

fixed with 65cf00374a9

PR #1388, opened 2017-11-19
Amend Shopware/Models/Order/Detail::getTaxRate

JoshuaBehrens:
Because it is a minor ~~upfuck~~ discomfort for IDEs and their users.

It just fixes a type in phpDoc.

```php
/** @var $model \Shopware\Models\Order\Detail */
$model = /* --- */;
$taxRate = $model->getTaxRate(); // this gets hinted as string
var_dump($taxRate); // this will dump float
```

None.

It is a documentation change.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Order/Detail.php
+++ b/engine/Shopware/Models/Order/Detail.php
@@ -746,7 +746,7 @@ class Detail extends ModelEntity
     }
 
     /**
-     * @param string $taxRate
+     * @param float $taxRate
      */
     public function setTaxRate($taxRate)
     {
--- a/engine/Shopware/Models/Order/Detail.php
+++ b/engine/Shopware/Models/Order/Detail.php
@@ -754,7 +754,7 @@ class Detail extends ModelEntity
     }
 
     /**
-     * @return string
+     * @return float
      */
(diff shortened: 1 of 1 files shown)


engine/Shopware/Models/Order/Detail.php
```diff
@@ -754,7 +754,7 @@ public function setTaxRate($taxRate)
     }
 
     /**
-     * @return string
+     * @return float
      */
     public function getTaxRate()
```
mitelg, line 759: the setter should also be changed then...

mitelg (changes requested):

engine/Shopware/Models/Order/Detail.php
JoshuaBehrens: Well I guess I did not think my ~~rant~~ amendment through.

mitelg (approved):

MarcelSchmaeing:
Hi,

thx. We merged this with this commit: https://github.com/shopware/shopware/commit/4089ead315e47fc80a52069c2a3b41700b07ba81
into the 5.4 Branch.

Greetings, 

Marcel

PR #1390, opened 2017-11-20
add a new snippet for decimal precision of article price

mnaczenski:
When using net prices it would be helpful to be able to add prices with a higher decimal precision than two. Especially when you want to get a rounded gross-price.

Change the decimal precision of the price column to a snippet, so you can change the precision in the backend.

Try to add 4,1256 as net price to an article

### 6. Checklist


mitelg:
why didn't you add these new snippets to the ini files?

mnaczenski:
@mitelg done

mitelg:
@mnaczenski you forgot the german "translation". also the value in the ini file defers from the value in the javascript files

mnaczenski:
@mitelg thank you for the hint.
I didn't recognize that there are two languages in the ini file.

PR #1392, opened 2017-11-20
Fix voucher tax model fields

windaishi:
When a voucher is configured with a fix tax group instead of the special choises 'automatic, 'default' or 'none', the tax rate is not preselected in the tax configuration combo box of the voucher configuration. Instead there is just the id in the combobox. 

In the ExtJS model, declare the id explicit as string and not as automatic. This fixes the problem and autoselects the correct entry when opening 

Creating a new vouche with (for example) 19 % tax:
image

Reopen the voucher:
image

The tax configuration is '1' instead of '19 %'
image

### 6. Checklist


cjuner:
@windaishi, please rebase for Shopware 5.5, or delegate it appropriately.

windaishi:
I rebased to 5.4 because this can be merged as a bugfix for 5.4

PR #1393, opened 2017-11-20
Fixed range slider values

shyim:
When i filter the price from 415€ to unlimited the range slider value is 416.55. It makes no sense. The value of the label and the value of the input should be the same, otherwise the results are wrong.

Uses the same rounding of label to the value in url, so it should be equal anytime.

Goto https://www.shopwaredemo.de/hoehenluft-abenteuer/ausruestung/ski/
Filter price from 415€ to unlimited and show the url it outputs ?p=1&o=1&n=12&min=416.55

should be ?p=1&o=1&n=12&min=415.00
😕 

### 6. Checklist


shyim:
Fixed with https://github.com/shopware/shopware/commit/bea330e9d28fbc36a1e4fa576a298da9fbc261ad

PR #1394, opened 2017-11-21
Add unique placeholders in condition parser trait

Scribblerockerz:
The condition is not unique if the same attribute is used for multiple
conditions.

The introduced DynamicConditionParserTrait has a bug in it's current implementation regarding placeholder name generation if multiple conditions on the same attribute are applied.

It generates a hash placeholder based on the `field`,  `operator` and `value` of the condition.

Add a new `ProductAttributeCondition ` with an "attribute" `OPERATOR_LT` and a second `ProductAttributeCondition` with the same "attribute" and `OPERATOR_GT operator`.

### 6. Checklist


mitelg:
thanks for your contribution @Scribblerockerz :+1: :tada:

PR #1398, opened 2017-11-22
Exclude hidden categories from sidebar menu

Amaridian:
Issue SW-16689
Categories marked as "hidden" for the main navigation will still be displayed in the offcanvas menu.

It adds the same if-clause as in /themes/frontend/Bare/frontend/index/main-navigation.tpl line 20

Look at any offcanvas menu and compare it to the corresponding main navigation when a category is set as "hidden"

https://issues.shopware.com/issues/SW-16689

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/index/sidebar-categories.tpl
+++ b/themes/Frontend/Bare/frontend/index/sidebar-categories.tpl
@@ -6,29 +6,31 @@
 
         {block name="frontend_index_categories_left_before"}{/block}
         {foreach $categories as $category}
-            {block name="frontend_index_categories_left_entry"}
-                <li class="navigation--entry{if $category.flag} is--active{/if}{if $category.subcategories} has--sub-categories{/if}{if $category.childrenCount} has--sub-children{/if}" role="menuitem">
-                    <a class="navigation--link{if $category.flag} is--active{/if}{if $category.subcategories} has--sub-categories{/if}{if $category.childrenCount} link--go-forward{/if}"
-                        href="{$category.link}"
-                        data-categoryId="{$category.id}"
-                        data-fetchUrl="{url module=widgets controller=listing action=getCategory categoryId={$category.id}}"
-                        title="{$category.description|escape}"
-                        {if $category.external && $category.externalTarget}target="{$category.externalTarget}"{/if}>
-                        {$category.description}
+            {if !$sCategory.hideTop}
+                {block name="frontend_index_categories_left_entry"}
+                    <li class="navigation--entry{if $category.flag} is--active{/if}{if $category.subcategories} has--sub-categories{/if}{if $category.childrenCount} has--sub-children{/if}" role="menuitem">
+                        <a class="navigation--link{if $category.flag} is--active{/if}{if $category.subcategories} has--sub-categories{/if}{if $category.childrenCount} link--go-forward{/if}"
+                            href="{$category.link}"
+                            data-categoryId="{$category.id}"
+                            data-fetchUrl="{url module=widgets controller=listing action=getCategory categoryId={$category.id}}"
+                            title="{$category.description|escape}"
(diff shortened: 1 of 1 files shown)


shyim:
Looks similar to #809

Amaridian:
@shyim yes, it is. But we are fixing this missbehavior in each and every customer installation of shopware. The offcanvas menu replaces the top navigation in mobile view so shown menu-items should be the same.

Another possibility would be to add a "hide in sidebar navigation" option beside the "hide in top navigation" checkbox in the backend.

Gugiman:
My PR #809 was closed because this config option "works as intended".
The second try on #975 got closed either...

I would be glad and confused at the same time if this PR get merged :D

MarcelSchmaeing:
Hi @Amaridian,

the initial intention of this config option was to hide a top level category element, in case you have to many  top level categories to display them properly.

If this option controls the offcanvas category menu, too. The user has no chance to display the category on a mobile device. So it comparable to simply deactivate the category at all. 
Of course there might be a few edge cases in which you can use this option for a different behavior. 

But we think it is not a benefit for the product to implement your PR. 

We hope you can understand the decision.

Therefore we will close this PR.

Greetings,

Marcel

MFCBS:
@MarcelSchmaeing perhaps you could suggest how we hide categories in the mobile menu then? Why is the functionality different between devices? In my eyes this is broken. As @Amaridian says, we too are fixing this in every customer installation, almost all of our customers report it as a bug

MarcelSchmaeing:
Hi @MFCBS,

ok, maybe your requirements are not clear. 
It makes no sense to me that this option will also hide the category in the mobile menu. Therefore the category will only be displayed in the sidebar on desktop devices?

Otherwise, if this option will also hide the category in the sidebar on desktop devices you can simply deactivate the category at all.

Which behavior is desired?

Greetings,

Marcel

oktupol:
> Otherwise, if this option will also hide the category in the sidebar on desktop devices you can simply deactivate the category at all.

I disagree. A category not shown in the menu can still be accessed if you know the category's exact url, which can be distributed by e.g. newsletters or shopping worlds. I don't have any examples, but I can see why this might be a desired feature. A disabled category can't be accessed at all.

MFCBS:
@MarcelSchmaeing thanks for coming back to me

An example from one of our customers, they're using a category with sub-categories for manufacturer promotions, e.g. buy this TV and get a DVD player for free. The customer does not want the category to appear in the top navigation but they still need it accessible via the URL for newsletters etc, therefore the category can not be deactivated. We've unticked "Display in top navigation" and it is correctly hidden from the top navigation on desktop, but shows (incorrectly) on the mobile menu.

I'm really struggling to understand how this checkbox is functioning correctly!

MarcelSchmaeing:
Ahh ok, 
so like I wrote, this case was never the intention of this field.
The right way is to work f.e. with Landingpages or something else.

The categories were never planned to work like this.

Of course you can always write a small plugin with a new attribute field in the category module to solve this individual case.

So right it is not planned to change this field to a general "Hide Category" field.
But you can of course write your requirements in this ticket and vote for it:

https://issues.shopware.com/issues/SW-16689

Greetings,

Marcel

MFCBS:
Thanks @MarcelSchmaeing, can you think of any other way we could structure categories to achieve the desired functionality? As far as I can see, if the intention of this checkbox is not to hide categories, there is no way in Shopware to achieve what we need to achieve?

MarcelSchmaeing:
You are right if the landingepages are not enough.
So it is more like a missing feature which I don't want to implement with this field.

MFCBS:
Thanks @MarcelSchmaeing I've voted for the ticket. In my eyes this is a higher priority than minor -
 customers get different experiences between desktop and mobile at the moment and that's pretty poor!

Thanks again

PR #1402, opened 2017-11-23
Bugfix: select-values not always showing in plugin config

EtienneBruines:
When not autoloading, the loading mechanism isn't always triggered. 

This fix makes sure the loading happens on time, in order to show
the options to the user. (By setting `autoLoad` to true.) This fixes some
issues that arose when `autoLoad` was removed two months ago. 

1. Have a plugin with config file like:

```xml
<element type="select">
            <name>standardCurrency</name>
            <label lang="en">Default Currency</label>
            <label lang="de">Standard Währung</label>
            <store>Shopware.apps.Base.store.Currency</store>
</element>
```

2. Install it
3. Attempt to configure this field (using the plugin manager). No options will be shown/visible/selectable. 
screenshot-2017-11-23_14 40 58

N/A

N/A

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/base/component/element/select.js
+++ b/themes/Backend/ExtJs/backend/base/component/element/select.js
@@ -133,7 +133,8 @@ Ext.define('Shopware.apps.Base.view.element.Select', {
         try {
             store = Ext.create(storeId, {
                 storeId: configSelectStoreId,
-                pageSize: 1000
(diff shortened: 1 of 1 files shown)


mitelg:
same as #1383

EtienneBruines:
Agreed. Moving discussion there.

PR #1403, opened 2017-11-23
fix emotion banner component

resslinger:
If a emotion banner has no retina source image, the image will not display on the site actually.

This change fix this issue.

Delete all retina images for a banner image.

### 6. Checklist


mitelg:
thanks for your contribution @lacodimizer   :+1: :tada:

PR #1406, opened 2017-11-24
fix incompatible event names with sw52 service decorating 

ghost:
fix incompatible event names with sw52 service decorating by services.xml in new plugin-structure

With the conversion of plugins to the new Shopware 5.2 plugin structure and the related service decoration via services.xml, there are many incompatibilities of plugins that decorate the same service, as the "old" plugins, e.g. listen to the event "Enlight_Bootstrap_AfterInitResource_shopware_storefront.list_product_service" even though the container is fired when decorating via services.xml the event "Enlight_Bootstrap_AfterInitResource_my_decorated_service.list_product_service".

fires a fallback event for all "old" plugins with the original servicename

1. run Shopware
2. install 2 plugins which decorate the ListProductService once by event and once by services.xml
3. see that the plugin which decorates by event wont run

---
### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/Container.php
+++ b/engine/Shopware/Components/DependencyInjection/Container.php
@@ -87,9 +87,10 @@ class Container extends BaseContainer
      */
     public function has($name)
     {
+        $fallbackName = $this->getFormattedId($name);
         $name = $this->getNormalizedId($name);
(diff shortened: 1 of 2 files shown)


viosys:
CLA are signed.

PR #1407, opened 2017-11-25
Add attributes to pagetree

magnolia4:

To have the ability to use the custom page attributes for the footer links

### 6. Checklist

- [ ]  I have squashed any insignificant commits

### DIFF
--- a/engine/Shopware/Components/SitePageMenu.php
+++ b/engine/Shopware/Components/SitePageMenu.php
@@ -151,6 +151,8 @@ class SitePageMenu
             }
 
             $site['childrenCount'] = count($site['subPages']);
+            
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/SitePageMenu.php
```diff
@@ -151,6 +151,8 @@ private function buildSiteTree($parentId, $sites, $activeId)
             }
 
             $site['childrenCount'] = count($site['subPages']);
+            
+            $site['attribute'] = Shopware()->Container()->get('shopware_attribute.data_loader')->load('s_cms_static_attributes', $id);
```
mitelg, line 155: please inject the service you need via the constructor

magnolia4:
Basically it's correct but at this point no translations are required because you have to create a separate tree for each language.

By the way all other fields aren't translatable, too:
image

PR #1408, opened 2017-11-27
fix static analysis errors in shopwares core files

jochenmanz:
shopware has numerous wrong typehints and even missing use statements. This PR tries to eliminate some of them

fixes a huge amount of wrong typehints, missing use statements, duplicate array key usage, missing parameters, unnecessary function calls, unused methods and some generic optimisations for less error prone static analysis of the shopware core

phpstorm should display most of the errors when comparing the files with the older version. For a more complete look you need to install phpstan and run a level=1 analysis. 

none

none

### 6. Checklist


engine/Shopware/Components/Api/Resource/Cache.php
```diff
@@ -55,13 +55,13 @@ class Cache extends Resource implements ContainerAwareInterface, BatchInterface
     /**
      * Sets the Container.
      *
-     * @param Container $container
+     * @param Container $Container
      */
-    public function setContainer(Container $container = null)
+    public function setContainer(Container $Container = null)
```
mitelg, line 60: why the variable name in uppercase?

engine/Shopware/Components/Api/Resource/Resource.php
```diff
@@ -98,11 +98,11 @@ public function getContainer()
     }
 
     /**
-     * @param $container
+     * @param Container $Container
      */
-    public function setContainer($container)
+    public function setContainer(Container $Container = null)
```
mitelg, line 103: why the variable name in uppercase?

engine/Shopware/Components/Emotion/Preset/PresetLoader.php
```diff
@@ -57,7 +57,7 @@ public function load($presetId)
         $preset = $this->modelManager->getRepository(Preset::class)->find($presetId);
 
         if (!$preset) {
-            throw new NoResultException('The preset with id ' . $presetId . ' could not be found.');
+            throw new RuntimeException('The preset with id ' . $presetId . ' could not be found.');
```
mitelg: this is a break, isn't it?

engine/Shopware/Models/Country/Repository.php
```diff
@@ -277,37 +277,4 @@ public function getStatesByCountryIdQueryBuilder($countryId)
...
-
-    /**
-     * Returns an instance of \Doctrine\ORM\Query object which selects a
-     * list of countries for the passed area id.
-     *
-     * @param $stateId
-     *
-     * @return \Doctrine\ORM\Query
-     */
-    public function getStateAttributesQuery($stateId)
```
mitelg: removing public methods is a break

engine/Shopware/Components/Api/Resource/Resource.php
jochenmanz: Shopwares ContainerAwareInterface says it needs to be uppercase :)

engine/Shopware/Components/Api/Resource/Cache.php
jochenmanz: Shopwares ContainerAwareInterface says it needs to be uppercase :)

engine/Shopware/Models/Newsletter/Address.php
```diff
@@ -179,22 +180,6 @@ public function getIsCustomer()
         return $this->isCustomer;
     }
 
-    /**
-     * @param int $lastMailingId
-     */
-    public function setLastMailingId($lastMailingId)
```
jochenmanz: Also a BC break then. But how to proceed with a function if the class property is missing?

engine/Shopware/Models/Newsletter/Newsletter.php
```diff
@@ -467,22 +470,6 @@ public function getId()
         return $this->id;
     }
 
-    /**
-     * @return \Doctrine\Common\Collections\ArrayCollection
-     */
-    public function getOrders()
-    {
-        return $this->orders;
```
jochenmanz: also a bc break, but unfortunatly the class proerty orders and alreadySentTo doenst exist. Should we warn via a trigger_error?

PR #1409, opened 2017-11-27
php-cs-fixer adjustments

jochenmanz:
The codebase has some open php-cs violations. This PR fixes them

Fixes all outstanding violations

run php-cs-fixer on the current codebase

### 6. Checklist

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -39,7 +39,6 @@ return PhpCsFixer\Config::create()
         'no_useless_return' => true,
         'ordered_class_elements' => true,
         'ordered_imports' => true,
-        'yoda_style' => false,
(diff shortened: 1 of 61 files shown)


.php_cs.dist
```diff
@@ -39,6 +39,7 @@ return PhpCsFixer\Config::create()
         'no_useless_return' => true,
         'ordered_class_elements' => true,
         'ordered_imports' => true,
+        'yoda_style' => false,
```
mitelg: afaik, the current php-cs-fixer version which is used in Shopware does not support this config. I think a composer update is required

jochenmanz: reapplied all changes with the old php-cs-fixer version.

PR #1413, opened 2017-11-30
Document Creation - Remove wrong UstID table relation

Gugiman:
The change improves the ability to edit data for only one particular order.

Changes dependency of UstID setting from "default Billing Adress" to "order billing adress"

When changing UstID settings of an order in the detail settings, Shopware still uses the UstID setting of the default billing address. So the change in the first steps doesn't take any effect on the document creation unless the default billing address is changed..

### 6. Checklist


Gugiman:
Done

PR #1415, opened 2017-12-01
Fixed exception message in TableMapping::getDependingTables

oktupol:
To get a more useful error message when having a typo in an attribute table

Currently, the `sprintf` function call has too few arguments, resulting in an E_WARNING and returning `false` instead of a useful exception message. This change adds the table name as argument.

```php
$container->get('shopware_attribute.crud_service')->update(
    'this_is_not_an_attribute_table',
    'column',
    'string'
);
```

Expected: Uncaught \Exception with the message "Table this_is_not_an_attribute_table is no supported attribute table"

Currently this yields an E_WARNING: "sprintf(): Too few arguments in [...]". If the interpreter is set not to halt on warnings, you'll additionally get an Exception without any message.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/AttributeBundle/Service/TableMapping.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Service/TableMapping.php
@@ -166,7 +166,7 @@ class TableMapping
     public function getDependingTables($table)
     {
         if (!$this->isAttributeTable($table)) {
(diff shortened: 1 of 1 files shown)


mitelg:
this PR also fixes this issue: https://github.com/shopware/shopware/pull/1408/files#diff-3501427db26683f6d29cb4e98fc118e3

PR #1417, opened 2017-12-01
[BUGFIX] Fixed ordering of blog entries in shopping world element

wolv-dev:
Right now the blog entries are ordered by ID ASCENDING. So the last one will be shown in first position.

Reorders the elements by display_date.

Add the blog emotion element to the shopping world, set it to display two or three. Add more blog entries to the blog category than displayed in the emotion element.

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/BlogService.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/BlogService.php
@@ -61,11 +61,16 @@ class BlogService implements Service\BlogServiceInterface
      */
     public function getList(array $ids, Struct\ShopContextInterface $context)
     {
+        $result = [];
         $blogs = $this->blogGateway->getList($ids, $context);
 
         $this->resolveMedias($blogs, $context);
 
-        return $blogs;
+        foreach($ids as $id) {
+            $result[$id] = $blogs[$id];
+        }
+
+        return $result;
     }
 
     /**


OliverSkroblin:
Hey @wolv-dev,

thanks for the PR. The storefront data gateways shouldn't sort the main entity. All sortings are handled outside the services and gateways. 
The provided id list should be used for sorting.
You can see an example here:

https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/ListProductService.php#L148

The service layer should iterate the provided id array and create a new result which sorted by the id.
The blog sorting is already handled at this point:
https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Bundle/EmotionBundle/ComponentHandler/BlogComponentHandler.php#L123

wolv-dev:
@OliverSkroblin So why isn't the sorting working when it's already implemented?

When the sorting of the blog Ids is correct, but the returned blog entries are not sorted by blog.display_date', 'DESC' than there's an error in the gateway returning entries not sorted by passed Ids.

wolv-dev:
@shopwareBot Okay, I understand. Will change it.

wolv-dev:
@shopwareBot @OliverSkroblin I changed it so that the ordering isn't happening in the gateway.

mitelg:
thanks for your contribution @wolv-dev  :+1: :tada:

PR #1418, opened 2017-12-01
Avoid requesting abstract controllers

shyim:
Avoid requesting /backend/application which fires a Fatal Error (see: https://shopwaredemo.de/backend/application). It can spam customers with error mails or other usages.

Makes Application Controller in backend abstract and handles abstract Controller as 404.

Request /backend/application

Shopware_Controllers_Backend_Application is now abstract

### 6. Checklist


shyim:
It works also on shopware.com (https://de.shopware.com/backend/application) 😂  😂 . 
I will create a issue, but i guess it would be never fixed.

PR #1420, opened 2017-12-02
use typecast instead of intval, strval, doubleval and floatval

uehler:
intval, strval, doubleval and floatval are old and slow typecasts

removes intval, strval, doubleval and floatval in the engine/Shopware directories and does the typecast via (int), (string), (double) and (float) which is much quicker.

### 6. Checklist


mitelg:
thanks for your contribution @uehler  :+1: :tada:

PR #1423, opened 2017-12-04
fix product layout in manufacturer action

uehler:
the manufacturer listing shows the wrong productbox layout
example: main category has layout "minimal" but the manufacturer listing shows the "basic" layout. If infinite scrolling is activated, the reloaded templates are the correct "minimal" ones

sets the productBoxLayout field in the $sCategoryContent array according to the selected layout of the main category

see 1.

https://issues.shopware.com/issues/SW-20234
I already did a pr for this but I accidentally deleted the old commit... (was https://github.com/shopware/shopware/pull/1306)

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Listing.php
+++ b/engine/Shopware/Controllers/Frontend/Listing.php
@@ -130,8 +130,9 @@ class Shopware_Controllers_Frontend_Listing extends Enlight_Controller_Action
         /** @var \Shopware\Bundle\StoreFrontBundle\Service\CustomSortingServiceInterface $service */
         $sortingService = $this->get('shopware_storefront.custom_sorting_service');
 
+        $mainCategory = $context->getShop()->getCategory();
+        $categoryId = $mainCategory->getId();
         if (!$this->Request()->getParam('sCategory')) {
-            $categoryId = $context->getShop()->getCategory()->getId();
 
             $this->Request()->setParam('sCategory', $categoryId);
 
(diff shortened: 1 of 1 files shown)


mitelg:
thanks for your contribution @uehler  :+1: :tada:

engine/Shopware/Controllers/Frontend/Listing.php
```diff
@@ -186,10 +187,13 @@ public function manufacturerAction()
         $this->View()->assign('manufacturer', $manufacturer);
         $this->View()->assign('ajaxCountUrlParams', [
             'sSupplier' => $manufacturerId,
-            'sCategory' => $context->getShop()->getCategory()->getId(),
+            'sCategory' => $categoryId,
         ]);
 
-        $this->View()->assign('sCategoryContent', $this->getSeoDataOfManufacturer($manufacturer));
+        $categoryContent = $this->getSeoDataOfManufacturer($manufacturer);
+        $categoryContent['productBoxLayout'] = $mainCategory->getProductBoxLayout();
```
shyim, line 194: This introduced a new bug, the config for productBoxLayout is now overwritten 😄, which is set in the getSeoDataOfManufacturer function

uehler: Oh, we both fixed the same issue...
Seems like your fix was declined for 5.2 but was eventually merged into 5.4, while I fixed it for 5.3

So my commit could be reverted

shyim: Fixed with https://github.com/shopware/shopware/commit/c1386f8dc48c65d3ba88b45ce6d9505a5cdc247d ❤️

PR #1426, opened 2017-12-07
Allow empty translates

hgiesenow:
To allow an empty string as fallback translation

Check if the default is not set (null) or an empty string

$namespace->get('my_snippet', '', true);

-

None

### 6. Checklist

- [-] I have written tests and verified that they fail without my change
- [-] I have read the contribution requirements and fulfil them.


hgiesenow:
> first of all, what do you want to achieve with your change?

To allow an empty string as fallback translation

> Also provide tests for your case.

I did not find any tests for this component at all. So my manual test case is `$namespace->get('my_snippet', '', true);` which will throw an error before and works afterwards.

PR #1431, opened 2017-12-10
simplified if-return-return statements

uehler:
there are some
if ($foo == $bar) { return true } return false;
statements in the code which can be simplified by using only one return statement

simplifies the statements to
return $foo == $bar

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Plugin/Bootstrap.php
+++ b/engine/Shopware/Components/Plugin/Bootstrap.php
@@ -145,11 +145,7 @@ abstract class Shopware_Components_Plugin_Bootstrap extends Enlight_Plugin_Boots
      */
     public function secureUninstall()
     {
-        if (empty($this->info->capabilities['secureUninstall']) || empty($this->info->capabilities['install'])) {
-            return false;
-        }
-
-        return true;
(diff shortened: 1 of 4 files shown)


engine/Shopware/Plugins/Default/Core/PaymentMethods/Components/SepaPaymentMethod.php
```diff
@@ -219,11 +219,7 @@ private function validateIBAN($value)
             $rest = intval($part) % 97;
         }
 
-        if ($rest != 1) {
-            return false;
-        }
-
-        return true;
+        return !($rest != 1);
```
teiling88, line 222: Is ```return $rest == 1``` not the same?

engine/Shopware/Components/Plugin/Bootstrap.php
```diff
@@ -145,11 +145,7 @@ public function uninstall()
      */
     public function secureUninstall()
     {
-        if (empty($this->info->capabilities['secureUninstall']) || empty($this->info->capabilities['install'])) {
-            return false;
-        }
-
-        return true;
+        return !(empty($this->info->capabilities['secureUninstall']) || empty($this->info->capabilities['install']));
```
janbuecker, line 148: IMHO, this is more difficult to read than before.

PR #1435, opened 2017-12-13
Fix RewriteMatcher for the same URLs in different subshops

aragon999:
The same issue as in this pull request: https://github.com/shopware/shopware/pull/1283
The problem occurs if there are the same SEO URLs from different subshops where the URL from the current shop is not the main URL (see testcase).

Changes the order of the `ORDER BY` statement.

See test.

### 6. Checklist


aragon999:
I added some more tests, but I don't understand what you exactly mean with the virtual urls, since they are not considered for the rewrite urls and the information is removed here from the `pathInfo`: https://github.com/shopware/shopware/blob/9a116f29f78c0005e031e34ffcce919b6a50e3a9/engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php#L465-L474

PR #1445, opened 2017-12-15
refactored sGetDownloads of sAdmin

dneustadt:

The amount of fetchable esd order positions is hard-capped at 500 orders (general orders, not even excluding those without esd positions). This would leave customers without any possibility to get access to their digital products once they ordered more than 500 times. A figure that could easily be reached within b2b environments or even b2c.
Furthermore the pagination is achieved by splicing an array that could possibly consist of thousands of rows, fetching 500 orders on each request. Performance bottleneck.

Query esd order positions only by joining orders on those positions. Work the pagination offset and limit into said query instead of using array_splice.
A little ugly but should be fully backwards compatible and is a lot more sensible. You get the idea.

Single User, more than 500 ESD orders, try getting your downloads for the first order.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -1173,81 +1173,94 @@ class sAdmin
     public function sGetDownloads($destinationPage = 1, $perPage = 10)
     {
         $userId = $this->session->offsetGet('sUserId');
-        $getOrders = $this->db->fetchAll(
-            "SELECT
-                id, ordernumber, invoice_amount, invoice_amount_net,
-                invoice_shipping, invoice_shipping_net,
-                DATE_FORMAT(ordertime, '%d.%m.%Y %H:%i') AS datum,
-                status, cleared, comment
-            FROM s_order WHERE userID = ? AND s_order.status >= 0
-            ORDER BY ordertime DESC LIMIT 500",
-            [$userId]
-        );
 
-        foreach ($getOrders as $orderKey => $orderValue) {
+        /** @var \Doctrine\DBAL\Query\QueryBuilder $query */
+        $query = Shopware()->Container()->get('dbal_connection')->createQueryBuilder();
+        $query->select([
+                'SQL_CALC_FOUND_ROWS orders.id as orderID',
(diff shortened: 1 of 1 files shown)


engine/Shopware/Core/sAdmin.php
```diff
@@ -1173,81 +1173,94 @@ public function sSaveRegisterSendConfirmation($email)
...
+                'SQL_CALC_FOUND_ROWS orders.id as orderID',
+                'orders.ordernumber as oOrdernumber',
+                'orders.invoice_amount',
+                'orders.invoice_amount_net',
+                'orders.invoice_shipping',
+                'orders.invoice_shipping_net',
+                'DATE_FORMAT(orders.ordertime, "%d.%m.%Y %H:%i") AS datum',
+                'orders.status as oStatus',
+                'orders.cleared',
+                'orders.comment',
```
hlohaus, line 1189: attributes would be nIce

dneustadt: Good idea. Now here's a question. For esd variants you could have either data from s_order_details_attributes and/or s_articles_esd_attributes. As far as I know the latter one isn't utilized anywhere in the frontend yet. This would be the ideal opportunity to pass them to the view. The data from s_order_details_attributes is probably more useful in the general orders listing anyway. What would you prefer?

hlohaus:
nice. but you don't squashed the commits

engine/Shopware/Core/sAdmin.php
hlohaus: you forgot s_order_attributes and yes, you can make aliases for it. orderArributes and esdAttributes.

dneustadt: I would consider it for another PR with some further refactoring. This one is more or less a hotfix.

hlohaus: okay :)

dneustadt:
@shopwareBot unrelated to this Pull Request: Is there any prediction to when the issue of having a hard-cap for the downloads listing will be solved?

dneustadt:
closed because I don't care anymore

PR #1448, opened 2017-12-19
[Labs] Fix XSS vulnerability on search result page

bcremer:

Userinput is injected into the DOM without escaping resulting in a non-persistent Cross-site scripting vulnerability.

## before patch

image

## after patch

image

Escapes the variable before passing into the twig function

Attack can be triggered via the `search` query parameter of the search:

http://172.18.0.3/search?search=%3Cscript%3Ealert(%22XSS%22)%3B%3C%2Fscript%3E

### 6. Checklist


FiveDigital:
Reminds my of #1095

janbuecker:
We haven't ported this change into labs right now. I guess we'll enable escaping by default and opt-out for parsing html.

soebbing:
Hi @bcremer, since Labs has moved to another repository, I'll close this PR here. Thanks for your contribution! 💙

PR #1450, opened 2017-12-20
Move jQuery header block back

hrombach:
At some point in the development of SW5.3 (specifically [this](https://github.com/shopware/shopware/commit/892860a071f5a5e9d68fdd9bae07607e68f224e4) commit) the block
"frontend_index_header_javascript_jquery" was moved below
"frontend_index_header_javascript_jquery_lib". This went undocumented
and breaks any script that relies on jQuery being present in that block
if asynchronous Javascript is disabled.

The block name suggests that jQuery should be available here, and it does not seem like moving it up was done intentionally.

The block positions are changed back to how they were before the commit linked above.

Put jQuery code in and disable asynchronous Javascript loading.

n/a

n/a

### 6. Checklist

- [ /] I have written tests and verified that they fail without my change
- [ /] This change has comments for package types, values, functions, and non-obvious lines of code

### DIFF
--- a/UPGRADE-5.3.md
+++ b/UPGRADE-5.3.md
@@ -30,6 +30,7 @@ This changelog references changes done in Shopware 5.3 patch versions.
 * Changed the `createValueListFacetResult` method of `ProductAttributeFacetHandler` to display translations of attribute values in facet list
 * Changed the snippet export to specify the fallback language
 * The tax id of the invoice address is now also used for deliveries to foreign countries if the invoice country and delivery country are marked to be tax free for companies and if the delivery address does not have a tax id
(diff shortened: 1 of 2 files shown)


Phil23:
Hey @hrombach,

thank you for your contribution. I have to confess that the naming of the block for the statistic call can be irritating, but the change was not made without a reason. We cannot simply switch this to the old behaviour, because it would implement another break. The better solution would be to rename the block for the statistic call and leave it where it is and then just create the wanted block as a new empty block below the jquery_lib block. For now you can simply get around this by appending the other block, so it should be no big deal. If you still think this is important you can create a new pull request with the changes I suggested targeting the next minor version 5.4.

Thx!

PR #1452, opened 2017-12-27
Fix clones of template manager with actived smarty security

windaishi:
This PR will fix many SmartySecurity problem with messages like 'directory [...] not allowed by security setting'

The problem is cloning instances of `Enlight_Template_Manager`. This is done at several places in the Shopware code. 

The security policy of the `Enlight_Template_Manager` instance holds a reference to the template manager instance. When cloning the template manager, the reference of the security_policy to the Smarty instance has to be updated to the new cloned Smarty instance.

Without doing this, every self::fetch() after a directory was added with self::addTemplateDir(), would lead to a SmartyException with message 'directory [...] not allowed by security setting'. This is because
the security_policy still holds a reference to the old Smarty instance that does not know this new directories as template sources.

After cloning instances of `Enlight_Template_Manager`, it clones the security_police object and updates the reference of the security_police to point to the correct, new, cloned instance of `Enlight_Template_Manager`.

The added UnitTest failes without the fix and is successfull with the fix.

None

None

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Template/Manager.php
+++ b/engine/Library/Enlight/Template/Manager.php
@@ -297,6 +297,8 @@ class Enlight_Template_Manager extends Smarty
      * SmartyException with message 'directory [...] not allowed by security setting'. This is because
      * the security_policy still holds a reference to the old Smarty instance that does not know this new directories
      * as template sources.
+     *
+     * The security_policy is also cloned so other instances of the Enlight_Template_Manager do not get affected.
      */
     public function __clone()
     {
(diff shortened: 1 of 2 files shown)


mitelg:
the build is failing: https://travis-ci.org/shopware/shopware/jobs/322279111#L1234

windaishi:
The tests should be fixed now.

soebbing:
Hi @windaishi! I just merged your PR, it will be part of the 5.4.1 release. Thank you for helping us make Shopware better! 💙

PR #1453, opened 2018-01-03
Fixed: Null values are not stored correctly

jahnch:
Values with empty string are not stored as null in database which causes that MySQL falls back to default value.

It fixes the problem, that empty strings are written to database which cause that the database falls back to default value instead of writing NULL on corresponding table column.

It corrects the wrong if statement which does not correctly check if value is changed to null in prior lines. The "||" ignores that value is null so value will be changed to an empty string in case of integer, float or decimal.

Add a custom attribute to an article as an integer which can be nullable. Change it inline in the backend in the article list(i.e.: activate or deactivate). Then in the database the value is not null. It has the value 0 because en empty string is send to database.

Ticket: SW-20826

-

### DIFF
--- a/engine/Shopware/Components/MultiEdit/Resource/Product/DqlHelper.php
+++ b/engine/Shopware/Components/MultiEdit/Resource/Product/DqlHelper.php
@@ -769,9 +769,11 @@ class DqlHelper
             $value = null;
         }
 
-        $type = $info['type'];
-        if ($value && $type == 'decimal' || $type == 'integer' || $type == 'float') {
-            $value = str_replace(',', '.', $value);
+        if($value !== null) {
+            $type = $info['type'];
+            if ($value && $type == 'decimal' || $type == 'integer' || $type == 'float') {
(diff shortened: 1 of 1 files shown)


jahnch:
Yes you are right. I removed value from if statement. Now it is correct.
The parentheses were wrong in the original shopware code. Thats why i added this update. I just forgot to remove value from if statement.

Pmuckel:
The PR's changes only influence the MultiEdit functionality of the article overview. Entered values in the custom attributes at the article detail page aren't affected. This results in two different behaviors for the custom attributes (possibility to create NULL values by editing via the MultiEdit funcionality but not via the article detail page). To achieve a consistent behavior, the AttributeData-Controller (action "saveData") inside the AttributeBundle has to be editet as well.

soebbing:
For the reason @Pmuckel gave we'll close this PR now. If the change is still important to you, feel free to create a new PR which incorporates the changes in all relevant places. Also some tests would be nice. :)

PR #1455, opened 2018-01-05
Fix wrong data hydration in AddressHydrator

barbieswimcrew:
Because $address->setAdditionalAddressLine1 will be used twice, once with the data from address_line1 and once with data from address_line2, the AddressGateway will retrieve invalid data. To put it in a nutshell the address_line1 data will be overridden by address_line2 so we need to use $address->setAdditionalAddressLine2

Just replace the second usage of $address->setAdditionalAddressLine1() by $address->setAdditionalAddressLine2() in the AddressHydrator's hydrate method.

When trying to retrieve a customers address by the AddressGateway it will return wrong data in additionalAddressLine1. Instead of the correct value there mistakenly appears the value of additionalAddressLine2.

-

-

### 6. Checklist


shyim:
thanks for your contribution @barbieswimcrew  +1 :tada: :blue_heart:

PR #1458, opened 2018-01-08
Add s_user_attributes to pdf-documents

mnaczenski:
Currently there are no s_user_attributes available on the pdf-documents

adds the user_attributes to pdf-documents

Create a new document/index.tpl in your own theme and try to use the user_attributes:
`
    {$User.attributes.attributeName}
`

https://github.com/shopware/shopware/pull/1457
I edited my PR to use the data_loader.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -349,18 +349,22 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
     }
 
     
-     /**
+    /**
      * Get user_attributes (s_user_attributes)
+     *
+     * @param int $userID
+     *
+     * @return array
+     * @throws Exception
      */
-
-    public function getUserAttributes($id)
+    public function getUserAttributes($userID)
     {
-        if (empty($id)) {
+        if (empty($userID)) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Document.php
```diff
@@ -348,6 +348,24 @@ public function getVoucher($id)
         return $getVoucher;
     }
 
+    
+     /**
+     * Get user_attributes (s_user_attributes)
+     */
+
```
aragon999: According to the PSR2 coding standards there should be no empty line :-)

engine/Shopware/Components/Document.php
```diff
@@ -348,6 +348,24 @@ public function getVoucher($id)
         return $getVoucher;
     }
 
+    
+     /**
+     * Get user_attributes (s_user_attributes)
```
aragon999: The PHPDoc is not correct (return value type is missing).

engine/Shopware/Components/Document.php
mnaczenski: Thank you for your hints. I am very unfamiliar with those things, but I updated my PR.

soebbing:
Thanks @mnaczenski! 💙

PR #1459, opened 2018-01-08
Add filter event to listing widget controller

hrombach:
Currently, it is impossible to manipulate the view parameters defined in `Shopware_Controllers_Widgets_Listing::fetchListing()` because the view is immediately fetched after assignment. 

Add a new event `Shopware_Controllers_Widgets_Listing_fetchListing_filterViewParams` to allow developers to manipulate the variables before they are assigned to the template. 

-/-

-/-

-/-

### 6. Checklist

### DIFF
--- a/UPGRADE-5.3.md
+++ b/UPGRADE-5.3.md
@@ -2,6 +2,14 @@
 
 This changelog references changes done in Shopware 5.3 patch versions.
 
+## 5.3.6
+
+[View all changes from v5.3.5...v5.3.6](https://github.com/shopware/shopware/compare/v5.3.5...v5.3.6)
+
+### Additions
+
+* Added new event `Shopware_Controllers_Widgets_Listing_fetchListing_filterViewParams` to `Shopware_Controller_Widgets_Listing::fetchListing()` in order to allow for filtering of template variables before the template is fetched by the function.
+
 ## 5.3.5
 
 [View all changes from v5.3.4...v5.3.5](https://github.com/shopware/shopware/compare/v5.3.4...v5.3.5)
--- a/engine/Shopware/Controllers/Widgets/Listing.php
+++ b/engine/Shopware/Controllers/Widgets/Listing.php
@@ -30,12 +30,12 @@ use Shopware\Components\Compatibility\LegacyStructConverter;
 use Shopware\Components\Routing\RouterInterface;
 
 /**
- * Shopware Listing Widgets
+ * Shopware Listing Widgets.
  */
 class Shopware_Controllers_Widgets_Listing extends Enlight_Controller_Action
 {
     /**
-     * product navigation as json string
+     * product navigation as json string.
      */
     public function productNavigationAction()
     {
@@ -102,7 +102,7 @@ class Shopware_Controllers_Widgets_Listing extends Enlight_Controller_Action
 
     /**
      * topseller action for getting topsellers
-     * by category with perPage filtering
+     * by category with perPage filtering.
      */
     public function topSellerAction()
     {
@@ -177,7 +177,7 @@ class Shopware_Controllers_Widgets_Listing extends Enlight_Controller_Action
     }
 
     /**
-     * tag cloud by category
+     * tag cloud by category.
      */
     public function tagCloudAction()
     {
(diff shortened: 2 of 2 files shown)


engine/Shopware/Controllers/Widgets/Listing.php
```diff
@@ -551,7 +558,7 @@ private function fetchPagination(ProductSearchResult $result)
         $this->View()->assign([
             'sPage' => $this->Request()->getParam('sPage'),
             'pages' => ceil($result->getTotalCount() / $sPerPage),
-            'baseUrl' => $this->Request()->getBaseUrl() . $this->Request()->getPathInfo(),
+            'baseUrl' => $this->Request()->getBaseUrl().$this->Request()->getPathInfo(),
```
aragon999: I think you are not supposed to remove the spaces here, see https://github.com/php-fig-rectified/fig-rectified-standards/blob/master/PSR-2-R-coding-style-guide-additions.md#strings-and-concatenation
The same is true for the other change in line 591

hrombach: These formatting changes are the result of the php-cs-fixer pre-commit hook. I did not make these myself.

hrombach: Therefore I am waiting for comment by a Shopware developer on this.

aragon999: Very strange, since according to the cs fixer configuration there should be one space: https://github.com/shopware/shopware/blob/5.3/.php_cs.dist#L45 and I never ran into this issue.

hrombach: That is indeed weird, I'll try re-running cs-fixer

hrombach: It looks like I'm missing something. I don't usually work with cs-fixer. The `.dist`-config is apparently not loaded, did I miss any steps before modifying my local shopware directory?

aragon999: Actually I don't think so, the config should be loaded automatically. I am not sure why this does not happen. 
Just for avoiding obvious problems, the file exists in your file system, right? :-)

hrombach: yeah, I have the shopware repo checked out and apart from my commit it is identical to upstream

mitelg: usually this happens, when you use a local installed version of the cs-fixer and not the one, which is in the vendor directory

aragon999: Okay, then I don't have any idea, the last time I made a commit the file was correctly loaded (if I remember correctly). Sorry. Maybe it is an OS issue?

hrombach: @mitelg I'm definitely calling the cs-fixer that comes shipped with shopware

hrombach:
@mitelg and @aragon999 I found my mistake, my vendor libraries were outdated, which is why php-cs-fixer was behaving wrong.

UPGRADE-5.3.md
```diff
@@ -2,6 +2,22 @@
...
 
+## 5.3.6
+
+[View all changes from v5.3.5...v5.3.6](https://github.com/shopware/shopware/compare/v5.3.5...v5.3.6)
+
+### Additions
+
+* Added new event `Shopware_Controllers_Widgets_Listing_fetchListing_filterViewParams` to `Shopware_Controller_Widgets_Listing::fetchListing()` in order to allow for filtering of template variables before the template is fetched by the function.
+
+### Changes
```
shopwareBot: Please remove these lines until there is content for them.

hrombach:
I removed the empty lines and decided to squash the commits while I was at it.

mitelg:
hey @hrombach 
please rebase your branch. Could you also add a new event in the `fetchPagination` method?

mitelg:
hey @hrombach 
we decide to close this PR in favor of #1519
the solution of @shyim provides more flexibility, as e.g. the whole controller context is available, etc

thanks anyway for your contribution and keep up the good work :blush:  :+1:

PR #1461, opened 2018-01-09
This event allows to change paper size of the document

waldicom:
This event allows to change paper size of the document. Now is only "A4" paper size available

This change creates an event to manipulate the paper size

Users can create an event listener and set their own paper size for the document

https://issues.shopware.com/issues/SW-20866

The new event should be documented in the README.MD

### 6. Checklist


soebbing:
Hi @waldicom, we're sorry you had to wait so long. But your change will be part of the next Shopware-release! 💙

PR #1462, opened 2018-01-10
Update for the FontFaceKit and using woff2

Hotte512:

Speeds up a bit... woff2 is smaler and used by modern Browsers like edge:) as of 04/06/2016

Updating the FontFaceKit from https://github.com/FontFaceKit/open-sans as of 2018-01-10
Adding .woff2 files in font.less
Adding .woff2 to the Expire Files in the .htaccess

Also doing some small changes in .htaccess, but i don't know how to separate the Pullrewquest ;-)

Slow... as webpagetest, dareboost or gtmetrix shows up fonts are big and slow.

### 6. Checklist


klarstil:
Hey @Hotte512 thanks for your contribution. Your PR adds source files directly in the repository. We're using `npm` to manage the third party dependencies for the frontend of shopware. The mapping from the `node_modules` into the `vendors` folder of the Responsive theme is handled by [this grunt tasks](https://github.com/shopware/shopware/blob/5.3/themes/Frontend/Responsive/Gruntfile.js#L106-L126)

We're using the following repository for the Open Sans Fontface: https://github.com/FontFaceKit/open-sans

It would be great when you would just update the version of the `open-sans-fontface` repository in the `package.json` and update the corresponding grunt tasks to reflect your changes. Otherwise your changes will be deleted when we're updating the frontend dependencies the next time.

Best regards,
klarstil

Hotte512:
as i'm more a user i do not exactly know what to do... (don't know github that well)
i'll tryout adding it to the package and make changes into the Gruntfile... will figure out i think.

But how can i testdrive that ? - SORRY - ;)

btw. do you thing adding eot makes sense?
I read that eot should be "first" but i don't think so...

klarstil:
Hey @Hotte512 
don't worry about it. Let me explain the process and how it works. We're using `npm`, the node package manager to download the third party dependencies for the Responsive theme. These dependencies will be placed in a folder named `node_modules` after running `npm install` in the Response theme directory.

We're using `Grunt`, a JavaScript task runner, to copy the files we want into the `frontend/public/vendors` directory.

The changes are looking good so far but I haven't tested them yet. Here's how you can test it:

```
cd themes/Frontend/Responsive
npm install
grunt
```

Best regards

Hotte512:
@klarstil could you tell me what would be the next step here? Just testreport or?

PR #1463, opened 2018-01-11
Fix ajax number search

mnaczenski:
When a product number matches the search term, the product search redirects directly to this product. You can activate this product matching with the setting "activateNumberSearch". The ajax search does no consider this setting and will show no products or unrelated products. This change will make the ajax-search behave like the product search.

Add a check for a direct match if the setting is activated

Activate number search and search for a valid article number in the ajax-search. 

https://issues.shopware.com/issues/SW-16189

Since I messed up with my branches, this is the same pull-request as https://github.com/shopware/shopware/pull/1424 that was closed by me.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/AjaxSearch.php
+++ b/engine/Shopware/Controllers/Frontend/AjaxSearch.php
@@ -61,11 +61,11 @@ class Shopware_Controllers_Frontend_AjaxSearch extends Enlight_Controller_Action
         $result = $this->get('shopware_search.product_search')->search($criteria, $context);
 
         //check if search-tearm is a valid product-number
-        if ((Shopware()->Config()->get('activateNumberSearch') == "1")) {
-        $directhit = Shopware()->Container()->get('shopware_storefront.list_product_service')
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Frontend/AjaxSearch.php
```diff
@@ -60,6 +60,25 @@ public function indexAction()
         /** @var ProductSearchResult $result */
         $result = $this->get('shopware_search.product_search')->search($criteria, $context);
 
+        //check if search-tearm is a valid product-number
+        if ((Shopware()->Config()->get('activateNumberSearch') == "1")) {
```
mitelg: try to avoid the using of the Shopwar() singleton.
as you are in a controller, you can use `$this->container->get('')` or `$this->get('')` instead

eg: `$this->get('config')->get()`

mnaczenski: thank you for your hint, i changed my PR according to your recommendation :+1:

PR #1464, opened 2018-01-11
Add unique document type key

niklasbuechner:

Plugin developers need to identify their own custom document types in order to provide the document-related functions of their plugin. Therefore, currently, they need to rely on users not changing the name of their document types. This is cumbersome especially to international customers who have to keep the German names in order to not break the plugin.

We propose to untangle the concern of identifying the document type from its user-visible name by introducing a unique identifier key for each document type - similar to the handling of payment methods.

This pull request adds a unique `key` column to `s_core_documents` and its corresponding model. The added key can be viewed and edited in the basic settings panel.

_Please note:_ Since the key column has a unique constraint, this is a breaking change. Once this is merged, all plugins which create a document type will have to provide a valid and unique key for their document types. Existing custom document types are migrated by this pull request.

The behaviour at hand is not a reproducible bug but rather an architectural flaw, from the point of view of a plugin developer, as described in 1.

How to test the pull request:
1. Open the backend.
2. Navigate to the "Basic settings".
3. Choose "PDF document creation" from the shop's settings.

None.

Corresponding documentation: http://community.shopware.com/PDF-Belegerstellung_detail_1086.html
* Update screenshots
* Add key field to the *Details* section

### 6. Checklist


engine/Shopware/Controllers/Backend/Config.php
```diff
@@ -637,6 +637,19 @@ public function saveValuesAction()
...
                     break;
+                case 'document':
+                    $exceptionMessage = $ex->getMessage();
+                    if (strpos($exceptionMessage, '1062 Duplicate entry') !== false
+                        &&
+                        strpos($exceptionMessage, 'for key \'key\'') !== false
+                    ) {
+                        $this->View()->assign([
+                            'success' => false,
+                            'message' => 'Please choose a different technical name since another document with the same technical name already exists.',
```
shyim: Should be a snippet or?

fixpunkt:
> Add the new column via migration file and make it optional as it can be possible that there are another existing types

The column cannot be optional (i.e. nullable) since it's `UNIQUE`. We'd have to remove the unique constraint, which would defeat the purpose of this PR. Also, existing document types get an auto-generated unique `key` as part of the migration, so this is not actually an issue:

https://github.com/shopware/shopware/pull/1464/files#diff-8eaeb4e19be1dfd0d75c70d5c793193fR37

engine/Shopware/Controllers/Backend/Config.php
fixpunkt: Fair point, although this message is never actually shown to the user because the ExtJS validations prevent this (and are internationalized).

niklasbuechner:
I applied the suggestions by removing the changes to `install/latest.sql`, adding the missing snippet and writing some tests.

> Add the new column via migration file and make it optional as it can be possible that there are another existing types

I fully agree with @fixpunkt here. The idea of the PR is to make the document types identifiable. Allowing multiple document types with the same key, negates a correct identification of a single document type.

Other existing document types will not break, since there is a migration mechanism in place.

mitelg:
@niklasbuechner please adjust the migration number. otherwise the builds fails

soebbing:
Hi @niklasbuechner, your PR was merged a while ago with https://github.com/shopware/shopware/commit/43cd973d10d and is already part of the Shopware Release 5.5 RC1 🎉

PR #1467, opened 2018-01-15
repair the Shopware_Form_Builder event according to the documentation

vitek499:

With the current splitting of events, it is not possible to capture events from the namespace Symfony\Component\Form\FormEvents from the plugins. You have `Shopware_Form_Builder`, but this event is based on `PRE_SET_DATA`. If programmer needs form event like `POST_SUBMIT`, he can't do that cleanly.

Something you need it, for example if constraints depending on another form field.

Probably repair bug, cause your documentation (here https://developers.shopware.com/developers-guide/address-management-guide/ , part `Create the Subscriber file`) said:

```
    /** @var \Symfony\Component\Form\FormBuilderInterface $builder */
    $builder = $event->getBuilder();
```

But thats not true, cause `$event->getBuilder()` returns `Symfony\Component\Form\FormInterface`

If you adding something into `Shopware\Bundle\AccountBundle\Form\Account\AddressFormType`, which is depending on `customer_type` filed.

If you accept the PR documentation will correspond to reality.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/FormBundle/Extension/EventExtension.php
+++ b/engine/Shopware/Bundle/FormBundle/Extension/EventExtension.php
@@ -51,7 +51,10 @@ class EventExtension extends AbstractTypeExtension
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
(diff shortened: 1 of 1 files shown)


vitek499:
Of course, thanks

PR #1468, opened 2018-01-15
Block save order when order has been modified by another user

windaishi:
If two people are working with the same order at the same time, the user who saved at last, may reset the changes made by the other user.

Blocks the update of the order if the order has been changed intermediately. (See GIF)
peek 2018-01-15 20-25

Open two orders at the same time in the backend, change something, save both. The second save of the changes will completely revert the first changes.

### 6. Checklist


samuelvogel:
Thank you for the quick answer!

This feature is especially useful for shops which also change orders via the API which applies to [Pickware](https://www.pickware.de) as well of course. When an order is picked while somebody also changes it's payment status in the backend for example, stock levels of the contained articles might become corrupted which is hard to figure out, tedious to fix and impossible to prevent from a user perspective (the user can not be "more careful" or do anything else to avoid the situation).

Therefore we really need this enhancement to land in the core and will send an enhanced PR shortly which also includes articles (same problem applies and `changetime` column is already present there) and maybe customers (probably not affected that often).

PR #1471, opened 2018-01-17
Update sendMailAction

maxout:
added the ability to add multiple receivers

you may want to inform more than one receiver about the Order state

loop through the param 'to' and perform for each receiver a $mail->addTo()

### DIFF
--- a/engine/Shopware/Controllers/Backend/Order.php
+++ b/engine/Shopware/Controllers/Backend/Order.php
@@ -815,7 +815,10 @@ class Shopware_Controllers_Backend_Order extends Shopware_Controllers_Backend_Ex
         }
 
         $mail->setFrom($this->Request()->getParam('fromMail', ''), $this->Request()->getParam('fromName', ''));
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Order.php
```diff
@@ -815,7 +815,10 @@ public function sendMailAction()
         }
 
         $mail->setFrom($this->Request()->getParam('fromMail', ''), $this->Request()->getParam('fromName', ''));
-        $mail->addTo($this->Request()->getParam('to', ''));
+        $addTo = explode(',',$this->Request()->getParam('to', ''));
+        foreach($addTo as $add){
```
Gugiman, line 819: I don't think you need to iterate the array elements to add every mailaddress separately.
addTo() is able to handle arrays and iterates them by itself.

`$addTo = explode(',',$this->Request()->getParam('to', ''));
$mail->addTo($addTo);`

PR #1473, opened 2018-01-19
Fix Uncaught TypeError when passing null to addFacet() method

barbieswimcrew:
Since Shopware allows to search (and open detail page) for articles that are mapped to inactive categories only an "Fatal error: Uncaught TypeError: Argument 1 passed to Shopware\Bundle\SearchBundle\ProductNumberSearchResult::addFacet() must implement interface Shopware\Bundle\SearchBundle\FacetResultInterface, null given, called in /var/www/shop/htdocs/engine/Shopware/Bundle/SearchBundleES/FacetHandler/CategoryFacetHandler.php on line 145 and defined in /var/www/shop/htdocs/engine/Shopware/Bundle/SearchBundle/ProductNumberSearchResult.php:87" occurs when trying to search for the article via Elasticsearch.

Wrapping the addFacet() methods call inside a instanceof FacetResultInterface condition will avoid that the fatal error occurs in case of $facet would be NULL.

First of all this error occurs only when using Elasticsearch. Create an article "foo" that is mapped to an inactive category. The article itself needs to be active. When trying to open "/search?sSearch=foo" the error described above should be thrown. This happens because the buildFacetResult() method of the CategoryTreeFacetResultBuilder returns both, an TreeFacetResult object or NULL. Therefor it's possible that a $facet of type NULL is going to be injected to the ProductNumberSearchResult's addFacet() method.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleES/FacetHandler/CategoryFacetHandler.php
+++ b/engine/Shopware/Bundle/SearchBundleES/FacetHandler/CategoryFacetHandler.php
@@ -31,6 +31,7 @@ use Shopware\Bundle\SearchBundle\Criteria;
 use Shopware\Bundle\SearchBundle\CriteriaPartInterface;
 use Shopware\Bundle\SearchBundle\Facet\CategoryFacet;
 use Shopware\Bundle\SearchBundle\FacetResult\CategoryTreeFacetResultBuilder;
+use Shopware\Bundle\SearchBundle\FacetResultInterface;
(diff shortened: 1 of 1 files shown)


FiveDigital:
Why is this not implemented yet?

barbieswimcrew:
Any update on that?

mitelg:
hey @barbieswimcrew 
please be patient :blush: 
As you have might seen, my colleagues are currently doing hard to work on every PR. I'm sure it will be your PR's turn soon :wink:

PR #1474, opened 2018-01-19
Prevent backend from overwriting orders, customers and articles that have been modified in the meantime

windaishi:
This is a new version of the closed PR #1468, which includes the requested improvements as well as tests.

When a backend user saves an entity like `Article`, `Order` or `Customer`, any changes that may have happened concurrently after the entity was opened in the backend up to the point in time where the user presses the "Save" button are overwritten.

Some users tend to open multiple orders, leave them open to do something else and then come back to edit them later. In these cases, write conflicts can and do happen. This may be the case when another backend user has modified the order in the meantime. However, something that we've seen quite often is that orders are modified over the API (e.g. when the order is being picked or shipped using Pickware Mobile), but these changes are then overwritten by another user saving a stale representation of the order from the backend. This causes problems which are very hard to debug and hard to recover without a lot of manual steps.

Because of this, getting this PR merged is a top priority for us (Pickware).

This is PR uses the known pattern of Optimistic Locking. See: https://en.wikipedia.org/wiki/Optimistic_concurrency_control

Blocks saving an order, article (item) or customer if the respective entity was modified by someone else after the entity was loaded by the backend. The user gets a message instructing them to close and re-open the entity in question and try again:

peek 2018-01-15 20-25

Open two orders at the same time in the backend, change something, save both. The second save of the changes will completely revert the first changes.

None.

None.

### 6. Checklist


cjuner:
@windaishi, can you please investigate the test failure? Your merge commit https://github.com/shopware/shopware/commit/cc36493753238ede9c48ca30466cba383b24377c fails, though it's mainline (5.4 branch) parent commit https://github.com/shopware/shopware/commit/9f0a7848aeac7556bf253208a291450c70491a42 passes the tests, see https://travis-ci.org/shopware/shopware/builds/343762069.

shyim:
thanks for your contribution @windaishi  :+1:  :tada: :blue_heart:.
Your PR has been merged with https://github.com/shopware/shopware/commit/369aa7f7afa9c8b7f2bead27bc8765fa0d61e53c

PR #1475, opened 2018-01-22
fix race condition on overdue cronjob list

ThomasOpiolka:
To fix a race condition in the cronjob execution logic. Longer running jobs and a short cron:run frequency might lead to running the same cronjob more than once.   

Before executing a cronjobs a added function checks, if the job is still overdue. If another thread already executed the cronjob it will skip and prevent double execution.

#### Setup:
System cronjob runs every minute
```
*/1 * * * * php /pathToShopware/bin/console sw:cron:run
```
3 or more Cronjobs setup in Shopware Backend.

#### Scenario
```
0:00 cron:run (Thread 1) will select all 3 jobs as overdue and keep them in memory.
0:01 Job1 runs in Thread 1 for 90 seconds.

1:00  cron:run (Thread 2) will select 2 jobs as overdue and keep them in memory.
1:01  Job2 runs in Thread 2 for 10 seconds.
1:11  Job3 runs in Thread 2 for 10 seconds.
**1:30 Job3 runs in Thread 1 for 10 seconds** (because it was selected earlier)
```
None?

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Components/Cron/Adapter/DBAL.php
+++ b/engine/Library/Enlight/Components/Cron/Adapter/DBAL.php
@@ -147,7 +147,12 @@ class Enlight_Components_Cron_Adapter_DBAL implements Enlight_Components_Cron_Ad
             $this->overdueJobsList = $this->getOverdueJobs();
         }
 
-        return array_pop($this->overdueJobsList);
+        while (($nextJob = array_pop($this->overdueJobsList)) !== null) {
+            if ($this->isJobStillOverdue($nextJob->getId())) {
+                return $nextJob;
+            }
+        }
+        return null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


engine/Library/Enlight/Components/Cron/Adapter/DBAL.php
```diff
@@ -147,7 +147,12 @@ public function getNextJob($force = false)
             $this->overdueJobsList = $this->getOverdueJobs();
         }
 
-        return array_pop($this->overdueJobsList);
+        while(($nextJob = array_pop($this->overdueJobsList)) !== null)
+        {
+            if($this->isJobStillOverdue($nextJob->getId()))
```
aragon999: This is not following the PSR-2 coding style, see here: http://www.php-fig.org/psr/psr-2/#5-control-structures

ThomasOpiolka:
I changed the coding style to psr-2 and squashed the commit.

ThomasOpiolka:
is there something I can do to make this happen anytime soon?

Wachert:
Mehr als sinnvolle Ergänzung, sollte umgehend aufgenommen werden.

PR #1476, opened 2018-01-23
Fix vouchers with conditions

mnaczenski:
Currently it is not possible to add a voucher with conditions to your basket, because $articles is always an array and this will lead to a wrong MySQL-query.

... 
AND ( ordernumber IN ()  OR s_articles.supplierID = '2' )
...

Because ordernumber IN () is an invalid MySQL-syntax

Change the checks from "is_array" to "!empty" in order to restore the functionality.

Restrict your voucher to a specific supplier and then add this voucher to your cart. You will see an error like this:

SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') OR s_articles.supplierID = '320' ) GROUP BY sessionID' at line 6 in engine/Library/Zend/Db/Statement/Pdo.php on line 224

https://issues.shopware.com/issues/SW-20965
https://forum.shopware.com/discussion/51237/gutscheineingabe-im-checkout-seit-5-3-5-fuehrt-zu-ups-ein-fehler-ist-aufgetreten#latest

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -237,12 +237,12 @@ class sBasket
      */
     public function sGetAmountRestrictedArticles($articles, $supplier)
     {
-        if (!is_array($articles) && empty($supplier)) {
+        if (empty($articles) && empty($supplier)) {
             return $this->sGetAmountArticles();
         }
 
         $extraConditions = [];
-        if (is_array($articles)) {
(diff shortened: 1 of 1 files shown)


mnaczenski:
the ticket is already resolved as part of 5.4.1

PR #1477, opened 2018-01-23
Fix applyDataAttributes() in jQuery plugin

buddhaCode:

The attr() method doesn't convert dash-case (html) attribute name to camel case (javascript) names like described in the dev docs.

Before this change a html attribute like ```data-myExample``` would be parsed into javascript variable ```myexample``` not in ```myExample``` as expected. An attribute ```data-my-example``` would be parsed into ```my-example```. Under this circumstances one can not use camel cased javascript variable names.

Documentaion is already right.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js
@@ -308,7 +308,7 @@
                     return;
                 }
 
-                attr = me.$el.attr('data-' + key);
+                attr = me.$el.data(key);
 
                 if (typeof attr === 'undefined') {
                     return true;


dneustadt:
+1 for the use of the .data() method. I would highly recommend to implement a fallback using the current concatenated attribute selector though, for backwards compatibility.

aragon999:
I am in favor with the idea of parsing the attributes in a "normal" way. 
There was another pull request implementing a fallback as well, which was closed due to inactivity, see here: https://github.com/shopware/shopware/pull/1382
You could start from there I guess.

Phil23:
Hey @buddhaCode,

this is just a code style preference. It would introduce a big major break for all jQurey plugins. So this has no benefit. It would be a much better solution to pass an optional parameter to the `applyDataAttributes()` method to trigger this, but also this is no guaranty that we will merge this anyway.

Thx!

PR #1478, opened 2018-01-24
Fixed note count on first click

shyim:
First click on mark product in note returns count 0, because response sUniqueID cookie cannot be found since 5.3.0 RC1.

SW-18393 added support for Cookies for paths and this broke that functionality (https://github.com/shopware/shopware/commit/e01d43cfc84b66f59c28540b3de3157f745996a2#diff-8ee74466675cdfac64b39296cf3020ccR127),

https://github.com/shopware/shopware/blob/5.3/engine/Shopware/Core/sBasket.php#L1330 does not contain "-/" at end of the key for the path.

Loop throught all cookies to find the right one without using the key.

Clear all Cookies, mark a product as note and you get 0 as count

### 6. Checklist

- [] This change has comments for package types, values, functions, and non-obvious lines of code

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -1325,10 +1325,10 @@ class sBasket
      */
     public function sCountNotes()
     {
-        $responseCookies = $this->front->Response()->getCookies();
+        $responseUniqueId = $this->getUniqueFromResponse();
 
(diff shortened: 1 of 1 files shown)


shyim:
Fixed with https://github.com/shopware/shopware/commit/40ec126c20d2b2e7177acfc3569aa1e62ffa9efd

PR #1481, opened 2018-01-25
Labs administration dump plugins command dump relative path

screeny05:
This PR changes the way the `bin/console administration:dump:plugins` command exports paths. The original behaviour was to dump paths absolute to the docker instance, but this doesn't work in case the user runs webpack from outside the docker machine. So in order to be able to use webpack more flexible we dump paths relative to the projectDir

In order to run webpack from outside docker

Add a new private function `getPathRelativeToProjectDir(string $absolute): string` to `AdministrationDumpPluginsCommand` in order to export universal relative paths.

```bash
./psh.phar docker:ssh
bin/console administration:dump:plugins
exit
npm run --prefix src/Administration/Resources/administration/ dev shopware.next:8088
```

-

-

### 6. Checklist


Phil23:
Thank you @screeny05,

I've made a ticket for this. For now the tickets of the labs project are not public.

Thx! :thumbsup:

Phil23:
Thank you  @screeny05,

was merged with 4e6ff07c639d4a4490dbd6cdac1e03cfa38c8eb2.

👍

PR #1484, opened 2018-01-25
Added automatic controller registration for namespaced controller

shyim:
Make controller registration easy like without namespaces

Searches also in Controller folder (Symfony Convention) for Controller and use them as namespaced controller.
I also added a new class Shopware\Components\Controller which basic extends from Enlight_Controller_Action it looks better in namespaced way :+1: 

Create Plugin Test and create a controller in Test/Controller/Frontend/Test.php with following content
```php
<?php

namespace Test\Controller\Frontend;

use Shopware\Components\Controller;

class Test extends Controller
{
    public function indexAction()
    {
        die('Ich mag bananen');
    }
}
```
and request page /test after plugin activation :heart: 

Document Controller registration also possible with namespaces in Controller folder instead Controllers

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Controller.php
+++ b/engine/Shopware/Components/Controller.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Components;
+
+class Controller extends \Enlight_Controller_Action
+{
+}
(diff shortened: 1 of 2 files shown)


engine/Shopware/Components/Controller.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Components;
+
+class Controller extends \Enlight_Controller_Action
```
oktupol, line 27: I'm not a big fan of that class. I know that `Enlight_Controller_Action` is ugly, but by defining another class that serves just as alias, we're just building ourselves another barrier when it comes to compatibility in later versions. This class assumes that a potential replacement for `Enlight_Controller_Action` in a later version of Shopware is located at `Shopware\Components\Controller`, which might not be the case.

shyim: Hey,
yea i had the same consideration, but I left it in the pull request to force a conversation :)

I had a another idea to define a simple alias using class_alias to provide the same possibility. 

But how @shopware thinks about that? Add this possibility or not?

fionera: You could write an ControllerInterface wich defines the Basic Functions and then implement these in the Controller Class which forward the calls to the old Enlight Class

shyim:
I will entire rewrite this PR so i close this.

Current state is registering controllers using DI with a new tag `shopware.controller`

Example Usage would be

```xml
<?xml version="1.0" ?>
<container xmlns="http://symfony.com/schema/dic/services"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">

    <services>
        <service id="shyim_controller.test" class="ShyimController\Test">
            <argument type="service" id="dbal_connection"/>
            <tag name="shopware.controller" module="frontend" controller="test"/>
        </service>
    </services>
</container>
```

```php
<?php

namespace ShyimController;

use Doctrine\DBAL\Connection;

class Test extends \Enlight_Controller_Action
{
    /**
     * @var Connection
     */
    private $connection;

    public function __construct(Connection $connection)
    {
        $this->connection = $connection;
    }

    public function indexAction()
    {
        var_dump($this->connection->fetchAll('SELECT * FROM s_articles'));
        die("YAY");
    }
}
```

teiling88:
I would love to see a test there ;-)

PR #1495, opened 2018-01-31
Upgrade mPDF to 7.0.3, install it via composer and unbundle mPDF 6

cjuner:
This is an evolution of https://github.com/shopware/shopware/pull/1252, adapted to `5.4` and upgrading mPDF to an even newer version. This version also removes mPDF's upstream fonts to reduce the release bundle size to address concerns raised in the [previous PR](https://github.com/shopware/shopware/pull/1252#issuecomment-361294477). Thanks to @samuelvogel for the initial version. This pull request's description is mostly identitical to the previous one.

mPDF is currently directly bundled into the Shopware repository in version 6.0.0, which is almost 3 years old (judging from the [tag on GitHub](https://github.com/mpdf/mpdf/releases/tag/v6.0.0)). It was last updated with Shopware 5.1.0 (a4ea89a) and is shipped with custom fonts and a minor config change (see 9f921df). Some changes where made afterwards to backport specific fixes for PHP 7 compatibility (see b560f11 and df075d6).

Unfortunately, there still remains a problem with PHP 7 and SVG usage on documents (a lot of people use a vector graphic to achieve sharply printed logos on their PDFs) when the server is configured to show `E_DEPRECATED` errors: `PHP Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; SVG has a deprecated constructor in /engine/Library/Mpdf/classes/svg.php on line 33`

Also, since mPDF version `6.1.0`, errors are thrown as exceptions, which leads to better usability in the backend when there is not enough disk space or file permissions are wrong (see https://github.com/mpdf/mpdf/commit/2d02c208b8d7b15c4208517375b9a10b0ea7cefa). The improved output for order documents is pending PR #1250.

Update mPDF to version 7.0.3 (the current stable release), which has much better compatibility. Also it includes mPDF via composer to clean up the source tree and ease future updates.

We checked carefully for this to not introduce any backwards incompatibility. We first diffed a4ea89a against a original mPDF 6.0.0 zip from http://www.mpdf1.com to make sure there are no hidden changes. We then reapplied all configuration options in a composer compatible way. The exact same 5 fonts are available (all fonts shipped with the mPDF distribution are disabled, as before) and even the backupSIPFont option is still deactivated (even though probably nobody knows what it actually does nowadays).

We also did something very similar for FPDF with #1009.

The custom settings and font configurations from previous Shopware version are retained for mPDF 7.

Just create an order document with a regular Shopware 5.4 instance, save it and then create one with the PR applied. They will both look 100 % the same, pixel perfect. The same applies  to SEPA debit authorization documents. Both types of document were verified with visual inspection using `diffpdf`. Fortunately, mPDF is not under a lot of development, so this was the expected result.

n/a

None

### 6. Checklist


mitelg:
merged with 38530e0b6ee on 5.5 branch

cjuner:
Please note that by squashing, the commit message has become non-reflective of what the commits actually do: https://github.com/shopware/shopware/commit/93a5aac14d8d137b78000a392669b1154d89c2b9 claims "For now, it is installed in parallel to mPDF 6, which is located in `engine/Library/Mpdf/`.", which was the case for one of the original commits, but not the (now squashed) patch set.

mitelg:
@cjuner you are right. 

@soebbing will you consider this? :smile:

PR #1496, opened 2018-01-31
Handle mail errors during order status changes to avoid stock-errors in Shopware ERP powered by Pickware

windaishi:
Batch processing of orders is a frequently used feature in Shopware. However, it has a serious problem:
When sending mail fails for just one order in the batch process, the batch processing gets canceled and the user doesn't even gets informed about what happened. Also, the AfterHooks of the `batchProcessAction` in the `Shopware_Controllers_Backend_Order` do not get called anymore. Together with Shopware ERP (Powered by Pickware) this leads to serious stock errors that are hard to identify.

Common errors which cause the mailing to fails are:
* Syntax errors in mail templates
* Errors in the mailer configuration

This adds a `try-catch` around the code block that sends the email in the `batchProcessAction`. With this, the batch processing does not stop anymore when a mail error appears during batch processing. Also, errors that appear during the batch processing now are passed to the backend and the user gets shown an error message.

image

Corrupt a mail template, for example by adding a '{if}' tag without a closing '{/if}' tag.
Use batch processing and configure like it would send an email with the corrupted mail templated.
The backend then looks like the following:
image

### 6. Checklist


cjuner:
@windaishi, please rebase to Shopware 5.5, or delegate appropriately.

windaishi:
The branch has been rebased to 5.5

windaishi:
I added a commit that addesses the same issue at another place:

When changing the status of an order to a status with an email template and this email template is corrupted, saving the order fails hard with an exception. The order gets saved correctly but no `after`-hooks are executed. Together with Shopware ERP (Powered by Pickware) this leads to serious stock errors that are hard to identify and recover from.

It improves the `Order/save` action so it catches any exceptions that occur during email generation and reports the problem to the backend user.

1. Corrupt a status mail template, for example by adding an `{if}` tag without a corresponding closing `{/if}` tag.
2. Change the status of an order that would send this mail template on status change.

Result after applying this changeset:
image

shyim:
Your change has been merged :)

PR #1497, opened 2018-02-01
Fixed combobox store loading with value null

shyim:
After plugin installation with a remote combobox, the store will be never loaded when the value is null

Load the combobox store also when the value is null

Create a new plugin use the example from wiki https://developers.shopware.com/developers-guide/plugin-configuration/#selectionfield-/-remote-combobox and see the combobox is empty

### 6. Checklist


shyim:
Has been meanwhile fixed in 5.4

PR #1498, opened 2018-02-01
Fix emotions not loading in listings with GET parameter p=1

Crease29:

If infinite scrolling is enabled, the jQuery plugin will instantly set the GET parameter `p=1` if the user scrolls.
If the user then reloads the page, there is no emotion anymore.

This fix changes the condition that disables emotions in listings if a page parameter is set in the request.

1. Enable infinite scrolling in your theme
2. Assign an emotion to a category that has more than 1 page in its frontend listing
    - Make sure the checkbox to show products under the emotion is **checked**!
3. Open the category in frontend
    - The emotion should be visible.
4. Scroll down a little on the page: The parameter `p=1` will be appended to your URL.
5. Reload the page now.
    - The emotion won't be visible.

:heavy_minus_sign: 

:heavy_minus_sign: 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Listing.php
+++ b/engine/Shopware/Controllers/Frontend/Listing.php
@@ -222,7 +222,7 @@ class Shopware_Controllers_Frontend_Listing extends Enlight_Controller_Action
      */
     protected function getEmotionConfiguration($categoryId, $withStreams = false, $streamId = null)
     {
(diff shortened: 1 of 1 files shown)


Crease29:
Sorry, I will close this PR, because products won't be visible on page 1 when products aren't shown below the emotion. I will work on a javascript solution and open a new PR.

PR #1501, opened 2018-02-01
Fix execution model of `replace` hooks

svenmuennich:
**This PR is an updated version of #1179 and fixes hooking of `protected` methods.**

Currently the execution model of `replace` hooks is broken, once more than one `replace` hook exists for the same method. The problem is that all `replace` hooks are executed sequentially and each of these hooks has the opportunity to call `executeParent()`. Hence, if all `replace` hooks do call `executeParent()`, the original (parent) implementation is executed more than once. This basically breaks `replace` hooks on methods that have side effects, because these side effects are applied for every call of `executeParent()`. As a result, it is currently very risky to have more than one `replace` hook on the same method. In fact, we at Pickware refrain from using `replace` hooks altogether because of that behaviour!

Example:

```php
class HookTarget
{
    public function fooBar()
    {
        echo "HookTarget: fooBar!\n";
    }
}

class FooSubscriber implements \Enlight\Event\SubscriberInterface
{
    //...

    /**
     * First hook
     */
    public function onHookTargetFooBar(\Enlight_Hook_HookArgs $args)
    {
        echo "FooSubscriber: before executeParent()\n";
        $args->getSubject()->executeParent($args->getMethod(), $args->getArgs());
        echo "FooSubscriber: after executeParent()\n";
    }
}

class BarSubscriber implements \Enlight\Event\SubscriberInterface
{
    //...

    /**
     * Second hook
     */
    public function onHookTargetFooBar(\Enlight_Hook_HookArgs $args)
    {
        echo "BarSubscriber: before executeParent()\n";
        $args->getSubject()->executeParent($args->getMethod(), $args->getArgs());
        echo "BarSubscriber: after executeParent()\n";
    }
}
```

Output:

```
FooSubscriber: before executeParent()
HookTarget: fooBar!
FooSubscriber: after executeParent()
BarSubscriber: before executeParent()
HookTarget: fooBar!
BarSubscriber: after executeParent()
```

The desired execution model of `replace` hooks prevents multiple calls of the original implementation, if all hooks call `executeParent()` only once. That said, the `replace` hooks should be parents of each other, based on the position they are sorted by in the event manager. This execution model is very similar to the one used by _ExtJS_ and hence should be more familiar and most importantly predictable to shopware developers.

This PR refactors the whole hook manager as well as changes the generated proxy classes (which is easy now thanks to #1174) to add an execution context, which executes the registered `replace` hooks as a chain in which each call to `executeParent()` calls the respective next hook, if available. Only the last (highest position) hook’s call to `executeParent()` calls the original implementation of the hooked method. Finally the respective return values can be passed down the chain again. That is, the `replace` hook with the lowest position is able to ultimately override the return value, before it is passed to any `after` hooks:

```php
class HookTarget
{
    public function fooBar()
    {
        echo "HookTarget: fooBar!\n";
    }
}

class FooSubscriber implements \Enlight\Event\SubscriberInterface
{
    //...

    /**
     * First hook
     */
    public function onHookTargetFooBar(\Enlight_Hook_HookArgs $args)
    {
        echo "FooSubscriber: before executeParent()\n";
        $args->getSubject()->executeParent($args->getMethod(), $args->getArgs());
        echo "FooSubscriber: after executeParent()\n";
    }
}

class BarSubscriber implements \Enlight\Event\SubscriberInterface
{
    //...

    /**
     * Second hook
     */
    public function onHookTargetFooBar(\Enlight_Hook_HookArgs $args)
    {
        echo "BarSubscriber: before executeParent()\n";
        $args->getSubject()->executeParent($args->getMethod(), $args->getArgs());
        echo "BarSubscriber: after executeParent()\n";
    }
}
```

Output:

```
FooSubscriber: before executeParent()
BarSubscriber: before executeParent()
HookTarget: fooBar!
BarSubscriber: after executeParent()
FooSubscriber: after executeParent()
```

Of course it is still possible to execute the original implementation more than once, by calling `executeParent()` in the same hook repeatedly. This also works for recursive methods, which would result in a quite complex call tree. Example with calling `executeParent()` twice in a recursive method:

```
  recMethod()
       |
    HOOK_R0 ----------------------------------
       |                                     |
executeParent()                       executeParent()
       |                                     |
   METHOD_R0                             METHOD_R0
       |                                     |
  recMethod()                           recMethod()
       |                                     |
    HOOK_R1 ---------------               HOOK_R1 ---------------
       |                  |                  |                  |
executeParent()    executeParent()    executeParent()    executeParent()
       |                  |                  |                  |
   METHOD_R1          METHOD_R1          METHOD_R1          METHOD_R1
```

On a side note, this fix kind of makes `before` and `after` hooks obsolete, since it is now possible to just subscribe `replace` hooks, perform custom steps _before_ the original implementation, call `executeParent()` and then perform some more steps _after_ that call (just like in _ExtJS_).

See 1. and 2.

n/a

The documentation on hooks should reflect

a) the execution model using a minimal example and
b) that `before` and `after` hooks are obsolete.

### 6. Checklist

### DIFF
--- a/UPGRADE-5.4.md
+++ b/UPGRADE-5.4.md
@@ -84,6 +84,7 @@ This changelog references changes done in Shopware 5.4 patch versions.
         - Frontend/Compare.php
         - Frontend/Note.php
         - Widgets/Listing.php
+* Changed the execution model of `replace` hooks to prevent multiple calls of the hooked method, if more than one `replace` hook on the same method exists and all of them call `executeParent()` once
 
 ### Removals
 
--- a/engine/Library/Enlight/Hook/HookExecutionContext.php
+++ b/engine/Library/Enlight/Hook/HookExecutionContext.php
@@ -1,0 +1,178 @@
+<?php
+/**
+ * Enlight
+ *
+ * LICENSE
+ *
+ * This source file is subject to the new BSD license that is bundled
+ * with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://enlight.de/license
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@shopware.de so we can send you a copy immediately.
+ *
+ * @category   Enlight
+ * @package    Enlight_Hook
+ * @copyright  Copyright (c) 2017, shopware AG (http://www.shopware.de)
+ * @license    http://enlight.de/license     New BSD License
+ * @version    $Id$
+ * @author     Heiner Lohaus
+ * @author     $Author$
+ */
+
+/**
+ * The Enlight_Hook_HookExecutionContext represents a single execution of a hook.
+ *
+ * In order to execute a proxy method and all its 'before', 'replace' and 'after' hooks, a new context must be created
+ * and executed.
+ *
+ * @category   Enlight
+ * @package    Enlight_Hook
+ * @copyright  Copyright (c) 2017, shopware AG (http://www.shopware.de)
+ * @license    http://enlight.de/license     New BSD License
+ */
+class Enlight_Hook_HookExecutionContext
+{
+    /**
+     * @var Enlight_Hook_HookManager
+     */
+    protected $hookManager;
+
+    /**
+     * @var Enlight_Hook_HookArgs
+     */
+    protected $args;
+
+    /**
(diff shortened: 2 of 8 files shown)


mitelg:
merged with c4b22d8bed9 on 5.5 branch

PR #1504, opened 2018-02-01
Removed deprecated label from widgets

shyim:
Its deprecated and snippets should be used to make it translateable

Removes setLabel, getLabel from Widget model and the database column

Document how to use snippets to set a label

### 6. Checklist


shyim:
Has been already fixed in 5.5 Branch

PR #1506, opened 2018-02-04
misc - fix issue with typ text2 and php7.1 and above

teiling88:
Without this change, you receive only the first character from input fields with typ text2. For more details see https://3v4l.org/Zduk7

Fix typing issue

### 6. Checklist


mitelg:
thanks @teiling88 :+1: :tada:

PR #1507, opened 2018-02-05
Publish app server ports in docker environment

FiveDigital:
The webserver of the docker container can't be reached at `localhost:8088` under MacOS because the ports are not published. Thanks to @Crease29 for helping me with the docker config!

I changed the `docker-compose.yml` to publish the ports of the app server. I'm not sure if it's possible to access the variables set in `.psh.yaml.dist` so that we can make this configurable via a new variable like `SW_HOST_PORT`.

Run the docker container with or without my change. Shopware can only be reached with my change oder MacOS.

No

No

### 6. Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,8 @@ services:
     volumes:
       - .:/app
       - ~/.composer/cache:/.composer/cache
+    ports:
+      - "8088:80"
 
   app_mysql:
     build: dev-ops/docker/containers/mysql


janbuecker:
Port mappings are removed from the base `docker-compose.yml` for a reason. Our CI platform should not bind ports as it may comes to concurrent builds and may break when the ports can not be mapped because they are already in use. 

For local development, the ports are defined in `dev-ops/docker/docker-compose.override.yml` which will be copied to root when running `./psh.phar docker:start`. So the container will be available on `localhost:18000` (mysql at `localhost:13306`, elasticsearch at `localhost:19200`). See [docker-compose.override.yml](https://github.com/shopware/shopware/blob/5.4/dev-ops/docker/docker-compose.override.yml) for reference.

This change will not be merged, but feel free to comment on this PR in case it does not work for you.

PR #1510, opened 2018-02-05
Added autocomplete to Enlight_Controller_Plugins

shyim:
Everytime i have to look in docs whats the name of the Plugin and which methodes, because there is no autocomplete.

Adds method for the Controller Plugins and improves autocomplete 

Type Shopware()->Front()->Plugins()-> and you see all Plugins :scream: 

### 6. Checklist


soebbing:
Thank you, @shyim 😆

PR #1511, opened 2018-02-05
Added more blocks to frontend/account/profile.tpl

shyim:
Try to remove the email change block without copying the entire file :thinking: 

Adds more blocks

### 6. Checklist


shyim:
Updated version of the PR #1715

PR #1513, opened 2018-02-05
Make subrequests easier useable for plugins

shyim:
Moved code from action tag plugin to Enlight_Controller_Front::subRequest

Makes easier subrequesting pages, without copying 50 lines logic from action tag. The action tag is now small and cute :smiley: 

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Controller/Front.php
+++ b/engine/Library/Enlight/Controller/Front.php
@@ -466,4 +466,77 @@ class Enlight_Controller_Front extends Enlight_Class implements Enlight_Hook
     {
         return $this->invokeParams;
     }
+
+    /**
+     * @param array $params
+     * @return Enlight_Controller_Response_Response|string
+     * @throws Enlight_Controller_Exception
+     * @throws Enlight_Exception
+     * @throws Exception
+     */
+    public function subRequest(array $params)
+    {
+        $dispatcher = clone $this->Dispatcher();
+
+        $modules = $dispatcher->getControllerDirectory();
+        if (empty($modules)) {
+            throw new Exception('Action helper depends on valid front controller instance');
+        }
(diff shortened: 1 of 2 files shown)


engine/Library/Enlight/Controller/Front.php
```diff
@@ -466,4 +466,77 @@ public function getParams()
...
+     * @throws Enlight_Exception
+     * @throws Exception
+     */
+    public function subRequest(array $params)
+    {
+        $dispatcher = clone $this->Dispatcher();
+
+        $modules = $dispatcher->getControllerDirectory();
+        if (empty($modules)) {
+            throw new Exception('Action helper depends on valid front controller instance');
```
mitelg, line 483: both exceptions still refer to the action smarty plugin

engine/Library/Enlight/Template/Plugins/function.action.php
```diff
@@ -36,75 +36,11 @@
  */
 function smarty_function_action($params, Enlight_Template_Default $template)
 {
-    /** @var $front Enlight_Controller_Front */
-    $front = Shopware()->Front();
-    $dispatcher = clone $front->Dispatcher();
+    $response = Shopware()->Front()->subRequest($params);
```
dneustadt, line 40: Maybe add a filter event while you're at it? Can't have enough of those.
`
Shopware()->Events()->filter(
    'Enlight_Template_Plugins_ActionFunction_FilterResponse',
    $response,
    ['params' => $params]
);
`

hlohaus: Or a service with an interface.

shyim:
I will rework this Pull Request and open it internal

PR #1521, opened 2018-02-10
Regarding Shopware AboCommerce/Subscription plugin installation issue

Novalnet-Technic:


fionera:
I think you should think about what you are trying to do and what a PullRequest is

PR #1523, opened 2018-02-13
Parse forms mails with smarty

shyim:
Maybe 5.5 Branch?

You cannot do something with the submited data like if conditions etc

Parses mail template as smarty 

That the field is now smarty and the Compatibility is not the best, so after update it should be updated 😈 

### 6. Checklist


soebbing:
Hi @shyim, thanks for your PR!

I'm sorry i have to close it, but we think it has limited benefit while adding a bit of a security risk and possible confusions for shop owners and other developers (because custom modifiers won't work here).

Sorry I don't have any better news for you! :(

PR #1524, opened 2018-02-14
fix sorting of customer overview

mnaczenski:
Since shopware 5.3 the newest customers won't be listed as first entries. The sorting was changed to customernumber. This is a problem since the customernumber is optional and can be deactivated in the basic settings.

Change the sorting to firstlogin (= creation date)

Add a few customers and change one customernumber to 99999, this entry will be listed as first customer in the grid

https://issues.shopware.com/issues/SW-19907

### 6. Checklist


soebbing:
Hi @mnaczenski, thank you for your PR! We'll close it as agreed!

PR #1526, opened 2018-02-14
Use the short array syntax for the generated config.php during installation

jochenmanz:
The old php array syntax is used for the generated config.php during the installation process. This PR changes this behaviour

The ConfigWriter uses the new and short array syntax for the config instead.

- install shopware without this change
- save the config.php
- install shopware with this change
- compare config.php

### 6. Checklist

### DIFF
--- a/recovery/install/src/Service/ConfigWriter.php
+++ b/recovery/install/src/Service/ConfigWriter.php
@@ -24,8 +24,8 @@
 
 namespace Shopware\Recovery\Install\Service;
 
+use RuntimeException;
 use Shopware\Recovery\Install\Struct\DatabaseConnectionInformation;
-use Zend\Code\Generator\ValueGenerator;
 
 /**
  * @category  Shopware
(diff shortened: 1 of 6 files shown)


dneustadt:
Does this really justify another dependency, especially when needed exactly once?

mitelg:
@dneustadt this is the `composer.json` of the installer. so maybe it is ok, because there is not that much code at all :wink:

dneustadt:
@mitelg well according to githubs api the zend-code repo weighs in at 761348 bytes of PHP (https://api.github.com/repos/zendframework/zend-code/languages)
That's a whole lot of code in service to what arguably is code quality at best, even more in a single configuration file. I'd rather see a different file format for the config settings than carry a rather large package for the little benefit it provides.

mitelg:
@dneustadt oh, I didn't know that :smile:  :see_no_evil: 
then you are totally right

bcremer:
Please keep in mind that `zend-code` gets already pulled as an transitive dependency anyway:

```
~/www/shopware (5.4) 
$ composer install --no-dev --ignore-platform-reqs
Loading composer repositories with package information
Installing dependencies from lock file
Nothing to install or update
Generating optimized autoload files
Warning: Ambiguous class resolution, "Doctrine\ORM\Persisters\Entity\BasicEntityPersister" was found in both "$baseDir . '/engine/Library/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php" and "/home/bcremer/www/shopware/vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php", the first will be used.
Warning: Ambiguous class resolution, "Doctrine\Common\Proxy\AbstractProxyFactory" was found in both "$baseDir . '/engine/Library/Doctrine/Common/Proxy/AbstractProxyFactory.php" and "/home/bcremer/www/shopware/vendor/doctrine/common/lib/Doctrine/Common/Proxy/AbstractProxyFactory.php", the first will be used.
~/www/shopware (5.4) 
$ composer show | grep zend-code                  
zendframework/zend-code         2.6.3     provides facilities to generate arbitrary code using an object oriented interface
```

Anyhow I'm also not seeing the benefit of using zend-code just for getting rid of old style arrays.

jochenmanz:
I've reworked the whole thing now as you have a valid point. I added the socket to the build vars and replaced the zend-code approach with a str_replace approach leveraging the .dist file.

recovery/install/src/Service/ConfigWriter.php
```diff
@@ -49,39 +50,31 @@ public function __construct($configPath)
...
         ];
 
-        foreach ($info as $key => $parameter) {
-            if ($key == 'port' && empty($parameter)) {
-                continue;
-            }
-            if ($key == 'socket' && empty($parameter)) {
-                continue;
-            }
+        $template = file_get_contents($this->configPath . '.dist');
```
shyim, line 66: The release.zip does not contain a config.php.dist, this would fail or? @shopwareBot

PR #1527, opened 2018-02-14
Change Content-Type header for statistics XHR

FiveDigital:
Because you will get the following error in Chrome: `Refused to execute script from '/widgets/index/refreshStatistic?requestPage=/&requestController=index' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.`

After a deeper investigation this issued has already been fixed in cb9355113550988176a48ab4290757a2e866c01c but got reverted in 8712c61e418ceab7ede1f7a19279784314f9e62c. Perhaps we should discuss here if the problem "Removed content header leading to endless request" could be fixed in an other way.

This PR adds the correct Content-Type header for the `refreshStatistic` action in the `Index` controller.

Activate Shopware statistics and browse a random page with Chrome having strict MIME type checking enabled.

I have found this ticket: https://issues.shopware.com/issues/SW-18506 The ticket status is "solved" but the problem remains.

No.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Widgets/Index.php
+++ b/engine/Shopware/Controllers/Widgets/Index.php
@@ -34,6 +34,7 @@ class Shopware_Controllers_Widgets_Index extends Enlight_Controller_Action
     {
         if ($this->Request()->getActionName() === 'refreshStatistic') {
             $this->Front()->Plugins()->ViewRenderer()->setNoRender();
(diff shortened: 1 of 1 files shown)


soebbing:
The problem was revisited in https://issues.shopware.com/issues/SW-20951 and was fixed in https://github.com/shopware/shopware/commit/f1d6c5d4a96b092db5172918807ed578d0d16506

PR #1528, opened 2018-02-15
Fix illogical count implementation in Enlight_Collection_ArrayCollection

jmatthiesen81:
This change is necessary, because the method returns a not expected value.

The implemented count method return in every case 0.

Add some elements to the collection and call the count method.

-

-

### 6. Checklist


soebbing:
Thank you for your contribution, @jmatthiesen81! 💙

PR #1531, opened 2018-02-15
misc - create a native init without ant for psh

teiling88:
### 6. Checklist


dev-ops/common/actions/init.sh
```diff
@@ -1,5 +1,26 @@
...
+php bin/console sw:theme:initialize
+
+php bin/console sw:admin:create --name="Demo user" --email="demo@example.com" --username="demo" --password="demo" --locale="de_DE"
+
+date +"%Y%m%d%H%M" | tee recovery/install/data/install.lock
+
+php bin/console sw:firstrunwizard:disable
+
+curl -k -L -o test_images.zip "http://releases.s3.shopware.com/test_images_since_5.1.zip"
+unzip test_images.zip
```
DanielBoettner, line 20: I suggest adding "-o" here.
If the script fails for some reason, or somebody wants to re-run the init script it would not fail because because of a missing input.

dev-ops/common/actions/.init_composer.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+php composer.phar install --no-interaction --optimize-autoloader --no-suggest
```
DanielBoettner, line 3: I suggest to change `composer.phar` to a more flexible implementation. A parameter would allow the user to use `/usr/bin/composer` or other paths to use their global composer.

dev-ops/common/actions/init.sh
```diff
@@ -1,5 +1,26 @@
 #!/usr/bin/env bash
 #DESCRIPTION: initialization of your environment
 
-cp /usr/local/bin/composer composer.phar
-ant -f build/build.xml build-unit
+INCLUDE: ./.init_composer.sh
+INCLUDE: ./cache.sh
+
+php bin/console sw:database:setup --steps=drop,create,import,setupShop --host=__SW_HOST__ --path=__SW_BASE_PATH__
```
DanielBoettner, line 7: Maybe a warning, when **SW_HOST** isn't set within [psh](https://github.com/shopwareLabs/psh) or this specific file would be useful. When testing this PR on my system I forgot to set the host and had to fix it afterwards in the database.

dev-ops/common/actions/.mink_disable-csrf.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+mkdir -p .git/hooks/
+ln -sf build/gitHooks/pre-commit .git/hooks/pre-commit
```
shyim, line 4: Excactly the same as .init_git-hooks.sh

dev-ops/common/actions/init.sh
```diff
@@ -1,5 +1,26 @@
...
+
+php bin/console sw:theme:initialize
+
+php bin/console sw:admin:create --name="Demo user" --email="demo@example.com" --username="demo" --password="demo" --locale="de_DE"
+
+date +"%Y%m%d%H%M" | tee recovery/install/data/install.lock
+
+php bin/console sw:firstrunwizard:disable
+
+curl -k -L -o test_images.zip "http://releases.s3.shopware.com/test_images_since_5.1.zip"
```
shyim, line 19: This should be a seperate command

teiling88:
Good point I will change it :-)

PR #1535, opened 2018-02-19
fix User model fluent interface

buddhaCode:

Fluent interface is broken.

Developer convenience.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/User/User.php
+++ b/engine/Shopware/Models/User/User.php
@@ -594,10 +594,14 @@ class User extends ModelEntity
 
     /**
      * @param string $encoder
+     *
+     * @return \Shopware\Models\User\User
      */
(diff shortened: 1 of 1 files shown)


engine/Shopware/Models/User/User.php
```diff
@@ -594,10 +594,14 @@ public function setAttribute($attribute)
 
     /**
      * @param string $encoder
+     *
+     * @return \Shopware\Models\User\User
```
oktupol: There is no need to specify the fully qualified class name in this case. `@return User` is sufficient, and consistent.

buddhaCode:
@shopwareBot Are there any chances for this PR to get merged?

soebbing:
Hey @buddhaCode, thank you for your contribution! 🎉

PR #1536, opened 2018-02-20
Fixed openPdf with composer version

shyim:
I cant open documents with composer version

This fixes it

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Order.php
+++ b/engine/Shopware/Controllers/Backend/Order.php
@@ -928,7 +928,7 @@ class Shopware_Controllers_Backend_Order extends Shopware_Controllers_Backend_Ex
     public function openPdfAction()
     {
         $name = basename($this->Request()->getParam('id')) . '.pdf';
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Order.php
```diff
@@ -928,7 +928,7 @@ public function createDocumentAction()
     public function openPdfAction()
     {
         $name = basename($this->Request()->getParam('id')) . '.pdf';
-        $file = Shopware()->DocPath('files/documents') . $name;
+        $file = Shopware()->Container()->getParameter('shopware.app.documentsdir') . '/' . $name;
```
mitelg, line 931: please also fix Line 780 in this file

shyim: Done ❤️

soebbing:
Hi @shyim! We merged your PR and changed some other places with similar problems. Thanks for pointing that out!

PR #1538, opened 2018-02-22
add article mode constants to models

buddhaCode:

I'm always looking for the numbers. Nobody would expect them in /themes/Frontend/Bare/frontend/checkout/cart_item.tpl. I think the right place is the model.

Reducing developer pain.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/CanceledOrder.php
+++ b/engine/Shopware/Controllers/Backend/CanceledOrder.php
@@ -22,6 +22,7 @@
  * our trademarks remain entirely with us.
  */
 
+use Shopware\Models\Article\Article as ProductModel;
 use Shopware\Models\Order\Order;
 
 /**
@@ -44,7 +45,7 @@ class Shopware_Controllers_Backend_CanceledOrder extends Shopware_Controllers_Ba
         // Get user, shipping and billing
         $builder = Shopware()->Models()->createQueryBuilder();
         $builder->select(['orders', 'customer', 'billing', 'payment', 'shipping'])
-            ->from(\Shopware\Models\Order\Order::class, 'orders')
+            ->from(Order::class, 'orders')
             ->leftJoin('orders.customer', 'customer')
             ->leftJoin('orders.payment', 'payment')
             ->leftJoin('customer.defaultBillingAddress', 'billing')
@@ -81,7 +82,7 @@ class Shopware_Controllers_Backend_CanceledOrder extends Shopware_Controllers_Ba
         $numberModel->setNumber($newOrderNumber);
 
         // Set new ordernumber to the order and its details
-        $orderModel = Shopware()->Models()->find(\Shopware\Models\Order\Order::class, $orderId);
+        $orderModel = Shopware()->Models()->find(Order::class, $orderId);
         $orderModel->setNumber($newOrderNumber);
         foreach ($orderModel->getDetails() as $detailModel) {
             $detailModel->setNumber($newOrderNumber);
@@ -166,11 +167,12 @@ class Shopware_Controllers_Backend_CanceledOrder extends Shopware_Controllers_Ba
         $params = [
             'endDate' => $endDate,
             'startDate' => $startDate,
+            'articleModeProduct' => ProductModel::MODE_PRODUCT,
         ];
 
         $sql = 'SELECT id
             FROM s_order_basket
-            WHERE modus = 0
+            WHERE modus = :articleModeProduct
             AND datum >= :startDate AND datum <= DATE_ADD(:endDate, INTERVAL 1 DAY)
             GROUP BY sessionID';
         $result = Shopware()->Db()->query($sql, $params);
@@ -216,7 +218,7 @@ class Shopware_Controllers_Backend_CanceledOrder extends Shopware_Controllers_Ba
             (
                 SELECT lastviewport, sessionID
                 FROM s_order_basket
-                WHERE modus = 0
+                WHERE modus = :articleModeProduct
                 AND datum >= :startDate AND datum <= DATE_ADD(:endDate,INTERVAL 1 DAY)
                 $filter
                 GROUP BY sessionID
@@ -525,6 +527,7 @@ class Shopware_Controllers_Backend_CanceledOrder extends Shopware_Controllers_Ba
         $params = [
             'endDate' => $endDate,
             'startDate' => $startDate,
+            'articleModeProduct' => ProductModel::MODE_PRODUCT,
         ];
 
         if (is_array($filter) && isset($filter[0]['value'])) {
@@ -571,7 +574,7 @@ class Shopware_Controllers_Backend_CanceledOrder extends Shopware_Controllers_Ba
         SELECT DATE_FORMAT(datum, '%Y-%m-%d') as date, sum(price) as price, AVG(price) as average,
                   COUNT(DiSTINCT sessionID) as number, YEAR(datum) as year, MONTH(datum) as month
                 FROM `s_order_basket`
-                WHERE s_order_basket.modus = 0
+                WHERE s_order_basket.modus = :articleModeProduct
                     AND datum >= :startDate AND datum <= DATE_ADD(:endDate,INTERVAL 1 DAY)
                     $filter
                 GROUP BY DAY(datum), MONTH(datum), YEAR(datum)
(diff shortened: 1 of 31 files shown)


buddhaCode:
I added the constants via an interface. This maybe breaks with the OO design. There are other ways:

- introduce a new abstract class and let the model classes extend this super class
- introduce a trait an use static varibales. Constants are not allowed in traits.

Personally, I think the interface is the most elegant way. But it's open for discussion.

pascalheidmann-bedarf (approved):

buddhaCode:
@shopwareBot No problem. I will do so. Any idea how to use the constants in smarty tpls?
`{if $mode eq constant('Shopware\Models\Article\Article::MODE_FOO')}` feels wrong and introduces dependencies between model and view.

mitelg:
@buddhaCode I would suggest simply assign the constants to the template in the index controller. so they are available everywhere

buddhaCode:
I'm on it. I just returned from vacation this week. So i couldn't spare some time on work.

buddhaCode:
I decided to use the constant function in templates. Loading the constants to view in the index controller does only help in the index controller. I think a Frontend Predispatch Event handler would work. But didn't know where to place it.

I was searching the code for the terms: modus, .modus, 'modus', "modus", getMode and setMode.

Hopefully, I didn't miss anything.

pthreat:
@buddhaCode Thanks a lot for your pull request, this looks really cool!
One small thing that would make it perfect, would be to use aliased class names, just like you did in:

engine/Shopware/Bundle/CustomerSearchBundleDBAL/Indexing/CustomerOrderGateway.php

Where you used a "use" statement and aliased the class name, I'm talking about the following:

use Shopware\Models\Article\Article; //If you could rename this to ProductModel it would be great.

Then on line 245: $query->setParameter(':articleModeProduct', Article::MODE_PRODUCT);

We would like that you alias in every other file just like you did it in this file.

Again, this is just great and we thank you for your contribution.

buddhaCode:
@pthreat as you wish ;)

themes/Frontend/Bare/frontend/checkout/ajax_cart_item.tpl
```diff
@@ -1,26 +1,19 @@
 {namespace name="frontend/checkout/ajax_cart"}
 
-{* Constants for the different basket item types *}
-{$IS_PRODUCT = 0}
```
shopwareBot, line 4: Removing those might break some plugins. Can you just fill those variables with the constants?

buddhaCode:
sure thing

themes/Frontend/Bare/frontend/checkout/ajax_cart_item.tpl
buddhaCode: sure thing

themes/Frontend/Bare/frontend/checkout/cart_item.tpl
```diff
@@ -1,35 +1,36 @@
 {* Constants for the different basket item types *}
-{$IS_PRODUCT = 0}
-{$IS_PREMIUM_PRODUCT = 1}
-{$IS_VOUCHER = 2}
-{$IS_REBATE = 3}
-{$IS_SURCHARGE_DISCOUNT = 4}
+{ @deprecated, use contants instead }
+{$IS_PRODUCT = constant('Shopware\Models\Article\Article::MODE_PREMIUM')}
```
mitelg: this should be `MODE_PRODUCT` shouldn't it?

buddhaCode: guess so! sorry.

engine/Shopware/Controllers/Backend/CanceledOrder.php
```diff
@@ -564,14 +565,15 @@ public function getBasketAction()
...
+        $articleModeProduct = ProductModel::MODE_PRODUCT;
         // if modus = 0 turns out to be wrong, this statement might me helpful
         //        WHERE s_order_basket.ordernumber = s_order.ordernumber AND s_order.status = -1
         $sql = "SELECT  date, price, average, number, year, month FROM
         (
         SELECT DATE_FORMAT(datum, '%Y-%m-%d') as date, sum(price) as price, AVG(price) as average,
                   COUNT(DiSTINCT sessionID) as number, YEAR(datum) as year, MONTH(datum) as month
                 FROM `s_order_basket`
-                WHERE s_order_basket.modus = 0
+                WHERE s_order_basket.modus = $articleModeProduct
```
pascalheidmann-bedarf: Why not using a prepared statement here?

engine/Shopware/Core/sAdmin.php
```diff
@@ -3952,7 +3957,7 @@ private function handlePaymentMeanSurcharge($country, $payment, $currencyFactor,
                     'netprice' => $surcharge_net,
                     'tax_rate' => $tax_rate,
                     'datum' => new Zend_Date(),
-                    'modus' => 4,
+                    'modus' => Shopware\Models\Article\ProductModel::MODE_PAYMENT_SURCHARGE_DISCOUNT,
```
pascalheidmann-bedarf: Does this even work? This should be just _ProductModel::MODE_PAYMENT_SURCHARGE_DISCOUNT_

engine/Shopware/Core/sBasket.php
```diff
@@ -1050,7 +1051,7 @@ public function sInsertSurchargePercent()
             'netprice' => $discountNet,
             'tax_rate' => $tax,
             'datum' => new Zend_Date(),
-            'modus' => 4,
+            'modus' => Shopware\Models\Article\ProductModel::MODE_PAYMENT_SURCHARGE_DISCOUNT,
```
pascalheidmann-bedarf: Again just _ProductModel::MODE_PAYMENT_SURCHARGE_DISCOUNT_

engine/Shopware/Core/sBasket.php
```diff
@@ -2326,7 +2327,7 @@ private function getBasketArticles(array $getArticles)
             }
 
             // Get additional basket meta data for each product
-            if ($getArticles[$key]['modus'] == 0) {
+            if ($getArticles[$key]['modus'] == Shopware\Models\Article\ProductModel::MODE_PRODUCT) {
```
pascalheidmann-bedarf: And again _ProductModel::MODE_PRODUCT_

themes/Frontend/Bare/frontend/checkout/ajax_cart_item.tpl
```diff
@@ -1,26 +1,27 @@
 {namespace name="frontend/checkout/ajax_cart"}
 
 {* Constants for the different basket item types *}
-{$IS_PRODUCT = 0}
-{$IS_PREMIUM_PRODUCT = 1}
-{$IS_VOUCHER = 2}
-{$IS_REBATE = 3}
-{$IS_SURCHARGE_DISCOUNT = 4}
+{ @deprecated, use contants instead }
```
pascalheidmann-bedarf: Why deprecating and not using those variables? Using those variables looks far more clean than using everytime a constant-lookup

pascalheidmann-bedarf (changes requested):
Several times this PR uses FQN for alias imported Models.
Also using constant() in templates (therefore having lots of function calls with identical input and output) looks not very clean.

engine/Shopware/Controllers/Backend/CanceledOrder.php
buddhaCode: Actually, i missed this.

engine/Shopware/Core/sAdmin.php
buddhaCode: This is really woking ;) Of course the FQM is not necessary.

themes/Frontend/Bare/frontend/checkout/ajax_cart_item.tpl
buddhaCode: I can see your point.

buddhaCode:
Thanks for your review @pascalheidman-bedarf. I made some changes.

cm-springlane:
Isn't mode 10 (in your code MODE_BUNDLE_ARTICLE) the bundle rebate? Perhaps the constant could be named accordingly to avoid confusion.

buddhaCode:
Yes, your absolutly right. Sorry for the mistake. Thanks for the review, @cm-springlane.

buddhaCode:
@shopwareBot Are there any chances for this PR to get merged?

PaddyS:
Hey @buddhaCode,

can you please rebase your branch to the current 5.5 and squash your commits?
We **might** have a look at it soon. :-)

buddhaCode:
@PaddyS Puh never rebased something before. But i'll try. I'm on vacation for the next 1,5 weeks. I'll give it a try when I returned.

buddhaCode:
@PaddyS I tried to squash and rebase today. But there are waaaaaaaay too many changes to fix them all. Sorry. I will close the PR.

PR #1539, opened 2018-02-22
assign glyph-names

tinect:
It makes it easier to change the font or creating another one (f.e. with icomoon) using shopware's glyph-names - without changing all the names in template.

adding glyph-names to svg

### 6. Checklist


tinect:
see #1540

PR #1541, opened 2018-02-23
Put icon in ajax_cart right as it is classed

tinect:
Icon isn't displayed.

it moves the icon right, as it is defined in class.

### 6. Checklist


tinect:
some more broken. will check and come back :-)

PR #1544, opened 2018-02-27
Added theme typehint to LessDefinition

shyim:
Currently you can pass everything, but the property only allows a Theme or null otherwise the LessCollector crashes.  (Call to a member function getDiscardedLessThemes() on array)

Adds a simple Typehint to avoid that

Do $definition->setTheme(['shopware die']); and theme compile is broken with message "Call to a member function getDiscardedLessThemes() on array" which does not help

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Theme/LessDefinition.php
+++ b/engine/Shopware/Components/Theme/LessDefinition.php
@@ -108,7 +108,7 @@ class LessDefinition
      * @param array $config          contains the less variables, has to be a key value array
      * @param array $files           contains the full file name paths
      * @param null  $importDirectory Full path to the import directory for less @import commands
-     * @param Theme $theme           the corresponding theme
+     * @param Theme|null $theme           the corresponding theme
      */
     public function __construct(array $config = [], array $files = [], $importDirectory = null, Theme $theme = null)
     {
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Theme/LessDefinition.php
```diff
@@ -107,9 +108,9 @@ class LessDefinition
      * @param array $config          contains the less variables, has to be a key value array
      * @param array $files           contains the full file name paths
      * @param null  $importDirectory Full path to the import directory for less @import commands
-     * @param null  $theme           the corresponding theme
+     * @param Theme $theme           the corresponding theme
```
mitelg: should be `Theme|null`

engine/Shopware/Components/Theme/LessDefinition.php
```diff
@@ -107,9 +108,9 @@ class LessDefinition
      * @param array $config          contains the less variables, has to be a key value array
      * @param array $files           contains the full file name paths
      * @param null  $importDirectory Full path to the import directory for less @import commands
-     * @param null  $theme           the corresponding theme
```
mitelg: @shyim could fix that typehint, too? should be `string|null`

and if you are on it anyway, would you mind to fix the `mixed` typehints below?

PR #1546, opened 2018-02-28
Fixing 0% Vat Problem

s-bessing:
If a user group is set to 0% VAT, the surcharge tax will set to 19%. This results in a negativ VAT and the Total is less then the price net.

### 6. Checklist


pascalheidmann-bedarf (changes requested):
Use !== comparator instead of ambiguous !=

s-bessing:
I tried it but it didn't worked. It seems, that the value is not an number. 
The part has to be fixed by shopware anyway, because the hardcoded 19% are not a really good solution.

PR #1547, opened 2018-02-28
Enable smarty 3 on codemirror mode

shyim:
The default mode is smarty 2, see https://github.com/shopware/shopware/blob/5.4/engine/Library/CodeMirror/mode/smarty/smarty.js#L21

Changes the default mode to smarty 3

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/base/component/Shopware.form.field.CodeMirror.js
+++ b/themes/Backend/ExtJs/backend/base/component/Shopware.form.field.CodeMirror.js
@@ -178,6 +178,15 @@ Ext.define('Shopware.form.field.CodeMirror',
 
         me.config = config;
 
+        if (me.config.mode === 'smarty') {
+            me.config.mode = {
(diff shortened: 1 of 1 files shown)


themes/Backend/ExtJs/backend/base/component/Shopware.form.field.CodeMirror.js
```diff
@@ -178,6 +178,15 @@ Ext.define('Shopware.form.field.CodeMirror',
 
         me.config = config;
 
+        if (me.config.mode === 'smarty') {
+            me.config.mode = {
+                name: 'smarty',
+                version: 3
+            }
+        } else if(me.config.mode.name === 'smarty') {
```
shopwareBot, line 186: this leads to errors if:

- config is not an object
- config has no property mode
- config.mode is not `smarty`
- config.mode is not an object

Please fix these issues and your code style :-)

PR #1550, opened 2018-03-01
Use individual join alias for closeout variant join.

kleinmann:
Currently a fatal error occurs if one uses a product stream with a closeout condition (Abverkauf-Filter) for a category and requests that category in the shop:

```Uncaught Doctrine\DBAL\Query\QueryException: The given alias 'variant' is not unique in FROM and JOIN clause table. The currently registered aliases are: product, variant, avoidCustomerGroup, productCategory.```

This happens because `\Shopware\Bundle\SearchBundleDBAL\QueryBuilderFactory::createQuery` adds a join on `s_articles_details` with the alias `variants` and `\Shopware\Bundle\SearchBundleDBAL\ConditionHandler\CloseoutConditionHandler::generateCondition` also tries to join with that exact alias.

I used an individual join alias in the condition handler (as that is more likely to not be in use) to prevent the collision.

An alternative solution would be to check if a join called `variant` is already present in the query builder and reuse that. Maybe a state can be introduced for that, like `\Shopware\Bundle\SearchBundleDBAL\ConditionHandler\SalesConditionHandler::STATE_INCLUDES_TOPSELLER_TABLE`. Please let me know if you prefer that solution. I wanted to fix the issue first and foremost. :-)

* Create a product stream and select the "Abverkauf-Filter"
* Create a category using that product stream
* Access the category in the shop frontend

no issue

none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/ConditionHandler/CloseoutConditionHandler.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/ConditionHandler/CloseoutConditionHandler.php
@@ -37,6 +37,8 @@ use Shopware\Bundle\StoreFrontBundle\Struct\ShopContextInterface;
  */
 class CloseoutConditionHandler implements ConditionHandlerInterface
 {
+    const STATE_INCLUDES_VARIANT_TABLE = 'variant';
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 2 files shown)


kleinmann:
I have changed the PR to use a query state. However, I am not that happy with the dependency on `CloseoutConditionHandler` in `QueryBuilderFactory`. Other STATE constants are all in various handlers, so for consistency it might be preferred this way.

I'm not sure, if the condition handler can even be used without creating the query in the `QueryBuilderFactory`. In that case the join in `CloseoutConditionHandler` may be fully redundant and not needed.

kleinmann:
Seems like this was closed in a later commit (f141c952a4cc2430c6a983ef8bbcb94a08f2b9af) for issue SW-21708. I'm closing this PR.

PR #1551, opened 2018-03-01
Add smarty block around "back to shop" button

hrombach:
Not a huge deal probably, but currently the "back to shop" button is appended to the block `frontend_index_logo_trusted_shops`, which means if you wanted to hide it, you'd have to hide the whole block. By default, the block is empty, but it might have been appended or edited, so I just feel that it would be a good idea to add an extra block here.

Wrap the code for the "back to shop" button in a block `frontend_register_back_to_shop_button`.

-/-

-/-

-/-

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/register/index.tpl
+++ b/themes/Frontend/Bare/frontend/register/index.tpl
@@ -13,14 +13,16 @@
 {* Back to the shop button *}
 {block name='frontend_index_logo_trusted_shops'}
     {$smarty.block.parent}
-    {if $theme.checkoutHeader && !$toAccount}
-        <a href="{url controller='index'}"
-           class="btn is--small btn--back-top-shop is--icon-left"
-           title="{"{s name='FinishButtonBackToShop' namespace='frontend/checkout/finish'}{/s}"|escape}">
-            <i class="icon--arrow-left"></i>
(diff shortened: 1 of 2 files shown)


themes/Frontend/Bare/frontend/register/index.tpl
```diff
@@ -13,14 +13,16 @@
...
     {$smarty.block.parent}
-    {if $theme.checkoutHeader && !$toAccount}
-        <a href="{url controller='index'}"
-           class="btn is--small btn--back-top-shop is--icon-left"
-           title="{"{s name='FinishButtonBackToShop' namespace='frontend/checkout/finish'}{/s}"|escape}">
-            <i class="icon--arrow-left"></i>
-            {s name="FinishButtonBackToShop" namespace="frontend/checkout/finish"}{/s}
-        </a>
-    {/if}
+    {block name='frontend_register_back_to_shop_button'}
```
shopwareBot: please rename it to `frontend_register_index_back_to_shop_button`

hrombach: Done :)

hrombach:
I rebased this PR since it seems to have gotten stale

soebbing:
Hi @hrombach, thank you for your help! Your change will be part of the upcoming 5.4.4 release! 🎉

PR #1553, opened 2018-03-02
Use the Minimum search term length for the Ajax Search

fmaasde:
ATM. You can set the minimum search term length in the backend but this value is not used for the Ajax search.

It adds the config value (MinSearchLenght) for the minimum search term length to the Ajax search using data-minLength in the  `shop-navigation.tpl`

Set the "Configuration > Basic settings > Frontend > Search > Maximum search term length"
Without this commit the the default value '3' will be used and the Ajax search starts with 3 characters.

With this commit the value from the config will be used to trigger the Ajax search.

-

-

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/index/shop-navigation.tpl
+++ b/themes/Frontend/Bare/frontend/index/shop-navigation.tpl
@@ -12,7 +12,7 @@
 
         {* Search form *}
         {block name='frontend_index_search'}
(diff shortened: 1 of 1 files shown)


pascalheidmann-bedarf (approved):

soebbing:
Hi @fmaasde, thank you for your contribution! 💙 💋

PR #1559, opened 2018-03-07
IMP allow to set a PropertyValues relations position via api

Guenzn:
To control the order of property values, it is necessary to set the position attribute of the values related \Shopware\Models\Property\Relation 

It introduces a new field to API resource **PropertyValue** named _relationPosition_ , which will be used to set the relations position when creating or updating propertyValues via the articles API.

There is no possibility to control the order of property values before this feature.

None

https://developers.shopware.com/developers-guide/rest-api/models/#property-value
Documentation of property value needs to be extended

### 6. Checklist

### 7. TODO (Sorry, it was to complicated to me, because I'm new to shopware)
The field _relationPosition_ should also be returned inside  **PropertyValue** by a GET request to articles API.


Guenzn:
closed,

PR #1561, opened 2018-03-07
Add number range for doc_3/storno bill

pascalheidmann-bedarf:
Add default number range for storno bills as this is lacking right now. Therefore storno bills title read _"Reversal invoice for invoice No. "_.

Add entry to s_order_number table to have a concrete number range for storno bills

Without this PR create a storno bill and you will see that the created document has no number range (therefore name _0.pdf_) and inside the document is the addressed title without a number.
With an added number range index the functionality can find a valid number.

- https://github.com/shopware/shopware/pull/1560 This PR does basically the same as the referenced PR just using migrations instead of direct editing the install sql.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1214-add-stornobill-numberrange.php
+++ b/_sql/migrations/1214-add-stornobill-numberrange.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


_sql/migrations/1214-add-stornobill-numberrange.php
```diff
@@ -0,0 +1,31 @@
...
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+class Migrations_Migration1214 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $this->addSQL('INSERT INTO `s_order_number`(`number`, `name`, `desc`) VALUES (20000, "doc_3", "Stornorechnungen")');
```
shyim: Some customers have the number range already. That would explode on updates 😄  (name is unique)

pascalheidmann-bedarf: Yeah but as those number ranges used for documents are autogenerated as doc_ + DOCID every other name wouldn't work.
Might only change the migration to INSERT IGNORE to prevent this problem.

shyim: insert ignore would be nice 👍

PR #1563, opened 2018-03-07
Fixed exception, when license key is not readable

shyim:
There is a empty RuntimeException without a message

Adds a message to exception

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLicenceService.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLicenceService.php
@@ -89,11 +89,6 @@ class PluginLicenceService
             throw new \RuntimeException('Could not read license key');
         }
 
-        $this->connection->delete('s_core_licenses', [
-            'module' => $info['module'],
-            'host' => $info['host'],
-        ]);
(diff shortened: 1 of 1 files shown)


JoshuaBehrens:
I absolutely had this scenario! The shop owner thinks at first it is a plugin problem not a license manager problem.

What about making module and host combined unique? So this scenario will be impossible by structure.

shyim:
That could be a better solution, but the license manager plugin needs also a update and that is closed source 😟 . Otherwise customers who are trying inserting a "duplicate license", seeing a sql error and think w000t my shop is broken.

JoshuaBehrens:
Unfortunately this is true.

@shopwareBot , pls fix

mitelg:
hey @shyim @JoshuaBehrens 
I'm currently working on the license manager plugin and I added a check before inserting into the DB, if there is already an entry for the module/host pair. If so, the entry will be updated. I think that change will make this PR obsolete. Additionally the license manager also supports older Shopware versions. 
What do you guys think about that?

shyim:
Sure it would be better in License Manager. I would leave this pr open, because the empty Exception 😄

PR #1566, opened 2018-03-09
Fix creating duplicate item export entries when running cached export

biforb4:
Stop creating duplicate entries in s_export

After calling `$this->generateExport($filePath);` the `$productFeed` becomes detached entity.
`$productFeed = $em->merge($productFeed);` attaches it back to the Entity Manager.
`$productFeed->setLastExport('now');` sets last_export which would be overridden after calling `EntityManager::merge()`

Singleton `Shopware()->Models()` is refactored to `$em = $this->container->get('models');`

Create an item export with Caching interval set to anyhing other than "Live" and "Only cron". Run the export, it will crate a duplicate entry in s_export table

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Export.php
+++ b/engine/Shopware/Controllers/Backend/Export.php
@@ -61,8 +61,10 @@ class Shopware_Controllers_Backend_Export extends Enlight_Controller_Action impl
     {
         $this->prepareExport();
         $this->sendHeaders();
-
-        $productFeed = Shopware()->Models()->getRepository('\Shopware\Models\ProductFeed\ProductFeed')->find((int) $this->Request()->feedID);
+        /** @var \Shopware\Components\Model\ModelManager $em */
+        $em = $this->container->get('models');
+        /** @var \Shopware\Models\ProductFeed\ProductFeed $productFeed */
+        $productFeed = $em->getRepository('\Shopware\Models\ProductFeed\ProductFeed')->find((int) $this->Request()->feedID);
 
         // Live generation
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Export.php
```diff
@@ -91,8 +93,9 @@ public function indexAction()
 
             // update last refresh
             $productFeed->setCacheRefreshed('now');
-            Shopware()->Models()->persist($productFeed);
-            Shopware()->Models()->flush($productFeed);
+            $productFeed->setLastExport('now');
+            $productFeed = $em->merge($productFeed);
```
shopwareBot, line 97: Is this merge really necessary?

biforb4: Yes, `$productFeed` entity is in detached state so persist creates a new row in s_export table every time feed runs. `$em->merge($productFeed)` prevents that.

biforb4:
Any idea when will this be merged?

shyim:
Hey @biforb4,

i tried many things to reproduce that on Shopware 5.5.1 and it does not work . I tried with 2 minutes cache time

* Multiple Requests to open the product feed in Web
* Multiple Requests and Cron simultaneously or Request then Cron
* Open, wait for cache expire and request again
* Run multiple very big import with 20k products simultaneously

Could you check that issue again on Shopware 5.5.1?

alea123:
we have the same issue on sw 5.4.3 ... but its here for many versions below 5.4.3 ... hope after updating to 5.5+ it will be gone

PR #1165, opened 2017-06-15
Respect defaultValue on Checkbox Attributes

shyim:
## Description

| Questions               | Answers |
|-------------------------|-------------------------------------------------------|
| Why?                    | I created a checkbox attribute on articles, while creating a new article the checkbox doesn't respect the defaultValue and was anytime unchecked. So the option is currently useless |
| BC breaks?              | no |
| Tests exists & pass?    | yes |
| Related tickets?        |  |
| How to test?            | $this->get('shopware_attribute.crud_service')->update('s_articles_attributes', 'singlearticle', 'boolean', [
            'displayInBackend' => true,
            'label' => 'Artikel nur einmal bestellbar'
        ], null, false, 1); |
| Requirements met?       | Yep |

PR #1568, opened 2018-03-12
Added defaultValue handling in attributes

shyim:
Respecting default values on created fields

Uses the defaultValue field in backend 😊 

```php
 foreach ($this->container->getParameter('shopware_attribute.table_entity_mapping') as $table => $item) {
            $this->container->get('shopware_attribute.crud_service')->update($table, 'test', TypeMapping::TYPE_BOOLEAN, [
                'displayInBackend' => true,
                'label' => 'LOL'
            ], null, false, true);

            $this->container->get('shopware_attribute.crud_service')->update($table, 'test2', TypeMapping::TYPE_STRING, [
                'displayInBackend' => true,
                'label' => 'LOL TEXT'
            ], null, false, 'Hi i am a Text');

            $this->container->get('shopware_attribute.crud_service')->update($table, 'test3', TypeMapping::TYPE_INTEGER, [
                'displayInBackend' => true,
                'label' => 'LOL INT'
            ], null, false, 1);

            $this->container->get('shopware_attribute.crud_service')->update($table, 'test4', TypeMapping::TYPE_FLOAT, [
                'displayInBackend' => true,
                'label' => 'LOL FLOAT'
            ], null, false, 1.5);

            $this->container->get('shopware_attribute.crud_service')->update($table, 'test5', TypeMapping::TYPE_HTML, [
                'displayInBackend' => true,
                'label' => 'LOL HTML'
            ], null, false, '<h1>Hii i am html</h1>');

            $this->container->get('shopware_attribute.crud_service')->update($table, 'test6', TypeMapping::TYPE_TEXT, [
                'displayInBackend' => true,
                'label' => 'LOL TEXTAREA'
            ], null, false, 'AREAA');
        }
```

Image

#1165 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/AttributeBundle/Service/CrudService.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Service/CrudService.php
@@ -332,6 +332,7 @@ class CrudService
             'tableName' => $table,
             'columnName' => $column,
             'columnType' => $unifiedType,
+            'defaultValue' => $defaultValue,
         ]);
 
         $configId = null;
--- a/themes/Backend/ExtJs/backend/base/attribute/Shopware.attribute.Form.js
+++ b/themes/Backend/ExtJs/backend/base/attribute/Shopware.attribute.Form.js
@@ -191,7 +191,7 @@ Ext.define('Shopware.attribute.Form', {
 
         Ext.each(fields.items, function(field) {
             try {
-                field.setValue(null);
+                field.setValue(typeof field.defaultValue === 'undefined' ? null : field.defaultValue);
             } catch (e) {
             }
         });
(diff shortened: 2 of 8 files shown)


FiveDigital:
The string type does not allow a default value: https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Bundle/AttributeBundle/Service/TypeMapping.php#L55

shyim:
@FiveDigital Good catch, i have removed it. Its useless when the backend view shows a defaultValue which is not defineable in table column. Thanks 😊

FiveDigital:
np! Although I have to say that I find it stupid that no default value can be used for string attributes.

OliverSkroblin:
@FiveDigital Thats because a mysql TEXT field do not support a default value. 
And for the following question why we use TEXT and not VARCHAR, this is because mysql has a max row size that is reached faster with a varchar than with a TEXT BLOB ;)

FiveDigital:
Perhaps we can add a small text or whatever mapped to a VARCHAR, too? Because we are in need of this in nearly every project so far.

mitelg:
thanks @shyim :blush: :blue_heart:

PR #1569, opened 2018-03-13
Added fileSize to media manager table

shyim:
Seen which file is the biggest in the albumn

Adds the fileSize column (default hidden)
bildschirmfoto 2018-03-13 um 10 11 16

### 6. Checklist


Hotte512:
This one is nice... THX!

As the ISSUE-Tracker seems to be kinda slow..
Would you mind to add a method to check for duplicate images and Filenames on Upload? :-)

Story:
Backend-User is creating an new Product wich uses the same picture as other Producst and as he is a "USER" he would not notify to search for that picture in Media-Library and instead Upload it again.

=>Shopware is going to Save the Pictre n-times with some hash in Filename. - BAD IDEA! (Storage / Bad Filenames etc.)

There Should be a check if the File (may be by Checksum, for performance save it to DB? ) and Filename already exists, if so the user should see Preview and be asked to use Picture A instead of saving the new one or rename it in a more sens-full way.
(Also casesensitiveness could bring trouble for those doing stuff on windows..)

The super duper function would be if all locations the Media is Used are saved and shown (like: "Image is Used in Product SW1000, SW1005, Blog ID2") in the MediaManager (also good for Garbage Collector!) 

By the Way, you name Media Manager once "Media Manager" and once "Medienverwaltung".
image

THX again - I will also create this as ISSUE.

soebbing:
Once again, thanks @shyim! 💙

PR #1573, opened 2018-03-15
Hide reference price if referenceUnit = 0

pascalheidmann-bedarf:
Without this PR price calculations per unit result in 0,00/unit prices.

Add check to check if sArticle.referenceunit is defined if not hide those calculations.

1. Create/edit an article to have purchaseunit set to 1 (or anything other) but leave referenceunit empty.
2. Load up articles detail page on frontend
3. Without this PR it will state something like _1 piece (0.00€ / 1 piece)_ below the price. With this PR this infobox won't show up.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/detail/block_price.tpl
+++ b/themes/Frontend/Bare/frontend/detail/block_price.tpl
@@ -4,7 +4,7 @@
         {* @deprecated *}
         {block name='frontend_detail_data_block_prices_headline'}{/block}
 
-        {$hasReferencePrice = ($sArticle.referenceprice > 0)}
+        {$hasReferencePrice = ($sArticle.referenceprice > 0 && $sArticle.referenceunit)}
 
         {block name="frontend_detail_data_block_prices_table"}
             <table class="block-prices--table">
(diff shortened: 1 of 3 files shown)


pascalheidmann-bedarf:
As an alternative you might set referenceunit=1 if referenceunit is not a valid number.

Mardl:
Missing 
themes/Frontend/Bare/frontend/detail/block_price.tpl:68
`{if $hasReferencePrice && $sArticle.referenceunit }`

themes/Frontend/Bare/frontend/detail/block_price.tpl
```diff
@@ -65,7 +65,7 @@
                                                         {$blockPrice.price|currency} {s name="Star" namespace="frontend/listing/box_article"}{/s}
                                                     </td>
                                                 {/block}
-                                                {if $hasReferencePrice}
+                                                {if $hasReferencePrice && $sArticle.referenceunit }
```
mitelg: would it make sense to add this to [Line 7](https://github.com/shopware/shopware/pull/1573/files#diff-f7419b3e9dcd913aafc5943b9c96bbb0R7)?

Mardl: oh, yes, sorry. Need in line 7, because it is also relevant for line 27

soebbing:
Hi @pascalheidman-bedarf, thank you for your contribution! 💙

PR #1574, opened 2018-03-16
Added basketId to sExport shipping

shyim:
We have used the basket id in custom shipping calculation and now the product feed exports to google broken, because the id is not defined

Adds id as null

### 6. Checklist


pascalheidmann-bedarf (approved):

shyim:
So i don't allowed to use basket attributes in custom shipping calculation? 😢

PR #1575, opened 2018-03-16
Require search terms keywords to have a minimum length to be indexed

pascalheidmann-bedarf:
We realised that (especially with SwagFuzzy search enabled) our search results were quite bad as the indexed keywords often contained only 1 or 2 characters and therefore produced bad alternatives or even worse ranked high as partial matches.
By reusing the already in place term min length config to filter search terms we could improve drastically the results as false positives got minimised a lot.

The PR extends the filter mechanism of the _TermHelper_'s _filterBadWordsFromString_ to also filter out terms that are shorter than the value in _minsearchlenght_ (that typo is already in Shopware sometime and would require its own PR). Those then won't turn up in s_search_keywords.
This approach is quite bruteforce but with default settings (min 3 chars) still very effective versus a hand made list of all _badwords_ which will miss most cases.

Check your s_search_keywords table without this PR and you might see lots of 2 digits numbers or english words etc. that aren't part of your _badwords_ list.
After truncating your keyword index and rebuilding it with this PR in place those low quality keywords should be history.

As this PR reuses _minsearchlenght_ the description of this field should mentions this functionality.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1448-add-minsearch-index-length.php
+++ b/_sql/migrations/1448-add-minsearch-index-length.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+use Shopware\Components\Migrations\AbstractMigration;
+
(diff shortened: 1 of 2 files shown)


dneustadt:
I'd like to recommend setting up a new/different config value for the required length of indexed keywords. I can see cases where the operator intends the user to enter more complex keywords but still wants parts of those to be recognized. Example: The user should enter "Taxation" but the keyword "Tax" should still be legible for indexation.

shyim:
Hey @pascalheidman-bedarf,

sorry for the late response :see_no_evil:  and thanks for the Contribution :blue_heart:.
I understand your point, but the current implementation could break shops after upgrading which changed min. search length. 
So could you please change your PR to:
- a new config option which is initial 0, to do the same as without your change
- use the new config option in that condition

pascalheidmann-bedarf:
@dneustadt @shyim good suggestions, I will to incorporate it soon™

_sql/migrations/1448-add-minsearch-index-length.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function up($modus)
+    {
+        $sql = <<<SQL
+INSERT INTO `s_core_config_elements` (`form_id`, `name`, `value`, `label`, `description`, `type`, `required`, `position`, `scope`, `options`)
+VALUES
+	((SELECT form.id FROM s_core_config_elements element JOIN s_core_config_forms form ON form.id = element.`form_id` WHERE form.`name` = "Search" AND element.name = "minsearchlenght" LIMIT 1), 'minSearchIndexLength', 's:1:\"0\";', 'Minimale Suchwortlänge', NULL, 'text', 0, 0, 0, NULL);
```
shopwareBot, line 36: Element translation missing. `s_core_config_element_translations`

PR #1579, opened 2018-03-21
Fix typo

Diu:
With `'` enclosed string doesn't expand variables

Replace `'` with `"` in shop resource

Send invalid request to shop resource, like `{ "id": 1, "categoryId": 0 }`

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Shop.php
+++ b/engine/Shopware/Components/Api/Resource/Shop.php
@@ -216,7 +216,7 @@ class Shop extends Resource
                 }
             } else {
                 if (isset($params[$param]) && empty($params[$param])) {
-                    throw new \Exception('param $param may not be empty');
+                    throw new \Exception("param $param may not be empty");
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Api/Resource/Shop.php
```diff
@@ -216,7 +216,7 @@ private function prepareShopData($params, $shop = null)
                 }
             } else {
                 if (isset($params[$param]) && empty($params[$param])) {
-                    throw new \Exception('param $param may not be empty');
+                    throw new \Exception("param $param may not be empty");
```
mitelg: please use something like [sprintf](http://php.net/manual/en/function.sprintf.php)

Diu: The other exceptions in the same file also don't use sprintf. To be consistent I would have to check all of them as well

soebbing:
Hey @diu! Thank you for your contribution! 💙

PR #1581, opened 2018-03-21
Added default table options for SchemaTool

shyim:
Doctrine should prefer defaultTableOptions instead of mysql server configuration

Adds defaultTableOptions to Doctrine connection (https://symfony.com/doc/master/bundles/DoctrineBundle/configuration.html search for default_table_options)

There was many forum threads, with illegal mix of collates because incorrect collates between default shopware tables and generated using SchemaTool. I hope that fixes it.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/Bridge/Db.php
+++ b/engine/Shopware/Components/DependencyInjection/Bridge/Db.php
@@ -103,6 +103,11 @@ class Db
         $options['user'] = $options['username'];
 
         unset($options['username'], $options['adapter']);
+        
+        $options['defaultTableOptions'] = [
+            'charset' => 'utf8',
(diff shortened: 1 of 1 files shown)


shyim:
Would that not collide with old installations, where the table definition is utf8? In my installed git version(5.4) all tables are still utf8 :)

mitelg:
@shyim you are right :wink: 
we had the plan, to change that, but discarded it again

shyim:
@mitelg Should that then not also utf8? https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Configs/Default.php#L93 😉

shyim:
relevant forum thread due utf8 issue https://forum.shopware.com/discussion/comment/219468#Comment_219468

PR #1583, opened 2018-03-23
Fix ProductSearchResult::__construct() error by Stream Listing

ahmad-saad:
because it breaks the listing streaming page by ajax request (listingCountAction)

after this [Commit](https://github.com/shopware/shopware/commit/ae2c77c32b18b9c09d0715615c141f72776be408#diff-086d23a1c2e8dd882fbfb04c40b8926f) the `\Shopware\Bundle\SearchBundle\ProductSearchResult::__construct` requires now the used Criteria and ShopContext object.

you had this changed in 'engine/Shopware/Bundle/SearchBundle/ProductSearch.php' but not in 'engine/Shopware/Controllers/Widgets/Listing.php' exactly in **fetchStreamListing** line [430](https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Controllers/Widgets/Listing.php#L430)

just add a product stream to a category then try to use the filter in the listing page this will fire the following error:

> Uncaught exception: Argument 4 passed to Shopware\Bundle\SearchBundle\ProductSearchResult::__construct() must be an instance of Shopware\Bundle\SearchBundle\Criteria, none given, called in ....../engine/Shopware/Controllers/Widgets/Listing.php on line xxxx

no

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Widgets/Listing.php
+++ b/engine/Shopware/Controllers/Widgets/Listing.php
@@ -430,7 +430,9 @@ class Shopware_Controllers_Widgets_Listing extends Enlight_Controller_Action
         return new ProductSearchResult(
             $result->getProducts(),
             $result->getTotalCount(),
-            $facetFilter->filter($result->getFacets(), $criteria)
+            $facetFilter->filter($result->getFacets(), $criteria),
+            $criteria,
+            $context
         );
     }
 


mitelg:
thanks for your contribution @ahmad-saad :+1: :tada:

PR #1585, opened 2018-03-23
wrong param sort in sAdmin::sUpdateNewsletter

resslinger:
In the controller frontend/newsletter will be called the method Shopware()->Modules()->Admin()->sNewsletterSubscription(Shopware()->System()->_POST['newsletter'], true);

The first parameter is the email, the second the status of the subscription.

But the sAdmin::sUpdateNewsletter has actually as first parameter the status and as the second parameter the email.
So it doesn't work correct.

This change edit the sort of the parameter in sAdmin::sUpdateNewsletter to the wrong call.

-

-

-

### 6. Checklist

- [- ] I have written tests and verified that they fail without my change

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -536,13 +536,13 @@ class sAdmin
     /**
      * Add or remove an email address from the mailing list
      *
-     * @param bool   $status   True if insert, false if remove
      * @param string $email    Email address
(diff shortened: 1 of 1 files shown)


shyim:
You should change your target to 5.5, its a breaking  change on all plugins which hooks that place

resslinger:
You are right. We should fix the newsletter subscription on the form page, because the newsletter subscription work not so well...

PR #1586, opened 2018-03-26
Bugfix wrong entity in from condition

barbieswimcrew:
Because there is an error that Shopware\Models\Site\Site model does not have a field "siteId" when using the repository's getAttributesQuery() method it came up, that the from condition was wrong. Instead of Shopware\Models\Site\Site we need to select Shopware\Models\Attribute\Site.

Therefor we need to replace the wrong entity in the from condition of the getAttributesQueryBuilder method.

Just try to call the Site Repository's getAttributesQuery($siteId) method to retrieve attributes for a specific static cms page. An exception will be thrown because the model has no field "siteId"

\-

\-

### 6. Checklist


soebbing:
Hi @barbieswimcrew, thank you very much for your contribution! 💙

PR #1592, opened 2018-03-29
Product Overview- MultiEdit Component: Fix BackupDir Path for products so that the prod…

StefanS-O:

The product overview in the backend didn't render, because the setupBackupDir method returned the wrong path.

Shopware()->Container()->getParameter('shopware.app.rootdir') returns the project path without trailing slash, so the when concating a filepath you need to begin with a slash.

1. Open Backend
2. Go to Articles / Overview

Without this fix, you get an exception that the backup dir can't be found.

does not apply

does not apply

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/MultiEdit/Resource/Product/Backup.php
+++ b/engine/Shopware/Components/MultiEdit/Resource/Product/Backup.php
@@ -96,7 +96,7 @@ class Backup
     public function setupBackupDir()
     {
         $projectDir = Shopware()->Container()->getParameter('shopware.app.rootdir');
-        $this->backupPath = $projectDir . 'files/backup/multi_edit';
+        $this->backupPath = $projectDir . '/files/backup/multi_edit';
         $this->backupPath = rtrim($this->backupPath, '/\\') . '/';
 
         if (!is_dir($this->backupPath)) {


StefanS-O:
This is also an issue in 5.5

soebbing:
Hi @StefanS-O! The parameter "shopware.app.rootdir" is being created in \Shopware\Components\ConfigLoader::DocPath and should always end in a slash. Are you sure this hasn't got something to do with your installation?

StefanS-O:
Hi @soebbing, i used composer to install Shopware, maybe it's related to that?

soebbing:
Good hint! In composer this parameter is being generated differently and i think we had to change that after the initial release.

At least right now it also has a slash at the end, maybe have a look at how the path is being generated here: https://github.com/shopware/composer-project/blob/master/app/config/config.php#L15

StefanS-O:
Ok so i will just change the projectDir path in my config and everything should be fine.
I probably used the installer before this commit:
https://github.com/shopware/composer-project/commit/b53a5082a560b9bc646d3105cebd143f9a3075b9#diff-ea24575eefced9021991b9113cfea18e

Sorry for the confusion :)

soebbing:
Sorry for the problems 😭 

Maybe you could watch the composer-repo to be notified if there is a change that might be relevant for you? I think this is still an issue that it is hard to keep track of commits there, though they aren't that often as on Shopware itself .

StefanS-O:
Will do!

PR #1593, opened 2018-03-31
Added assertRequiredPluginsPresent to Shopware\Components\Plugin ...

Mardl:
… like in legacy Shopware_Components_Plugin_Bootstrap

The same like in legacy Shopware_Components_Plugin_Bootstrap

Added in Upgrade-5.4.md

### 6. Checklist

### DIFF
--- a/UPGRADE-5.4.md
+++ b/UPGRADE-5.4.md
@@ -8,6 +8,7 @@ This changelog references changes done in Shopware 5.4 patch versions.
 
 ### Additions
 
+* Added assertRequiredPluginsPresent to Shopware\Components\Plugin like in legacy Shopware_Components_Plugin_Bootstrap
 * Added possibility to enable/disable forms without having to delete them
 * Added pagination to the attribute filter of the product stream configurator 
 * Added `json` attribute for snippets in `Enlight_Components_Snippet_Resource`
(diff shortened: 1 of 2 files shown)


aragon999:
I am not sure if this function is needed, since it can be easily implemented but if, it should get implemented, I think the the container parameter `active_plugins` should be used to check if the required plugins are active, to avoid database queries:
```php
protected function assertRequiredPluginsPresent(array $plugins)
{
    return !(bool) array_diff(
        $plugins,
        array_keys($this->container->getParameter('active_plugins'))
    );
}
``` 
Maybe if such a function is implemented there are still some more things to be considered, the first things which came into my mind are: 
- Check for the correct versions of each plugin
- Maybe it would be better not to return a Boolean value, but an array containing all the missing plugins 
- Maybe it is better to put the functionality in a service, since some logic might differ if a certain plugin is present or not

mitelg:
hey @Mardl 
thanks for your first PR! 
In the new plugin system the required plugins should be declared in the `plugin.xml` file of your plugin. 
See the XSD file https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Components/Plugin/schema/plugin.xsd#L17

You can find an example here: https://developers.shopware.com/developers-guide/plugin-quick-start/#require-other-plugins

PR #1595, opened 2018-04-02
Addming IDs to login and register form for better JS usage

Isengo1989:
Forms can be better recognized by simple JS (like for Google Invisible ReCaptcha)

It adds an id to the login and register form

It simple adds two IDs

### 6. Checklist


soebbing:
Hi @Isengo1989, thank you very much for your contribution! 💙

PR #1599, opened 2018-04-03
Added language switcher for products

shyim:
Enables redirect between languages in product detail page without landing on home page or 404

Redirects to language seo url of the new shop (language switcher)

Go to any product and switch the language to english

### 6. Checklist


shyim:
Problem cannot be reproduced in 5.5 branch and works without this patch.

PR #1601, opened 2018-04-03
Use same session for different languages of same shop

shyim:
Add some items to cart, change to a another language everything is gone

Use same session for different languages of same shop. Also checks for categories in sBasket

https://issues.shopware.com/issues/SW-17499

https://issues.shopware.com/issues/SW-17499

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/Bridge/Session.php
+++ b/engine/Shopware/Components/DependencyInjection/Bridge/Session.php
@@ -86,11 +86,11 @@ class Session
 
         /** @var $shop \Shopware\Models\Shop\Shop */
         $shop = $container->get('Shop');
+        $mainShop = $shop->getMain() ?: $shop;
 
-        $name = 'session-' . $shop->getId();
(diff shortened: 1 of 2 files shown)


mitelg:
Hi @shyim 
I added your PR to the ticket you mentioned. therefore the "scheduled" label. The ticket has the status "not scheduled" 

This is a very breaking change and must of course be discussed detailed. 

Thanks for your contribution anyway!

nlubisch:
@shyim @mitelg How about the ability to configure the behaviour of the session handling?

A combobox with following options would be nice:
1) Every sub and language shop has it's own session
2) Every sub shop (with language shops) has it's own session
3) The complete shop has one session

shyim:
The solution to make the entire shop same session cannot work with different domains. So i would suggest only these two options :blush:

nlubisch:
You are right :) and to avoid a breaking change the first option is the default.

shyim:
Can be done using this plugin https://github.com/shyim/ShyimLanguageShopSession

PR #1603, opened 2018-04-03
Allow admins to show inactive articles

shyim:
Allow to show inactive articles like in shopware 4 in admin mode.

Allow to show inactive articles like in shopware 4 in admin mode.

Set a article to inactive and show the article with loggedin backend

### 6. Checklist


shyim:
Tests are now included, CI fails because of BenchmarkBundle

shyim:
Does not work with HttpCache. Will rework it later :blush:

PR #1605, opened 2018-04-04
Update jquery.emotion.js

tyurderi:
Do not remove me.$overlay when the emotion never shown me.$overlay yet

It fixes a visibility bug.

It prevents the emotion loader overlay from beeing hidden by another emotion worlds which will not be shown for the current device.

You need to create an emotion for two devices - e.g. desktop and mobile and associate it to the same category. Now when you open the emotion world in the frontend. The loading indicator/overlay is removed too early - even before the emotion world was loaded - because of the emotion world for mobile is hiding it before.

none

none

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.emotion.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.emotion.js
@@ -129,7 +129,6 @@
              * Hide the emotion world if it is not defined for the current device.
              */
             if (devices.indexOf(types[state]) === -1) {
-                me.$overlay.remove();
                 me.hideEmotion();
                 return;
             }


tyurderi:
With all due respect, but I don't really understand what this is needed for. 

Unfortunately all emotion-loaders refer to the same overlay html element, see here:
[5.4@themes/Frontend/Reponsive/frontend/_public/src/js/jquery.emotion.js#L85](https://github.com/shopware/shopware/blob/5.4/themes/Frontend/Responsive/frontend/_public/src/js/jquery.emotion.js#L85)

That's why the emotion-loaders who never placed the overlay somewhere else should not remove it, when the emotion should not be shown.

shyim:
Hey @tyurderi,

sorry for the Late response :see_no_evil: 
I have tested your change and found out a new Issue introduced with that.

* Create a Desktop only emotion for a Category with products
* Visit the page with a mobile viewport
* The emotion is still loading and the mobile user cannot access the category page

PR #1610, opened 2018-04-05
Use short description in ajax loaded listing boxes

danielwolf1:
When listings are loaded by ajax, the long description is loaded instead of the short description if the config flag useShortDescriptionInListing is set to true

The config value 'useShortDescriptionInListing' only
was implemented in initial loaded listing boxes. If listing
gets loaded with ajax, the LegacyStructConverter is used
to build the article array. This class did ignore the config value.

Make sure, category load on scroll is activated. Load category page, make sure to scroll to "page 2" - in the new loaded listing boxes, the long description will appear.

### 6. Checklist


soebbing:
With your implementation of the feature, you should also remove the then duplicate check of the config "useShortDescriptionInListing" in \sArticles::getListing() that is responsible for the replacement of the text on the initial load.

The check there also makes sure the "short description" exists before using it instead of the long one, you might want to implement that check as well in your PR.

PR #1612, opened 2018-04-05
Added banner config to english emotion demo data

shyim:
That place is clickable in german emotion, but in english its does not work 😢 . I want to buy a fancy Ibiza Beach Towel

Adds the missing banner config

### 6. Checklist


soebbing:
Thanks @shyim  👍

PR #1613, opened 2018-04-05
Clean up shopware

hlohaus:
would nice

clean up shopware :)

no

no

no

### 6. Checklist

- [x ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ c] I have read the contribution requirements and fulfil them.


soebbing:
Hi @hlohaus! Awesome to hear from you! 💙  I think there also are some backend templates associated with this controller, I think they should be removed as well.

PR #1614, opened 2018-04-09
Use find in ajax variant for footer javascript inline

shyim:
I have no idea, why i get with filter null, but with find it works 🤣 , every call about that place uses find, but that one uses filter. Find can also handle selector so i changed it to that.

Changes filter to find on javascript inline element search.

I have no idea why i have this issue 😟 and what introduces that.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.ajax-variant.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.ajax-variant.js
@@ -134,7 +134,7 @@
 
                     // Update global variables
                     window.controller = window.snippets = window.themeConfig = window.lastSeenProductsConfig = window.csrfConfig = null;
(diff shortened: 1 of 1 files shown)


larsbo:
@shyim maybe the reason could be white space/comments in the html code of the ajax response? I had problems with that line some days ago and after removing the FroshProfiler (which adds html comments and white spaces) it worked :)

shyim:
That could be, but its on production 😟  and on development the issue can't be reproduced 🙈 . .filter('script') gives me lot of scripts but the javascript inline is not listed.

larsbo:
what about `Grundeinstellungen > SEO/Router-Einstellungen > HTML-Kommentare entfernen`?

shyim:
I have already disabled all plugins which generates html comments. Also the plugin which embedds svg on detail page without success. I don't have time to debug it anymore, because with find its working by my customer...

shyim:
Okay i have also issues with find on other products. I will come with more informations and a better solution 😢

PR #1615, opened 2018-04-09
Add cache clear hint

afoeder:
Just installing a plugin via CLI without clearing the cache afterwards does not pick up
possible changes in e.g. service configurations, leading to possible confusion when installing.

Just print a hint to the console suggesting clearing the cache via `sw:cache:clear`.

Have a plugin that introduces service container modifications; `sw:plugin:active` or `sw:plugin:install --activate` it, try it and realise the service container modifications don't affect.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Commands/PluginInstallCommand.php
+++ b/engine/Shopware/Commands/PluginInstallCommand.php
@@ -95,6 +95,6 @@ EOF
 
         $pluginManager->activatePlugin($plugin);
 
-        $output->writeln(sprintf('Plugin %s has been activated successfully.', $pluginName));
+        $output->writeln(sprintf('Plugin %s has been activated successfully. Consider sw:cache:clear to enable possible behaviors that come with the plugin.', $pluginName));
     }
 }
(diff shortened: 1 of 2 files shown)


engine/Shopware/Commands/PluginInstallCommand.php
```diff
@@ -95,6 +95,6 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $pluginManager->activatePlugin($plugin);
 
-        $output->writeln(sprintf('Plugin %s has been activated successfully.', $pluginName));
+        $output->writeln(sprintf('Plugin %s has been activated successfully. Consider sw:cache:clear to enable possible behaviors that come with the plugin.', $pluginName));
```
mitelg, line 98: this message should be added in line 91. otherwise it will only be displayed if the `--activate` parameter is set

afoeder: this was intended because only installing without activating it won't expect to imply changes that need cache clearing no? (not sure though)

mitelg: mmmmh :thinking: ok, yes your are right :blush: :+1:

mitelg:
merged with b8b529516eecb92ce68114a5b10e26174f5278c7
thanks for your contribution @afoeder :+1: :tada: :blue_heart:

PR #1616, opened 2018-04-09
Fixed missing attributes in sORDER mail

FiveDigital:
Without this fix the variables `{$billingaddress.attributes}` and `{$shippingaddress.attributes}` are always `null` because `html_entity_decode([])` returns `null`.

Recursively apply `html_entity_decode()` to the passed data.

Place an order in the frontend. After that, look at the variables for the sORDER mail.

-

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sOrder.php
+++ b/engine/Shopware/Core/sOrder.php
@@ -1918,15 +1918,9 @@ EOT;
      */
     private function getUserDataForMail($userData)
     {
-        foreach ($userData['billingaddress'] as $key => $value) {
-            $userData['billingaddress'][$key] = html_entity_decode($value);
-        }
-        foreach ($userData['shippingaddress'] as $key => $value) {
-            $userData['shippingaddress'][$key] = html_entity_decode($value);
-        }
-        foreach ($userData['additional']['country'] as $key => $value) {
-            $userData['additional']['country'][$key] = html_entity_decode($value);
-        }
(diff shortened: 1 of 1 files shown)


FiveDigital:
Sorry, I had to fix it twice after some more testing. I will squash commits.

FiveDigital:
I will have a look at the failing test.

FiveDigital:
This bug is still present in 5.4 and 5.5

engine/Shopware/Core/sOrder.php
```diff
@@ -1918,22 +1918,30 @@ private function getOrderForStatusMail($orderId)
...
-        foreach ($userData['billingaddress'] as $key => $value) {
-            $userData['billingaddress'][$key] = html_entity_decode($value);
-        }
-        foreach ($userData['shippingaddress'] as $key => $value) {
-            $userData['shippingaddress'][$key] = html_entity_decode($value);
-        }
-        foreach ($userData['additional']['country'] as $key => $value) {
-            $userData['additional']['country'][$key] = html_entity_decode($value);
-        }
+        $userData['billingaddress'] = $this->htmlEntityDecodeRecursive($userData['billingaddress']);
```
shopwareBot: Wouldn't  array_walk_recursive here be better than that solution?

FiveDigital: Yes why not? If this speeds up merging I'll change the code ;)

PR #1617, opened 2018-04-09
[BUGFIX] Explicit exception when class could not be found

afoeder:
Using an incorrectly configured class as a console command leads to a confusing error message because it's wrongly assumed that the command would not extend the base command class when it actually is.

Introduce an additional check whether the given class is actually present.

* Configure a service tagged as `console.command`
* Make the service some arbitrary class name that can not be located/autoloaded
* When running anything, the compiler pass will kick in and claim the service would not extend `\Symfony\Component\Console\Command\Command` no matter that the actual class can be found.

### 6. Checklist


shyim:
Hey @afoeder,

thank you for your contribution.
we have removed our custom AddConsoleCommandPass.php in commit https://github.com/shopware/shopware/commit/35629108d3d6f92fe35bd521065dbc35b84eb629 and now use the symfony default implementation where your Bug is fixed.

afoeder:
yip fair enough, using Sf's default is better anyways.

PR #1619, opened 2018-04-09
add parallel mode to cache warmer

buddhaCode:

The cache warmer calls each seo URL to warm up the cache. Calling one URL at a time is really slow on a big shopware installation (> 10k products, > 40k variants, 3 sub shops). Firing more than on request in parallel will speed things up.

This change adds two options to the console command "sw:warm:http:cache".
* -p let you use the parallel mode
* -b let you change the batch size

Current behavior will not be touched. So no breaking changes.

I'd not include this option to the backend GUI. It should be available for pro's only, since one can easily block the whole webserver with a high batch size.

I didn't find any page description the options for this command. Maybe a hint at https://developers.shopware.com/sysadmins-guide/shopware-5-performance-for-sysadmins/#shopware-http-cache would be usefull.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Commands/WarmUpHttpCacheCommand.php
+++ b/engine/Shopware/Commands/WarmUpHttpCacheCommand.php
@@ -24,11 +24,11 @@
 
 namespace Shopware\Commands;
 
+use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Helper\ProgressBar;
 
 class WarmUpHttpCacheCommand extends ShopwareCommand
 {
--- a/engine/Shopware/Commands/WarmUpHttpCacheCommand.php
+++ b/engine/Shopware/Commands/WarmUpHttpCacheCommand.php
@@ -24,11 +24,11 @@
 
 namespace Shopware\Commands;
 
-use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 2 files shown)


engine/Shopware/Commands/WarmUpHttpCacheCommand.php
```diff
@@ -68,7 +70,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $cacheWarmer = $this->container->get('http_cache_warmer');
 
         foreach ($shopIds as $shopId) {
-            $limit = 10;
+            $limit = intval($input->getOption('batch-size'));
```
mitelg: please use the `(int)` cast instead (much faster)

engine/Shopware/Components/HttpCache/CacheWarmer.php
```diff
@@ -209,16 +211,39 @@ public function callUrls($urls, $shopId)
...
         foreach ($urls as $url) {
             $request = $this->guzzleClient->createRequest('GET', $url, $guzzleConfig);
-            try {
-                $this->guzzleClient->send($request);
-            } catch (\Exception $e) {
-                $this->logger->error(
-                    'Warm up http-cache error with shopId ' . $shopId . ' ' . $e->getMessage()
-                );
+
+            if($parallelMode) {
```
mitelg: please also check your code style. the easiest way would be to install the githook which checks the code style right before the commit. 

```
cd build
ant install-git-hooks
```

buddhaCode: actually, not sure to what line you're referring to, but i applied the hooks now.

soebbing:
Hi @buddhaCode! Your PR got merged (https://github.com/shopware/shopware/commit/98f169561b26523925d56dc20c5b53310e6e4350) and will be part of the Shopware release 5.5.0! 🎉 
Thank you for your help! 💙

PR #1622, opened 2018-04-10
Fixed image thumbnail href

shyim:
Google does not index all detail images

Adds missing href from image thumbnails (current variable is empty)

### 6. Checklist


MBDealer:
Great commit, increases SEO Visibility in Google Image Search for all Shops with more than one product image per product.

soebbing:
Thanks @shyim! 💙

PR #1633, opened 2018-04-11
Allow plugin confirmation cancel

shyim:
I have clicked on plugin uninstall accidentally (mouse is too fast and so) and see this wonderful confirm message.
image
and pressed ESC and my plugin got uninstalled :scream: 

Do not uninstall plugin on ESC (cancel) confirmation callback :worried: 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/view/plugin_helper.js
+++ b/engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/view/plugin_helper.js
@@ -331,13 +331,16 @@ Ext.define('Shopware.apps.PluginManager.view.PluginHelper', {
         var me = this;
 
         Ext.MessageBox.confirm(title, message, function (apply) {
+            if (apply === 'cancel') {
+                return;
+            }
             if (apply !== 'yes') {
+
                 me.hideLoadingMask();
-
                 if (Ext.isFunction(declineCallback)) {
-                     declineCallback();
-                }
+                    declineCallback();
 
+                }
                 return;
             }
             callback();
--- a/engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/view/plugin_helper.js
+++ b/engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/view/plugin_helper.js
@@ -328,30 +328,34 @@ Ext.define('Shopware.apps.PluginManager.view.PluginHelper', {
     },
 
     confirmMessage: function(title, message, callback, declineCallback) {
-        var me = this,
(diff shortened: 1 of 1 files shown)


Mardl:
@shyim Luck. If you had pressed return, then the data would have been deleted. I think the DefaultButton should always set to "no" on delete or uninstall operations rather to "yes". But only my opinion

larsbo:
There should also be a „cancel“ button. Not everybody would try to press the (undocumented) escape key ;-)

shyim:
I tweaked the window little bit. 
image

engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/view/plugin_helper.js
```diff
@@ -328,20 +328,30 @@ Ext.define('Shopware.apps.PluginManager.view.PluginHelper', {
     },
 
     confirmMessage: function(title, message, callback, declineCallback) {
-        var me = this;
+        var me = this,
+            cfg = {
```
mitelg: It would be nice, if you put the config in an own function, so it could easily be extended

t2oh4e: And I would recommend to not use `cfg` as variable name.... wth does that mean? :p

mitelg: @t2oh4e
dzoeschmaster

shyim: config

t2oh4e: then use config. I would even specify which config it is. `messageBoxConfig` or something like that

shyim: Hmm okay, i would left the extjs feeling http://docs.sencha.com/extjs/4.1.0/source/MessageBox.html#Ext-window-MessageBox-method-confirm. I change it later

mitelg: "left the extjs feeling" lol :joy:

soebbing:
Sorry @shyim, we'll close this PR as this creates some inconsistencies in the interface and makes the popup hard to understand. Feel free to create a new PR, adding a YESNOCANCEL in every case of uninstalling a plugin and improving the descriptive texts.

PR #1635, opened 2018-04-12
Use getRawBody() instead php://input

shyim:
I am working currently on a shopware php-ppm compability, php-ppm does not set php://input

So i changed it to $request->getRawBody() which can be manipulated with right data. (Example usage: https://github.com/shyim/shopware-ppm/commit/6202a67222dab31e45689ff45251d6b4faba4d89#diff-c23da96dc8986a4cee89980f3aad30e0R80)

**Addional idea for a improvement:**
A addional setter to Enlight Request something like setRawBody, to set the content from Symfony request at that place https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Kernel.php#L250
If i get a go it would implement that too in this pull request :smile: 

Have a JSON in Symfony Request getContent, shopware does not transform that in Enlight request

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Controller/Plugins/JsonRequest/Bootstrap.php
+++ b/engine/Library/Enlight/Controller/Plugins/JsonRequest/Bootstrap.php
@@ -85,7 +85,7 @@ class Enlight_Controller_Plugins_JsonRequest_Bootstrap extends Enlight_Plugin_Bo
             $this->parseInput === true
             && ($contentType = $request->getHeader('Content-Type')) !== false
             && strpos($contentType, 'application/json') === 0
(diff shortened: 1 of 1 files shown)


buddhaCode:
php-ppm compatibility would be absolutely AWSOME!

shyim:
@buddhaCode feel free to contribute https://github.com/shyim/shopware-ppm 😄

buddhaCode:
@shyim I would love to. But i guess my manager wouldn't let me spare some time. He would love the speed, but i think it's not realistic to use this in production. We have to many encrypted 3rd party plugins. I guess we will never be able to fix their memory leaks.

PR #1639, opened 2018-04-19
SW-21637 - Only search active products

fmaasde:
Atm inactive products are searched as well, but will not be displayed in the frontend.

This change limits the search result to active prodcuts

-

https://issues.shopware.com/issues/SW-21637

-

### 6. Checklist


philipreinken:
Since inactive products are filtered by the `ProductNumberSearch` already, we could finally integrate this change - thank you for your contribution @fmaasde ! 🥳

PR #1640, opened 2018-04-21
fix fallback translation for config store options element

ahmad-saad:
without this change the option label translation will not work and return always the first translation in the array

change the fallback lang string from 'en' to 'en_GB' , because the store option label user the full lang str not the short one.

just check the example hier https://forum.shopware.com/discussion/53062/mehrsprachigkeit-bei-select-options-in-config-xml

no changes

### 6. Checklist


mitelg:
hey @ahmad-saad 
there is already a PR for that issue: https://github.com/shopware/shopware/pull/1289

PR #1641, opened 2018-04-22
Collect width and height for SVG

xf-:
Shopping worlds and other components need width and height for correct calculations like image mapping.

Add to supported type image also type vector with extension svg. Svg is most common supported vector format in browsers. Svg is XML based and has different ways to add width and height. Directly width and height attribute or viewbox (second and third value).

Add a shopping world banner and try to add an SVG with a image map.

Is closed

SVG support is improved for example in shopping world. This allows modern light weight vector images for highDPI displays with less traffic and no need thumbnails.

### 6. Checklist

- [ ] I have written tests and verified that they fail without my change -
-> No existing tests for migrateMeta

### DIFF
--- a/engine/Shopware/Models/Media/MediaSubscriber.php
+++ b/engine/Shopware/Models/Media/MediaSubscriber.php
@@ -93,32 +93,62 @@ class MediaSubscriber implements EventSubscriber
     {
         $media = $eventArgs->getEntity();
 
-        if (!($media instanceof Media)) {
-            return;
-        }
-
-        if ($media->getType() !== Media::TYPE_IMAGE) {
+        if (!$this->isFormatSupported($media)) {
             return;
         }
 
         $mediaService = $this->container->get('shopware_media.media_service');
 
         if ((!$media->getHeight() || !$media->getWidth()) && $mediaService->has($media->getPath())) {
-            list($width, $height) = getimagesizefromstring($mediaService->read($media->getPath()));
 
-            if ($media->getId()) {
-                $eventArgs->getEntityManager()->getConnection()->executeUpdate(
-                    'UPDATE s_media SET width = :width, height = :height WHERE id = :id',
-                    [
-                        ':width' => $width,
-                        ':height' => $height,
-                        ':id' => $media->getId(),
-                    ]
-                );
+            switch ($media->getType()) {
+                case Media::TYPE_IMAGE:
+                    list($width, $height) = getimagesizefromstring($mediaService->read($media->getPath()));
+                    break;
+
+                case Media::TYPE_VECTOR:
+                    if ($media->getExtension() === 'svg') {
+                        $xml = simplexml_load_string($mediaService->read($media->getPath()));
+                        $attr = $xml->attributes();
+
+                        if ($attr->width > 0 && $attr->height > 0) {
+                            $width = (int)$attr->width;
+                            $height = (int)$attr->height;
+                        } elseif ($attr->viewBox && count($size = explode(' ', $attr->viewBox)) === 4) {
+                            $width = (int)$size[2];
+                            $height = (int)$size[3];
+                        }
+                    }
             }
 
-            $media->setWidth($width);
-            $media->setHeight($height);
+            if (!empty($height) && !empty($width)) {
+                if ($media->getId()) {
+                    $eventArgs->getEntityManager()->getConnection()->executeUpdate(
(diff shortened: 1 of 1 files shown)


engine/Shopware/Models/Media/MediaSubscriber.php
```diff
@@ -97,28 +97,41 @@ private function migrateMeta(LifecycleEventArgs $eventArgs)
             return;
         }
 
-        if ($media->getType() !== Media::TYPE_IMAGE) {
```
shopwareBot: Please reimplement this early return and add the vector type here as well

engine/Shopware/Models/Media/MediaSubscriber.php
```diff
@@ -97,28 +97,41 @@ private function migrateMeta(LifecycleEventArgs $eventArgs)
...
-                $eventArgs->getEntityManager()->getConnection()->executeUpdate(
-                    'UPDATE s_media SET width = :width, height = :height WHERE id = :id',
-                    [
-                        ':width' => $width,
-                        ':height' => $height,
-                        ':id' => $media->getId(),
-                    ]
-                );
+
+            if ($media->getType() === Media::TYPE_IMAGE) {
```
shopwareBot: Please change this to a switch statement with the TYPE_IMAGE as the default instead of if / elseif construct

engine/Shopware/Models/Media/MediaSubscriber.php
xf-: I did that because only SVG is supported and needs more conditions. And if videos supported in future you need to extend the `if` again and maybe check if a PHP module is present.

without avoids duplicate checks and usdes correct type or passes the code. No expensive compares after in switch and if height/width is filled.

```php
// You want
if ($media->getType() !== Media::TYPE_IMAGE && !($media->getType() === Media::TYPE_VECTOR && $media->getExtension() === 'svg')) {
```

soebbing: I understand your reasoning, maybe we could work around it by abstracting the condition in a function.
```
if (!$this->isFormatSupported($media)) {
    return;
}
```
This function then could contain the ugly condition, nicely formatted.

xf-: i don't understand why do you want an return? This way in case you have an image you need more conditions evaluated as with switch case and height/width at end. Also both are really simple.

engine/Shopware/Models/Media/MediaSubscriber.php
```diff
@@ -93,32 +93,62 @@ private function migrateMeta(LifecycleEventArgs $eventArgs)
...
         }
     }
+
+    /**
+     * Test file for instance Media and has supported types
+     *
+     * @param Media $media
+     * @return bool
+     */
+    private function isFormatSupported($media)
```
soebbing: You can add the type to the function signature

hlohaus: now scrutinizer are falling :)

engine/Shopware/Models/Media/MediaSubscriber.php
```diff
@@ -93,32 +93,62 @@ private function migrateMeta(LifecycleEventArgs $eventArgs)
...
         }
     }
+
+    /**
+     * Test file for instance Media and has supported types
+     *
+     * @param object $media
+     * @return bool
+     */
+    private function isFormatSupported($media)
```
xf-, line 148: @soebbing i changed the doc-header it to object, because getEntity returns an object and instance of is in this function.

shopwareBot: There'll be a Media object anyway. Can you please change the typehint to Media?

oktupol: This is not true. The event listener listens to prePersist-Events of all entities.

mitelg: @shopwareBot there was an early return which is now part of this method

xf-: I thought one complete check is easiest with early return.

xf-:
@soebbing any chance for 5.5?

soebbing:
@xf- I'd like to do that, but it will have to be after 5.5.0, sorry :(

PR #1644, opened 2018-04-24
Add ProductNumberSearch from SearchBundleDBAL as own service

aragon999:
Add `Shopware\Bundle\SearchBundleDBAL\ProductNumberSearch` as own service to the service container to access it, even if the SearchBundleES is activated. 

Add  `Shopware\Bundle\SearchBundleDBAL\ProductNumberSearch` as `shopware_searchdbal.product_number_search` to the DI container.

Request service `shopware_search.product_number_search` when you have the ElasticSearch activated 

-

-

### 6. Checklist

### DIFF
--- a/UPGRADE-5.4.md
+++ b/UPGRADE-5.4.md
@@ -13,6 +13,7 @@ This changelog references changes done in Shopware 5.4 patch versions.
   * `Shopware_Controllers_Widgets_Listing_fetchPagination_preFetch`
 * Added a new category for the basic setting containing privacy options
 * Added possibility to enable/disable data protection information texts
+* Registed class `Shopware\Bundle\SearchBundleDBAL\ProductNumberSearch` additionally as service `shopware_searchdbal.product_number_search`
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


soebbing:
I see two issues with this solution:
- Anyone can do this in his/her own plugin if necessary
- You don't have all relevant decorations you might want to have

I think a better solution would be to work with an alias.
- We create two services, e.g. `product_number_search.dbal` and `product_number_search.es`
- We then make the current service `shopware_search.product_number_search` an alias which just points to the service being used

That way the `dbal` and `es` service both still are accessible directly while the default service can be swapped out. The only question would be if it is possible to decorate an alias - I couldn't find a direct answer to that. We would need some tests to make sure everything works as intended.

aragon999:
Yes, as discussed yesterday, these are real issues, I like the setup you are proposing. I will try to find out the behavior when an alias is decorated. 
I also like the idea, if it works, we will have a mechanism only to decorate the `DBAL` product search or the `ES` product search, and do not have to create a switch in the decoration process.

aragon999:
Yes, I had no time yet to create an working test example. If it helps, this pull request can be closed and I can create a new pull request, when I am finished?

soebbing:
I don't think that is necessary. As long as you say you're still planning on looking into the topic, that is fine with us! :)

PR #1645, opened 2018-04-24
Allow attributes be readonly in backend

pascalheidmann-bedarf:
Sometimes you want to add a attribute that should not be editable for your users but still visible in backend. This allows to make that attribute be readonly.

Adding a new attribute configuration option that marks that attribute as readonly. This setting is then used to disable the input field in extjs. Therefore changes by code are of course still possible

New option "readonly" for attribute cru(d).

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/AttributeBundle/Service/ConfigurationStruct.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Service/ConfigurationStruct.php
@@ -234,7 +234,7 @@ class ConfigurationStruct implements \JsonSerializable
     /**
      * @return bool
      */
-    public function readonly()
+    public function isReadonly()
     {
         return $this->readonly;
     }
--- a/engine/Shopware/Bundle/AttributeBundle/Service/ConfigurationStruct.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Service/ConfigurationStruct.php
@@ -60,6 +60,10 @@ class ConfigurationStruct implements \JsonSerializable
      * @var bool
      */
     private $displayInBackend = false;
+    /**
+     * @var bool
+     */
+    private $readonly = false;
 
     /**
      * @var bool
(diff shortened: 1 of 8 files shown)


engine/Shopware/Bundle/AttributeBundle/Service/ConfigurationStruct.php
```diff
@@ -227,6 +231,22 @@ public function setDisplayInBackend($displayInBackend)
         $this->displayInBackend = $displayInBackend;
     }
 
+    /**
+     * @return bool
+     */
+    public function readonly()
```
shyim: Should the name of the function not isReadOnly?

pascalheidmann-bedarf: You are right. Copy/pasting displayInBackend properties ;)

buddhaCode:
Love your PR, @pascalheidman-bedarf. I had the same idea and have the same thing in one of my stashes :)

FiveDigital:
Perhaps this can be merged into 5.5 if @pascalheidman-bedarf changes the target branch?

pascalheidmann-bedarf:
@shopwareBot have fixed the row editor (but forgot to sync latest changes that use `FQN::class` instead of classnames directly as strings).
Will look into the HTML editor later but it seems this is a shortcoming in the `Shopware.form.field.TinyMCE` class I have no clue about yet ;)

engine/Shopware/Components/MultiEdit/Resource/Product/DqlHelper.php
```diff
@@ -1027,4 +1035,23 @@ private function showColumnInGrid($column)
 
         return $show;
     }
+
+    private function loadReadOnlyAttribute(array $fields) {
```
shyim, line 1039: PHP Docs and the code style is wrong :wink:

shyim:
Tinymce can be easily fixed with a if condition here https://github.com/shopware/shopware/blob/c75f8af607857acafe07c0374bed59c37363828b/themes/Backend/ExtJs/backend/base/attribute/field_handler/Shopware.attribute.HtmlFieldHandler.js#L39

if attribute has disabled true set field.readOnly = true; (https://github.com/shopware/shopware/blob/5.5/themes/Backend/ExtJs/backend/base/component/Shopware.form.field.TinyMCE.js#L333)

PR #1648, opened 2018-04-25
Emit events during plugin lifecycle

fivetide:
Currently it is very hard to hook into plugin installation process in order to provide cross-plugin-functionality. 
In addition, further plugin-related functionality could be made self-contained (i am looking at you, PluginInstaller).

During plugin lifecycle changes, events are emitted indicating these. 

No issue

none

Maybe
https://github.com/shopware/devdocs/blob/master/source/developers-guide/event-guide/index.md
or in a seperate section.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
@@ -25,6 +25,18 @@
 namespace Shopware\Bundle\PluginInstallerBundle\Service;
 
 use Doctrine\DBAL\Connection;
+use Enlight_Event_EventManager;
+use Shopware\Bundle\PluginInstallerBundle\Events\PluginEvent;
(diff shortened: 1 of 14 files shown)


engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
```diff
@@ -135,6 +155,8 @@ public function installPlugin(Plugin $plugin)
 
             $pluginBootstrap->install($context);
 
+            $this->events->notify(PluginEvent::POST_INSTALL, new PrePluginInstallEvent($context, $pluginBootstrap));
```
mitelg: this should be a PostPluginInstallEvent, shouldn't it?

fivetide: You're absolutely right

PR #1650, opened 2018-04-26
Make search in orders hookable

vanWittlaer:
The searchInOrder function controlls the selection of orders in the backend order listing pane. Making it hookable will facilitate customization of the search without affecting larger parts of the order repository class. There is a customer request to be able to search for article ordernumbers in the backend orders list.

Just changes visibility of searchInOrders from private to protected. 

Method searchInOrders not hookable.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Order/Repository.php
+++ b/engine/Shopware/Models/Order/Repository.php
@@ -662,6 +662,30 @@ class Repository extends ModelRepository
      *
      * @return int[]
      */
+    protected function searchInOrders($term)
+    {
+        $query = $this->getEntityManager()->getConnection()->createQueryBuilder();
+        $query->select('orders.id');
+        $query->from('s_order', 'orders');
+        $builder = Shopware()->Container()->get('shopware.model.search_builder');
+        $builder->addSearchTerm($query, $term, [
+            'orders.ordernumber^3',
+            'orders.transactionID^1',
+            'orders.comment^0.2',
+            'orders.customercomment^0.2',
+            'orders.internalcomment^0.2',
+        ]);
+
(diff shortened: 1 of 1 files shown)


vanWittlaer:
Sorry für die uneccessary commits - räum ich noch auf.

engine/Shopware/Models/Order/Repository.php
```diff
@@ -756,7 +756,7 @@ private function searchAddressTable($term, $table, array $excludedOrderIds = [])
      *
      * @return int[]
      */
-    private function searchInOrders($term)
+    protected function searchInOrders($term)
```
shopwareBot: Can you move the method to be above the other private methods?

vanWittlaer:
OK, moved now protected function searchInOrders in front of other private functions.

engine/Shopware/Models/Order/Repository.php
vanWittlaer: Yep, changed.

mitelg:
@vanWittlaer @shopwareBot wouldn't it make sense to splitt up the method in one which returns the `QueryBuilder` object, which would be hookable, and another one which returns the actual result? that's how we do it in other places.

vanWittlaer:
Hm, yes in general - but then searchInOrders makes use of the Shopware\Model\SearchBuilder, whilst the other query builders use the em query builder ... I'd say searchInOrders is such a small and self-contained routine that I'd just leave it as is. If you still want to split let me know, it is possible of course.

vanWittlaer:
@shopwareBot Muss ich da selbst was machen um das incomplete zurückzusetzen??

soebbing:
Hi @vanWittlaer, thank you very much for your contribution! 💙

PR #1651, opened 2018-04-26
Fix price rounding to prevent errors in PriceCount

screeny05:
This change ensures, that when retrieving the distinct PriceCount of an article the prices will always be rounded. This prevents comparision errors when using prices with more than 2 decimals.

This way a Shop can use more than two digits of precision for prices while still ensuring, that the "starting at" badge for prices works correctly.

Our current workaround is to activate the `calculateCheapestPriceWithMinPurchase` config, even though this is not the most optimal solution.

Modify `ListProductGateway::getPriceCountQuery()` to add a `ROUND` to the query.

1. Create a new article with variants
2. Edit the prices in the database to have small differences, e.g.:
    Article A: 50.378151260504
    Article B: 50.378151260503
3. Visit a Category containing the article.
4. The Article will show a "starting at" next to the price.

-

-

### 6. Checklist


screeny05:
I thought i've already done that. I've signed the CLA again, but i'm not sure if it was the correct link.

PR #1653, opened 2018-04-27
Added possibility to set sTarget for Detail/ratingAction

SvenHerrmann:
I'd like to give a user the possibility to rate an already bought article in his account. After rating he should be routed back to the account and not to the article's detail page.

If sTarget and sTargetAction are set, the user is forwarded.
If not, it works as before.

### 6. Checklist


soebbing:
Hi @SvenHerrmann, thank you very much for your contribution! 💙

PR #1656, opened 2018-05-07
change docID column type from integer to varchar

nicoanders:
The change is needed to provide compatibility for customized number ranges and the resulting document ids. Other identifiers, like the column `ordernumber` in table `s_order` or the column `customernumber` in the table `s_user`, has already the right datatype for the requirement.

It makes the column `docID` compatible for letters. So a docID like 'INV_12345' is possible.
The resulting filename will be 'INV_12345.pdf'.

Create a Document-ID (bid) with letters in. This leads to a "0" value for the column `docID` in the table `s_order_documents`.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1227-change-minimum-order-surcharge.php
+++ b/_sql/migrations/1227-change-minimum-order-surcharge.php
@@ -21,7 +21,7 @@
  * trademark license. Therefore any rights, title and interest in
  * our trademarks remain entirely with us.
  */
-class Migrations_Migration1226 extends Shopware\Components\Migrations\AbstractMigration
+class Migrations_Migration1227 extends Shopware\Components\Migrations\AbstractMigration
 {
     public function up($modus)
     {


_sql/install/latest.sql
```diff
@@ -5523,7 +5523,7 @@ CREATE TABLE IF NOT EXISTS `s_order_documents` (
   `userID` int(11) NOT NULL,
   `orderID` int(11) unsigned NOT NULL,
   `amount` double NOT NULL,
-  `docID` int(11) NOT NULL,
+  `docID` varchar(30) COLLATE utf8_unicode_ci NOT NULL,
```
shopwareBot: Please undo this change and add a migration instead.

nicoanders: @shopwareBot Done

nicoanders:
Hi @shopwareBot,
it´s failing because of this [part of the test](https://github.com/shopware/shopware/blob/55447d7658de7791cc783ce2b893ec42ddea73be/tests/Unit/Components/BasketSignature/BasketSignatureCreatorTest.php#L380-L387).
Because of this, the index 'sCurrencyId' is missing in the called [function](https://github.com/shopware/shopware/blob/d7481242e8198853a5285ee67e288dd33485e697/engine/Shopware/Components/BasketSignature/BasketSignatureGenerator.php#L52).
But its not being caused by my PR, it goes with the commit e6dd472.

PR #1658, opened 2018-05-15
SW-21760 - Backend: Shipping Costs countries to use position column

shoobyban:
Shipping Costs Countries lists don't use position field, with full country list it's nearly impossible to use the controls as the list will be unordered.

Sets sorting for the lists in the Backend ExtJS to the position field in countries table.

From Backend -> Configuration -> Shipping Costs choose a shipping cost and go to countries tab, when having not only the standard set of countries it's apparent that it's not ordered by position field.

### 6. Checklist
Well, I had no time for all of these (just about to launch Mywalit, sorry)


soebbing:
Hi @shoobyban, thank you for your contribution! 🎉

PR #1659, opened 2018-05-17
Fix TemplateMails createMail for shops without templates

hanneswernery:
With the recent changes in the `TemplateMail` component (https://github.com/shopware/shopware/commit/ed2e80e974470eb3451f1a9afd9d3dd1e34d2c14#diff-e8b4c5662b6e1a932d05d94187a75b70) a theme is added to the mail context that is built from the shop's template. But this theme addition will crash if the given shop has no own template (i.e. it is a language sub shop).
See https://github.com/shopware/shopware/commit/ed2e80e974470eb3451f1a9afd9d3dd1e34d2c14#diff-e8b4c5662b6e1a932d05d94187a75b70R167 

This PR adds checks before adding a theme to the mail context: If the passed shop has no own template, the shop's main shop template is used. If even that shop has no template, no theme is added to the context at all (which was the behavior before https://github.com/shopware/shopware/commit/ed2e80e974470eb3451f1a9afd9d3dd1e34d2c14).

Create a sub shop that has no own template. Call `createMail` of the `TemplateMail` component with this shop as parameter. The mail creation will fail when trying to build the shop theme.

None.

None.

### 6. Checklist

### DIFF
--- a/UPGRADE-5.4.md
+++ b/UPGRADE-5.4.md
@@ -11,7 +11,7 @@ This changelog references changes done in Shopware 5.4 patch versions.
 * Changed Tinymce editor to resolve placeholder images on initialization
 * Changed product notification to match the documented feature
 * Changed AJAX search to respect the basic setting for direct number searches and show the same results as the normal search 
-* Changed TemplateMail to work without shop context
+* Changed TemplateMail to work without shop context or shops without templates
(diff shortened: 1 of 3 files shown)


mitelg:
thanks for your contribution @hanneswernery :+1: :tada: :blue_heart:

PR #1662, opened 2018-05-22
Add Terminate Event to Shopware Kernel

fionera:
Symfony brings a Terminate Kernel Event so you can execute expensive post-response code. This Pull-Request adds it to the Shopware Kernel

Adding the 

Adding the TerminableInterface to the Shopware Kernel and notify the listener

### 6. Checklist


engine/Shopware/Kernel.php
```diff
@@ -230,6 +233,17 @@ public function handle(SymfonyRequest $request, $type = HttpKernelInterface::MAS
         return $response;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function terminate(Request $request, Response $response)
```
shopwareBot: Please change this to "SymfonyRequest" and "SymfonyResponse" and remove the duplicated imports.

fionera:
@shyim Implemented it in 5.5 <3

PR #1663, opened 2018-05-22
Inherit highDpi and quality settings from Album

EtienneBruines:
People create sub-albums all the time, without realizing their settings with respect to highDpi thumbnails are being lost. Effectively disabling highDpi thumbnails for everyone ever using sub-albums, with the exception of the few that check the settings manually. 

Not only 'thumbnail size' and 'should we create a thumbnail' are being inherited upon creation from the parent, but also the 'thumbnail quality', 'should we create a highDpi thumbnail' and the 'highDpi thumbnail quality'. 

- configure something for some album
- create a subalbum
- notice the highDpi-settings are not inherited from the parent.

N / A

N / A

### 6. Checklist


EtienneBruines:
The current 5.5 build is broken; unrelated to this PR.

PR #1670, opened 2018-06-06
bugfix and perfomance improve

dasistweb-gmbh:
The main problem was that media files, that where supposed to get collected by the garbage collector, where NOT collected properly.
Also the debugging process was pretty slow due to the missing sql index for the mediaId column inside of the temp table.

- added a index to the temporary media_used table for increasing performance
- use str_array instead of a int array at the select by path query

Execute the `sw:media:cleanup` command on a strictly configured sql environment.
Depending on your configuration the cleanup will be EXTREMLY slow or not working at all because the files are not getting indexed correctly.

- 

-

### 6. Checklist

- [ / ] I have written tests and verified that they fail without my change
- [ / ] I have squashed any insignificant commits

### DIFF
--- a/engine/Shopware/Bundle/MediaBundle/GarbageCollector.php
+++ b/engine/Shopware/Bundle/MediaBundle/GarbageCollector.php
@@ -110,7 +110,7 @@ class GarbageCollector
      */
     private function createTempTable()
     {
-        $this->connection->exec('CREATE TEMPORARY TABLE IF NOT EXISTS s_media_used (id int auto_increment, mediaId int NOT NULL, PRIMARY KEY pkid (id))');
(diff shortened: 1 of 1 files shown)


Gugiman:
Similar to:
https://github.com/shopware/shopware/pull/1290

shyim:
Is that not a duplicate PR of #1290 ?

martinweinmayr:
@gugiman
Thats correct. But in this case speed is only a nice to have addition.
The real problem is that the query does not work properly on some servers with int_array, in fact that the params are strings ...

In addition i don‘t underdtand why shopware has not talen the quick win from #1290 since 8 months ...

soebbing:
Hi @dasistweb-gmbh, this issue has been fixed with #1290. Thank you for your contribution anyway! 💙

PR #1680, opened 2018-06-18
Add event to filter generated HTML of \Shopware_Components_Document

windaishi:
Currently the generated HTML code for document cannot be manipulated before it is put in the PDF renderer. That's why some plugins use replace-hooks for the method `\Shopware_Components_Document::render` to manipulate the HTML.
Replace hooks should be avoided in any case because multiple replace hooks of the same method lead to unexpected results in Shopware (doubled executions).

Adds a filter event `Shopware_Components_Document_Render_FilterHtml` to allow manipulation of the generated document HTML before rendering it to PDF.

N/A

N/A

N/A

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -249,10 +249,16 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
             $this->_template->setTemplateDir($inheritance);
         }
 
-        $data = $this->_template->fetch('documents/' . $this->_document['template'], $this->_view);
+        $html = $this->_template->fetch('documents/' . $this->_document['template'], $this->_view);
+
+        /** @var \Enlight_Event_EventManager $eventManager */
+        $eventManager = Shopware()->Container()->get('events');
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Document.php
```diff
@@ -249,10 +249,16 @@ public function render($_renderer = '')
             $this->_template->setTemplateDir($inheritance);
         }
 
-        $data = $this->_template->fetch('documents/' . $this->_document['template'], $this->_view);
+        $html = $this->_template->fetch('documents/' . $this->_document['template'], $this->_view);
+
+        /** @var \Enlight_Event_EventManager $eventManager */
+        $eventManager = Shopware()->Container()->get('events');
+        $html = $eventManager->filter('Shopware_Components_Document_Render_FilterHtml', $html, [
+            'document' => $this,
```
shopwareBot: Can you rename this to 'subject' for consistency reasons?

windaishi: No problem :)

windaishi:
Fixed @shopwareBot

soebbing:
Hi @windaishi, thank you again for your contribution! 🎉

PR #1682, opened 2018-06-20
Add support for specifying Symfony Trusted Header options in config.php

mintopia:

When Shopware is deployed to some environments behind managed proxy servers such as OpenShift of Kubernetes Ingress servers, the application fails to run due to conflicting Forwarded IP headers. 

```
PHP Fatal error: Uncaught Symfony\\Component\\HttpFoundation\\Exception\\ConflictingHeadersException: The request has both a trusted "FORWARDED" header and a trusted "X_FORWARDED_PORT" header, conflicting with each other. You should either configure your proxy to remove one of them, or configure your project to distrust the offending one.
```

In many environments, it is not possible to change the configuration of the proxy, so instead it is requird to pass the correct trusted header name options to Symfony.

Working in a similar manner to the existing functionality to specify trusted proxy IPs, in config.php you can specify the keys for headers and their values, these will then be passed through to Symfony's Request::setTrustedHeaderName() function.

1. Deploy Shopware to an environment behind a proxy server that passes through X-Forwarded-For, X-Forwarded-Port and X-Forwarded headers. An example would be OpenShift Origin.
2. Finish the installation wizard
3. Try and visit the backend or frontend of Shopware. You will get a blank page/error and a message in the error log as above.

https://issues.shopware.com/issues/SW-17908

There is some basic documentation for trusted proxy IPs on the page for configuring Varnish for Shopware, it could be added to that page.

### 6. Checklist

### DIFF
--- a/UPGRADE-5.4.md
+++ b/UPGRADE-5.4.md
@@ -9,6 +9,7 @@ This changelog references changes done in Shopware 5.4 patch versions.
 ### Changes
 
 * Changed behaviour of the notification plugin, so that it will only notify the customer if the stock is at least as high as the minimal purchase amount
+* Added support in config.php to specify TrustedHeader options for Symfony.
 
 ## 5.4.5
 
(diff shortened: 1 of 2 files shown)


mintopia:
Travis CI tests failed, will fix and re-submit

mintopia:
See #1683

PR #1683, opened 2018-06-20
Configurable trusted headers

mintopia:

When Shopware is deployed to some environments behind managed proxy servers such as OpenShift of Kubernetes Ingress servers, the application fails to run due to conflicting Forwarded IP headers. 

```
PHP Fatal error: Uncaught Symfony\\Component\\HttpFoundation\\Exception\\ConflictingHeadersException: The request has both a trusted "FORWARDED" header and a trusted "X_FORWARDED_PORT" header, conflicting with each other. You should either configure your proxy to remove one of them, or configure your project to distrust the offending one.
```

In many environments, it is not possible to change the configuration of the proxy, so instead it is required to pass the correct trusted header name options to Symfony.

Working in a similar manner to the existing functionality to specify trusted proxy IPs, in config.php you can specify the keys for headers and their values, these will then be passed through to Symfony's Request::setTrustedHeaderName() function.

1. Deploy Shopware to an environment behind a proxy server that passes through X-Forwarded-For, X-Forwarded-Port and X-Forwarded headers. An example would be OpenShift Origin.
2. Finish the installation wizard
3. Try and visit the backend or frontend of Shopware. You will get a blank page/error and a message in the error log as above.

https://issues.shopware.com/issues/SW-17908

There is some basic documentation for trusted proxy IPs on the page for configuring Varnish for Shopware, it could be added to that page.

### 6. Checklist


mintopia:
Resolved the conflict in the changelog.

soebbing:
Hi @mintopia, thank you for your contribution! 

We're currently trying to integrate your PR into the upcoming Shopware 5.5 release. Since 5.5 uses Symfony 3.4, I stumbled upon your change since it uses the `setTrustedHeaderName` API, which is deprecated since Symfony 3.3.

I created a little Gist with an updated change: https://gist.github.com/soebbing/cc9e3dca65727dfe0603410eeed72cfc#file-kernel-php-L174 - It uses the bitfield parameter in the `setTrustedProxies` method which is described in more depth here: https://symfony.com/doc/current/deployment/proxies.html The parameter would be initialized in `Default.php` as `0` and could be overridden using the constants from the `Request` class in the `config.php`.

Would that change help cover your use case as well? Thank you very much in advance for your feedback!

mintopia:
Hi @soebbing,

Yes, adding support for setting that second parameter as per your gist looks like it will resolve the issue as well.

Setting the the value in config.php would be ideal for for us as we can then set it correctly at deploy-time.

Thanks,
Jessica

soebbing:
Hi @mintopia, thank you for your contribution and your feedback! 💙

soebbing:
Your PR was merged in 0d34fa6cb1230cd14212cc2535d0f51b8eb95de9 and modified as discussed. 👍

PR #1684, opened 2018-06-21
Add Filter event to backend variant generation

larsbo:

Without the event it is not easy to change variant details during backend generation programmatically.

This PR adds a new filter event `Shopware_Controllers_Article_CreateConfiguratorVariants_FilterData` which allows the modification of article variants during backend generation.

### 6. Checklist


larsbo:
@shopwareBot The use case is to set a unique value to each generated variant, e.g. in our case an unique EAN number.

larsbo:
sounds legit 😉 I'll update my PR!

larsbo:
@shopwareBot Using the `$data` array instead of the model has the drawback that there is no possibility to check if the variant is newly generated or not. 
When using the `Detail` model you can check if `$detail->getId()` is zero but inside the `$data` array every variant got the `id` copied from the main detail. This could be solved by resetting the id inside `prepareVariantData()`. I updated my PR with that change and hope this is no breaking change 😃

larsbo:
That's also fine. I updated the PR.

larsbo:
@shopwareBot the changelog entry of the PR should be moved from 5.5.0 to 5.5.2 😉

shyim:
@larsbo We are collecting new entries / changes for UPGRADE.md and changes that all in one commit before a release happens. That save us time and makes the work with pull requests faster 😊 .

So we have it on the list to update that 😉

larsbo:
alright, makes sense 😄

PR #1685, opened 2018-06-21
Removed check for sTAXAUTOMODE

jlaute:
removes an unnecessary if clause

Just removes the call for `sTAXAUTOMODE` and the following if clause. The same stuff is already made 6 lines above
```
$taxMode = $this->config->get('sTAXAUTOMODE');
if (!empty($taxMode)) {
     $tax = $this->getMaxTax();
}
```
### 6. Checklist


soebbing:
Hi @jlaute, thank you for your contribution! 💙

PR #1687, opened 2018-06-22
Calculate NET of shipping cost with 4 digits

s-bessing:
When creating a document for an order, the tax for the shipping is calculated with the difference between invoice_shipping and invoice_shipping_net. In CH the tax is 7.7% and it result in an rounding error with a shipping cost of 2.9 CHF. The PDF is generated with a tax of 8%.

Rounding error in CH

Change the presission to 4 digits for the shipping net cost

Set the tax to 7.7% and add a shipping cost of 2.9. Place an order and create an invoice PDF.

### 6. Checklist


s-bessing:
Related with this PR: https://github.com/shopwareLabs/SwagBackendOrder/pull/17

PR #1688, opened 2018-06-25
automatically add attributes for customers created via the rest api

jochenmanz:
When a customer is created via the API, the attributes table is not populated automatically with a new record for the new customer. This leads to problems with Plugins depending on a existing attributes record for the new customer. For Example https://github.com/portrino/shopware-hybrid-auth

it adds a empty attribute record for the new customer

create a new customer via the api, try to login with the same customer via the hybridAuth Plugin

### 6. Checklist


soebbing:
Hi @jochenmanz, thank you for your contribution! 🎉

PR #1689, opened 2018-06-25
Add article details to api variant enpoint

larsbo:
The `/variants` endpoint response doesn't contain all relevant data to work effectively with the API. This PR adds some useful additions from the `article` entity to the variant list and detail queries.

The `/variants` list result will be extended by a partial join containing
- article.id
- article.taxId
- article.name
- article.active

The `/variants/[id]` detail result will be extended by a full article join.

A use case could be the query of active variants. Since both the variant and the article have an `active` state it is necessary to filter for both the variant.active and variant.article.active flag.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Variant.php
+++ b/engine/Shopware/Components/Api/Resource/Variant.php
@@ -88,7 +88,8 @@ class Variant extends Resource implements BatchInterface
         }
 
         $builder = $this->getRepository()->getVariantDetailQuery();
-        $builder->andWhere('variants.id = :variantId')
+        $builder->addSelect('article')
+                ->andWhere('variants.id = :variantId')
                 ->addOrderBy('variants.id', 'ASC')
                 ->addOrderBy('customerGroup.id', 'ASC')
                 ->addOrderBy('prices.from', 'ASC')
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Api/Resource/Variant.php
```diff
@@ -126,10 +126,11 @@ public function getList($offset = 0, $limit = 25, array $criteria = [], array $o
         /** @var \Doctrine\DBAL\Query\QueryBuilder */
         $builder = $this->getRepository()->createQueryBuilder('detail');
 
-        $builder->addSelect(['prices', 'attribute', 'customerGroup'])
+        $builder->addSelect(['prices', 'attribute', 'partial article.{id,name,active,taxId}', 'customerGroup'])
                 ->leftJoin('detail.prices', 'prices')
                 ->innerJoin('prices.customerGroup', 'customerGroup')
                 ->leftJoin('detail.attribute', 'attribute')
+                ->leftJoin('detail.article', 'article')
```
shopwareBot: Can you change this to an innerJoin? This relation should ___always___ exist.

larsbo: fixed!

engine/Shopware/Models/Article/Repository.php
```diff
@@ -2820,6 +2820,7 @@ public function getVariantDetailQuery()
         $builder->select([
             'variants',
             'attribute',
+            'article',
```
mitelg: why is this necessary?

larsbo: I thought the article select was forgotten because there is already a join on `variants.article` in the query. Or is the inner join only to ensure that an article entry exists for this variant?

Nevertheless, making some article "base data" available directly inside the variant result could halve the number of API requests when you need to know variant base data like the tax id.

shopwareBot: Hey @larsbo,

as this might be a fairly dangerous change, we'd like you to remove that line again.
Since the overlying method returns the builder, you could just call the method and adjust the builder afterwards to add your desired 'article' data without impacting everything else.

larsbo: good point, I moved the select to the variant resource.

PR #1690, opened 2018-06-26
calculateVoucherValues with tax 7.7

ener-Dev:
This creates incorrect amounts with vouchers in the shopping cart.

As a result, the tax is calculated not with 7% but with 7.7%.

Just create the tax of 7.7% and select it on a voucher. As a result, you should be able to recognize the wrong calculation with 7%.

https://issues.shopware.com/issues/SW-22036

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2161,7 +2161,7 @@ class sBasket
                     [$temporaryTax]
                 );
                 $taxRate = $getTaxRate;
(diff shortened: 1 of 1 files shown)


ener-Dev:
Hello,

what does my bug-report have to do with the shipping code? 

My error refers to the tax calculation for coupons. I could not detect a fix in the commit.

thanks for your reply!

Edit: correct commit: https://github.com/shopware/shopware/commit/1da309a7d622cbde0e30f79b2c226ac41646a5e4

thanks!

PR #1691, opened 2018-06-27
account orders unrelated to showEsd setting

dneustadt:
sorry, if there already is a pr, just to be sure... 

The showEsd config setting affects whether or not the link to both, esd downloads AND regular orders
are shown in the account sidebar. Previously it would only affect whether the link to the downloads would be shown in the account sidebar, which is the expected behavior.

Remove the condition on showEsd in the template for the account orders menu entry.

Enable showEsd, try to access previous orders in storefront account.

### 6. Checklist


mitelg:
thanks for your contribution @dneustadt :+1: :tada: :blue_heart:

PR #1693, opened 2018-06-28
Removed incorrect return type annotation

chriskonnertz:
This removes an incorrect return type annotation.

It removes the "@return array" annotation, because the method does not return anything - it does not make use of the "return" keyword.

Search for the "return" keyword in the code of the loadStoresAction() method - you will not find it so there is no return value.

None

None

### 6. Checklist

- [x] * I have written tests and verified that they fail without my change
- [x] * I have squashed any insignificant commits
- [x] * This change has comments for package types, values, functions, and non-obvious lines of code

`*` = Not relevant

### DIFF
--- a/engine/Shopware/Controllers/Backend/Article.php
+++ b/engine/Shopware/Controllers/Backend/Article.php
@@ -514,8 +514,6 @@ class Shopware_Controllers_Backend_Article extends Shopware_Controllers_Backend_
     /**
      * The loadStoresAction function is an ExtJs event listener method of the article backend module.
      * The function is used to load all required stores for the article detail page in one request.
-     *
-     * @return array
      */
     public function loadStoresAction()
     {


chriskonnertz:
Well  that's fine but don't expect me to spend time again for another pull request, if clicking on a merge button for an obviously correct pr isn't worth your time.

ndzoesch:
Hey @chriskonnertz,

I'm sorry we upset you by not merging. It takes more time for us than just pressing the merge button, we would have to import the branch to our internal git and then stick to our normal workflow, including QA and testing.

A tiny change can (and in the past did) break things that no one thought of, be it in the Shopware core or for plugin developers.

We are thankful for people like you putting time and effort into Shopware. So if in the future you need a change to the core that is beneficial for others as well we would be glad if you submitted it as a PR.

chriskonnertz:
@ndzoesch

Hello,

thank you for this explanation.

However,

> A tiny change can (and in the past did) break things

this does not seem to be a valid argument in this case, since I did not change any 
executable code. 

>  So if in the future you need a change to the core that is beneficial for others as well we would be glad if you submitted it as a PR.

Unfortunately this is no longer possible.

PR #1694, opened 2018-06-28
Adds snippet for times sign in ajax cart

jkrzefski:
There should be no text present in the template at all. All of it should be moved into snippets. Also it would be semantically correct to use the times sign instead of a normal `x`.

Extract the `x` from the quantity indicator in the ajax cart into snippets and replaces it by the `&times;` sign.

Open the ajax cart while at least one product is in your cart.

None.

None.

### 6. Checklist


soebbing:
Hi @jkrzefsk! Thank you very much for your contribution! 🎉

PR #1695, opened 2018-06-28
Add dummy listStore to "Showpare.apps.Order.view.detail.Window" if none exists

pascalheidmann-bedarf:
If you open an order by searching for it or by clicking on order history in a customer in backend, you can't send documents via mail.

Adds listStore to _Showpare.apps.Order.view.detail.Window_ by extending its controller.

1. In backend: Open a customer account with at least 1 order
2. Go to orders and open one
3. Try sending a document via mail
-> JS error

Same goes for orders that are opened via search.

### 6. Checklist


buddhaCode:
Pascal du Schufft. Warst wohl schneller ;)

PR #1699, opened 2018-07-03
Added missing type hints for Order\Document\Document::$type

JoshuaBehrens:
Auto completion is missing therefore there is developer experience missing.
We need more developer experience :blue_heart: 

Add type hints.

Type `getType` on an instance of Models\Order\Order and :sob: because the autocompletion fails.

### 4. Checklist


soebbing:
Hey @JoshuaBehrens, thanks for this nice little addition! 💙

JoshuaBehrens:
It`s the small things that annoy when they are missing :)

PR #1700, opened 2018-07-03
Extracts ESD download logic into new bundle

jkrzefski:
To have nicer code and be able to access esd-related functions from everywhere.

It moves some esd-related business logic from a controller into a dedicated bundle.

Log into an account and download a purchased esd product.

None.

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/EsdBundle/Service/Core/DownloadService.php
+++ b/engine/Shopware/Bundle/EsdBundle/Service/Core/DownloadService.php
@@ -43,7 +43,7 @@ class DownloadService implements DownloadInterface
      */
     public function getAbsoluteFilePath($file)
     {
-        return $this->shopwareRootDirectory . $this->getRelativeFilePath($this);
+        return $this->shopwareRootDirectory . $this->getRelativeFilePath($file);
     }
 
     /**
--- a/engine/Shopware/Bundle/EsdBundle/Service/Core/DownloadService.php
+++ b/engine/Shopware/Bundle/EsdBundle/Service/Core/DownloadService.php
@@ -1,0 +1,110 @@
+<?php
+
+namespace Shopware\Bundle\EsdBundle\Service\Core;
+
+use Enlight_Controller_Front;
+use Shopware\Bundle\EsdBundle\Service\DownloadInterface;
+use Shopware_Components_Config;
+
+class DownloadService implements DownloadInterface
+{
+    /**
+     * @var string
+     */
+    private $shopwareRootDirectory;
+
+    /**
+     * @var Shopware_Components_Config
+     */
+    private $config;
+
+    /**
+     * @var Enlight_Controller_Front
+     */
+    private $front;
+
+    /**
+     * @param Enlight_Controller_Front $front
+     */
+    public function __construct(
+        $shopwareRootDirectory,
+        Shopware_Components_Config $config,
+        Enlight_Controller_Front $front
+    ) {
+        $this->shopwareRootDirectory = $shopwareRootDirectory;
+        $this->config = $config;
+        $this->front = $front;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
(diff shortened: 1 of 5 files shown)


engine/Shopware/Bundle/EsdBundle/Service/Core/DownloadService.php
```diff
@@ -0,0 +1,110 @@
...
+    public function existsFile($file)
+    {
+        return file_exists($this->getAbsoluteFilePath($file));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function sendFile($file)
+    {
```
JoshuaBehrens, line 69: I'd like to see an optional parameter to give a display filename. Imagine a download of a file that has a hash as filename but you want to download an instruction manual for product 12345. One will not find the right file in the download directory afterwards.

For the customer experience!

engine/Shopware/Bundle/EsdBundle/Service/Core/DownloadService.php
```diff
@@ -0,0 +1,110 @@
...
+        $relativeFilePath = $this->getRelativeFilePath($file);
+        $absoluteFilePath = $this->getAbsoluteFilePath($file);
+        $fileName = basename($absoluteFilePath);
+
+        switch ($this->config->get('esdDownloadStrategy')) {
```
JoshuaBehrens, line 74: As this strategy chosing method depends on the environment this should be a separate service itself for downloading anything the right way for any downloads. Maybe like downloading documents in the backend although they benefit the least from performance boosted downloads this should be normalized through out :blue_heart: Shopware.

JoshuaBehrens:
I need this as a service. Please! This is a PR leading in the right direction.

For the developer experience.

engine/Shopware/Bundle/EsdBundle/Service/Core/DownloadService.php
jkrzefski: I see your point. This would add a lot of flexibility to this service. But now that I think of it, it might be even better to make two separate services: One for the download itself and one for an ESD-related use-case.

engine/Shopware/Bundle/EsdBundle/Service/Core/DownloadService.php
shyim: We have removed esdDownloadStrategy support in 5.5. See https://github.com/shopware/shopware/commit/0497631dc76c3ad1a72dc20b587a397f5f7fa0a5

JoshuaBehrens: Still no public function to say: Download this file. Mixed feelings about this.

engine/Shopware/Bundle/EsdBundle/Service/Core/DownloadService.php
```diff
@@ -0,0 +1,112 @@
+<?php
+
+namespace Shopware\Bundle\EsdBundle\Service\Core;
```
shopwareBot, line 3: A bundle for a single service would be kinda overkill.

Can you try to move your service to another existent bundle or just into Shopware/Components.

jkrzefski: @shopwareBot I was looking for a fitting bundle to implement this into, but I did not find anything. Do you have a suggestion into which bundle I could implement this into?

Also I am now considering to make a few more services for even more modularity. Maybe it will be worth its own bundle then.

I am looking into porting this pr to 5.5 branch when I find time for it.

mitelg: if you create more services related to ESD, a bundle would be fine :+1:

jkrzefski:
I have made a new PR based on branch 5.5 and with a bit more flexibility. See #1704 for details. I will close this PR then as is is now obsolete.

PR #1703, opened 2018-07-06
Allow 0 as ID in state of Order API

Isengo1989:
The stateId is necessary on create of an order. This should not be requiered. Also see the ISSUE SW-21511 I also fixed the typo that states that shipping state is missing instead of billing state

The user can register without a stateId/state, why should the API disallow it.

It checks if the stateId is 0 - if so it does not check if the state exists in the list

Create order with stateId 0 fails

https://issues.shopware.com/issues/SW-21511

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Order.php
+++ b/engine/Shopware/Components/Api/Resource/Order.php
@@ -703,12 +703,14 @@ class Order extends Resource
             ));
         }
 
-        $state = $this->getContainer()->get('models')->find(State::class, $billing['stateId']);
-        if (!$state) {
-            throw new ApiException\NotFoundException(sprintf(
-                'Shipping State by id %s not found',
-                $billing['stateId']
-            ));
+        if($billing['stateId'] != 0){
+            $state = $this->getContainer()->get('models')->find(State::class, $billing['stateId']);
+            if (!$state) {
+                throw new ApiException\NotFoundException(sprintf(
+                    'Billing State by id %s not found',
+                    $billing['stateId']
+                ));
+            }
         }
 
         $billingAddress = new Billing();
@@ -742,12 +744,14 @@ class Order extends Resource
             ));
         }
 
-        $state = $this->getContainer()->get('models')->find(State::class, $shipping['stateId']);
-        if (!$state) {
-            throw new ApiException\NotFoundException(sprintf(
-                'Shipping State by id %s not found',
-                $shipping['stateId']
-            ));
+        if($shipping['stateId'] != 0) {
+            $state = $this->getContainer()->get('models')->find(State::class, $shipping['stateId']);
+            if (!$state) {
+                throw new ApiException\NotFoundException(sprintf(
+                    'Shipping State by id %s not found',
+                    $shipping['stateId']
+                ));
+            }
         }
 
         $shippingAddress = new Shipping();


engine/Shopware/Components/Api/Resource/Order.php
```diff
@@ -703,12 +703,14 @@ private function prepareCreateAddresses(array $params, \Shopware\Models\Order\Or
             ));
         }
 
-        $state = $this->getContainer()->get('models')->find(State::class, $billing['stateId']);
-        if (!$state) {
-            throw new ApiException\NotFoundException(sprintf(
-                'Shipping State by id %s not found',
-                $billing['stateId']
-            ));
+        if($billing['stateId'] != 0){
```
shopwareBot, line 706: Please check if the array key "stateId" exists instead.

If the array key is set, then it should be loadable.

engine/Shopware/Components/Api/Resource/Order.php
```diff
@@ -742,12 +744,14 @@ private function prepareCreateAddresses(array $params, \Shopware\Models\Order\Or
             ));
         }
 
-        $state = $this->getContainer()->get('models')->find(State::class, $shipping['stateId']);
-        if (!$state) {
-            throw new ApiException\NotFoundException(sprintf(
-                'Shipping State by id %s not found',
-                $shipping['stateId']
-            ));
+        if($shipping['stateId'] != 0) {
```
shopwareBot, line 747: Same here.

engine/Shopware/Components/Api/Resource/Order.php
FiveDigital: Also there are spaces missing: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#51-if-elseif-else

Isengo1989: @shopwareBot if Array Key exists is already above. Please see the issue on the tracker.

Isengo1989: Should I create a new PR with a branch instead of a commit?

mitelg: @Isengo1989 I think your intention was to make the `stateId` optional. so maybe we should remove the throwing of the exception there. instead of changing this logic here

soebbing: The Exception should still be thrown if an id was given that couldn't be found in the database. But I think the `if($billing['stateId'] != 0){` should be changed to `if (!empty($billing['stateId'])) {`, which would allow `null`, 0 or empty strings. Also feel free to change the text in the wrong Exception message.

You don't have to create a new PR, just change the current commit and force push into your branch.

Isengo1989: @mitelg would work for me - though it would mean if someone really sets a non existing stateId he would never know.

With no exception it would just go through and set it to a non existing ID - with the != 0 it would throw the "State ID does not exist message". Also I don't know how the Backend would like it on the order or customer window.

Both ways work for me - just let me know.

soebbing: If we set the the stateId to 0 on registrations in the Frontend, I would recommend to replace the check for the array key and the throwing of the exception above with the following code:
```
        if (empty($billing['stateId'])) {
            $billing['stateId'] = 0;
        }
```
It just initializes the stateId with 0 if it doesn't contain any valid data by itself.

Please remember to add a test for this case, it should just be a simple copy/paste from e.g. `\Shopware\Tests\Functional\Components\Api\OrderTest::testCreateOrderFailsOnUnknownStateIdInDetails` by changing the invalid statusId to an unset of the variable and an assertion that everything works.

soebbing: Are you still working on this?

PR #1705, opened 2018-07-09
 Remove Composer platform fake 

hranicka:

In order to install libraries requiring PHP >= 7.0, platform fake is an obstacle.
Furthermore, it's a lie and we should rely on "require" and real running version.

Removes specific PHP version faking.

Add for example frosh/shopware-profiler (`php composer.phar require --dev frosh/shopware-profiler`).

It results into this error:

```
  Problem 1
    - This package requires php ^7.0 but your PHP version (5.6.4) does not satisfy that requirement.

```

None.

None.

### 6. Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -115,10 +115,7 @@
     },
     "config": {
         "autoloader-suffix": "Shopware",
-        "optimize-autoloader": true,
-        "platform": {
-            "php": "5.6.4"
-        }
+        "optimize-autoloader": true
     },
     "scripts": {
         "cs-check": "php-cs-fixer fix --dry-run -v",


bcremer:
@hranicka The fixed platform PHP version cannot be removed as this is the minimum PHP Version that is supported. 

During the packaging process when the install/update ZIP files are created it must be ensured, that composer does not pull any libraries that are not supported on this PHP version.

If you want more control please have a look into https://github.com/shopware/composer-project.
When pulling the shopware/shopware as a dependency the platform constraint is ignored.

Installation of additional composer packages is not supported in the "traditional" Shopware version that is mainly deployed using a release zip file.

hranicka:
Hello @bcremer , thanks for explanation.

shopware/composer-project is really welcomed step, thanks for that.

On projects where we don't have Shopware 5.4 as a composer dependency, we have no other choice than to remove the platform fake out of composer.json manually and deal with it during further upgrades, right?

Thanks.

As of Shopware can be installed as a dependency, I don't have the issue anymore. Anyway I would go away from PHP 5 constraint since its EOL is this year :)

bcremer:
> On projects where we don't have Shopware 5.4 as a composer dependency, we have no other choice than to remove the platform fake out of composer.json manually and deal with it during further upgrades, right?

Right.

> Thanks.

You're welcome.

> Anyway I would go away from PHP 5 constraint since its EOL is this year 

Absolutely. 
Nudge Nudge @soebbing.

soebbing:
Oh how much I will love changing that to 7.1 😆

hranicka:
Please see https://github.com/shopware/composer-project/pull/13

If we don't want to do changes in this repository, we should really enforce people to use PHP 7.0+ on new projects.

PR #1707, opened 2018-07-10
Update Forms.php

cm-springlane:
make the custom form attributes available to the frontend template

A user should be able to use the freetext fields they configure in the backend, in the frontend

Makes freetext fields for Forms available in the frontend template

1) add a freetext field to Forms (s_cms_support_attributes)
2) try to find the freetext field in the template variables, it's not there

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Forms.php
+++ b/engine/Shopware/Controllers/Frontend/Forms.php
@@ -240,6 +240,7 @@ class Shopware_Controllers_Frontend_Forms extends Enlight_Controller_Action
         ];
 
         return array_merge($formData, [
+            'attribute' => Shopware()->Container()->get('shopware_attribute.data_loader')->load('s_cms_support_attributes', $formId),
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Frontend/Forms.php
```diff
@@ -240,6 +240,7 @@ protected function getContent($formId)
         ];
 
         return array_merge($formData, [
+            'attribute' => Shopware()->Container()->get('shopware_attribute.data_loader')->load('s_cms_support_attributes', $formId),
```
uehler, line 243: you should use
$this->get('shopware_attribute.data_loader')
or
$this->container->get('shopware_attribute.data_loader')

mitelg:
hey @cm-springlane 
this is already fixed in SW 5.5 -> https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Controllers/Frontend/Forms.php#L242

PR #1708, opened 2018-07-10
change docID column type from integer to varchar

nicoanders:
The change is needed to provide compatibility for customized number ranges and the resulting document ids. Other identifiers, like the column `ordernumber` in table `s_order` or the column `customernumber` in the table `s_user`, has already the right datatype for the requirement.

It makes the column `docID` compatible for letters. So a docID like 'INV_12345' is possible. The resulting filename will be 'INV_12345.pdf'.

Create a Document-ID (bid) with letters in. This leads to a "0" value for the column `docID` in the table `s_order_documents`.

#1656 mistakenly closed by myself

### 6. Checklist

### DIFF
--- a/_sql/migrations/1231-change-docid-column-type.php
+++ b/_sql/migrations/1231-change-docid-column-type.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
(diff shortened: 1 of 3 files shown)


shyim:
Thanks for your contribution! :blue_heart:.

PR #1709, opened 2018-07-11
Support all numbers

hlohaus:

Support ordernumbers like "000/001/002"

Remove the regex check

no

### 6. Checklist

- [x ] This change has comments for package types, values, functions, and non-obvious lines of code

### DIFF
--- a/engine/Shopware/Models/Article/Detail.php
+++ b/engine/Shopware/Models/Article/Detail.php
@@ -142,7 +142,6 @@ class Detail extends ModelEntity
     /**
      * @var string
      * @Assert\NotBlank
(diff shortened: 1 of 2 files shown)


JoshuaBehrens:
I expect some documentation changes. Did you check the code on spots where the code relies on the fact, that an ordernumber can not contain a weird unicode character? I can imagine that the product feeds just build the xml without escaping xml special characters like <.

uehler:
Now you are supporting ordernumbers with spaces again...

hlohaus:
mhh. i know how to change the backend check, but i can't remove the model check by plugin. :/
And why is not it possible to change that?

PR #1710, opened 2018-07-12
Used interface instead of class for CustomerServiceInterface

JoshuaBehrens:
Used interface instead of class to enable decoration via DI. The class hint is flawed as this will throw an exception if an other interface implementation is given.

Change type hints/constructor signature.

Use a different interface implementation (e.g. a decorator) as constructor parameter.

### 4. Checklist

### DIFF
--- a/engine/Shopware/Bundle/CustomerSearchBundleDBAL/Indexing/CustomerProvider.php
+++ b/engine/Shopware/Bundle/CustomerSearchBundleDBAL/Indexing/CustomerProvider.php
@@ -24,12 +24,12 @@
 
 namespace Shopware\Bundle\CustomerSearchBundleDBAL\Indexing;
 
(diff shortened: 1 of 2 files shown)


mitelg:
would be cool, if you could also fix https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Bundle/CustomerSearchBundleDBAL/Indexing/SearchIndexer.php#L45 :blush: :+1:

JoshuaBehrens:
I will do so for the all holy @mitelg

pascalheidmann-bedarf (approved):

soebbing:
Thank you for your contribution, @JoshuaBehrens! 💙

JoshuaBehrens:
Thanks to @shyim for managing my branches all the time :blue_heart:

PR #1712, opened 2018-07-16
Make Zend Session check compatible with PHP 7.1+

t2oh4e:
To make Shopware compatible with PHP 7.1+ configurations

Check wether the code is run with PHP 7.1+, if so use `session.sid_bits_per_character` instead of `session.hash_bits_per_character` which was removed with PHP 7.1

Run shopware on PHP 7.1+ with `session.sid_bits_per_character` set to 4 or 6.
Login to backend or frontend. Shopware will throw an exception that the session has already been started.
After this fix you will be able to login correctly.

Since Shopware is currently not fully compatible with PHP 7.1 you may want to note this somewhere.

### 6. Checklist

### DIFF
--- a/engine/Library/Zend/Session.php
+++ b/engine/Library/Zend/Session.php
@@ -558,7 +558,8 @@ class Zend_Session extends Zend_Session_Abstract
             }
         }
 
-        $hashBitsPerChar = ini_get('session.hash_bits_per_character');
+        $hashBitsPerChar = ini_get('session.sid_bits_per_character') ?: ini_get('session.hash_bits_per_character');
+
         if (!$hashBitsPerChar) {
             $hashBitsPerChar = 5; // the default value
         }


soebbing:
Hi @t2oh4e!

Wouldn't it be better to check for the old ini setting and if it is empty, try the new one instead of checking for a fixed PHP version?

t2oh4e:
Hey @soebbing,
thanks for your reply.
Since that ini setting was replaced in PHP 7.1 the outcome would be the same.
I preferred checking the PHP version instead of checking if the value is set two times but I can change it if you prefer that solution.

jkrzefski:
@t2oh4e I also would prefer a check for existence of the ini settings. It seems a bit hacky to me to check for specific versions of php.

t2oh4e:
I (actually @jkrzefski, but I recomitted to keep the history clean, sorry bro) removed the fixed check for php 7.1 and switched to checking wether `session.sid_bits_per_character` exists and fall back to `session.hash_bits_per_character` if the first one is not present.

soebbing:
Hey @jkrzefski and @t2oh4e, thank you both for your support! 💙

PR #1713, opened 2018-07-18
Fix sorting of emotion elements

Crease29:

The EmotionElementGateway sorts by the deprecated columns `s_emotion_element.start_row` and `s_emotion_element.start_col`. These columns are always filled with `1` and causes elements to be returned in the order in which they have been created in the database.
The user doesn't see this because the elements are sorted the right way in the frontend later.

Use `s_emotion_element_viewports.start_row` and `s_emotion_element_viewports.start_col` for sorting instead.

1. Create a new emotion shopping world
2. Place some text elements there with an increasing number
3. Flip the order of the elements
4. Open the XHR request that loads this emotion shopping world in the frontend (e.g. http://localhost/widgets/emotion/index/emotionId/183/controllerName/index)
5. Open the source code of this request and see that the elements arent returned in the order you placed them in the backend. They will be shown in the order of how you created the elements initially.

:heavy_minus_sign: 

:heavy_minus_sign: 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/EmotionBundle/Service/Gateway/EmotionElementGateway.php
+++ b/engine/Shopware/Bundle/EmotionBundle/Service/Gateway/EmotionElementGateway.php
@@ -124,6 +124,8 @@ class EmotionElementGateway
         $builder->addOrderBy('emotionElementViewport.start_row', 'ASC')
                 ->addOrderBy('emotionElementViewport.start_col', 'ASC');
 
+        $builder->groupBy('emotionElementViewport.elementID');
+
         return $builder;
     }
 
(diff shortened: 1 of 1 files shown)


buddhaCode (approved):

soebbing:
As you say, the elements are sorted the right way in the frontend, which makes sense when you think about it: Only in the browser can be decided which viewport will be shown to the user which decides wich viewport should be first.

I would argue it makes sense to just drop the sorting in the SQL all together as it might cost a little bit performance and only gives a false sense of order.

Crease29:
In my case I had to add an attribute to article boxes in emotions for tracking purposes. It shall be tracked which position converts better. Without this change the information would just be wrong because the elements are echoed to the HTML in the order they have been created in the database...

soebbing:
But won't your tracking information be wrong anyway, even with your change? The elements are ordered by javascript in regards to the viewport after rendering the HTML.

You could attach the element-id as a data-attribute to the HTML and determine the position of the element on click.

Crease29:
Alright, I guess it's just an edgecase for me and the fix works pretty good for me. 
I will just wrap this in a plugin for this project.
Thank you anyways for having a look at my PR!

PR #1714, opened 2018-07-19
[SW-22002] Fix ambiguous column names for birthday cron

sobyte:
See ticket description.

Add table prefix for select statement and reorder for legibility.

See ticket description.

[SW-22002](https://issues.shopware.com/issues/SW-22002)

None.

### 6. Checklist


pascalheidmann-bedarf (changes requested):
While in general a good change to improve code quality the left join doesn't require a matching s_user_address row and those fields can be null.
Therefore filtering is implicit by where clause and not already on JOIN.
I suggest using a `JOIN s_user_addresses ub` instead of the left join.

sobyte:
@pascalheidman-bedarf Thanks for your comment. I've adjusted the code to your suggestion.

soebbing:
Hi @sobyte, thank you for your contribution! 💙

PR #1716, opened 2018-07-20
SW-20629: Fix endless loop for cronjobs with interval of 0

t2oh4e:
It fixes an endless loop with cronjobs that have an interval of zero

Check whether the interval is zero and if it is just use the current timestamp as next execution time instead of trying to calculate it.

Create a cronjob, set the interval to zero and execute the cronjobs e.g. via console:
php bin/console sw:cron:run
The command will be stuck after running through your cronjob and get stuck while calculating the next execution timestamp. After this fix it will finish correctly.

https://issues.shopware.com/issues/SW-20629

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Components/Cron/Manager.php
+++ b/engine/Library/Enlight/Components/Cron/Manager.php
@@ -323,9 +323,13 @@ class Enlight_Components_Cron_Manager
         $now = $now->getTimestamp();
         $interval = $job->getInterval();
         $next = $job->getNext()->getTimestamp();
-        do {
-            $next += $interval;
-        } while ($now >= $next);
+        if ($interval > 0) {
+            do {
+                $next += $interval;
+            } while ($now >= $next);
+        } else {
+            $next = $now;
+        }
         $job->setNext($next);
         $job->setEnd($now);
 


jkrzefski:
@shopwareBot I do not understand, what you mean by `"0" option`. If you mean the possibility to set an interval to 0 should be removed, then I wonder why. That would basically render this pr useless.

pascalheidmann-bedarf (commented):
I suggest to remove the loop completely as this will always be some kind of possible endless loop.
Say I have a job running every second and just has a `sleep(1);` this would also loop forever.

I understand the thought behind using a loop but nonetheless it possibly hinders every other cron running after it, killing it effectively.

jkrzefski:
@pascalheidman-bedarf I think, you misunderstand the purpose of the loop. It does not execute the job. It merely exists to increment the timestamp by a multiple of the given interval. The only reason for this loop is to "calculate" the fitting multiple of the value given for the interval. So this would not run a specific job over and over again.

@shopwareBot I still fail to see a reason for your suggestion. If it is still possible to have an interval of 0 via database or cronjob.xml file, why should this option be artificially limited in the backend?

pascalheidmann-bedarf:
@jkrzefski oh, yes, you are right. But the problem I described is somewhere else in Shopware's code nonetheless.

jkrzefski:
@pascalheidman-bedarf That might be true. But if that is an issue for you, I believe you should make another PR to fix that.

t2oh4e:
Actually I think it's not necessary to remove the zero option from the backend as @jkrzefski already stated. Is there a reason you want it to be removed @shopwareBot? If you insist on removing it I will change the PR accordingly.

PR #1717, opened 2018-07-21
Product stream as config element

hlohaus:
It isn't

Adds product_stream as config element

-

-

Yes, add the new type.

### 6. Checklist


uehler:
You can just use this snippet to create a stream select in the plugin config:
   
    <element type="select">
        <name>test</name>
        <label>test</label>
        <options>
            <xtype>productstreamselection</xtype>
        </options>
    </element>

PR #1718, opened 2018-07-23
Added block to change product box image link attributes

JoshuaBehrens:
The product-image.tpl generates a embracing a-tag. This a-tag is not modifiable yet.

* adds an optional variable to change the target
* adds a block to add even more attributes
* uses `/attributes` naming to support the [FroshProfiler](https://github.com/FriendsOfShopware/FroshProfiler)s block annotator

1. Try to change the target to _blank
2. Copy whole file and overwrite it completely with the old content for changing one line without using the power of ~~grayskull~~ inheritance
3. :man_facepalming: 
4. no profit
5. (╯ರ ~ ರ）╯︵ ┻━┻

### 4. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/listing/product-box/product-image.tpl
+++ b/themes/Frontend/Bare/frontend/listing/product-box/product-image.tpl
@@ -2,7 +2,7 @@
 <a href="{$sArticle.linkDetails}"
    title="{$sArticle.articleName|escape}"
    class="product--image"
-    {block name="frontend_listing_box_article_image/attributes"}
+    {block name="frontend_listing_box_article_image_attributes"}
         {if $target}target="{$target}"{/if}
     {/block}
    >
--- a/themes/Frontend/Bare/frontend/listing/product-box/product-image.tpl
+++ b/themes/Frontend/Bare/frontend/listing/product-box/product-image.tpl
@@ -2,9 +2,7 @@
 <a href="{$sArticle.linkDetails}"
    title="{$sArticle.articleName|escape}"
    class="product--image"
-    {block name="frontend_listing_box_article_image_attributes"}
-        {if $target}target="{$target}"{/if}
-    {/block}
+    {block name="frontend_listing_box_article_image_attributes"}{/block}
    >
     {block name='frontend_listing_box_article_image_element'}
         <span class="image--element">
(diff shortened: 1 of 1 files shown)


themes/Frontend/Bare/frontend/listing/product-box/product-image.tpl
```diff
@@ -1,7 +1,11 @@
 {* Product image - uses the picture element for responsive retina images. *}
 <a href="{$sArticle.linkDetails}"
    title="{$sArticle.articleName|escape}"
-   class="product--image">
+   class="product--image"
+    {block name="frontend_listing_box_article_image/attributes"}
```
shyim: frontend_listing_box_article_image_attributes :worried: 
See https://github.com/FriendsOfShopware/FroshProfiler/commit/cd0d8874ffb2a656f46b5b32d9c5d8ffb0181719

JoshuaBehrens: @shyim y u do dis to me ლ(ಠ益ಠლ)

themes/Frontend/Bare/frontend/listing/product-box/product-image.tpl
```diff
@@ -1,7 +1,11 @@
 {* Product image - uses the picture element for responsive retina images. *}
 <a href="{$sArticle.linkDetails}"
    title="{$sArticle.articleName|escape}"
-   class="product--image">
+   class="product--image"
+    {block name="frontend_listing_box_article_image_attributes"}
+        {if $target}target="{$target}"{/if}
```
shopwareBot: Is this necessary?
We'd rather have the empty block, so you're able to add attributes to that link in your custom theme.

Can you remove the target please?

JoshuaBehrens: When I remove it I lose the ability to easily customize the target on a smarty include. Then I have to override the block just for a target. But I will have to do it for any attribute. So this is acceptable for me.

JoshuaBehrens: This change shall be written in the books of history as this action was done in the name of equality of optional attributes.

JoshuaBehrens:
I am :confused: 
It is incomplete yet it is picked and scheduled. Anything more I have to fix?

PR #1719, opened 2018-07-24
Move upperParams assignment lower to contain the productBoxLayout parameter

JoshuaBehrens:
When using a custom product box layout in an infinite listing it does not load the products by layout option.

It copies the upperParams after assigning the productBoxLayout property. It is important not to move the historyParams as this will result in the productBoxLayout option being part of the URL.

1. Set attribute `data-productBoxLayout` to a different product box layout on a element with the `swInfiniteScrolling` jquery plugin registered on it
2. Load that listing on a page > 1
3. Click on "load previous"
4. See products in box basic
5. 😤 :anger: 

### 4. Checklist

### 5. Quickfix
```js
$.overridePlugin('swInfiniteScrolling', {
    init: function() {
        var me = this;
        me.superclass.init.apply(this, arguments);
        
        if (me.opts.productBoxLayout) {
            me.upperParams.productBoxLayout = me.opts.productBoxLayout || null;
        }
    }
});
```

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.infinite-scrolling.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.infinite-scrolling.js
@@ -163,13 +163,13 @@
             };
 
             me.params = parseQueryString(window.location.href);
-            me.upperParams = $.extend({}, me.params);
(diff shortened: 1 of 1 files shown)


JoshuaBehrens:
Props to @jkrzefski for quickfix

PR #1720, opened 2018-07-24
Fetches usergroup id from sSYSTEM not ShopContext

SC-cmarquardt:
ShopContext Variable is not updated during Login Process. sSYSTEM contains the correct usergroup of the user during login process.
See Issue https://issues.shopware.com/issues/SW-22229

ShopContext Variable is not updated during login process if known but not logged in customer puts articles into the basket and logs-in afterwards.

Fetches usergroup from sSYSTEM as in other locations in this class and not from ShopContext.

Issue occurs together with plugin https://store.shopware.com/atsd00790/warenkorb-von-registrierten-benutzern-speichern.html if not "public available" articles are stored in the basket and the user performs a new login.

https://issues.shopware.com/issues/SW-22229

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -1414,7 +1414,7 @@ class sBasket
         list($queryAdditionalInfo, $quantity) = $this->getAdditionalInfoForUpdateArticle($id, $quantity);
         $queryNewPrice = $this->getPriceForUpdateArticle($id, $quantity, $queryAdditionalInfo);
 
-        $customerGroupId = $this->contextService->getShopContext()->getCurrentCustomerGroup()->getId();
+        $customerGroupId = $this->sSYSTEM->sUSERGROUPDATA['id'];
         if (in_array($customerGroupId, $queryAdditionalInfo['blocked_customer_groups'])) {
             // if blocked for current customer group, delete article from basket
             $this->sDeleteArticle($id);


jkrzefski:
I understand the problem you describe, but I do not think that this solution will be sufficient. Since [sSYSTEM is deprecated](https://github.com/shopware/shopware/blob/5.4/engine/Shopware/Core/sSystem.php#L30) the problem should instead be solved by updating the ShopContext returned by the ContextService at the exact same time the login is performed.

shyim:
Hey, @SC-cmarquardt,

i had a similar issue also last weeks in the proportional tax calculation ticket. Can you check that the fix from https://github.com/shopware/shopware/commit/413e404e5bbb0f67c881e596dd891085894db031 also fixes your problem?

Thanks! :blush:

SC-cmarquardt:
Hi @jkrzefski,
I also see your point but this would mean a bigger rewrite of the login handling. As fas as I have seen the creation of a ShopContext is rarely used. Alternatively I can fetch the information from the user session, but I wanted to stick to the same process as it is currently used in the file.

Hi @shyim,
as stated above an alternative way would be fetching the information from the session and not from sSYSTEM. But your solution does not fix my problem as the fetching of information is not changed where I need the change.

Kind regards,
Christopher

shyim:
@SC-cmarquardt the sRefreshBasket call updates all articles and also calls the sUpdateArticle method. So there should be the updated data after loggedin :relaxed:

SC-cmarquardt:
@shyim now I see your point. But according to https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Bundle/StoreFrontBundle/Service/ContextServiceInterface.php#L93 `initializeContext` is deprecated and to be removed in 6.0

shyim:
Good catch, i change that tomorrow 👍. So this PR can be closed then?

SC-cmarquardt:
As far as I can see your change is on the 5.5 branch. I would like to have the fix for the 5.4 branch.
Either we use yours for 5.5 and mine quickfix for 5.4 or can you also do a backport for 5.4?

jkrzefski:
@SC-cmarquardt I believe the release of 5.5 is rather close. I do not know this for sure, but if the release actually is close, it makes no sense to have this change in the 5.4 branch.

SC-cmarquardt:
@jkrzefski we are currently working with a project using the 5.4 branch. So if there will be another 5.4. release before 5.5 becomes stable I would be very happy to see this fixed.

PR #1726, opened 2018-07-26
Added support for multilingual product name sorting

JoshuaBehrens:
Sorting products by its name is not working properly when using translations on products to serve them in two different languages. When one sort products by name in a category listing they are ordered by their untranslated name. 

It joins the s_core_translations table for sorting by the suitable translation.

The common shopware installation serves a German product set with English translations. If you have an 'Igelmütze' in this shop it will be ordered wrong in an English listing as the translation says 'Hedgehog hat'. The listing displays 'Hedgehog hat' but it was sorted by 'Igelmütze'. This will be subtle as H and I are next to each other. But imagine a 'Strandtuch' product and its translation 'Beach towel'.

I recognized it in the Shopware 4 demo data in the category 'Genusswelten'/'World of Indulgence' with the product 'Wacholder Premium extra mild 32%' with its translation 'Juniper premium extra mild 32%'.

### 4. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/SortingHandler/ProductNameSortingHandler.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/SortingHandler/ProductNameSortingHandler.php
@@ -37,25 +37,21 @@ use Shopware\Bundle\StoreFrontBundle\Struct\ShopContextInterface;
  */
 class ProductNameSortingHandler implements SortingHandlerInterface
 {
-    const TRANSLATION_IDENTIFIER = 's:10:"txtArtikel";s:';
+    const TRANSLATIONS_TABLE = 's_articles_translations';
 
     const TRANSLATION = 'productTranslationName';
 
-    const TRANSLATION_OBJECT_DATA = self::TRANSLATION . '.objectdata';
+    const TRANSLATION_NAME = self::TRANSLATION . '.name';
 
-    const TRANSLATION_OBJECT_KEY = self::TRANSLATION . '.objectkey';
+    const TRANSLATION_PRODUCT_ID = self::TRANSLATION . '.articleID';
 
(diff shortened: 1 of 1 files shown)


JoshuaBehrens:
My code contains a lot of static expression building which can be extracted into a common sql query builder. If this is required to be merged I am open for suggestions.
If constants or variable interpolation are unwanted in this project I can change this too.

engine/Shopware/Bundle/SearchBundleDBAL/SortingHandler/ProductNameSortingHandler.php
```diff
@@ -54,6 +74,101 @@ public function generateSorting(
         ShopContextInterface $context
     ) {
         /* @var ProductNameSorting $sorting */
-        $query->addOrderBy('product.name', $sorting->getDirection());
+        $query->leftJoin(
+            self::PRODUCT,
+            self::S_CORE_TRANSLATIONS,
```
shyim: Its not necessary to use s_core_translations for that. Please use the s_articles_translations table instead

JoshuaBehrens: This is redundant, isn't it? Which is the leading table?

shyim: s_articles_translations is identical to the s_core_translations and is currently only used for search indexing for performance reasons.

I guess the usage of the table here is much faster than your extract sql solution

JoshuaBehrens: It has a way better performance hence to the missing string operations. Therefore I am glad to know that.

JoshuaBehrens:
Everything fine for me as it eventually got merged :blue_heart:

PR #1727, opened 2018-07-26
Allow empty stateId in Order API

Isengo1989:
The stateId is necessary on create of an order. This should not be requiered. Also see the ISSUE SW-21511 I also fixed the typo that states that shipping state is missing instead of billing state

The user can register without a stateId/state, why should the API disallow it.

It checks if the stateId is empty - if so it sets it to 0. If it is not empty it follows the check if the state actually exists and still throws an error

Create order with empty stateId fails

https://issues.shopware.com/issues/SW-21511

### 6. Checklist


Isengo1989:
@soebbing sorry I lost track of the other PR (https://github.com/shopware/shopware/pull/1703) because of work.

I added two tests and verified that this solution works with an empty stateId and also throws an error on not matching stateIds.

shyim:
Hey @Isengo1989 ,

Thank you for your contribution! Your change has been merged with https://github.com/shopware/shopware/commit/e90491689bf2d31fadf2259504b20f202f1ca5f4
:blue_heart: :tada:

PR #1728, opened 2018-07-27
Relax access modifiers in SearchBundleDBALs VariantHelper

JoshuaBehrens:
If you want to derive from VariantHelper you cannot gain advantage from its own small utility functions. This makes a child class useless and one has to implement the same small utility functions again.

Mui :clipboard: :spaghetti: 

Replace private with protected

1. Derive from VariantHelper
2. Try to make an other group by in `joinPrices`
3. Have to completely re.implement `joinPrices` without the help of the utility functions
4. :clipboard: 
5. :spaghetti: 

### 4. Checklist


soebbing:
Thanks again for your contribution, @JoshuaBehrens! 🎉

PR #1730, opened 2018-07-27
fixed unescaped post param

uehler:
unexcaped post param is used as value in a template form. All other params are escaped, so this one should be escaped too?
@flynorc found it in the conexco bootstrap theme. We compared the template file with the one in your bare theme and found the same behaviour.

escapes that param

look into the template file

### 6. Checklist


pascalheidmann-bedarf (commented):
While definitely a good change, IMO this code should never access those globals in the first place. Can you change the handling inside the controller?

PR #1731, opened 2018-07-30
Add filter event for product price conversion.

kleinmann:
`convertProductPriceStruct` is the only conversion method without a filter event. This prevents a clean extension of this conversion and should be fixed.

It adds a filter event in a similar manner as in all other conversion methods.

- Try to extend the product price conversion
- Fail to do so because there is no filter event

There is no issue yet.

Change in UPGRADE-5.5.md is included. I'm not aware of any other necessary changes.

### 6. Checklist


soebbing:
Hi @kleinmann, thank you very much for your contribution! 💙

PR #1733, opened 2018-07-31
Make 'Shopware\Bundle\PluginInstallerBundle\Service\SubscriptionService::getPluginInformationFromApi()' public

svenmuennich:

Currently information about plugin subscriptions can only be retrieved by calling `Shopware\Bundle\PluginInstallerBundle\Service\SubscriptionService::getPluginInformation()`, which only allows to fetch new information from the Showpare store API every 24 hours (based on a browser cookie). Hence it is currently not possible to arbitrarily get the latest subscription data.

This PR changes the visibility of `Shopware\Bundle\PluginInstallerBundle\Service\SubscriptionService::getPluginInformationFromApi()` from `private` to `public`. Furthermore it removes its only parameter `$secret` and instead directly loads the required value as needed.

**Note:** The shop needs to be logged into your Shopware store account.

Call `Shopware()->Container()->get('shopware_plugininstaller.subscription_service')->getPluginInformationFromApi()` from anywhere inside Shopware to always retrieve the latest subscription data.

n/a

n/a

### 6. Checklist

- [ ] ~I have written tests and verified that they fail without my change~ (cannot be tested without a valid store account)

### DIFF
--- a/UPGRADE-5.5.md
+++ b/UPGRADE-5.5.md
@@ -167,6 +167,7 @@ This changelog references changes done in Shopware 5.5 patch versions.
        | --variantswitch       | -d    | Warm up variant switch of configurators  |
        | --productwithnumber   | -x    | Warm up products with number parameter   |
        | --productwithcategory | -y    | Warm up producss with category parameter |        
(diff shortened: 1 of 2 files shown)


shyim:
thanks for your contribution @svenmuennich  :+1: :tada: :blue_heart:

PR #1732, opened 2018-07-31
Adds new shopware domains to "SEO follow backlinks"

jkrzefski:
### 1. Why is this change necessary?
It is not. Just to have the domains up to date.

### 2. What does this change do, exactly?
Add the new domains `de.shopware.com` and `en.shopware.com` to the default value of the configuration option `SEO follow backlinks`.

### 3. Describe each step to reproduce the issue or behaviour.
Open the backend and go to `Basic settings > Frontend > SEO / router settings` and find the option `SEO follow backlinks`. See that the new shopware domains are missing in the default value.

### 4. Please link to the relevant issues (if any).
None.

### 5. Which documentation changes (if any) need to be made because of this PR?
None.

### 6. Checklist

- [ ] I have written tests and verified that they fail without my change
- [x] I have squashed any insignificant commits
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfil them.

PR #1735, opened 2018-07-31
Adds new shopware domains to "SEO-Follow Backlinks"

jkrzefski:
It is not. Just to have the domains up to date.

Add the new domains `de.shopware.com` and `en.shopware.com` to the default value of the configuration option `SEO follow backlinks`.

Open the backend and go to `Basic settings > Frontend > SEO / router settings` and find the option `SEO follow backlinks`. See that the new shopware domains are missing in the default value.

A few hours ago I made a similar PR #1732 which does not add a migration but instead edits the `latest/install.sql` file. This PR aims to correct that mistake.

None.

### 6. Checklist


_sql/migrations/1420-add-new-shopware-domain-to-seo-follow-backlinks.php
```diff
@@ -0,0 +1,39 @@
...
+    public function up($modus)
+    {
+        $sql = <<<SQL
+UPDATE s_core_config_elements
+SET s_core_config_elements.value = 's:54:"www.shopware.de,
+www.shopware.ag,
+www.shopware-ag.de,
+de.shopware.com,
+en.shopware.com";'
+WHERE s_core_config_elements.name = 'seobacklinkwhitelist';
```
shyim: Please make a md5 of the old value and equal this here. You would currently overwrite customer settings :worried: 
Also it would be awesome if you change the target to 5.5

jkrzefski: I thought custom settings were stored in `s_core_config_values` instead of `s_core_config_elements` and that this is the place to set a default value. I'll get to work as soon as I can.

shyim: Oh you are right, `s_core_config_elements.value` is just the default value. I would also change the values in s_core_config_values with a md5 check :)

When the customer saves the seo form in basic settings, these unchanged values would be also saved in s_core_config_values :blush:

pascalheidmann-bedarf (commented):
This PR does a lot of things at the same time that do not feel like they should be in only one PR. You might want to split those...

jkrzefski:
@pascalheidman-bedarf I guess that's what I get for listening to @shyim 😉. I only added one file and made one change to that file. But switching the target from 5.4 to 5.5 added a lot of noise in the git log.

shyim:
@jkrzefski You rebase to 5.5 got wrong :wink:

jkrzefski:
@shyim You said nothing about a rebase. Only to change the target. Hence I did just that. But I will do a rebase when I find the time for it.

mitelg:
@jkrzefski isn't it kind of clear, that you need to update your source branch, if you change the target branch? :smile:

PR #1737, opened 2018-08-01
Enable search for address-list in backend view

roman-1983:
Having users with a lot of addresses makes it hard to find the right address in the backend-view for the addresses

It enables the extjs search function for the address-list in the user-view in the backend

Add more than 20 addresses to a user account and try to find the right address in the backend view

`null`

none

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/customer/view/address/list.js
+++ b/themes/Backend/ExtJs/backend/customer/view/address/list.js
@@ -159,7 +159,6 @@ Ext.define('Shopware.apps.Customer.view.address.List', {
 
         return {
             columns: columns,
-            searchField: true,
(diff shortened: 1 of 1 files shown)


themes/Backend/ExtJs/backend/customer/view/address/list.js
```diff
@@ -159,7 +159,7 @@ Ext.define('Shopware.apps.Customer.view.address.List', {
 
         return {
             columns: columns,
-            searchField: false,
+            searchField: true,
```
mitelg: true is the default, so you could just delete the line

roman-1983:
I removed the line `searchField` completly because the default value is `true`.

PR #1738, opened 2018-08-02
Allow to configure Shopware::DocPath()

hranicka:

Currently Shopware determines its root directory from location of `engine/Shopware/Application.php` file.

Unfortunately this location can be nested in `vendor/shopware/shopware` which results into invalid paths in code relying on `Shopware::DocPath()`.

After provided change, it's possible to easily set correct `DocPath` from `Kernel::initializeShopware()` which can be overridden (and it is in mentioned composer-project).
Then there is no need of any special proxy or adapter object to fix behavior when using Shopware as a Composer dependency.

Install Shopware's composer-project and call `Shopware()->DocPath()`. It's suppose to return project root path which is not within `vendor/shopware/shopware`.

This problem is related to https://github.com/shopware/composer-project/pull/15

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Application.php
+++ b/engine/Shopware/Application.php
@@ -64,15 +64,18 @@ class Shopware extends Enlight_Application
 
     /**
      * @param Container $container
+     * @param string|null $docPath
      */
-    public function __construct(Container $container)
+    public function __construct(Container $container, $docPath = null)
     {
         // Initialize global Shopware function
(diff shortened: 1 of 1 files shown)


hranicka:
`kernel.root_dir`, thanks :+1: 

The fix in composer-project: https://github.com/shopware/composer-project/pull/15

PR #1741, opened 2018-08-07
Hotfix: CVE-2018-14773

hranicka:

There is a security issue in symfony/http-foundation <2.8.44, see https://security.sensiolabs.org/database?package=symfony/http-foundation

shopware/shopware requires vulnerable 2.8.41

Updates all Symfony components to their latest patch versions. They should be without any breaking changes, except security related things.

There is also added library which checks for vulnerable dependencies: https://github.com/Roave/SecurityAdvisories

Not necessary, Shopware requires vulnerable version of a library.

-

-

### 6. Checklist


hlohaus:
the security issue is not there, shopware use zend aka enlight as mvc framework:

https://symfony.com/blog/cve-2018-14773-remove-support-for-legacy-and-risky-http-headers
and:

https://github.com/shopware/shopware/blob/5.4/engine/Library/Enlight/Controller/Request/RequestHttp.php#L681

hranicka:
Thank you @hlohaus , but Shopware uses vulnerable version of that library and then it's not possible to add https://github.com/Roave/SecurityAdvisories into a project based on Shopware (even if it's project based on shopware/composer-project).

Tests are passing after Symfony components upgrade so there should not be any problem with that. Or do you see any?

Thank you.

mitelg:
hey @hranicka 
afaik it is not planned to release another 5.4.x version, as we just released the 5.5.0 RC1 and soon we will release the final 5.5.0 version. This is already fixed in Shopware 5.5.

hranicka:
Hello, we're using shopware/composer-project and there we cannot change version of Symfony components since shopware/shopware requires specific versions.

Also we cannot add roave/security-advisories to any project built on Shopware. We cannot even install any library which includes roave/security-advisories as a dependency.

If you use something to check security issues in the build process, it's not used in projects build on Shopware since they may not use the same build process.

For me it's also quite surprising you've written there won't be any new Shopware 5.4.x release since 5.5.x final release is comming. I created a ticket for that: https://issues.shopware.com/issues/SW-22321

Thank you for your support.

soebbing:
If adding roave/security-advisories to the composer-project is important to you, I'd recommend to fork Shopware for the moment, change the dependencies there and require that fork in the composer-project.

Security fixes are provided to older Shopware versions using a Security Plugin https://store.shopware.com/en/swag575294366635f/shopware-security-plugin.html

hranicka:
As I wrote, `roave/security-advisories` could be used also in any other library required in project based on Shopware. There is a solution you suggested and it's to wait and again upgrade to Shopware 5.5 which has not been released yet and (probably) will cause some back-compatibility issues with older code bases.

:(

soebbing:
Hi again @hranicka! I discussed your problem with some collegues internally and if it would be feasable for us to relax the version constraints in the `composer.json` of Shopware. For the moment we decided against it as it could lead to unspecific bugs which we would be unable to reproduce or identify.

As a possible workaround maybe you want to take a look at the [Symfony Security Checker](https://symfony.com/doc/current/security/security_checker.html) which could serve as an alternative for `roave/security-advisories` while not preventing you from actually using possibly insecure libraries.

hranicka:
Hello, like I've written - there may be a package which requires `roave/security-advisories`, not necessary the application itself. Furthermore it uses the same database as mentioned Symfony Security Checker.

So if we would use Symfony Security Checker in CI, our build would be failing anyway. We should not ignore these build failures so it would prevent us from using possibly insecure libraries.

Thanks there are newer Symfony components in Shopware 5.5. But it has no stable release yet, brings some more effort while upgrading bigger projects based on Shopware and that's why I proposed this merge request upgrading Symfony components to newer patch versions.

PR #1742, opened 2018-08-07
Changed repository method params to be allowed to be emtpy

Gugiman:
getListQueryBuilder can't be called without optional params because of false type of the filter param which should be array

Method can be called without parms to get all results

Changed default value of optional param 

call getListQueryBuilder() without params

none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Form/Repository.php
+++ b/engine/Shopware/Models/Form/Repository.php
@@ -51,22 +51,23 @@ class Repository extends ModelRepository
      * Helper function to create the query builder for the "getListQuery" function.
      * This function can be hooked to modify the query builder of the query object.
      *
-     * @param null|array $filter
+     * @param null $filter
      * @param null $orderBy
      *
      * @return \Doctrine\ORM\QueryBuilder
      */
-    public function getListQueryBuilder($filter = [], $orderBy = null)
+    public function getListQueryBuilder($filter = null, $orderBy = null)
     {
         $builder = $this->getEntityManager()->createQueryBuilder();
         $builder->select(['form', 'attribute'])
             ->from($this->getEntityName(), 'form')
             ->leftJoin('form.attribute', 'attribute');
         
-        $this->addFilter($builder, $filter);
-
-        if($orderBy !== null)
-        {
+        if($filter !== null) {
+            $this->addFilter($builder, $filter);
+        }
+        
+        if($orderBy !== null) {
             $this->addOrderBy($builder, $orderBy);
         }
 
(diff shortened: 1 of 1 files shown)


engine/Shopware/Models/Form/Repository.php
```diff
@@ -51,20 +51,24 @@ public function getListQuery($filter, $orderBy, $offset, $limit)
...
-    public function getListQueryBuilder($filter = null, $orderBy = null)
+    public function getListQueryBuilder($filter = [], $orderBy = null)
     {
         $builder = $this->getEntityManager()->createQueryBuilder();
         $builder->select(['form', 'attribute'])
             ->from($this->getEntityName(), 'form')
             ->leftJoin('form.attribute', 'attribute');
-
+        
         $this->addFilter($builder, $filter);
```
shopwareBot: Can you add here also a if condition that the $filter is not null and keep the default value in the parameter.

Gugiman: Ok but before one question for my comprehension:

If i want to call the method without any filtering, I have to call it with an empty array as param. Otherwise an exception is thrown.
Why not setting the default param value to array immidiately? (I see that this has to be done also for $orderBy)

soebbing: Changing the default parameter could be a potential break for others developers

Gugiman:
I made the requested changes

engine/Shopware/Models/Form/Repository.php
```diff
@@ -51,22 +51,23 @@ public function getListQuery($filter, $orderBy, $offset, $limit)
...
         $builder = $this->getEntityManager()->createQueryBuilder();
         $builder->select(['form', 'attribute'])
             ->from($this->getEntityName(), 'form')
             ->leftJoin('form.attribute', 'attribute');
         
-        $this->addFilter($builder, $filter);
-
-        if($orderBy !== null)
-        {
+        if($filter !== null) {
```
FiveDigital: Please add a space after the `if`

engine/Shopware/Models/Form/Repository.php
```diff
@@ -51,22 +51,23 @@ public function getListQuery($filter, $orderBy, $offset, $limit)
...
+            $this->addFilter($builder, $filter);
+        }
+        
+        if($orderBy !== null) {
```
FiveDigital: A space after the `if` is missing here, too.

Gugiman: argh...
changed

PR #1743, opened 2018-08-08
Add regex validation to registration

t2oh4e:
To force customers to use certain characters in their password.

It adds a new config option to define a regular expression. If it is set passwords will be validated against the regular expression

Set the new config option in the basic settings at "Registration / login" to
`/^(?=.*\d)(?=.*[a-z])(?=.*[A-Z])(?=.*[^a-zA-Z0-9]).+/`.
Now you will have to provide a password with at least one uppercase letter, one lowercase letter, a number and a special character.
Otherwise you will not be able to register and get an error message.

-

-

### 6. Checklist

### DIFF
--- a/UPGRADE-5.5.md
+++ b/UPGRADE-5.5.md
@@ -101,6 +101,7 @@ This changelog references changes done in Shopware 5.5 patch versions.
         - `DynamicCacheTimeService`
     * Added `invalidation_date_provider` tag to the DIC
 * Added parameter mode to Log module, to directly open the systemlogs tab
+* Added a config option to set a regular expression to validate passwords during registration
 
 ### Changes
 
--- a/_sql/migrations/1434-add-password-regex.php
+++ b/_sql/migrations/1434-add-password-regex.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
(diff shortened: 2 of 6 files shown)


FiveDigital:
Perhaps you could also add something like `pattern="[A-Za-z]{3}"` to the `<input />` for the password to also have client side validation?

t2oh4e:
@FiveDigital Actually a good idea but I think it has some drawbacks since the input pattern does not exactly work like the PHP `preg_match`. I think there are some minor differences one has to consider when using a RegEx in both cases.
Additionally the password will be validated via Ajax by default giving instant feedback on the client side. Therefore I would leave out the pattern on the input field.

FiveDigital:
You're right 👍

soebbing:
I see some issues with this change :/

For one: The initial regex (an empty one) has nothing to do with the error message you provide. That could lead to a wrong error message and incorrect password requirements being shown.

Also: I am not yet sure if it is a good idea to provide shop owners with such a basic tool like a regex. It is far too easy even for tech savy shop owners to shoot themselves in the foot. I think at best we would have to provide a list of validations from which a shop owner can choose from, like "no validation", "small and capital letters", "small and capital letters and a number", "small and capital letters, a number and special char", etc. which also automatically shows the correct error in the frontend.

Bottom line: I can see the need for agencies and developers for an easy way to implement a project's custom password requirements in a straight-forward way, but I would very much prefer something like e.g. a custom `RegexConstraint` which only needs to be implemented by a plugin. But I am pretty sure the Symfony Form Components have all that is necessary for such a feature. Maybe a little DevDocs-article on how to best change password requirements would already be enough?

t2oh4e:
I can understand you concerns and will therefore switch to decorating the `PasswordValidator` to my needs with a plugin :)

soebbing:
Thank you for your answer, we'll close the PR then.

Maybe you want to create a PR in the devdocs with your plugin as an example?

PR #1749, opened 2018-08-09
Update AppCache.php

wesolowski:

If you move over the page and go back by using the history back of your browser, some contents are displayed cached by your browser. For example by opening a list page, clicking on a product and then going back by hitting the history back on your browser, some widgets like basket or notes are displayed from the browser cached.

The change adds a “no-store” to the response. 

Activate HTTP Cache and move shop into production mode. Open a listing page, move on a detail page, add a product to basket and then use the history back to go to the listing page.
The basket is not updated because it was loaded from the browser cache.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/HttpCache/AppCache.php
+++ b/engine/Shopware/Components/HttpCache/AppCache.php
@@ -116,7 +116,7 @@ class AppCache extends HttpCache
         $response = parent::handle($request, $type, $catch);
 
         $response->headers->remove('cache-control');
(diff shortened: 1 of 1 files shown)


buddhaCode:
I guess that setting should be a bit more selective. Maybe just apply no store only for ajax requests? I think it's a good think, that the listing page is not coming from browser cache.

soebbing:
I agree with @buddhaCode. There are ESI widgets that shouldn't be reloaded, e.g. the language menu. Maybe just add this header to the basket and notes widget?

PR #1750, opened 2018-08-09
Model corrections

jochenmanz:
better developer experience

corrected some misleading typehints, removed a bit of broken code and broken funktions, added missing properties

no actual issue, just some code optimizations

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Article/Configurator/Template/Template.php
+++ b/engine/Shopware/Models/Article/Configurator/Template/Template.php
@@ -70,6 +70,7 @@ class Template extends ModelEntity
      * @ORM\JoinColumn(name="unit_id", referencedColumnName="id")
      */
     protected $unit;
+
     /**
      * @var int
      *
@@ -87,94 +88,98 @@ class Template extends ModelEntity
     private $articleId;
 
     /**
-     * @var int
+     * @var int|null
      *
      * @ORM\Column(name="unit_id", type="integer", nullable=true)
      */
-    private $unitId = null;
+    private $unitId;
 
     /**
      * @var string
+     *
      * @Assert\NotBlank
      * @Assert\Regex("/^[a-zA-Z0-9-_. ]+$/")
      *
-     * @ORM\Column(name="order_number", type="string", nullable=false, unique = true)
+     * @ORM\Column(name="order_number", type="string", nullable=false, unique=true)
      */
     private $number = '';
 
     /**
-     * @var string
+     * @var string|null
      *
      * @ORM\Column(name="suppliernumber", type="string", nullable=true)
      */
-    private $supplierNumber = null;
+    private $supplierNumber;
 
     /**
-     * @var string
+     * @var string|null
      *
      * @ORM\Column(name="additionaltext", type="string", nullable=true)
      */
-    private $additionalText = null;
+    private $additionalText;
 
     /**
-     * @var int
+     * @var bool
      *
      * @ORM\Column(name="active", type="integer", nullable=false)
      */
     private $active = false;
 
     /**
-     * @var int
+     * @var int|null
      *
      * @ORM\Column(name="instock", type="integer", nullable=true)
      */
-    private $inStock = null;
+    private $inStock;
 
     /**
-     * @var int
+     * @var int|null
      *
      * @ORM\Column(name="stockmin", type="integer", nullable=true)
      */
-    private $stockMin = null;
+    private $stockMin;
 
     /**
-     * @var int
+     * @var bool
      *
      * @ORM\Column(name="laststock", type="boolean", nullable=false)
      */
-    private $lastStock = null;
+    private $lastStock;
 
     /**
-     * @var float
+     * @var float|null
      *
      * @ORM\Column(name="weight", type="decimal", nullable=true, precision=3)
      */
-    private $weight = null;
+    private $weight;
+
+    /**
+     * @var float|null
+     *
+     * @ORM\Column(name="width", type="decimal", nullable=true, precision=3)
+     */
+    private $width;
+
+    /**
+     * @var float|null
+     *
+     * @ORM\Column(name="length", type="decimal", nullable=true, precision=3)
+     */
+    private $len;
 
     /**
      * @var float
      *
-     * @ORM\Column(name="width", type="decimal", nullable=true, precision=3)
-     */
-    private $width = null;
-
-    /**
-     * @var float
-     * @ORM\Column(name="length", type="decimal", nullable=true, precision=3)
-     */
-    private $len = null;
-
-    /**
-     * @var float
      * @ORM\Column(name="height", type="decimal", nullable=true, precision=3)
      */
-    private $height = null;
+    private $height;
 
     /**
-     * @var float ean
+     * @var float
+     *
      * @ORM\Column(name="ean", type="string", nullable=true)
      */
-    private $ean = null;
+    private $ean;
 
     /**
      * @var float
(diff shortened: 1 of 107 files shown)


engine/Shopware/Models/Country/Repository.php
```diff
@@ -278,21 +278,6 @@ public function getStatesByCountryIdQueryBuilder($countryId)
...
-     * Returns an instance of \Doctrine\ORM\Query object which selects a
-     * list of countries for the passed area id.
-     *
-     * @param $stateId
-     *
-     * @return \Doctrine\ORM\Query
-     */
-    public function getStateAttributesQuery($stateId)
-    {
-        $builder = $this->getShopsQueryBuilder($stateId);
```
jochenmanz, line 291: the function getShopsQueryBuilder is only available in the Analytics Repository

engine/Shopware/Models/Dispatch/Repository.php
```diff
@@ -152,12 +152,12 @@ public function getShippingCostsQueryBuilder($dispatchId = null, $filter = null,
                 ->leftJoin('dispatch.holidays', 'holidays')
                 ->leftJoin('dispatch.attribute', 'attribute')
                 ->leftJoin('dispatch.payments', 'payments');
-        if (null !== $dispatchId) {
+        if ($dispatchId !== null) {
```
jochenmanz: shopware's own php-cs-fixer applied some missing changes

engine/Shopware/Models/Document/Order.php
```diff
@@ -66,6 +66,7 @@ class Shopware_Models_Document_Order extends Enlight_Class implements Enlight_Ho
      * @var ArrayObject
      */
     protected $_positions;
+    protected $positions;
```
jochenmanz: this positions is very weired, could be duplicate but some code relies on it. I suspect that some functionality depending on positions and _positions could be broken as well.

engine/Shopware/Models/Emotion/PresetTranslation.php
```diff
@@ -21,27 +21,35 @@
...
+ *
  * @copyright  Copyright (c) shopware AG (http://www.shopware.de)
  *
  * @ORM\Entity
  * @ORM\Table(name="s_emotion_preset_translations")
  */
 class PresetTranslation extends ModelEntity
 {
+    /**
+     * @var \Shopware\Models\Emotion\Preset
```
jochenmanz: shopware's php-cs-fixer again

engine/Shopware/Models/Newsletter/Newsletter.php
```diff
@@ -467,22 +468,6 @@ public function getId()
         return $this->id;
     }
 
-    /**
-     * @return \Doctrine\Common\Collections\ArrayCollection
-     */
-    public function getOrders()
-    {
-        return $this->orders;
```
jochenmanz: properties do not exists, could be replaced with a return null if these functions need to be here

engine/Shopware/Models/Document/Order.php
shopwareBot: Could you make for this change a seperate Pull request? It looks like the only change which could introduce a break.

jochenmanz: i have removed the addition of the $position property :)

jochenmanz:
@mitelg i have merged the additional model corrections changes into this one and also fixed even more typehints, phpdocs, doctrine annotations and broken code

engine/Shopware/Models/Article/Configurator/Template/Template.php
```diff
@@ -87,94 +88,98 @@ class Template extends ModelEntity
...
      */
-    private $additionalText = null;
+    private $additionalText;
 
     /**
-     * @var int
+     * @var bool
      *
-     * @ORM\Column(name="active", type="integer", nullable=false)
+     * @ORM\Column(name="active", type="boolean", nullable=false)
```
shyim: That is a Breaking Change, please check also the extjs model type definition

engine/Shopware/Models/Media/Media.php
```diff
@@ -203,21 +211,23 @@ class Media extends ModelEntity
...
+     *
      * @ORM\Column(name="created", type="date", nullable=false)
      */
     private $created;
 
     /**
      * Internal container for the uploaded file.
      *
-     * @var UploadedFile
+     * @var File
```
shyim: You have forgotten the import for File

engine/Shopware/Models/Newsletter/Address.php
```diff
@@ -192,15 +193,15 @@ public function getIsCustomer()
      */
     public function setLastMailingId($lastMailingId)
     {
-        $this->lastMailingId = $lastMailingId;
+        $this->lastNewsletterId = $lastMailingId;
```
shyim, line 196: Could you create additional methodes setLastNewsletterId. These are needed for the fromArray functionality of models

engine/Shopware/Models/ProductStream/ProductStream.php
```diff
@@ -190,23 +192,25 @@ public function getAttribute()
     /**
      * @param \Shopware\Models\Attribute\ProductStream|array|null $attribute
      *
-     * @return \Shopware\Models\Attribute\ProductStream
+     * @return ProductStream
      */
     public function setAttribute($attribute)
     {
-        return $this->setOneToOne($attribute, '\Shopware\Models\Attribute\ProductStream', 'attribute', 'productStream');
+        $this->setOneToOne($attribute, '\Shopware\Models\Attribute\ProductStream', 'attribute', 'productStream');
```
shyim: This change makes no difference. setOneToOne returns the current entity.

engine/Shopware/Models/Property/Option.php
```diff
@@ -196,10 +198,12 @@ public function getAttribute()
     /**
      * @param \Shopware\Models\Attribute\PropertyOption|array|null $attribute
      *
-     * @return \Shopware\Models\Attribute\PropertyOption
+     * @return Option
      */
     public function setAttribute($attribute)
     {
-        return $this->setOneToOne($attribute, '\Shopware\Models\Attribute\PropertyOption', 'attribute', 'propertyOption');
+        $this->setOneToOne($attribute, '\Shopware\Models\Attribute\PropertyOption', 'attribute', 'propertyOption');
```
shyim: Also here

engine/Shopware/Models/ProductStream/ProductStream.php
jochenmanz: Acutally the declared return type differs. The setOneToOne returns a ModelEntity, which is not the same as the setAttribute function wants (true for all the other locations as well). That way even phpstorm does not throw a warning at this line.

shyim: I dont like the solution. I would just fix the return value of ModelEntitiy instead to @return $this
That fixes the warning and nothing must be changed here :wink: 

```
Index: engine/Shopware/Components/Model/ModelEntity.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- engine/Shopware/Components/Model/ModelEntity.php	(revision b5ac0f0812dc75f347b804e6c9ced7db2c123ed9)
+++ engine/Shopware/Components/Model/ModelEntity.php	(date 1536134771000)
@@ -95,7 +95,7 @@
      * @param string                 $property  Name of the association property, example: 'orders'
      * @param string|null            $reference Name of the reference property, example: 'customer'
      *
-     * @return ModelEntity
+     * @return $this
      */
     public function setOneToOne($data, $model, $property, $reference = null)
     {
```

jochenmanz: thats also a possible fix :) i will implement that instead

.php_cs.dist
```diff
@@ -31,10 +31,11 @@ our trademarks remain entirely with us.
 EOF;
 
 return PhpCsFixer\Config::create()
-    ->setUsingCache(false)
+    ->setUsingCache(true)
     ->setRules([
         '@PSR2' => true,
         '@Symfony' => true,
+            '@DoctrineAnnotation'                       => true,
```
shyim: Indentation

jochenmanz: i will revert that one - this should not be in this pr :D

engine/Shopware/Models/Article/Configurator/Template/Template.php
jochenmanz: i have also changed the type in the configurator_template.js or should i revert that change?

engine/Shopware/Models/Newsletter/Address.php
jochenmanz: added

engine/Shopware/Models/Document/Order.php
```diff
@@ -532,9 +517,9 @@ public function processPositions()
                         // Fix defined tax
                         $temporaryTax = (int) $ticketResult['taxconfig'];
                         $getTaxRate = Shopware()->Db()->fetchOne("
-                        SELECT tax FROM s_core_tax WHERE id = $temporaryTax
+                            SELECT tax FROM s_core_tax WHERE id = $temporaryTax
                         ");
-                        $position['tax'] = $getTaxRate['tax'];
+                        $position['tax'] = $getTaxRate;
```
jochenmanz, line 522: possible bugfix

jochenmanz:
commits squashed :)

PR #1752, opened 2018-08-11
Plugin commands clear caches using normalized cache clearing workflow

JoshuaBehrens:
There is [a hint](https://github.com/shopware/shopware/blob/f32eba00c30de2f4098a6f5083879b63ad6e597c/engine/Shopware/Commands/PluginActivateCommand.php#L90) that I should clear the cache after installing a plugin. Why not clear it directly?

The scheduled caches from the plugin installation change are passed to the backend cache controller. This is replicated in the commands to change a plugin installation state. Both workflows are using the same methods, which are now extracted into the cache service.

1. Activate a plugin that adds a feature to the storefront via command line
2. Reload page
3. See no changes
4. Facepalm because one forgot the cache clear which is done by the backend automatically
5. `sw:cache:clear`

`--clear-cache` is a new option for some plugin commands to enable cache clearing as this is now part of the routines. This might change scripts in the composer-project. There is a new cache tag: search.

### 5. Checklist

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/InstallerService.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/InstallerService.php
@@ -24,6 +24,7 @@
 
 namespace Shopware\Bundle\PluginInstallerBundle\Service;
 
+use Shopware\Components\CacheManager;
 use Shopware\Components\Model\ModelManager;
 use Shopware\Components\Plugin\ConfigReader;
 use Shopware\Components\Plugin\ConfigWriter;
@@ -341,7 +342,21 @@ class InstallerService
         }
 
         if (array_key_exists('invalidateCache', $result)) {
-            $context->scheduleClearCache($result['invalidateCache']);
+            foreach ($result['invalidateCache'] as $cacheTag) {
+                if ($cacheTag === 'frontend') {
+                    $cacheTags[] = CacheManager::CACHE_TAG_CONFIG;
+                    $cacheTags[] = CacheManager::CACHE_TAG_TEMPLATE;
+                    $cacheTags[] = CacheManager::CACHE_TAG_THEME;
+                    $cacheTags[] = CacheManager::CACHE_TAG_HTTP;
+                } elseif ($cacheTag === 'backend') {
+                    $cacheTags[] = CacheManager::CACHE_TAG_CONFIG;
+                    $cacheTags[] = CacheManager::CACHE_TAG_TEMPLATE;
+                } else {
+                    $cacheTags[] = $cacheTag;
+                }
+            }
+
+            $context->scheduleClearCache($cacheTags);
         }
 
         if (array_key_exists('message', $result)) {
--- a/engine/Shopware/Commands/PluginActivateCommand.php
+++ b/engine/Shopware/Commands/PluginActivateCommand.php
@@ -25,11 +25,8 @@
 namespace Shopware\Commands;
 
 use Shopware\Bundle\PluginInstallerBundle\Service\InstallerService;
-use Shopware\Components\CacheManager;
-use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
@@ -37,13 +34,15 @@ use Symfony\Component\Console\Output\OutputInterface;
  *
  * @copyright Copyright (c) shopware AG (http://www.shopware.de)
  */
-class PluginActivateCommand extends ShopwareCommand
+class PluginActivateCommand extends PluginCommand
 {
     /**
      * {@inheritdoc}
      */
     protected function configure()
     {
+        parent::configure();
+
         $this
             ->setName('sw:plugin:activate')
             ->setDescription('Activates a plugin.')
@@ -51,12 +50,6 @@ class PluginActivateCommand extends ShopwareCommand
                 'plugin',
                 InputArgument::REQUIRED,
                 'Name of the plugin to be activated.'
-            )
-            ->addOption(
-                'clear-cache',
-                'cc',
-                InputOption::VALUE_NONE,
-                'Clear any caches that are requested by update routines'
             )
             ->setHelp(<<<'EOF'
 The <info>%command.name%</info> activates a plugin.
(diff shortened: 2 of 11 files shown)


JoshuaBehrens:
I am not sure whether the travis errors are part of my PR as they seem to be unrelated.

t2oh4e:
As this differs from the current implementation and from the backend workflow (you have to explicitly clear the caches. It does not happen automatically.) I would prefer to turn it around: Add a flag to explicitly clear the necessary caches instead of adding a flag to disable an automatic cache clear.
Furthermore the linked hint does not tell you that you should clear the caches but that you may want to do it ;)
Any other opinions on that?

JoshuaBehrens:
@t2oh4e Good point that cache clearing is optional in the backend. This change seems reasonable to me. Going to change that.

engine/Shopware/Components/CacheManager.php
```diff
@@ -447,6 +461,61 @@ public function encodeSize($bytes)
...
+     * @param string[] $tags
+     *
+     * @return int
+     */
+    public function clearByTags(array $tags)
+    {
+        $count = 0;
+
+        foreach (array_unique($tags) as $tag) {
+            $count += (int) $this->clearByTag($tag);
```
hlohaus: Die Rückgabe von clearByTag ist void oder false.

mitelg:
@JoshuaBehrens nice work! :+1: 
but please adjust your initial description of the PR, as you did some more changes, than just the commands :wink:

engine/Shopware/Components/CacheManager.php
JoshuaBehrens: I don't see a mixed return type. All I see that I always return false which is wrong. The last return should've been true.

JoshuaBehrens:
Is everyone fine with the recent changes?

jkrzefski:
I added a summary of which caches have been cleared successfully and which have failed to be cleared. Also I included some minor bug-fixes and made a PR to this PR, so my changes are isolated and can be reviewed separately. Take a look at this one: https://github.com/niemand-online/shopware/pull/1

engine/Shopware/Commands/PluginActivateCommand.php
```diff
@@ -85,8 +93,16 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
-        $pluginManager->activatePlugin($plugin);
+        $activationContext = $pluginManager->activatePlugin($plugin);
+        $output->writeln(sprintf('Plugin %s has been activated.', $pluginName));
 
-        $output->writeln(sprintf('Plugin %s has been activated. Consider sw:cache:clear to enable possible behaviors that come with the plugin.', $pluginName));
+        if (!empty($input->getOption('clear-cache'))) {
```
soebbing: You could turn this into an early return

jkrzefski: @soebbing Would you keep the message about the scheduled cache clearing in case it is not dealt with automatically like so?

```php
if (empty($input->getOption('clear-cache'))) {
    $output->writeln(sprintf('Plugin %s has been activated. Consider sw:cache:clear to enable possible behaviors that come with the plugin.', $pluginName));
    return 0;
}
```

soebbing: @jkrzefski Personally, I think that is a nice addition

engine/Shopware/Components/CacheManager.php
hlohaus: Ich meine nur, macht das dann mit den Zahlen addieren sinn?

jkrzefski: The initial idea of this was to return the amount of successfully cleared caches. Personally I do not think of this feature as very useful. That's why I changed the return of this method from an integer (how many caches were cleared successfully) to an array of strings (which caches were cleared successfully) in my pr to this pr (see https://github.com/niemand-online/shopware/pull/1 ).

TL;DR: I do agree with you that this implementation does not make much sense.

JoshuaBehrens:
I added more changes heavily inspired by @jkrzefski PR (which does not render obsolete) and @soebbing recent comment.

jkrzefski:
Before this PR gets merged, please take a look at https://github.com/niemand-online/shopware/pull/1 which will add a summary output to console commands. Also that PR will add support for legacy plugins so I would strongly suggest to include it in this PR.

JoshuaBehrens:
Done @shopwareBot

engine/Shopware/Controllers/Backend/Cache.php
```diff
@@ -131,29 +131,26 @@ public function clearCacheAction()
...
-        }
-        if ($cache['proxy'] === 'on') {
-            $this->cacheManager->clearProxyCache();
-            $this->cacheManager->clearOpCache();
+        $cacheTags = [];
+
+        foreach ($cache as $cacheTag => $value) {
+            if ($value === 'on') {
+                if ($cacheTag === 'frontend') {
+                    $cacheTags[] = CacheManager::CACHE_TAG_CONFIG;
```
hlohaus, line 139: I saw it already above? Duplicate Code

JoshuaBehrens: Pardon, to what is it a duplicate to?

engine/Shopware/Bundle/PluginInstallerBundle/Service/InstallerService.php
```diff
@@ -341,7 +342,21 @@ private function applyLegacyResultToContext($result, InstallContext $context)
         }
 
         if (array_key_exists('invalidateCache', $result)) {
-            $context->scheduleClearCache($result['invalidateCache']);
+            foreach ($result['invalidateCache'] as $cacheTag) {
```
hlohaus, line 345: here. maybe you can do this in clearByTag

jkrzefski: Similar indeed, but not the same. To combine these in a common method seems possible, but I am unsure if more changes from us would have any effect now, since the pr is already closed.

hlohaus: mhh. Shopware seems to have forgotten it:
https://github.com/shopware/shopware/commit/99369a56b0bc5cc472c22c174ead25e02d2504de
Is that bad?

jkrzefski: As long as it works, I don't think it is bad. It is just one thing that *could* be better. Maybe you can add another PR on top of [this commit](https://github.com/shopware/shopware/commit/99369a56b0bc5cc472c22c174ead25e02d2504de)?

JoshuaBehrens: These are two different sources of tags. A backend cache manager way of handling tags and a plugin manager way of handling tags. Although they are similar they are not equal yet. I would be happy if they are the same.

PR #1754, opened 2018-08-12
let php-cs-fixer fix all the files

jochenmanz:
shopware is shipping a .php_cs.dist file to fix the code style automatically according to shopwares own defined rules. But unfortunately these fixes were not applied a long time now. To increase the DX of shopware, this feature is quite nice and helpful and simple to make it working again.

let php-cs-fixer fix all the files...

run ./vendor/bin/php-cs-fixer in the root directroy of shopware

### 6. Checklist


engine/Shopware/Components/DependencyInjection/LegacyPhpDumper.php
```diff
@@ -23,29 +23,28 @@
  * our trademarks remain entirely with us.
```
shopwareBot: Can you revert this file and add this file as exclude in the php_cs.dist file?

jochenmanz: done. Should i create a second PR with only that file?

jkrzefski:
This issue has recently been bothering me as the pre-commit-hook of this repository requires all changed files not to conflict with the .php_cs.dist file. So I ran `php-cs-fixer fix` and ended up with a bunch of changes completely unrelated to what I was doing. So this pr is actually necessary to reduce noise in future commits.

mitelg:
@jkrzefski limit the fixing to your specific files in the future :wink: 
the pre-commit hook should also suggest you some commands to fix only the files, which you have changed

jkrzefski:
@mitelg Ok, I will keep that in mind. But nevertheless some noise could be created if I want to edit a file that does not comply with the .php_cs.dist file, because I will have to fix the style for the entire file instead of only the changed lines. Don't get me wrong though. I like the idea of a mandatory style fix, but for that to work well, this pr is necessary.

soebbing:
Hi @jochenmanz, thank you for your contribution! 💙

PR #1755, opened 2018-08-13
do not set basketdiscount if it's 0

tinect:
currently if the basketdiscount result 0 it'll be displayed as "FREE" in Basket. Which makes no sense in my eyes.
adjust naming of basket-discount in sAdmin to sBasket`s.

check if $discount is set valid

define a product-price and basket discount which results <0.01€

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -4003,52 +4003,56 @@ SQL;
             [$this->sSYSTEM->sUSERGROUPDATA['id'], $amount]
         );
 
-        if (!empty($basket_discount)) {
-            $percent = $basket_discount;
-            $basket_discount = round($basket_discount / 100 * ($amount * $currencyFactor), 2);
+        if (empty($basket_discount)) {
+            return;
+        }
 
-            if($basket_discount) {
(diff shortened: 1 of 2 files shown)


engine/Shopware/Core/sBasket.php
```diff
@@ -375,6 +375,10 @@ public function sInsertDiscount()
         $discount *= -1;
         $discount = round($discount, 2);
 
+        if(!$discount) {
```
FiveDigital: Please add a space after the `if`

engine/Shopware/Core/sAdmin.php
```diff
@@ -4004,50 +4003,56 @@ private function handleBasketDiscount($amount, $currencyFactor, $discount_tax)
             [$this->sSYSTEM->sUSERGROUPDATA['id'], $amount]
         );
 
-        if (!empty($basket_discount)) {
-            $percent = $basket_discount;
-            $basket_discount = round($basket_discount / 100 * ($amount * $currencyFactor), 2);
+        if (empty($basket_discount)) {
```
shyim, line 4006: This condition is know duplicate see line 4013

tinect: Correct, but the variable is changed between

tinect:
@shopwareBot I'll try these days 👍

PR #1758, opened 2018-08-14
Keep numeric amounts for basket items.

kleinmann:
Those amounts are currently being overwritten with formatted strings, losing the
numeric values. If you need the item-specific amounts in any feature, you would have to duplicate all calculation.

It copies the amounts to separate `xNumeric` keys before converting the values to formatted strings. This is identical to the current handling of the single price.

1. Be in basket
2. Try to fetch calculated amounts for a basket item in a usable numeric way.
3. Fail to do so, because there's no numbers, only strings.

-

See accompanying changes in `UPGRADE-5.5.md`.

### 6. Checklist


kleinmann:
Rebased for 5.5.2.

PR #1763, opened 2018-08-15
Adding "shopware mail template name" to zend mail object

Isengo1989:
There is only one real Event for checking the Mail before sending it `Enlight_Components_Mail_Send` - other events are only notify-events. Therefore it would be nice to know which Mail will be sent (sORDER / sREGISTERCONFIRMATION etc.) in case you want to send a mail with third party (mailchimp, emarsys etc.)

Right now there is no indicator whatsoever which mail will be sent. You can only differ these mails by their subject, which will eventually change (subshop, some admin etc.)

It sets the MessageId ot the Zendmail object, which is never used anyways. It also catches Zend_Mail_Exceptions of setFrom.

Set setMessageId -> have the Shopware Template Name (sORDER / sREGISTERCONFIRMATION etc.) in the mail Object which can be accessed in `Enlight_Components_Mail_Send` 

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/base/component/Shopware.notification.SubscriptionWarning.js
+++ b/themes/Backend/ExtJs/backend/base/component/Shopware.notification.SubscriptionWarning.js
@@ -156,7 +156,7 @@ Ext.define('Shopware.notification.SubscriptionWarning', {
                     Shopware.Notification.createStickyGrowlMessage({
                         title: me.snippets.importantInformation,
                         text: me.snippets.noShopSecretWarning,
-                        width: 330,
+                        width: 460,
                         btnDetail: {
                             text: me.snippets.login,
                             callback: function () {


engine/Shopware/Components/TemplateMail.php
```diff
@@ -274,9 +274,20 @@ public function loadValues(\Enlight_Components_Mail $mail, Mail $mailModel, $ove
...
         } elseif (!empty($fromMail)) {
-            $mail->setFrom($fromMail);
+            try {
+                $mail->setFrom($fromMail);
+            } catch (Zend_Mail_Exception $e) {
+            }
+        }
+
+        try {
+            $mail->setMessageId($mailModel->getName());
```
tinect: There are shops using this... ;-)
Generally, I would suggest not to _misuse_ any existing fields/props or something else. 

Please extend Mail-Object or something else.

tinect (changes requested):

engine/Shopware/Components/TemplateMail.php
Isengo1989: @tinect sure, you're right. I found no usage in the core, so I thought this could be done quick. I will extend the Mail Object.

Isengo1989: @tinect adding the template to the header would not interfere with any functunality
auswahl_103

Also it would allow to match the email later on with rich tools like emarsys. What do you think?

tinect: Yes, great idea. Sounds good :-)

Isengo1989:
Removed **setMessageId** and added the value to the **header** instead.

engine/Shopware/Components/TemplateMail.php
```diff
@@ -274,9 +274,20 @@ public function loadValues(\Enlight_Components_Mail $mail, Mail $mailModel, $ove
         }
 
         if (!empty($fromMail) && !empty($fromName)) {
-            $mail->setFrom($fromMail, $fromName);
+            try {
+                $mail->setFrom($fromMail, $fromName);
+            } catch (Zend_Mail_Exception $e) {
```
shopwareBot: Please don't make empty try catches

engine/Shopware/Components/TemplateMail.php
```diff
@@ -274,9 +274,20 @@ public function loadValues(\Enlight_Components_Mail $mail, Mail $mailModel, $ove
...
         } elseif (!empty($fromMail)) {
-            $mail->setFrom($fromMail);
+            try {
+                $mail->setFrom($fromMail);
+            } catch (Zend_Mail_Exception $e) {
+            }
+        }
+
+        try {
+            $mail->addHeader('Template', $mailModel->getName());
```
shopwareBot: We don't want to expose the used template name to the recipient.

Isengo1989: @shopwareBot too bad :( I will fix the Exceptions according to this https://framework.zend.com/manual/1.10/en/zend.exception.using.html

Would this work for you guys?
```php
    /**
     * Property for the template. Can be filled by setTemplate function.
     *
     * @var null
     */
    protected $_swTemplate = null;

   /**
     * @return null
     */
    public function getSwTemplate()
    {
        return $this->_swTemplate;
    }

    /**
     * @param null $swTemplate
     */
    public function setSwTemplate($swTemplate)
    {
        $this->_swTemplate = $swTemplate;
    }
```

shopwareBot: Please remove the "Sw" prefix and the underscore from the property.

Isengo1989:
@shopwareBot done.

engine/Library/Enlight/Components/Mail.php
```diff
@@ -70,6 +70,13 @@ class Enlight_Components_Mail extends Zend_Mail
      */
     protected $_plainSubject = null;
 
+    /**
+     * Property for the template. Can be filled by setTemplate function.
+     *
+     * @var null
+     */
+    protected $Template = null;
```
shyim: properties starts with lower 😊

Isengo1989: @shyim argh this PR is cursed - well next try tmrw :laughing:

shyim:
@Isengo1989 please create with the newest commit a new pull request :)

shyim:
Wrong rebase? :D

Isengo1989:
> Wrong rebase? :D

WTF - was hab ich getan o.O

Isengo1989:
Sorry my bad, I am not sure what I did. But I guess it was wrong. It was no rebase tho. I fetched the upstream as usual, merged the upstream/5.5 in my fork and then.

Well anyways - new PR is done ^^

Isengo1989:
@mitelg what happend to the PR?

mitelg:
@Isengo1989 I don't know what you have done 🙂 you closed the PR somehow and I only closed the related things 😉
maybe you could just reopen a new clean PR

Isengo1989:
> @Isengo1989 I don't know what you have done  you closed the PR somehow and I only closed the related things 
> maybe you could just reopen a new clean PR

I think I forced annother PR/Change and overwrote my changes here -.- The changes are nowhere to be found - damn it

Well I will have to create a new one I guess

mitelg:
yeah I think so :frowning_face: 
Unfortunately I cannot even re-open the PR

PR #1764, opened 2018-08-15
Fix order state change and mail window when opened via quick search

buddhaCode:

The order state change and the mail window depends on the order list. Both try to update/reload the order list, when state is changed or a document is added. This dependency throws an error, if the order isn't oped via the order list window, but from the quick search bar in the blue top bar. When one opens an order via quick search there is simply no order list window available.

This change introduces a try/catch block, when accessing the order list object. It will simply not reloaded, if it's not available.

Working example:

1. Open in backend "Kunden -> Bestellungen"
2. Edit an order
3. Change order or payment state, to a state with an email associated. E.g. "Bestellstatus: In Bearbeitung" (not sure if this is really standard or our customization ;)
4. Save order
5. Growl notification: Succesfull 
5. Mail window opens

Not working example:

1. Look for existing order number in quick search
2. Pick order to open
3. Change order or payment state, to a state with an email associated. E.g. "Bestellstatus: In Bearbeitung" (not sure if this is really standard or our customization ;)
4. Save order
5. Growl notification: Succesfull 
5. Mail window does NOT open. There is an error on JavaScript console.

Don't know.

None.

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/order/controller/attachment.js
+++ b/themes/Backend/ExtJs/backend/order/controller/attachment.js
@@ -184,9 +184,14 @@ Ext.define('Shopware.apps.Order.controller.Attachment', {
     callStoreReload: function(attachmentGrid, addAsAttachment, listStore) {
         var me = this;
 
-        listStore.reload({
-            callback: Ext.bind(me.applyNewDocument, me, [attachmentGrid, addAsAttachment])
-        });
+        try {
+            listStore.reload({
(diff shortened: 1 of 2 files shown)


buddhaCode:
Yes, same problem. Different solution ;) But should work. But I guess Pascal's solution will reload the store. This should be unnecessary because the list windows is not loaded anyway.

PR #1772, opened 2018-08-21
WIP: Use correct Attribute data types in PHP

hranicka:

If I specify some attribute as a boolean, I expect boolean type also in PHP (e.g. $category->getAttribute()->getMyBoolean() should really return PHP `bool`).

For now, there are just tests as a proof of the problem.

Fix has to be added.

Create a new attribute with type "boolean", for instance like:

```php
        $attributes->update('s_categories_attributes', 'my_boolean', 'boolean', [
            'label' => 'Is it true?',
        ], null, false, false);
```

Then when accessing that attribute in PHP via generate Attribute class, value is integer, not bool.

### 6. Checklist


shyim:
I guess the main problem will be that this fix will introduce a break for plugins which depends on a integer value.

hranicka:
So it's correct that `boolean` attributes have `integer` data type in PHP?

PR #1773, opened 2018-08-21
Ask password with a hidden input in sw:admin:create

JoshuaBehrens:
There is currently no way to create an administration account without a visible password.

In the interactive way the command asks for missing information. Now if the missing information is the password there is a input used that does not print what you type and therefore a way to enter a password without being visible to others seeing the output of the terminal.

1. Execute `sw:admin:create`
2. Fill in all the questions being asked
3. Wonder why you see the password while you type it

### 4. Checklist


shyim:
Duplicates #1675

JoshuaBehrens:
Welp. As #1675 asks twice for the password which is better (although it uses a workaround that I do not like by using a second option). My work here was done.

PR #1775, opened 2018-08-21
Benchmark bundle corrections

jochenmanz:
enhance the developer experience :)

corrects misleading typehints, wrong types and wrong parameters

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/BenchmarkBundle/Provider/ShopwareProvider.php
+++ b/engine/Shopware/Bundle/BenchmarkBundle/Provider/ShopwareProvider.php
@@ -123,7 +123,7 @@ class ShopwareProvider implements BenchmarkProviderInterface
     }
 
     /**
-     * @return string
+     * @return string|false
      */
     private function getPhpVersion()
     {
@@ -147,7 +147,7 @@ class ShopwareProvider implements BenchmarkProviderInterface
     }
 
     /**
-     * @return string
+     * @return int
      */
     private function getMemoryLimit()
     {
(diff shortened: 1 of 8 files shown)


engine/Shopware/Bundle/BenchmarkBundle/Provider/ShopwareProvider.php
```diff
@@ -123,7 +123,7 @@ private function getServerSoftware()
     }
 
     /**
-     * @return string
+     * @return string|false
```
shyim, line 126: Why should that give false back?

jochenmanz: because phpversion can return a false in some cases. Alternatively a (string) in front would fix the type issue :)

shyim: phpversion never give false with no arguments 😉

teiling88: https://3v4l.org/hYfUZ

jochenmanz: would it be ok if we put a (string) in front? the declared return type of that function is string|false as these cannot be conditional :(

shopwareBot: A string cast would be fine! :-)

shopwareBot: @jochenmanz Still working on it?

jochenmanz: jep, just started again :) will be ready shortly

jochenmanz:
merged into the new pr https://github.com/shopware/shopware/pull/1792

PR #1776, opened 2018-08-21
Search bundle corrections

jochenmanz:
better developer experience

fixes some incorrect phpdoc type hints

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/VariantHelper.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/VariantHelper.php
@@ -28,7 +28,7 @@ use Doctrine\DBAL\Connection;
 use Shopware\Bundle\SearchBundle\Condition\VariantCondition;
 use Shopware\Bundle\SearchBundle\Criteria;
 use Shopware\Bundle\SearchBundle\Facet\VariantFacet;
-use Shopware\Bundle\StoreFrontBundle\Gateway\CustomFacetGatewayInterface;
+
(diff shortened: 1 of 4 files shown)


engine/Shopware/Bundle/SearchBundleDBAL/VariantHelper.php
```diff
@@ -45,7 +45,7 @@ class VariantHelper implements VariantHelperInterface
     protected $connection;
 
     /**
-     * @var CustomFacetGatewayInterface
+     * @var CustomListingHydrator
      */
     protected $customFacetGateway;
```
mitelg, line 50: how about a renaming? :blush:

teiling88: It would be a breaking change in my opinion.

mitelg: wasn't the visibility changed a few days ago? of course this is a BC, but would be ok imho, if this change was made in the RC

jochenmanz: not according to git blame. Should i include this change in the new PR?

mitelg: that were other properties. so I think we have to stick with the name

jochenmanz:
merged into the new pr https://github.com/shopware/shopware/pull/1792

PR #1777, opened 2018-08-22
Fix smarty flink function when rendered in cli context

jkrzefski:
When a template is rendered a cli environment there is no request object. Since the request is essentially null, the `flink` function will not render a base url even is `fullPath` is set to `true`. To initialize a shop context in a cli environment, one may call `\Shopware\Models\Shop\Shop::registerResources()` which will set a context to the router. That would be sufficient to let the router know about the base url and include it in generated urls. However that does not help at all when the `flink` function simply prepends the scheme and http host of the current request to the rendered url. The request is still null at this point, remember? So I suppose to use the router's `assemble` method to generate the base url of the given context.

When the smarty function `flink` is called with `fullPath` and the request is null, the base url is retrieved from the router's `assemble` method instead of *not at all*.

1. Make a simple template that uses `{link file="..." fullPath}`.
2. Call the router's `setContext` method to initialize it with a certain shop.
3. For comparison call the router's `assemble` method.
4. Execute the script in a cli environment.
5. See that `fullPath` has no effect. *Also a leading slash is missing.*
6. See that `assemble` as succeeded in rendering the base url.

None. (yet)

Maybe some, maybe none...

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Template/Plugins/function.flink.php
+++ b/engine/Library/Enlight/Template/Plugins/function.flink.php
@@ -61,7 +61,7 @@ function smarty_function_flink($params, $template)
 
         // Some cleanup code
         if (strpos($file, $docPath) === 0) {
-            $file = substr($file, strlen($docPath));
+            $file = '/' . substr($file, strlen($docPath));
         }
 
         // Make sure we have the right separator for the web context
@@ -82,8 +82,14 @@ function smarty_function_flink($params, $template)
         $file = $request->getBasePath() . '/';
     }
 
-    if ($request !== null && !empty($params['fullPath']) && strpos($file, '/') === 0) {
-        $file = $request->getScheme() . '://' . $request->getHttpHost() . $file;
+    if (!empty($params['fullPath']) && strpos($file, '/') === 0) {
+        if ($request === null) {
+            $baseUrl = Shopware()->Container()->get('router')->assemble();
+        } else {
+            $baseUrl = $request->getScheme() . '://' . $request->getHttpHost();
+        }
+
+        $file = $baseUrl . $file;
     }
 
     return $file;


engine/Library/Enlight/Template/Plugins/function.flink.php
```diff
@@ -61,7 +61,7 @@ function smarty_function_flink($params, $template)
 
         // Some cleanup code
         if (strpos($file, $docPath) === 0) {
-            $file = substr($file, strlen($docPath));
+            $file = '/' . substr($file, strlen($docPath));
```
shopwareBot, line 64: Is there any way to implement this change only in your new case?

Otherwise, can you specify why this change is necessary?

jkrzefski: As I understand this section of code it should trim the docPath off the beginning of the file path. So I assumed that in a web context that equals to a relative path with a leading slash, since it relates to the document root.

soebbing: Background of the question is: It doesn't look directly related to your fix and might break some behavior. I think it might be saver to make changes only in the case you're trying to fix.

btw: Do you see a way to add a test to this case?

jkrzefski: @soebbing It does relate to my fix since I only prepend the scheme and host later on. Without this slash it would miss a slash between the host and the uri. That's why I wanted to include a leading slash and I figured, it would fit here best, because only in this case it is made sure that the relative path relates to the document root. But if that is too unsafe I will update my PR to add the slash further down.

Honestly I have not yet created a custom test for shopware so I do not feel comfortable adding one here. If someone else would be willing to include a test for this, feel free to contribute to this PR.

jkrzefski: @soebbing What has initially lead me to adding this slash here is the condition in this line: https://github.com/shopware/shopware/blob/70f7c569660b08e93eaf9fe6784bab3d9144e26a/engine/Library/Enlight/Template/Plugins/function.flink.php#L85
The condition `strpos($file, '/') === 0` suggests that a leading slash is indeed valid in the file path and does directly correlate to the `fullPath` parameter. I have not found any possibility for a leading slash to exist in the file path prior to my change so I figured it made sense to add it when the path relates to the document root.

shyim:
Hey @jkrzefski,

thanks for the Contribution :blue_heart:
i have tested your Pull Request with multiple shop configurations

* Shop with Virtual URL => Link is broken, cause contains the virtual url
* Shop with Base Path => Links is broken, contains two times the base path, (once from $file, and once from assembled url)

Maybe the better way would be here instead using assemble from router, to just building the link by own with the 'shop' from the container.

jkrzefski:
Oh damn, I didn't think of those scenarios. But I do not like your suggestion to just build the link by hand. I will try to find a way that covers every possible configuration.

shyim:
I guess using the router should be impossible there, cause it generates links to controllers, and there is always basepath or virtual url considered :)

MrTschi:
As written by @shyim 
> Shop with Base Path => Links is broken [...]

the problem still exists within the latest Shopware version :(

PR #1778, opened 2018-08-23
consider active flag of shop-pages for sitemap-generation

mnaczenski:
Currently inacitive shop-pages will appear in the sitemap, that's bad for SEO and the user experience.

consider the activte flag while fetching the shop-pages

add a content-group to your shop and deativate one of the shop pages
 -> check the sitemap.xml

### 6. Checklist


mitelg:
@shopwareBot we should implement some events within this PR at the end of each private Method. so it is easier for plugin developers to change the sitemaps

shyim:
That class is not used anymore in 5.5 and is there because of compability reasons for plugins. Please add your change also to https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Bundle/SitemapBundle/Provider/StaticUrlProvider.php

mnaczenski:
@shyim 
thank you for the hint. I will add it to the new StaticUrlProvider tomorrow!

PR #1779, opened 2018-08-23
Always set given config value in config writer

JoshuaBehrens:
Imagine a boolean config element which defaults to falsey and is changable for every sub-/language shop. The loading hierarchy of the config reader is *requested shop*, *parent shop* (if given), *shop with id = 1*, *element default*.

This change allows to set the config value for the requested shop to the default value without always falling back to shop with id 1 setted value.

Always write a given value except it is null to match the coalasce in the reader.

Have the following shop tree.
```
|--* 1 Pan cake mixes
|--+ 2 Flux compensators
   |--* 3 Flux compensatoren (NL)
```

Have plugin (maybe for [Regulation (EU) 1169/2011](https://eur-lex.europa.eu/legal-content/DE/TXT/?uri=CELEX:32011R1169) better known as LMIV in :de:). This plugin has configuration element which allows being changed for every sub-/language shop (scope = 1) and a default value of null. For example the setting for being enabled.
One sets the enabled configuration value to true for the pancake shop (id: 1). Therefore it is enabled everywhere. Afterwards you disable it for the flux compensators shop (id: 2). This is denied by the config writer as the false value is loosely equals to the null default value. Therefore it skips the writing of the value. The reader now falls back to shop with id 1 and reads the value true.

Maybe some comments in the documentation about the config reader/writer have to be changed.

### 5. Checklist

### DIFF
--- a/engine/Shopware/Components/Plugin/ConfigWriter.php
+++ b/engine/Shopware/Components/Plugin/ConfigWriter.php
@@ -106,7 +106,7 @@ class ConfigWriter
         $valueModel = $this->valueRepository->findOneBy(['shop' => $shop, 'element' => $element]);
 
         if (!$valueModel) {
-            if ($value == $defaultValue || $value === null) {
+            if ($value === null) {
                 return;
             }
 
@@ -121,7 +121,7 @@ class ConfigWriter
             return;
         }
 
-        if ($value == $defaultValue || $value === null) {
+        if ($value === null) {
             $this->em->remove($valueModel);
         } else {
             $valueModel->setValue($value);


JoshuaBehrens:
The error is not related to the PR.

```
check-code-style:
     [exec] Loaded config default from "/home/travis/build/shopware/shopware/.php_cs.dist".
     [exec]    1) shopware/engine/Shopware/Core/sAdmin.php (method_argument_space, header_comment, no_whitespace_in_blank_line, no_blank_lines_after_phpdoc)
```

JoshuaBehrens:
What is the expected behaviour? If I know this I can work on that issue. If I am right the current situation is not the expected behaviour. Please provide a case that I break.

If the result of my work fixes a bug which exists since two years I am not allowed to fix this as everyone already had to live with it as it is? :confused:

This is getting to be a difficult task as no one can foreshadow how many people are using the cli and expected this behaviour. And if they expect this behaviour how did they not wonder why this behaviour is like it is and PRd before?

dneustadt:
> unfortunately this would result in a BC break.

Seriously? This PR fixes a bug. In fact it's a sloppy beginner's trap to not use the identity operator when the type of a value is meaningful. Since config values should be a generic way of storing data, they should fit every use case and that includes exact type comparison.
It's frustrating to see contributions regarding serious issues are being ignored or disregarded in favour of convenience.

soebbing:
Unfortunately it's not that easy. This change might fix the bug at hand, but it would also break the default value inheritance chain, leading to a whole set of new problems.

@JoshuaBehrens and I had a long chat on gitter over this issue on Friday. We discussed a possible solution but we'll have to see if that is feasible.

jkrzefski:
@soebbing Correct me if I misunderstand this, please. I just want to know if I understand this correctly.

The inheritance chain stays intact, because this pr only touches the writer and not the reader. So the only breaking case I can come up with is a script setting a config value of a subshop and *expecting for it to have no effect* on purpose. When a script sets a config value of a subshop to its default value that script would have to *expect this to have no effect* for this to be a breaking change.

Am I missing something here? This seems just too obvious to me.

soebbing:
The new writer writes the defaults as values for subshops. Now image you change the default on the parent element. That won't have any effect on the child as it has the old default written as a value. This is a change in behavior.

jkrzefski:
@soebbing The key part for my confusion is that the default value is only saved in the database if the writer puts it there which imho could never happen accidentally. As I see it, the only way for the writer to save the default value to the database is when the writer is explicitly asked to do so. And that is also the reason why I fail to see the problem here.
But maybe I‘m missing something crutial here. Is there a case where the writer would save the default value while not explicitly being asked to do so?

soebbing:
> The key part for my confusion is that the default value is only saved in the database if the writer puts it there which imho could never happen accidentally

Maybe I am missing something, but Isn't that what this PR is doing? The condition
`if ($value == $defaultValue || $value === null) {` should prevent default values from being written to the database, but that check got removed.

jkrzefski:
A default value is only written to the database if the writer is requested to do so, e.g. by the `sw:plugin:config:set` command. Personally I cannot imagine why the expected behaviour should be for this command to have no effect at all if the value I tried to set with it equals the default value. The only logic I can come up with is:
„I tell the writer to set the config element foo to the value bar. Afterwards I expect the config element foo to have the value bar stored in the database.“

Also I believe it is no expected behaviour either to let a change in the default value have an impact on a configuration. That would imply that when I leave a config value for a subshop on its default, it is subject to change if the plugin manufacturer decided to change to default value. That does not feel right.

JoshuaBehrens:
> mfw I peek into this PR again

peek into thread

JoshuaBehrens:
I just pushed my local tests that majorly succeed and some who fail to see the problem

https://github.com/niemand-online/shopware/tree/no/bugfix-config-writer-inheritance

JoshuaBehrens:
@shopwareBot I am still working on a solution that makes the tests :green_heart: and :blue_heart: This will eventually be a new PR.

soebbing:
Thank you @JoshuaBehrens!

JoshuaBehrens:
Followup is out now https://github.com/shopware/shopware/pull/1802

PR #954, opened 2017-02-05
Refactored xml reader + tests + misc stuff

nlubisch:
## Description
* Why is it necessary?
Refactored xml readers because every programmer does his own thing.
There were some redundant functions which i moved to a abstract base class.
All readers are now unified and 100% tested.

screenshot from 2017-02-05 19-37-41
screenshot from 2017-02-05 19-38-09

* What does it improve?
It does improve code, tests and xsd files.

**BC breaks?**
-> config.xsd
storeType has changed. PHP got removed due to when inserting an php array as CDATA you have to eval it to get the data and that is IMO a security issue.
There is now an explicit way to declare which type of store it is.

-> plugin.xsd
pluginName-attribute in "requiredPlugin" is now required.
complexType "scope" got removed because its not used.

**Tests pass?**
no, because there are 9 failures in non related tests
> There were 9 failures:
> 
> 1) Shopware\Tests\Functional\Bundle\AccountBundle\Controller\AddressTest::testChangeOfBillingAddressReflectsInAccount
> Failed asserting that two strings are equal.
> --- Expected
> +++ Actual
> @@ @@
>  'Herr
>  Shop ManMusterstr. 5555555 Musterhausen
> -Nordrhein-WestfalenDeutschland'
> +Nordrhein-WestfalenGermany'
> 
> /home/vagrant/www/tests/Functional/Bundle/AccountBundle/Controller/AddressTest.php:315
> 
> 2) sAdminTest::testsGetCountryList
> Failed asserting that actual size 1 matches expected size 0.
> 
> /home/vagrant/www/tests/Functional/Core/sAdminTest.php:894
> 
> 3) Shopware\Tests\Functional\Modules\Category\sGetCategoriesTest::testWithMainId
> Failed asserting that actual size 9 matches expected size 8.
> 
> /home/vagrant/www/tests/Functional/Modules/Category/sGetCategoriesTest.php:77
> 
> 4) Shopware\Tests\Functional\Modules\Category\sGetCategoriesTest::testWithSecondLevel
> Failed asserting that actual size 9 matches expected size 8.
> 
> /home/vagrant/www/tests/Functional/Modules/Category/sGetCategoriesTest.php:98
> 
> 5) Shopware\Tests\Functional\Modules\Category\sGetCategoriesTest::testWithThirdLevel
> Failed asserting that actual size 9 matches expected size 8.
> 
> /home/vagrant/www/tests/Functional/Modules/Category/sGetCategoriesTest.php:135
> 
[...]

PR #1781, opened 2018-08-26
Refactored xml parser + tests

nlubisch:
Reopened as the topic arised in Gitter.im.
Old PR: #954
Ticket: https://issues.shopware.com/issues/SW-17837

## Description
* Why is it necessary?
Refactored xml readers because every programmer does his own thing.
There were some redundant functions which i moved to a abstract base class.
All readers are now unified and 100% tested.

screenshot from 2017-02-05 19-37-41
screenshot from 2017-02-05 19-38-09

* What does it improve?
It does improve code, tests and xsd files.

**BC breaks?**
-> config.xsd
storeType has changed. PHP got removed due to when inserting an php array as CDATA you have to eval it to get the data and that is IMO a security issue.
There is now an explicit way to declare which type of store it is.

-> plugin.xsd
pluginName-attribute in "requiredPlugin" is now required.
complexType "scope" got removed because its not used.

**Tests pass?**
No, but only non related tests fail.


shyim:
Code-Style check fails

nlubisch:
Sure, i can. Give me some time :)

nlubisch:
Removed the misc stuff and opening a new PR with them in a minute.
I'm looking for the failing test this evening.

nlubisch:
The test succeeds locally. Don't see the problem here :(

shyim:
Same problem also local https://dpaste.de/i0yR

PR #1785, opened 2018-08-30
Make plugin loading order predictable

htuscher:

When loading the plugins, the DirectoryIterator which scans for plugins is not working
in a predictable order, because it uses the ctime of the directories as sorting order.

It makes the loading order of the plugins predictable (simply alphabetically).

This is just a workaround, there should be a meaningful plugin loading order according
to their dependencies (either composer or requiredPlugins from XML).

If you try to override a symfony parameter or service of another plugin, you can't be sure
that it works consistently due to deviations of the ctime values of the plugin directories.
This can be caused by deployment, etc.

https://issues.shopware.com/issues/SW-22399

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInitializer.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInitializer.php
@@ -68,9 +68,8 @@ class PluginInitializer
         $classLoader = new Psr4ClassLoader();
         $classLoader->register(true);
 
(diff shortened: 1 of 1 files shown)


uehler:
Why don't you sort them in the query?

htuscher:
I could, but it wouldn't make that much difference. The really necessary part is sorting after the DirectoryIterator scanned for directories. I just used it in both places for consistency.

htuscher:
I don't think it would break anything because it's currently simply unpredictable. Of course I would prefer a solution which respects plugin dependencies, but as this should be fixed fast and in my opinion in 5.4 (or even 5.3), I would like you to discuss it internally.

PR #1787, opened 2018-09-03
Improve ES Indexing Performance

htuscher:

The `\Shopware\Bundle\ESIndexingBundle\Product\ProductListingVariationLoader::fetchPrices` method took over 5 seconds for 10 products on MariaDB 10.3.

It helps the QueryOptimizer to reduce the rows returned by a subquery to the necessary. This speeds up the constructed query from 5s down to 20ms.

Try to index products with variants and multiple customergroups.


soebbing:
Hi @hhoechtl, thank you for your contribution! 💙

PR #1788, opened 2018-09-03
redesign comment in script-async-ready.tpl

tinect:
The comment breaks javascript in different scenario. F.e. html-Compressor, Browser-Addons, too.
literal will prevent any smarty-function to remove comment.

Maybe we could delete the comment completly?

it changed the comment-style.

short way to reproduce: {strip} around. The comment-line will be in one line with following stuff, so it won't be executable.

### 6. Checklist


soebbing:
Hi @tinect, thanks for your contribution!

PR #1790, opened 2018-09-03
Add notification plugin querybuilder event

larsbo:
With this event you can add more conditions that must be met before a product will be considered in the notifiation loop.

It adds a new `notify` event which allows the modification of the variant query builder.

E.g. adding a new attribute which should affect if a notification mail should be send or not.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
@@ -290,7 +290,9 @@ class Shopware_Plugins_Frontend_Notification_Bootstrap extends Shopware_Componen
             ->fetchAll();
 
         foreach ($notifications as $notify) {
-            $product = $conn->createQueryBuilder()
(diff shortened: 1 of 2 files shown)


engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
```diff
@@ -305,9 +307,16 @@ public static function onRunCronJob(Shopware_Components_Cron_CronJob $job)
                 ->where('d.ordernumber = :number')
                 ->andWhere('d.instock > 0')
                 ->andWhere('d.minpurchase <= d.instock')
-                ->setParameter('number', $notify['ordernumber'])
-                ->execute()
-                ->fetch(\PDO::FETCH_ASSOC);
+                ->setParameter('number', $notify['ordernumber']);
+
+            Shopware()->Events()->notify(
```
shopwareBot: Please use the Event Manager from the DI Container instead.

larsbo:
Done 👍

PR #1794, opened 2018-09-05
prevent display of "price starting from" because of price differences…

mkreusch:
… less than 0.01

Without this change the product price will show the "price starting from" (="ab ...") text even if the price difference between variants is lower than 0.01 and therefore not visible to the customer. 

The price will be rounded in the SQL query that is responsible for counting the number of different prices.

Create a product with two variants and set the prices in s_articles_prices for instance to 10.0001, 10.0002

### 6. Checklist

- [+] I have written tests and verified that they fail without my change
- [+] I have squashed any insignificant commits
- [+] This change has comments for package types, values, functions, and non-obvious lines of code
- [+] I have read the contribution requirements and fulfil them.


soebbing:
Hi @mkreusch, unfortunately this PR is a duplicate of https://github.com/shopware/shopware/pull/1651

mkreusch:
Thanks, @soebbing 
I did not find that one but good to know that it is in progress.

PR #1795, opened 2018-09-10
Create own function to push state.

sebastianlenz:
This is useful if someone wants to change the new pushState function.
In my case i would like to override the new function to replace window.history.pushState(stateObj.state, stateObj.title, location); with window.history.replaceState(stateObj.state, stateObj.title, location);

This seperates a function part to a seperate function.

-

-

-

### 6. Checklist


themes/Frontend/Responsive/frontend/_public/src/js/jquery.ajax-variant.js
```diff
@@ -151,20 +151,29 @@
...
                     $.loadingIndicator.close();
                 }
             });
         },
+        
+        /**
+         * Push state to browser history.
+         */
+        pushState: function(stateObj) {
+            var me = this;
```
shyim: You dont need me = this

sebastianlenz:
I changed from 5.4 to 5.5.
Is this correct or do i have to do something else?

PR #1796, opened 2018-09-11
Split SQL Query to avoid BNL Join in RiskManagement

Isengo1989:
It fixes an important Risk Management (RM) performance issue. When using the attribute filter in RM the query will use the Join buffer (Block nested loop) which will take significatly more time in the checkout. Because it will be used on every product in the basket the problem will get bigger the bigger the basket is. Splitting the query in two seperate queries will decrease the time by factor 4-5! 

It splits the one query in two and check the values programmaticly in PHP instead of using a join buffer in MYSQL.

Use the attribute filter in risk management, have at least a few products in the basket and +10.000 products/variants. In our setup (60.000 products / 15 products in the basket) it took almost 400-500ms per query and the basket loaded in 15-20 seconds.

-

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -1892,26 +1892,35 @@ class sAdmin
             if (!empty($value[0]) && isset($value[1])) {
                 $number = (int) str_ireplace('attr', '', $value[0]);
 
-                $sql = "
-                    SELECT s_articles_attributes.id
-                    FROM s_order_basket, s_articles_attributes, s_articles_details
-                    WHERE s_order_basket.sessionID = ?
-                    AND s_order_basket.modus = 0
-                    AND (
-                        s_order_basket.ordernumber = s_articles_details.ordernumber
-                        OR (s_order_basket.articleID = s_articles_details.articleID AND s_articles_details.kind = 1)
-                    )
-                    AND s_articles_details.id = s_articles_attributes.articledetailsID
-                    AND s_articles_attributes.attr{$number} = ?
-                    LIMIT 1
-                ";
+                $sql_productOrdernumber = $this->connection->createQueryBuilder()
+                   ->select(['s_articles_attributes.id'])
+                   ->from('s_order_basket, s_articles_attributes, s_articles_details')
+                   ->where('s_order_basket.sessionID = :sessionID')
+                   ->andWhere('s_order_basket.modus = 0')
+                   ->andWhere('s_order_basket.ordernumber = s_articles_details.ordernumber')
+                   ->andWhere('s_articles_details.id = s_articles_attributes.articledetailsID')
+                   ->andWhere('s_articles_attributes.attr' . $number . ' = :attrValue')
+                   ->setParameters([
+                       'attrValue' => $value[1],
+                       'sessionID' => $this->session->offsetGet('sessionId'),
+                   ])
+                   ->execute()->fetch(\PDO::FETCH_ASSOC);
 
-                $checkArticle = $this->db->fetchOne(
-                    $sql,
-                    [$this->session->offsetGet('sessionId'), $value[1]]
-                );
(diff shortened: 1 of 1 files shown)


Isengo1989:
auswahl_136

Here is a test ran with DataGrip

mitelg:
would be nice, if you could refactor the code to using the DBAL query builder :blush:

engine/Shopware/Core/sAdmin.php
```diff
@@ -1892,26 +1892,38 @@ public function sRiskATTRIS($user, $order, $value)
...
-                    )
+                    AND s_order_basket.ordernumber = s_articles_details.ordernumber
                     AND s_articles_details.id = s_articles_attributes.articledetailsID
                     AND s_articles_attributes.attr{$number} = ?
                     LIMIT 1
                 ";
 
-                $checkArticle = $this->db->fetchOne(
-                    $sql,
+                $sql_articleID = "
```
mitelg: please avoid "article" where possible. it is just the wrong translation. use "product" instead

Isengo1989: Sure thing, will do. Can you tell me why the Travis build failed? 

https://travis-ci.org/shopware/shopware/jobs/427367984

shyim: Codestyle Check fails, just run php-cs-fixer on the sAdmin file :blush:

Isengo1989:
> would be nice, if you could refactor the code to using the DBAL query builder

@mitelg I am not sure how this would work with doctrine tho. In the query there is $number set on the left side of the where.

`AND s_articles_attributes.attr{$number} = ?
`

engine/Shopware/Core/sAdmin.php
```diff
@@ -1892,26 +1892,35 @@ public function sRiskATTRIS($user, $order, $value)
...
-                    LIMIT 1
-                ";
-
-                $checkArticle = $this->db->fetchOne(
-                    $sql,
-                    [$this->session->offsetGet('sessionId'), $value[1]]
-                );
-
-                return (bool) $checkArticle;
+                $sql_ordernumer = $this->connection->createQueryBuilder()
```
mitelg: typo in number

engine/Shopware/Core/sAdmin.php
```diff
@@ -1892,26 +1892,35 @@ public function sRiskATTRIS($user, $order, $value)
...
+                   ->andWhere('s_order_basket.ordernumber = s_articles_details.ordernumber')
+                   ->andWhere('s_articles_details.id = s_articles_attributes.articledetailsID')
+                   ->andWhere('s_articles_attributes.attr' . $number . ' = :attrValue')
+                   ->setParameters([
+                       'attrValue' => $value[1],
+                       'sessionID' => $this->session->offsetGet('sessionId'),
+                   ])
+                   ->execute()->fetch(\PDO::FETCH_ASSOC);
+
+                $sql_articleID = $this->connection->createQueryBuilder()
```
mitelg: please avoid `article` as variable name, the correct translation is product :wink:

engine/Shopware/Core/sAdmin.php
```diff
@@ -1934,29 +1943,35 @@ public function sRiskATTRISNOT($user, $order, $value)
...
-                $checkArticle = $this->db->fetchOne(
-                    $sql,
-                    [
-                        $this->session->offsetGet('sessionId'),
-                        $value[1],
-                    ]
-                );
-
-                return (bool) $checkArticle;
+                $sql_productOrdernumer = $this->connection->createQueryBuilder()
```
mitelg: typo in number

Isengo1989:
@mitelg it should meet all requierments now, right?

mitelg:
@shopwareBot @PaddyS @soebbing did you forget this PR? :blush:

shyim:

PR #1798, opened 2018-09-13
notifcation attribute and event

dasistweb-gmbh:
Because nobody was able to take affect to the notification function by valid subscription or custom attributes

Add a Table s_articles_notification_attributes.
Add a Event to take affect on the Notification database subscription
Pass through the attributes for the notifications to the mail tempalate

### 6. Checklist

- [x] I have written tests and verified that they fail without my change - there were tests already


engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
```diff
@@ -227,13 +227,21 @@ public static function onNotifyConfirmAction(Enlight_Event_EventArgs $args)
...
             $notificationConfirmed = false;
+            $json_data = [];
             if (!empty($getConfirmation['hash'])) {
                 $notificationConfirmed = true;
                 $json_data = unserialize($getConfirmation['data']);
                 Shopware()->Db()->query('DELETE FROM s_core_optin WHERE hash=?', [$action->Request()->sNotificationConfirmation]);
             }
             if ($notificationConfirmed) {
-                $sql = '
+                $eventManager = Shopware()->Container()->get('events');
```
shyim: Don't use Shopware() global

engine/Shopware/Models/Article/Notification.php
```diff
@@ -58,6 +59,17 @@ class Notification extends LazyFetchModelEntity
      * @ORM\JoinColumn(name="mail", referencedColumnName="email")
      */
     protected $customer;
+
+    /**
+     * INVERSE SIDE
+     *
+     * @var \Shopware\Models\Attribute\ArticleNotification
+     *
+     * @Assert\Valid
```
shyim, line 68: I am not sure with this valid. After your Migration is runned, all notification would have no attributes. So this should fail then?

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
```diff
@@ -246,12 +254,14 @@ public static function onNotifyConfirmAction(Enlight_Event_EventArgs $args)
...
-                    $json_data['sShopPath'],
-                ]);
+                    Shopware()->Db()->query($sql, [
+                        $json_data['notifyOrdernumber'],
+                        $json_data['sNotificationEmail'],
+                        $json_data['sLanguage'],
+                        $json_data['sShopPath'],
+                    ]);
+                }
+
```
shyim: Maybe here a Event to get the lastInsertId to insert/update the new attributes or better a filter to provide the values of the attribute data?

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
dasistweb-gmbh: this is a static function theres no other possibitly i know

engine/Shopware/Models/Article/Notification.php
dasistweb-gmbh: this does not fail and is used like at the other models

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
dasistweb-gmbh: therefore i added a event to make the insert action yourself, you could use the modle for example

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
shyim: You can remove the static from the method. Listener must not be static  😊

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
shyim: Shopware_Notification_Save_Notification allows only to cancel the insert operation. So a plugin must cancel currently the operation and insert s_articles_notification and then into s_articles_notification_attributes.
When multiple plugins doing that on that event that would be uggly 🙈

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
dasistweb-gmbh: ok, done :D

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
dasistweb-gmbh: i refactored it after your recommendation

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
```diff
@@ -318,6 +335,10 @@ public static function onRunCronJob(Shopware_Components_Cron_CronJob $job)
                 continue;
             }
 
+            /** @var Shopware\Bundle\AttributeBundle\Service\DataLoader $attributeLoader */
+            $attributeLoader = Shopware()->Container()->get('shopware_attribute.data_loader');
```
mitelg: $this->get(

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
```diff
@@ -271,14 +287,15 @@ public static function onNotifyConfirmAction(Enlight_Event_EventArgs $args)
      *
      * @param Shopware_Components_Cron_CronJob $job
      */
-    public static function onRunCronJob(Shopware_Components_Cron_CronJob $job)
+    public function onRunCronJob(Shopware_Components_Cron_CronJob $job)
     {
         $modelManager = Shopware()->Container()->get('models');
```
mitelg: $this->get(

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
```diff
@@ -271,14 +287,15 @@ public static function onNotifyConfirmAction(Enlight_Event_EventArgs $args)
...
 
         $conn = Shopware()->Container()->get('dbal_connection');
```
mitelg: $this->get(

dasistweb-gmbh:
@mitelg what do i have to do for scheduling the pr?

mitelg:
@dasistweb-gmbh please be patient :blush:  
normally we meet on a daily base for the GitHub things, but today we skipped the meeting :disappointed:

dasistweb-gmbh:
@mitelg ok, alright :) i only wanted to get sure there's nothing i can do

PR #1808, opened 2018-09-24
add “not in” product stream filter

dasistweb-gmbh:
To filter product streams with the operator "NOT IN"

Filter Attributes in the backend Product Stream overview with the operator "NOT IN".

Backend -> Article -> Product Streams -> "create Filter stream" -> add filter "attribute filter" and search for the operator "NOT IN". It's not available in the extJs part.

/

/

### 6. Checklist


themes/Backend/ExtJs/backend/product_stream/view/condition_list/field/attribute.js
```diff
@@ -184,6 +185,8 @@ Ext.define('Shopware.apps.ProductStream.view.condition_list.field.Attribute', {
             }
         } else if (value.operator == 'IN') {
             value.value = value.value.split(",");
+        } else if (value.operator === 'NOT IN') {
+          value.value = value.value.split(",")
```
pascalheidmann-bedarf, line 189: I see code duplication here.
You might merge both IN and NOT IN conditions:
```
} else if (value.operator === 'IN' || value.operator === 'NOT IN') {
value.value = value.value.split(",")
}
```

shyim: I have merged it :wink:. See https://github.com/shopware/shopware/commit/d5712685075cae12ee208081df8988d30336acb0

PR #1811, opened 2018-09-26
Order api tests + attribute fix

Isengo1989:
When posting and updating orders it was not possible to set attributes from s_order_details_attributes. This PR will fix that and also create null attributes for the s_order_details_attributes if none given.

Also I created a Shopware_Tests_Api_OrderTest Class and added a few API tests for better codecoverage

It adds attribute to the whitelist and create the attribute on put and post of orders.

Create an order -> No s_order_details_attributes is created
Update an order -> No values will be inserted in s_order_details_attributes

https://issues.shopware.com/issues/SW-22139

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Order.php
+++ b/engine/Shopware/Components/Api/Resource/Order.php
@@ -517,6 +517,9 @@ class Order extends Resource
                 $detailModel->setShipped($detail['shipped']);
             }
 
+            // Set attributes
+            (isset($detail['attribute'])) ? $detailModel->setAttribute($detail['attribute']) : $detailModel->setAttribute([null]);
+
             $violations = $this->getManager()->validate($detailModel);
             if ($violations->count() > 0) {
                 throw new ApiException\ValidationException($violations);
@@ -610,6 +613,7 @@ class Order extends Resource
             'status',
             'shipped',
             'id',
+            'attribute',
         ];
 
         $details = $params['details'];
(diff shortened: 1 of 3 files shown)


engine/Shopware/Components/Api/Resource/Order.php
```diff
@@ -517,6 +517,9 @@ public function prepareCreateOrderDetailsData(array $params)
                 $detailModel->setShipped($detail['shipped']);
             }
 
+            // Set attributes
+            (isset($detail['attribute'])) ? $detailModel->setAttribute($detail['attribute']) : $detailModel->setAttribute([null]);
```
shyim, line 521: When i am updating my order status, would this not remove the detail attributes, because they are not given?

tests/Api/OrderTest.php
```diff
@@ -0,0 +1,511 @@
+<?php
+/**
```
shyim, line 2: Double header

engine/Shopware/Components/Api/Resource/Order.php
Isengo1989: @shyim true - I removed the else setting it to null from the prepareOrderDetailsData Method which is only used by put

tests/Api/OrderTest.php
Isengo1989: Removed it from OrderTest.php and also from CustomerTest.php - where I copied it in the first place :laughing:

Isengo1989:
@shopwareBot I am on vacation for +10 days - won't be able till then to add this. Will it be at least a Quick Pick then ? :)

PR #1816, opened 2018-09-28
ntr - ensure plugin load priority

teiling88:

Because of this DirectoryIterator (https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInitializer.php#L83) it is possible that 2 App Servers have a different chain how the plugins are loaded. It can cause to very hard to debug issues if the priority of the plugin loading is necessary.

With this small change it is ensured that the plugins are loaded and processed with an alphabetical sorting. 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Configs/Default.php
+++ b/engine/Shopware/Configs/Default.php
@@ -366,6 +366,12 @@ return array_replace_recursive([
             ],
         ],
     ],
+    /*
+     * @deprecated since 5.5, will be set to false with Shopware 5.6
+     */
+    'backward_compatibility' => [
+        'disable_plugin_sorting' => true,
+    ],
     'logger' => [
         'level' => $this->Environment() !== 'production' ? Logger::DEBUG : Logger::ERROR,
     ],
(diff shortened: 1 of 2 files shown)


uehler:
this is the same problem as #1785, isn't it?

teiling88:
Yes

shyim:
It's also the same code 😉

teiling88:
And the issue is still there...

t2oh4e:
Then you should maybe reopen the old PR instead of closing one because it may break things and then opening a new one with the exact same change...

shyim:
@t2oh4e that does not solve the entire problems. We want a loading which respects the dependencies and not just alphabetically. 

Just wait for the answer from shopwareBot

teiling88:
@t2oh4e - before the comment of @uehler I didn't know that the pr #1785 exist. I'm not working in the core development so I know not every pull request for Shopware itself. 

@shyim Your described fix is a much bigger solution and is not needed to fix the original problem. I have two different app server with the exact same sourcecode and they act different. 

This one liner fix that.

teiling88:
I have no idea if I take good config names but the functionality is now there :-)

PR #1814, opened 2018-09-27
Add not-writable path to exception message

Crease29:
### 1. Why is this change necessary?
To give better feedback what exact destination is invalid.

### 2. What does this change do, exactly?
It adds the path that is not writable to an exception message.

### 3. Describe each step to reproduce the issue or behaviour.
Remove the directory files/downloads and try to download the SwagDemoData plugin via plugin manager

### 4. Please link to the relevant issues (if any).

:heavy_minus_sign: 

### 5. Which documentation changes (if any) need to be made because of this PR?

:heavy_minus_sign: 

### 6. Checklist

- [X] I have written tests and verified that they fail without my change
- [X] I have squashed any insignificant commits
- [X] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I have read the contribution requirements and fulfil them.

PR #1817, opened 2018-09-28
Clean up and unify exception message generation

Crease29:

To provide a cleaner way of exception message generation.

- Clean up exception messages
- Replace string concatenation with sprintf in exception messages
- Use single quoted strings wherever possible

➖ 

It may be related to #1814

➖ 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/AttributeBundle/Service/DataPersister.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Service/DataPersister.php
@@ -81,7 +81,7 @@ class DataPersister
             throw new \Exception(sprintf('Table %s is no attribute table', $table));
         }
         if (!$foreignKey) {
-            throw new \Exception(sprintf('No foreign key provided'));
+            throw new \Exception('No foreign key provided');
         }
         $data = $this->filter($table, $data);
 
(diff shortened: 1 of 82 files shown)


engine/Shopware/Bundle/AttributeBundle/Service/DataPersister.php
```diff
@@ -81,7 +81,7 @@ public function persist($data, $table, $foreignKey)
             throw new \Exception(sprintf('Table %s is no attribute table', $table));
         }
         if (!$foreignKey) {
-            throw new \Exception(sprintf('No foreign key provided'));
+            throw new \Exception('No foreign key provided');
         }
```
shyim, line 85: Maybe add which table does not have a foreign key?

Crease29: I think that could confuse more than it would help. Here the foreignKey is used to determine the foreign entity of an attribute dataset... If I would change the exception message to e.g. `No foreign key provided for table %s` users could think that there is a missing foreign key in the table structure or so. What do you think?

Crease29:
Do you have any hint for me why the travis pipeline is failing?

shyim:
@Crease29 Code-Style check fails. Run local ant check-code-style

PR #1822, opened 2018-10-02
No product self reference in product streams

larsbo:
When a customer views a product detail page it makes no sense to show this product in a linked product stream in the cross selling section.

A new product stream condition + handler allows to remove articles from the product stream criteria. If a product stream gets loaded from the detail page the current article id will be automatically added as `SkipProductIdCondition` to remove the product from the stream.
The PR does not alter the behavior of product stream when loaded inside the emotion context.

* Create a product stream and add it to a product that potentially appear in the stream.
* Open the product page and look into the linked product stream. The currently open product will appear there.

https://issues.shopware.com/issues/SW-16157

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/ConditionHandler/SkipProductIdConditionHandler.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/ConditionHandler/SkipProductIdConditionHandler.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Bundle\SearchBundleDBAL\ConditionHandler;
+
+use Doctrine\DBAL\Connection;
+use Shopware\Bundle\SearchBundle\Condition\SkipProductIdCondition;
+use Shopware\Bundle\SearchBundle\ConditionInterface;
+use Shopware\Bundle\SearchBundleDBAL\ConditionHandlerInterface;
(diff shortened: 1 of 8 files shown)


engine/Shopware/Bundle/SearchBundleDBAL/ConditionHandler/SkipProductIdConditionHandler.php
```diff
@@ -0,0 +1,68 @@
...
+     * {@inheritdoc}
+     */
+    public function generateCondition(
+        ConditionInterface $condition,
+        QueryBuilder $query,
+        ShopContextInterface $context
+    ) {
+        $key = ':productIds' . md5(json_encode($condition));
+
+        $query->andWhere('variant.articleId NOT IN (' . $key . ')');
```
shyim, line 59: Please use instead sprintf

engine/Shopware/Controllers/Widgets/Emotion.php
```diff
@@ -163,7 +164,9 @@ public function productStreamArticleSliderAction()
             $limit = 0;
         }
 
-        $values = $this->getProductStream($streamId, $offset, $limit);
+        $articleId = $this->Request()->getParam('articleId');
```
mitelg: please use the correct translation `productId` for the variable name

themes/Frontend/Bare/frontend/detail/tabs/product_streams.tpl
```diff
@@ -4,7 +4,7 @@
         {include file="frontend/_includes/product_slider.tpl"
                  sliderMode="ajax"
                  sliderInitOnEvent="onShowContent-productStreamSliderId-{$relatedProductStream.id}"
-                 sliderAjaxCtrlUrl="{url module=widgets controller=emotion action=productStreamArticleSlider streamId=$relatedProductStream.id productBoxLayout="slider"}"
+                 sliderAjaxCtrlUrl="{url module=widgets controller=emotion action=productStreamArticleSlider streamId=$relatedProductStream.id articleId=$sArticle.articleID productBoxLayout="slider"}"
```
mitelg: please use the correct translation `productId` for the parameter

mitelg: additionally I would like to rename the parameter into `skippedProductId` or something like that.  with the name `productId` the intention behind that is not really clear. could also mean, only use this product

engine/Shopware/Bundle/SearchBundleDBAL/ConditionHandler/SkipProductIdConditionHandler.php
larsbo: Can you give an example how it should look like? This is the way all the other condition handler solve it.

engine/Shopware/Controllers/Widgets/Emotion.php
```diff
@@ -163,7 +164,9 @@ public function productStreamArticleSliderAction()
             $limit = 0;
         }
 
-        $values = $this->getProductStream($streamId, $offset, $limit);
+        $skippedArticleId = $this->Request()->getParam('skippedArticleId');
```
mitelg, line 167: why don't you want to name it `skippedProductId` ? :smile:

PR #1825, opened 2018-10-04
use css-flags, instead of ico-flags.png

tinect:
We should save unnecessary traffic and loads.

It replaces the ico-flags.png with css-version. The only exception is en_GB: the css-version is bigger and harder to handle in current environment, so we use encoded svg here.
it doesn't use calculated values like `calc(~"100% / 3")` due to known problems with IE/Edge.

Tested in:
image
+ Current safari of iPhone X, 6 and 6s
+ Chrome 69, Android 8.1

-

-

Language-Plugins have to be adjusted. Currently they are overwriting language-flags from responsive-theme with single pngs.
Example:
image

### 6. Checklist


themes/Frontend/Responsive/frontend/_public/src/less/_components/flags.less
```diff
@@ -2,22 +2,24 @@
...
+
+inspired by: https://github.com/dhanishgajjar/css-flags
+
+en_GB: we need to use svg for en_GB, cause it's smaller than the CSS-Version
 */
 
 .language--flag {
     .unitize-height(11);
     .unitize-width(14);
-    background: url("../../img/ico-flags.png") no-repeat 0 0;
```
mitelg: this file could be removed

soebbing:
Hi @tinect, thank you for your contribution! 💙

PR #1826, opened 2018-10-04
New Basket Discount Filter Event

jochenmanz:
currently it is not possible to change the way global basket discounts are processed. This PR adds a filter event to the appropriate location.

add's a filter event

none

none

Upgrade.md

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -455,6 +455,17 @@ class sBasket
                 )
             );
         } else {
+            $params = $this->eventManager->filter(
+                'Shopware_Modules_Basket_InsertDiscount_FilterParams',
+                $params,
+                [
+                    'subject' => $this,
+                    'getDiscounts' => $getDiscounts,
+                    'basketAmount' => $basketAmount,
+                    'basketDiscount' => $basketDiscount,
+                ]
+            );
+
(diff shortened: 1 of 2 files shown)


composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
```
shyim: This change is not necessary for your change

shyim:
Code-Style Check fails

engine/Shopware/Core/sBasket.php
```diff
@@ -455,6 +455,14 @@ public function sInsertDiscount()
                 )
             );
         } else {
+            $params = $this->eventManager->filter(
+                'Shopware_Modules_Basket_InsertDiscount_FilterParams',
```
kleinmann, line 459: Filter events need a second parameter of the return value, in your case `$params`. You can then omit the `discount` key in the additional arguments.

jochenmanz: I have corrected that and added more params to the filter event. @kleinmann thanks for the hint :)

shyim:
Hey @jochenmanz ,
thank you for your Contribution :blue_heart:

PR #1828, opened 2018-10-04
Not be able to select a non-existing variant

JoshuaBehrens:
Not created or afterwards deleted product variants may be selectable in the storefront which lead to a selection of the default product variant instead of the expected configurated variant.

The logic changes from:
> does configuration option x has a variant with this option in this set

to

> is there a possible one step change to an other configuration that result in an existing product variant

This is majorly done via SQL and less with storefront structs than before. A lot of joins are used that are optimized to use existing indices to maintain execution speed.

* have a product with a set that contains at least three groups with at least two options
configuration
* generate variants
generated-variants
* delete a variant that is not the default variant
delete-variant
* open the product in storefront
* choose configuration that just got deleted
say-wut
* get forwarded to default variant
see-ordernumber
wrong-item-in-card
* questionaningans
* install [this plugin](https://github.com/niemand-online/NoBugfixToDisableNonExistingVariants) that hotfixes this PR
first-plugin-that-fixes-a-pr-in-human-history
* retry
image
* success

The interface documentation already changed but as this changes the interface signature there is a need for a upgrade hint.

### 5. Checklist

- [x] Attribute @yarisade for finding this bug, keeping our sanity during the night where we solved this

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Gateway/ConfiguratorGatewayInterface.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Gateway/ConfiguratorGatewayInterface.php
@@ -72,28 +72,13 @@ interface ConfiguratorGatewayInterface
 
     /**
      * Returns all possible configurator combinations for the provided product.
-     * The returned array contains as array key the id of the configurator option.
-     * The array value contains an imploded array with all possible configurator option ids
-     * which can be combined with the option.
-     *
-     * Example (written with the configurator option names)
-     * array(
-     *     'white' => array('XL', 'L'),
-     *     'red'   => array('S', ...)
-     * )
-     *
-     * If the configurator contains only one group the function has to return an array indexed
-     * by the ids, which are selectable:
-     *
-     * Example (written with the configurator option names)
-     * array(
-     *     'white' => array(),
-     *     'red'   => array()
-     * )
+     * The returned array contains all combinations that are reachable by
+     * changing one of the group options in the given selection.
      *
      * @param Struct\BaseProduct $product
+     * @param int[]              $currentSelection
      *
      * @return array Indexed by the option id
      */
-    public function getProductCombinations(Struct\BaseProduct $product);
+    public function getProductCombinations(Struct\BaseProduct $product, array $currentSelection);
 }
--- a/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/ConfiguratorGateway.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/ConfiguratorGateway.php
@@ -28,6 +28,7 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Query\QueryBuilder;
 use Shopware\Bundle\StoreFrontBundle\Gateway;
 use Shopware\Bundle\StoreFrontBundle\Struct;
+use Shopware_Components_Config;
 
 /**
  * @category  Shopware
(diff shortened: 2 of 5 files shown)


engine/Shopware/Bundle/StoreFrontBundle/Gateway/ConfiguratorGatewayInterface.php
```diff
@@ -72,28 +72,13 @@ public function getConfiguratorMedia(Struct\BaseProduct $product, Struct\ShopCon
...
+     * The returned array contains all combinations that are reachable by
+     * changing one of the group options in the given selection.
      *
      * @param Struct\BaseProduct $product
+     * @param int[]              $currentSelection
      *
      * @return array Indexed by the option id
      */
-    public function getProductCombinations(Struct\BaseProduct $product);
+    public function getProductCombinations(Struct\BaseProduct $product, array $currentSelection);
```
mitelg, line 83: I appreciate your work :blush: :+1: but this is unfortunately a very breaking change

JoshuaBehrens: Serious bugs can take breaking changes with them. Do I get it right, that I need to alter my solution that it does not affect the interface signature? If so then I guess I have to change to content that is returned by the function. Doing this will result in confusion by anyone not knowing about this change. A change in a signature will at least result in exceptions.

But I give it a go

mitelg: it is ok, to make a BC, but you should be aware, that this will not be merged in 5.5.x and I'm not sure about 5.6 either :wink: 
A change without breaking the interface would be nice. how would the return value differ from now?

JoshuaBehrens: Selection in the storefront: size: 40, color: pink.
Current return value:
```jsonc
{
    "91": ["93", "94", "95", "96"], /* 38: Yellow, Pink, Cloth, Plastic */
    "92": ["93", "94", "95", "96"], /* 40: Yellow, Pink, Cloth, Plastic */
    "93": ["91", "92", "95", "96"], /* Yellow: 38, 40, Cloth, Plastic */
    "94": ["91", "92", "95", "96"], /* Pink: 38, 40, Cloth, Plastic */
    "95": ["91", "92", "93", "94"], /* Cloth: 38, 40, Yellow, Pink */
    "96": ["91", "92", "93", "94"] /* Plastic: 38, 40, Yellow, Pink */
}
```
PRs return value:
```jsonc
[
    {
        "15": 91, /* Size: 38 */
        "16": "94" /* Color: Pink */
    },
    {
        "15": "92", /* Size: 40 */
        "16": 93 /* Color: Yellow */
    },
    {
        "15": "92", /* Size: 40 */
        "16": "94", /* Color: Pink */
        "17": 96 /* Laces: Plastic */
    }
]
```

Whereas the ids are the option ids or group ids:

###### groups

id | name
---|------
15 | Size
16 | Color
17 | Laces

###### options

id | name
---|------
91 | 38
92 | 40
93 | Yellow
94 | Pink
95 | Cloth
96 | Plastic

I see a unwanted type inconsistency that I will fix. But there is a huge content difference while remaining the return type. Hooray for non-generic "collections" :tada:

JoshuaBehrens:
> make the **choice configurable**.

Where would one configure this? What will be the default? It will fix an error so is my PR the new default or is it to BCy? If I can fix this Problem without altering the signatures and return value structure I do not need an extra service, do I?

JoshuaBehrens:
I see. So there has to be a new service. I will investigate whether it is a viable solution to simple copy the service or to have a new service with more benefits.
Were should one configure the choice of service?

JoshuaBehrens:
After chatting with @soebbing about a deadlock issue that will arise with this change I came to the conclusion that I will work on a different point to solve this issue. I keep you posted

PR #1829, opened 2018-10-05
change rand() and mt_rand() to random_int()

teiling88:

image

### 6. Checklist

### DIFF
--- a/engine/Library/Zend/Mail.php
+++ b/engine/Library/Zend/Mail.php
@@ -1087,7 +1087,7 @@ class Zend_Mail extends Zend_Mime_Message
             $user = getmypid();
         }
 
-        $rand = mt_rand();
+        $rand = random_int();
 
         if ($this->_recipients !== array()) {
             $recipient = array_rand($this->_recipients);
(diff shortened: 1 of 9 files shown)


engine/Library/Zend/Mail.php
```diff
@@ -1087,7 +1087,7 @@ public function createMessageId() {
             $user = getmypid();
         }
 
-        $rand = mt_rand();
+        $rand = random_int();
```
tinect, line 1090: there are 2 parameters needed 😱

engine/Library/Zend/Mail/Transport/File.php
```diff
@@ -126,6 +126,6 @@ protected function _sendMail()
      */
     public function defaultCallback($transport)
     {
-        return 'ZendMail_' . $_SERVER['REQUEST_TIME'] . '_' . mt_rand() . '.tmp';
+        return 'ZendMail_' . $_SERVER['REQUEST_TIME'] . '_' . random_int() . '.tmp';
```
tinect, line 129: there are 2 parameters needed 😱

engine/Library/Zend/Mail.php
teiling88: upsi

engine/Library/Zend/Mail/Transport/File.php
teiling88: upsi

jenskueper: Shopware has a random helper library implemented. You could use https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Components/Random.php
instead. If something changes in the future, we only have to change it once.

PR #1833, opened 2018-10-09
performance fix variant filter

dasistweb-gmbh:
Every Listing and especially Filter queries took minutes to give a result.

We changed expensive left Joins to more handsome innerJoins for using db-indizes, so we can see a massive better performance. Admittedly we can not verify if the query is still serving the correct content in any case. @shopware can you verify this and give feedback?

we have different customergroups but only one fallback "EK" price. we use the variant filter to filter size of clothings. For the left joins will be used no index.
You can use the attached query to test the slow query.
[query.txt](https://github.com/shopware/shopware/files/2459296/query.txt)

103209

### 6. Checklist

- [x] I have written tests and verified that they fail without my change -> sql queries are not tested

### DIFF
--- a/engine/Shopware/Bundle/ESIndexingBundle/Product/ProductListingVariationLoader.php
+++ b/engine/Shopware/Bundle/ESIndexingBundle/Product/ProductListingVariationLoader.php
@@ -422,8 +422,8 @@ class ProductListingVariationLoader
         $subPriceQuery->addSelect('IFNULL(prices.articledetailsID, onsalePriceList.articledetailsID) as articledetailsID');
         $subPriceQuery->addSelect('IFNULL(prices.price, onsalePriceList.price) as price');
         $subPriceQuery->from('s_articles_details', 'details');
-        $subPriceQuery->leftJoin('details', '(' . $priceListingQuery . ')', 'prices', 'details.id = prices.articledetailsID');
-        $subPriceQuery->leftJoin('details', '(' . $onSalePriceListingQuery . ')', 'onsalePriceList', 'details.id = onsalePriceList.articledetailsID');
+        $subPriceQuery->innerJoin('details', '(' . $priceListingQuery . ')', 'prices', 'details.id = prices.articledetailsID');
+        $subPriceQuery->innerJoin('details', '(' . $onSalePriceListingQuery . ')', 'onsalePriceList', 'details.id = onsalePriceList.articledetailsID');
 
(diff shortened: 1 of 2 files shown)


engine/Shopware/Bundle/ESIndexingBundle/Product/ProductListingVariationLoader.php
```diff
@@ -422,8 +422,8 @@ private function fetchPrices(array $products, ShopContextInterface $context)
         $subPriceQuery->addSelect('IFNULL(prices.articledetailsID, onsalePriceList.articledetailsID) as articledetailsID');
         $subPriceQuery->addSelect('IFNULL(prices.price, onsalePriceList.price) as price');
         $subPriceQuery->from('s_articles_details', 'details');
-        $subPriceQuery->leftJoin('details', '(' . $priceListingQuery . ')', 'prices', 'details.id = prices.articledetailsID');
-        $subPriceQuery->leftJoin('details', '(' . $onSalePriceListingQuery . ')', 'onsalePriceList', 'details.id = onsalePriceList.articledetailsID');
+        $subPriceQuery->innerJoin('details', '(' . $priceListingQuery . ')', 'prices', 'details.id = prices.articledetailsID');
```
shyim, line 425: That is breaking the select usage
```IFNULL(prices.articledetailsID, onsalePriceList.articledetailsID)``

File: engine/Shopware/Bundle/SearchBundleDBAL/VariantHelper.php [hunk 2]
```diff
@@ -316,8 +316,8 @@ protected function joinSalePrices(QueryBuilder $query, ShopContextInterface $con
             'IFNULL(listing_price.product_id, onsale_listing_price.product_id) AS product_id',
         ]);
 
-        $subQuery->leftJoin('variant', '(' . $priceTable->getSQL() . ')', 'listing_price', implode(' AND ', $variantCondition));
-        $subQuery->leftJoin('variant', '(' . $onSalePriceTable->getSQL() . ')', 'onsale_listing_price', implode(' AND ', $variantOnSaleCondition));
+        $subQuery->innerJoin('variant', '(' . $priceTable->getSQL() . ')', 'listing_price', implode(' AND ', $variantCondition));
```
shyim, line 319: Here the same see the select above

dasistweb-gmbh:
@mitelg We were creating a issue first to verify our change. We would accept another solution to make the query faster. This is really needed for a new urgent lunch for our customer with shopware 5.5.
can any of the shopware developers take care of this problem?

PR #1836, opened 2018-10-11
backend order document search

jochenmanz:
allows searching for orders via their order document numbers

changes the order backend search indexer for ES and the default dbal search to include the order document numbers

add a document to a order and try to search the definied document id.

https://issues.shopware.com/issues/SW-22627

none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/AttributeBundle/Repository/Reader/OrderReader.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Repository/Reader/OrderReader.php
@@ -24,6 +24,8 @@
 
 namespace Shopware\Bundle\AttributeBundle\Repository\Reader;
 
+use Doctrine\DBAL\Connection;
+use PDO;
 use Shopware\Components\Model\ModelManager;
 use Shopware\Models\Order\Order;
 
@@ -56,11 +58,13 @@ class OrderReader extends GenericReader
     public function getList($identifiers)
     {
         $data = parent::getList($identifiers);
+        $documents = $this->getDocuments($data);
 
         $namespace = $this->snippets->getNamespace('backend/static/order_status');
 
         foreach ($data as &$row) {
             $row['orderStateName'] = $namespace->get($row['orderStateKey']);
+            $row['orderDocuments'] = $this->getOrderDocuments($documents, $row);
         }
 
         return $data;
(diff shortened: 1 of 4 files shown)


engine/Shopware/Bundle/AttributeBundle/Repository/Reader/OrderReader.php
```diff
@@ -121,4 +125,35 @@ protected function createListQuery()
...
     }
+
+    /**
+     * @param array $orders
+     *
+     * @return array
+     */
+    private function getDocuments(array $orders)
+    {
+        $orders = array_unique(array_keys($orders));
```
Crease29: Aren't keys unique by its definition?

engine/Shopware/Bundle/AttributeBundle/Repository/Reader/OrderReader.php
```diff
@@ -121,4 +125,35 @@ protected function createListQuery()
...
+
+    /**
+     * @param array $documents
+     * @param array $row
+     *
+     * @return array
+     */
+    private function getOrderDocuments(array $documents, array $row)
+    {
+        return array_values(array_filter(array_column(array_filter($documents, function (array $document) use ($row) {
```
mitelg, line 153: but why

engine/Shopware/Bundle/AttributeBundle/Repository/Reader/OrderReader.php
PaddyS: I guess he wanted to use `array_keys(array_flip())` instead, as this is just like `array_unique()`, but faster.

engine/Shopware/Models/Order/Repository.php
```diff
@@ -772,4 +775,23 @@ private function searchAddressTable($term, $table, array $excludedOrderIds = [])
 
         return $query->execute()->fetchAll(\PDO::FETCH_COLUMN);
     }
+
+    private function searchDocumentsTable($term, $table, array $excludedOrderIds = [])
+    {
+        $query = $this->getEntityManager()->getConnection()->createQueryBuilder();
+        $query->select('documents.orderID');
+        $query->from($table, 'documents');
+        $builder = Shopware()->Container()->get('shopware.model.search_builder');
```
shopwareBot, line 791: Can you please get rid of the Shopware() call?

Also: This method needs documentation :-)

jochenmanz: I have added the Documentation, but the Shopware() call is a bit complicated. As this code is inside a Model Repository, the Container is not injected automatically.

jochenmanz:
commits squashed :)

PR #1839, opened 2018-10-16
Only show active stores for article preview

pascalheidmann-bedarf:
If you have lots of shops but not all of them are active they are still shown in the shop selection list for article preview. By filtering for active only you can make the list shorter while having the same functionality as disabled shops automatically redirect to the home page of the default shop.

Extending `Shopware_Controllers_Backend_Article::loadStores` to filter the shop store to include only active shops.

\-

None known

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Article.php
+++ b/engine/Shopware/Controllers/Backend/Article.php
@@ -522,7 +522,7 @@ class Shopware_Controllers_Backend_Article extends Shopware_Controllers_Backend_
         $id = $this->Request()->getParam('articleId');
         $priceGroups = $this->getRepository()->getPriceGroupQuery()->getArrayResult();
         $suppliers = $this->getRepository()->getSuppliersQuery()->getArrayResult();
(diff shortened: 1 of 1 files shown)


pascalheidmann-bedarf:
@shopwareBot AFAIK the seo grid does NOT use the same store.

The preview gets its shop data from that big data set, but inside the seo grid view [`/backend/article/view/category/seo.js`](https://github.com/shopware/shopware/blob/5.5/themes/Backend/ExtJs/backend/article/view/category/seo.js) another `shopStore` gets created and subsequently another request to `[SHOP]/backend/base/getShops` gets fired.
One might suggest a change here in reusing the component but right now it isn't and therefore not breaking it.

PaddyS:
Hmpf, true that.
Must have been [layer 8](https://en.wikipedia.org/wiki/Layer_8) problem, just tested it once more and didn't see any breaks.
Also, you're right about it having its "own" store instance.

Guess we'll schedule it later today or tomorrow, sorry for any inconvenience!

PR #1840, opened 2018-10-17
Add event `plugin/swAjaxVariant/onRequestDataComplete`

mschop:
This event enables one, to react when the request loading the variants
through ajax is completed fully. In my case this was needed, because I
started another action that triggered the loadingIndicator to be
reopened. Opening multiple loadingIndicator is not supported or does not
work well. This new event enables one, to start an action after the
loading indicator is closed.

Resolves a bug.

See commit message.

```
$.subscribe('plugin/swAjaxVariant/onRequestData', function() {
   // start another action that uses loading-indicator
});
```

None

None

### 6. Checklist

- [ Are there any tests for the theme? ] I have written tests and verified that they fail without my change


shyim:
Hey,

we have already a Pull Request for that. See #1537

PR #1841, opened 2018-10-17
Added two numbers

lx-wnk:
Is there a Reason why "0" and "1" are missing in this string?
Added 0 and1 to 'd'-set

All numbers in string

----

----

----

### 6. Checklist

- [x] I have written tests and verified that they fail without my change - not necessary on my opinion


oktupol:
According to the function's phpdoc:

> The available characters in each set are user friendly - there are no ambiguous
> characters such as i, l, 1, o, 0, etc. This makes it much easier for users to manually
> type or speak their passwords.

lx-wnk:
@oktupol you are absolutely right!
Only saw the missing numbers and didn't read the DOC.

PR #1842, opened 2018-10-17
Update sAdmin.php

sebastianlenz:
If this line is removed, riskmanagement works for all payments and does not exclude users default payment address. This is hardly neccessary, otherwise riskmanagement is worthless.

Activates riskmanagement for users default payment.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -290,9 +290,7 @@ class sAdmin
         }
 
         // Check additional rules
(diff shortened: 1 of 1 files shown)


mnaczenski:
Since this option is used in many shops to provide a payment method for a specific group of customers that is deactivated or restricted  (e.g. invoice only for staff members) it is not possible to remove this check entirely. You could add a config option for it if neccessary.

The "paymentpreset" is a parameter that is only set via backend (not frontend) when a payment is explicit added to the customer. It is not possible to add a payment id to this field via frontend logic.

PR #1843, opened 2018-10-17
Adjust tracking code column type

niklasbuechner:
Splitting an order into multiple shipments is currently limited due to the tracking code column `s_order.trackingcode` allowing at most 255 characters. (If a tracking code has 20 characters, you can create a maximum of 12 shipments.) 

This pull request changes the type of the tracking code field `s_order.trackingcode` from `VARCHAR(255)` to `TEXT` so that a unlimited number of tracking codes can be saved. This will help e.g. the plugin _DHL Adapter_ to be able to remove its current limit of 9 tracking codes/shipments per order.

Try to save more than 12 tracking codes of 20 characters each in an order.

--

--

### 6. Checklist

### DIFF
--- a/_sql/migrations/1443-adjust-tracking-code-column-type.php
+++ b/_sql/migrations/1443-adjust-tracking-code-column-type.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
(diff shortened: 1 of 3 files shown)


niklasbuechner:
I updated the Doctrine model `Shopware\Models\Order\Order` as well. 👍

shyim:
And now squash commits  :smile:

niklasbuechner:
@shyim Done 😉

PR #1848, opened 2018-10-18
[SW-20698] Fix emotion element identifier for duplication

sobyte:
Duplicating an emotion item will cause problems to the element's translation when `start_row`, `start_col`, `end_row` and `end_col` are the same for multiple elements in `s_emotion_element`.
The concatenated string of these values is used to create an identifier between the original element and the cloned element for duplicating its translation.

Changing the identifier to the actual viewport values.

1. Create an emotion item with two text elements
2. Translate both elements
3. Duplicate emotion item
3. Check the translation of text element #1 (should be the translation of #2)

[SW-20698](https://issues.shopware.com/issues/SW-20698)

None

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Emotion.php
+++ b/engine/Shopware/Controllers/Backend/Emotion.php
@@ -27,6 +27,7 @@ use Shopware\Components\Emotion\EmotionExporter;
 use Shopware\Components\Emotion\Exception\MappingRequiredException;
 use Shopware\Components\Model\ModelManager;
 use Shopware\Components\Random;
+use Shopware\Models\Emotion\Data;
 use Shopware\Models\Emotion\Element;
 use Shopware\Models\Emotion\Emotion;
 use Shopware\Models\Emotion\Library\Field;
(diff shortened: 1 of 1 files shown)


sobyte:
Hi @shopwareBot 

i was able to reproduce your behaviour when not all elements are represent in every viewport. I've added an additional element identifiert for duplicating the translations and this should fix the problem now. Can you confirm this?

Note: I also already squashed my commits ;)

sobyte:
Hi @shopwareBot 

you're right. I thought i had this case covered but i obviously didn't so i've changed the identifier now completely to its data so it should definitely work now.
Can you please test again?

PaddyS:
Hey @sobyte,

just by looking at your code, I figured that it might break in the following case:
- Create an emotion
- Add a text-element with value "Example 1"
- Translate the element with value "Example Foo 1"
- Add another text-element, again with value "Example 1"
- Add a translation for that second element, with value "Example Bar 1"

I'm pretty sure that in this case, both **translations** would have the same value after duplicating.
Your element identifier is actually not unique by just adding "fieldName" and "value", since you can have multiple elements having the same value and the same name in just a singe emotion.
Please ensure you're building a unique identifier.

Can you please test that and fix it properly?
Please make sure this **actually** works. :-)

Also: Since there seem to be some edgecases, it would be very helpful to have tests for those cases.

FiveDigital:
One week for adding tests and changing the implementation while it took you month (at least in the past) to have a look at a PR seems not fair for me.

teiling88:
@FiveDigital he can create at every time a new one with the new implementation and tests. 

If the core guys would let this kind of prs open they would never come to 0 open PRs.

PR #1849, opened 2018-10-19
cache plugin and namespace config for less database load

dasistweb-gmbh:
The multiple database queries coused a high database load on heavy used shops

We cache the array result of the database requests and use it further than from cache.

Call homepage and trace your db queries.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Plugin/Namespace.php
+++ b/engine/Shopware/Components/Plugin/Namespace.php
@@ -556,8 +556,16 @@ class Shopware_Components_Plugin_Namespace extends Enlight_Plugin_Namespace_Conf
             WHERE namespace=?
         ';
 
-        $connection = $this->Application()->Container()->get('dbal_connection');
-        $rows = $connection->fetchAll($sql, [$this->name]);
+        /** @var Zend_Cache_Core $cache */
+        $cache = $this->Application()->Container()->get('cache');
+        $cacheKey = 'globalAllPlugins' . $this->name;
+        if ($cache->test($cacheKey)) {
+            $rows = (array) $cache->load($cacheKey);
+        } else {
+            $connection = $this->Application()->Container()->get('dbal_connection');
+            $rows = $connection->fetchAll($sql, [$this->name]);
+            $cache->save($rows, $cacheKey, ['Shopware_Plugin'], 86400);
+        }
 
         $plugins = [];
         foreach ($rows as $row) {
@@ -624,7 +632,15 @@ class Shopware_Components_Plugin_Namespace extends Enlight_Plugin_Namespace_Conf
              WHERE ce.type=0
              ORDER BY name, position
         ';
-        $listeners = $this->Application()->Db()->fetchAll($sql, [$namespace]);
+        /** @var Zend_Cache_Core $cache */
+        $cache = $this->Application()->Container()->get('cache');
+        $cacheKey = 'globalListenersPlugins' . $namespace;
+        if ($cache->test($cacheKey)) {
+            $listeners = (array) $cache->load($cacheKey);
+        } else {
+            $listeners = $this->Application()->Db()->fetchAll($sql, [$namespace]);
+            $cache->save($listeners, $cacheKey, ['Shopware_Plugin'], 86400);
+        }
 
         foreach ($listeners as $listenerKey => $listener) {
             if (($position = strpos($listener['listener'], '::')) !== false) {
--- a/tests/Functional/Components/Plugin/NamespaceTest.php
+++ b/tests/Functional/Components/Plugin/NamespaceTest.php
@@ -1,0 +1,89 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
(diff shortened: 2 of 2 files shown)


shyim:
Code-Style check fails in Travis

engine/Shopware/Components/Plugin/Namespace.php
```diff
@@ -556,8 +556,17 @@ protected function initStorage()
...
+
+        /** @var Zend_Cache_Core $cache */
+        $cache = $this->Application()->Container()->get('cache');
+        $cacheKey = 'globalAllPlugins' . $this->name;
+        if ($cache->test($cacheKey)) {
+            $rows = (array)$cache->load($cacheKey);
+        }else {
+            $connection = $this->Application()->Container()->get('dbal_connection');
+            $rows = $connection->fetchAll($sql, [$this->name]);
+            $cache->save($rows, $cacheKey);
```
shyim: Please consider using Cache Times and Tags for automatic invalidation 
https://github.com/shopware/shopware/blob/7a65778d8ddaa9021b2833e95c9ad3443cb2f2dd/engine/Shopware/Components/CacheManager.php#L183-L189

engine/Shopware/Components/Plugin/Namespace.php
```diff
@@ -624,7 +633,15 @@ protected function loadListeners($namespace)
...
         ';
-        $listeners = $this->Application()->Db()->fetchAll($sql, [$namespace]);
+        /** @var Zend_Cache_Core $cache */
+        $cache = $this->Application()->Container()->get('cache');
+        $cacheKey = 'globalListenersPlugins' . $namespace;
+        if ($cache->test($cacheKey)) {
+            $listeners = (array)$cache->load($cacheKey);
+        }else {
+            $listeners = $this->Application()->Db()->fetchAll($sql, [$namespace]);
+            $cache->save($listeners, $cacheKey);
```
shyim: Same here

dasistweb-gmbh:
i added tests even if nearly any other test ensures the functionality implicitly

dasistweb-gmbh:
why does my test fails? in other tests there is used the shopware kernel as well. what do i have to change

shyim:
You wrote a Unit test, the dependencies have to be mocked up.
See https://stackoverflow.com/questions/2741832/unit-tests-vs-functional-tests

engine/Shopware/Components/Plugin/Namespace.php
```diff
@@ -556,8 +556,16 @@ protected function initStorage()
...
 
-        $connection = $this->Application()->Container()->get('dbal_connection');
-        $rows = $connection->fetchAll($sql, [$this->name]);
+        /** @var Zend_Cache_Core $cache */
+        $cache = $this->Application()->Container()->get('cache');
+        $cacheKey = 'globalAllPlugins' . $this->name;
+        if ($cache->test($cacheKey)) {
+            $rows = (array) $cache->load($cacheKey);
+        } else {
+            $connection = $this->Application()->Container()->get('dbal_connection');
```
bcremer, line 565: Is it really worth having yet another cache layer here?
The above query should run in no time since it is a simple index lookup.

dasistweb-gmbh:
@shyim i moved the test to the function tests, what do have to do else to test this without errors?

engine/Shopware/Components/Plugin/Namespace.php
dasistweb-gmbh: this database query was recognized as database bottelneck on heavy load shopware systems.

shyim:
@dasistweb-gmbh Can you rebase your branch to the latest 5.5?

engine/Shopware/Components/Plugin/Namespace.php
bcremer: Can you post the EXPLAIN of this query on the system?
Maybe it's just a missing index on `namespace`.

tests/Unit/Components/Plugin/NamespaceTest.php
```diff
@@ -0,0 +1,89 @@
+<?php
```
shopwareBot: You probably wanted to remove this file?

engine/Shopware/Components/Plugin/Namespace.php
dasistweb-gmbh: bildschirmfoto 2018-10-25 um 09 14 39
bildschirmfoto 2018-10-25 um 09 15 48

dasistweb-gmbh: the best way to refactor would propabbly be to load the data once per request from the db and inject to the Configs

bcremer: Both Quries are not using any sensible keys.
I have no access to a running Shopware DB at the moment but it should be reasonable easy find and add the missing indices.

That will fix the root issue and not add more complexity to the bootstrapping prozess.

bcremer: > load the data once per request from the db and inject to the Configs

That also sounds reasonable. But maybe this is not needed after all after the query optimizations.

n3rdg45m: i think there should be an index on s_core_plugins for active and namespace, just try to set it and check again

dasistweb-gmbh: we have testet the different cases again. the cache is on heavy loads much more performant as the database queries. regardless of the database index

PR #1852, opened 2018-10-23
Fix time zone conversion bug

windaishi:

Shopware has some issues when the web server and the browser are in different time zones. This is because ExtJS submits date fields without the timezone. For explanation imagine the following scenario:

The web server runs on a server with GMT as timezone. In Shopware there is an order with order time "12:00:00 UTC (+0000)". This order is now shown in a Browser that runs in a different time zone, e.g. Europe/Berlin. The browser displays the time as "14:00 (CEST +0200)". When the user now hits the save button, the order time gets changed to "14:00 UTC (+0000)".

oct-16-2018 18-04-50 (Server runs in Europe/Berlin, SW 5.4.6)

This also leads to false positives with the optimistic locking / overwrite protection added in #1474. When an order is edited by a client with a different time zone it cannot be saved at all because the overwrite protection is always triggered. Even when clicking "yes" in the confirmation prompt, this will just cause the prompt to be shown again on the next try.

oct-23-2018 14-16-35 (Server runs in Europe/Berlin, SW 5.5.1)

**The problem result from the following issue:**
When ExtJS formats a date object before it gets submitted to the server, it is converted to a string representation. Unfortunately, this representation is a local time string without a timezone specification, e.g. `2007-12-24T18:21`. When this time string is interpreted by the server, it assumes it to be specified in the server's default timezone. As the time zone of client and server differ, the time is now shifted by the relative offset between the client and the server, which is wrong.

Every date field in all ExtJS models suffer from this bug. This PR is general fix for all models with a date field. It is fully backwards compatible.

Every date representation that is formatted for JSON serialization by ExtJs will be transformed into the local time zone of the server before being sent to the server. 

Unfortunately, just formatting the datetime in ISO 8601 format is insufficient. This is because a PHP DateTime object created from such an ISO string "lives" in the wrong time zone (UTC). When this date is saved by Doctrine it is not converted into the timezone of the server and therefore leads to a wrong date-time written to the database. See the following code example:

```php
<?php

// Create a DateTime object for 18:21 GMT
$dateTime = new \DateTime('2007-12-24T18:21Z');
// Our server lives in Europe/Berlin, therefore we need to persists 19:21 to the database.
// What doctrine persists:
echo $dateTime->format('Y-m-d H:i:s'); // 2007-12-24 18:21:00
echo PHP_EOL;
// What would be correct to persists: 
$dateTime->setTimeZone(new \DateTimeZone('Europe/Berlin'));
echo $dateTime->format('Y-m-d H:i:s'); // 2007-12-24 19:21:00
echo PHP_EOL;
```

This PR also reverts the commit d389efd because that fix lead to massive problem with the overwrite protection where every attempted order save caused the confirmation prompt to be shown on every Shopware installation, even if the order was not modified in the mean time. This is because ExtJS cannot handle the option `dateFormat: 'c'` in the ExtJS models correctly. (I could not find what the actual problem here was, but is is not necessary anymore with my fix).
Also this fix is not necessary anymore at this PR is a general solution for the problem and also solved problems with the overwrite-protection.

Access the Shopware backend with a client that runs in a different time zone as the web server. Open an order in the backend and save it. (You do not need to change anything.) After the order has been saved, you can see that the "created" time stamp has been shifted by the time zone offset between client and server. The more often you save the order, the further the "created" time stamp is shifted.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Article.php
+++ b/engine/Shopware/Controllers/Backend/Article.php
@@ -43,80 +43,80 @@ class Shopware_Controllers_Backend_Article extends Shopware_Controllers_Backend_
      *
      * @var \Shopware\Models\Article\Repository
      */
-    protected $repository;
+    protected $repository = null;
 
     /**
      * Repository for the shop model
      *
      * @var Repository
      */
-    protected $shopRepository;
+    protected $shopRepository = null;
 
     /**
      * Repository for the customer model
      *
      * @var \Shopware\Models\Customer\Repository
      */
-    protected $customerRepository;
+    protected $customerRepository = null;
 
     /**
      * Repository for the category model
      *
      * @var \Shopware\Models\Category\Repository
      */
-    protected $categoryRepository;
+    protected $categoryRepository = null;
 
     /**
      * @var \Shopware\Components\Model\ModelRepository
      */
-    protected $articleDetailRepository;
+    protected $articleDetailRepository = null;
 
     /**
      * @var \Shopware\Components\Model\ModelRepository
      */
-    protected $customerGroupRepository;
+    protected $customerGroupRepository = null;
 
     /**
      * Entity Manager
      *
      * @var null
      */
-    protected $manager;
+    protected $manager = null;
 
     /**
      * @var \Shopware\Components\Model\ModelRepository
      */
-    protected $configuratorDependencyRepository;
+    protected $configuratorDependencyRepository = null;
 
     /**
      * @var \Shopware\Components\Model\ModelRepository
      */
-    protected $configuratorPriceVariationRepository;
+    protected $configuratorPriceVariationRepository = null;
 
     /**
      * @var \Shopware\Components\Model\ModelRepository
      */
-    protected $configuratorGroupRepository;
+    protected $configuratorGroupRepository = null;
 
     /**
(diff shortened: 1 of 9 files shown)


engine/Shopware/Controllers/Backend/Article.php
```diff
@@ -43,80 +43,80 @@ class Shopware_Controllers_Backend_Article extends Shopware_Controllers_Backend_
      *
      * @var \Shopware\Models\Article\Repository
      */
-    protected $repository;
+    protected $repository = null;
```
shyim: Why.......

mitelg: @shyim did you even read the description? :smile: he said, that he reverted your commit completely

shyim:
Hey @windaishi ,

we don't want a another Extjs Core Hack. This override could potential break many existing code.
Please implement instead a new Extjs Model Type like following which sets the right client offset or makes the check in two side in UTC.

```javascript
Ext.apply(Ext.data.Types, {
    TIMESTAMP: {
        sortType: Ext.data.SortTypes.asDate,
        type: 'timestamp',
        convert: function(value) {
            // conversion code
        }
    }
});
```

So we could minimize the risk that we potential break existing code depending on the old value.

PR #1853, opened 2018-10-24
Changes HttpCache behaviour for checkout/info widget

jkrzefski:
Since Shopware 5.3 the widget `checkout/info` can display the name of the currently logged in customer. This name would then be saved in the http cache and would be the same for every customer with an empty cart and notepad. With this change, every customer will only see their respective names in the widget.

The Change makes the CacheControl disregard the state of the cart and notepad for the `checkout/info` widget, so it is never cached.

1. Have the basic setting for `Create Shopware Login Cookie` (found under `Login / registration`) set to true.
2. Log into the storefront.
3. Have an empty cart and notepad.
4. Log out
5. Log in with another customer with an empty basket and notepad.

You will be greeted as the customer you logged in first.

None.

I don't know.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Core/HttpCache/CacheControl.php
+++ b/engine/Shopware/Plugins/Default/Core/HttpCache/CacheControl.php
@@ -168,7 +168,7 @@ class CacheControl
 
         $controller = $this->cacheRouteGeneration->getControllerRoute($request);
 
-        return $controller === 'widgets/checkout' && (!empty($this->session->offsetGet('sBasketQuantity')) || !empty($this->session->offsetGet('sNotesQuantity')));
+        return $controller === 'widgets/checkout';
     }
 
     /**


hlohaus:
Schuld the SLT Cookie that not prevent this?

jkrzefski:
@hlohaus I'm afraid, I don't follow. Can you explain to me, what you mean?

jkrzefski:
@hlohaus Maybe it is intended to do so, but I did not find any condition for the SLT cookie. Can you tell me where this SLT cookie disables the cache?

n3rdg45m:
Just  replace the values on non caching controller tags for widgets/checkout with checkout,slt,userInfo and your are fine :-)

jkrzefski:
@n3rdg45m Could you explain, where you take those values from and what exactly they do?

n3rdg45m:
@jkrzefski I had the same problems und just check the cache settings (https://community.shopware.com/Cache-Performance-Modul_detail_845.html "Definition der Cache-Zeiten") and the add the vars that should not be cached

mnaczenski:
I think @n3rdg45m is referring to this migration: https://github.com/shopware/shopware/blob/5.5/_sql/migrations/932-add-http-cache-route.php

The noCache-Tag "slt" prevents the widget to be cached, when the slt cookie is present. 

slt

jkrzefski:
Interesting. That didn't do it in our case. But maybe we had some other problems. What I can say for sure is that decorating the CacheControl did the trick for our case. If I have the time, I might investigate a little further on why that was necessary in the first place.

PR #1855, opened 2018-10-25
Fix order overwrite protection false positives

windaishi:
Since Shopware 5.5.2 the overwrite protection dialog is triggered on every save of any order.

This is because ExtJS could not parse the "changed" timestamp because there still was a hidden "changed" field in the oder detail window. Because of this field the value for "changed" was always `null` in the JSON transferred to the server.

This PR removes this field.

Save an order in Shopware 5.5.2.

### 6. Checklist


soebbing:
Hi @windaishi, thank you for your contribution! 💙

PR #1856, opened 2018-10-26
Refactored xml parser + tests + misc stuff

nlubisch:
Reopened as the topic arised in Gitter.im, the test from my last PR is fixed.
Old PR: #1781
Ticket: https://issues.shopware.com/issues/SW-17837

## Description
* Why is it necessary?
Refactored xml readers because every programmer does his own thing.
There were some redundant functions which i moved to a abstract base class.
All readers are now unified and 100% tested.

screenshot from 2017-02-05 19-37-41
screenshot from 2017-02-05 19-38-09

* What does it improve?
It does improve code, tests and xsd files.

**BC breaks?**
-> config.xsd
storeType has changed. PHP got removed due to when inserting an php array as CDATA you have to eval it to get the data and that is IMO a security issue.
There is now an explicit way to declare which type of store it is.

-> plugin.xsd
pluginName-attribute in "requiredPlugin" is now required.
complexType "scope" got removed because its not used.

**Tests pass?**
Yes


nlubisch:
I've updated my branch for now. Hope this will be the last one :smiley:

shyim:
Hey @nlubisch,
your Change has been merged in 5.6. Thanks for the contribution! :blue_heart:

PR #1859, opened 2018-10-30
fix type annotation: ean is not a float

michz:

The field `ean` is stored into database as a string. Therefore it will be handled by Doctrine as a string, too. The annotation `float` would be misleading.

It may be discussed if `string` is the best choice here. But `float` does not make any sense to me.

Nothing in runtime. Just for correct IDE support and less confusion.

\-

\-

\-

### 6. Checklist

- [ ] ~~I have written tests and verified that they fail without my change~~ Not possible
- [ ] ~~This change has comments for package types, values, functions, and non-obvious lines of code~~ Fixes them!

### DIFF
--- a/engine/Shopware/Models/Article/Detail.php
+++ b/engine/Shopware/Models/Article/Detail.php
@@ -224,7 +224,7 @@ class Detail extends ModelEntity
     private $height = null;
 
     /**
-     * @var float ean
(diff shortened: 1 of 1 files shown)


michz:
> Please remember to sign the CLA.

Done.

PR #1860, opened 2018-10-30
show company in widget if it's not empty

tinect:
current statement doesn't work and shows always first- and lastname

add check for empty company

### 6. Checklist

### DIFF
--- a/tests/Functional/Controllers/Backend/WidgetsTest.php
+++ b/tests/Functional/Controllers/Backend/WidgetsTest.php
@@ -1,0 +1,210 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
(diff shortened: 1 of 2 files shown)


tinect:
My first test is added. I hope everything is okay. Please let me know, if anything is missing

tests/Functional/Controllers/Backend/WidgetsTest.php
```diff
@@ -0,0 +1,210 @@
...
+        $this->assertArrayHasKey('customers', $jsonBody['data']);
+
+        //first customer should be the one we added, ass there isn't any other process adding any s_statistics_currentusers
+        $this->assertEquals($this->userId, $jsonBody['data']['customers'][0]['userID']);
+        $this->assertEquals($addressData['company'], $jsonBody['data']['customers'][0]['customer']);
+
+        /*
+         * cleanup
+         */
+        $this->db->delete('s_statistics_currentusers', ['id = ?' => $this->statisticsId]);
```
shopwareBot, line 92: Please use transaction with rollback instead. See https://github.com/shopware/shopware/blob/a49b005e84206f788f911bf3adf156f1274b40d4/tests/Functional/Components/Emotion/Preset/EmotionToPresetDataTransformerTest.php#L51 as example

tinect: will be back! Will fix it after vacation. Please don't close :-)

PR #1862, opened 2018-10-30
fix missing help and support text for rich text config fields

buddhaCode:

HTML input fields in plugin configuration forms doesn't display support and help text.

Deep down in the rabbit hole `Ext.form.Field` is decorated to add elements for displaying help and support text for form inputs. Basically, most input elements derives form this class. Unfortunately, the HTML input component doesn't. It derives from `Ext.Component`. Thus, the decorations will not be applied for html inputs.
I moved the declaration stuff into a mixin and applied this to `Ext.form.Field` and `Ext.form.field.HtmlEditor`. If any ExtJs guru knows a more compliant and better way, I would be happy to change it ;)

Create plugin with config element of type `html` and add a `<description>` tag

Before:
pluginoptinfoinfobubble

After:
bildschirmfoto 2018-10-30 um 11 33 17

Guess, the same will apply for attribute fields.

Not aware of.

None.

### 6. Checklist

### DIFF
--- a/engine/Library/ExtJs/components/Enlight.form.mixin.HelpSupportElements.js
+++ b/engine/Library/ExtJs/components/Enlight.form.mixin.HelpSupportElements.js
@@ -1,0 +1,191 @@
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


engine/Library/ExtJs/components/Enlight.form.mixin.HelpSupportElements.js
```diff
@@ -0,0 +1,191 @@
...
+    helpIconEl: null,
+
+    /**
+     * Using the "afterRender()"-method we're injecting
+     * the support and help texts into the form elements.
+     *
+     * @private
+     * @return void
+     */
+    initHelpSupportElemes: function () {
```
shopwareBot: Shouldn't this be "initHelpSupportElements"? Please remember to change the other occurences.

htkassner: I guess you wanted to "say" Elements. Small typo. The idea of a mixin is not bad. We´re gonna evaluate this one.

buddhaCode:
I was just shortcutting elements ;)

buddhaCode:
Damn, it. 🙈 I made a new one #1871

PR #1863, opened 2018-10-31
Add DISTINCT to priceListingQuery to improve ES indexing performance

Pmuckel:
The ES indexing is very slow if the variant filter has been activated. This is amongst other things caused by the priceListingQuery and the related onSalePriceListingQuery. The results of these queries contain duplicate rows which makes them really slow. 

The DISTINCT of the `articledetailsID` leads to one result per detailID, which is enough for these subqueries. We tested this change and went from 9 hours to 1 minute indexing time for 7000 articles (inclusive variants). 

Create a bunch of articles with variants (more than 1000 for comparable results) and try to index them with activated variant filter. You can see the performance improvement with this change compared to the old query without the DISTINCT. 

Nope. 

None. 

### 6. Checklist


soebbing:
Hi @Pmuckel, thank you very much for your contribution! 💙 🎉

PR #1868, opened 2018-11-03
Extracted snippets enclosed in quotes into variables

JoshuaBehrens:
If you use quotes in snippets generated smarty php will either syntactically or logically break.

It will also improve parsing errors like this:
image
image

Use variables instead of generated literals.

* Set an error message to something like: `This error can be solved <a href="/">here</a>.`
* The quotes will break the generated smarty php up into a `string divided by string` and renders as `NaN`.
* One could use single quotes instead although it is not easier to maintain or explain to end users.

A best practice hint to [smarty snippet assign](https://shyim.me/faq/smarty-assign-tpl/) would be nice.

### 5. Checklist


JoshuaBehrens:
Squashed it :tennis:

PR #1874, opened 2018-11-08
State description translation in customer orders

pascalheidmann-bedarf:
With the removal of the description of states you are forced to use the translation service. The customer backend controller hasn't seen the required changes and therefore just return the internal name which (at least) will be displayed as order and payment state instead.

Translate both payment and order states loaded from _Shopware_Controllers_Backend_Customer::loadStores()_

Go to a customer and switch to the orders tab. There you will see only the internal state names.

### 6. Checklist


shyim:
Travis Fails

PR #1875, opened 2018-11-08
Added removal of both temporary file and directory which are created …

SieGeL2k16:
…automatically during upload of media data via API

If media data is uploaded via API Shopware creates first a tempory file, removes this file immediatly and creates a temporary directory under the same name. The media is then copied into this directory for further processing. However after media data is added to Shopware, these temporary directories are not removed afterwards! This is a massive problem if you bulk-add a lot of articles with images, as the temporary directory gets filled quickly with orphaned files. One of our servers had more than 350.000 temporary directories which occupied nearly the complete hd storage!

It first unlink() the temporary file and afterwards rmdir() the created temporary directory.

Add an article via API with at least one image included and watch your temporary directory. You can see directories being created for every image. But these directories won't disappear after the article was uploaded.

No issue opened (yet), we do require a fast solution to this problem.

No documentation changes required.

### 6. Checklist


engine/Shopware/Components/Api/Resource/Media.php
```diff
@@ -277,6 +281,8 @@ public function internalCreateMediaByFileLink($link, $albumId = -1)
             $manager->createMediaThumbnail($media, [], true);
         }
 
+        unlink($file);
+        rmdir(dirname($file));
```
oktupol: You could combine the latter two changes in a `finally` block to avoid duplicate code

SieGeL2k16: Thanks for the hint, I've added the finally block.

SieGeL2k16:
sorry, looks like I've forgot to pull first the latest changes :( I'm a bit lost here, what should i do to fix it finally?

soebbing:
I think the best way to fix this would be to copy the Sha of the relevant commit, delete the branch, create a new branch from 5.5 and cherry-pick the Sha of your commit. Then do a force push on the branch of the pr.

SieGeL2k16:
@soebbing I've re-added the pull-request, hopefully now everything is fine.

PR #1876, opened 2018-11-08
Attributes in several AttributeBundle Searchers 

pascalheidmann-bedarf:
You can't let existing Searchers  search for things that are stored in its attributes. Also the switch to the new search broke search by document number for orders.

- Adds attribute relations to following searchers: ProductSearcher, OrderSearcher, CustomerSearcher.
- Adds document search by document number to OrderSearcher.

The added attributes are more a preparation as I couldn't find a good way to extend the getSearchFields without changing the interface yet.

https://issues.shopware.com/issues/SW-22914

### 6. Checklist


shyim:
Document Searching is already fixed with https://github.com/shopware/shopware/pull/1836

I don't see the point why we should leftJoin the attributes

pascalheidmann-bedarf:
@shyim haven't seen it but good to see that I'm not the only one.

Regarding attributes: We have several metadata used to find customers, orders and products that are stored in attributes, and I would guess we aren't the only ones.
As I can't find any good way to decorate those searchers to add fields anyway you can close this PR for now.

PR #1878, opened 2018-11-09
Readonly attributes in Backend

pascalheidmann-bedarf:
Sometimes you want to add a attribute that should not be editable for your users but still visible in backend. This allows to make that attribute be readonly.

Adding a new attribute configuration option that marks that attribute as readonly. This setting is then used to disable the input field in extjs. Therefore changes by code are of course still possible

PR https://github.com/shopware/shopware/pull/1645
https://issues.shopware.com/issues/SW-21676

New option "readonly" for attribute cru(d).

### 6. Checklist


_sql/migrations/1444-add-readonly-to-attribute-field.php
```diff
@@ -0,0 +1,36 @@
...
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+class Migrations_Migration1444 extends Shopware\Components\Migrations\AbstractMigration {
+    public function up($modus) {
+        $this->addSQL('ALTER TABLE `s_attribute_configuration` ADD `readonly` TINYINT NOT NULL DEFAULT 0 AFTER `label`;');
+        $sql = <<<EOT
+SET @locale_de_DE = (SELECT id FROM s_core_locales WHERE locale = 'de_DE');
+SET @locale_en_GB = (SELECT id FROM s_core_locales WHERE locale = 'en_GB');
+INSERT IGNORE INTO s_core_snippets (namespace, shopID, localeID, name, value, created, updated, dirty) VALUES
```
shyim: Why.. Add them to the ini file

pascalheidmann-bedarf: You are right. Copy & pasted from https://github.com/shopware/shopware/pull/1645 but using snippets directly is way better and even felt strange when copy & pasting the migration 🙈

themes/Backend/ExtJs/backend/base/attribute/Shopware.attribute.Form.js
```diff
@@ -348,7 +348,8 @@ Ext.define('Shopware.attribute.Form', {
                 supportText: attribute.get('supportText'),
                 helpText: attribute.get('helpText'),
                 labelWidth: 155,
-                value: attribute.get('defaultValue')
+                value: attribute.get('defaultValue'),
+                disabled: attribute.get('readonly')
```
shopwareBot: Any specific reason why you're using "disabled" here instead of [readOnly](https://docs.sencha.com/extjs/4.1.3/#!/api/Ext.form.field.Base-cfg-readOnly)?

pascalheidmann-bedarf: Wasn't aware of the differences between those. Changed it to the more adequate _readOnly_.

pascalheidmann-bedarf:
@shopwareBot migrations added, tests now run as they should

pascalheidmann-bedarf:
While all elements now work more or less, I feel like that the multi selection grid isn't the cleanest solution as there is no way to disable individual columns inside the grid (sorting & deleting). Those are set right in the `initComponent()` function.
I played a bit around with providing a `setReadOnly()` function (the `Ext.form.FieldContainer` component has none), but that is never invoked on itself and has the problem not being able to access the child elements in `initComponent()` - therefore you can't use it to create a valid initial state.
Any ideas?

Oh and Travis CI fails with a completely irrelevant file...

mitelg:
@pascalheidman-bedarf ups, there went something wrong :wink:

pascalheidmann-bedarf:
@mitelg yeah... gladly I made a backup before. I feel like I will never get rebase right in the first try 😞 

@shopwareBot the multi selection contains both the selection/search field as well as the grid with all selected elements. Former is easy as it is an real input but latter is a complex thing that has controls for ordering and deleting elements which can't disabled on the fly. It feels not complete to not being able to change their state later.

pascalheidmann-bedarf:
@shopwareBot good point with the batch processing. I will take a look at it!

Regarding `allowSorting`: I played around with that property as well as `allowDelete` and `allowAdd` which will disable all those functionality. Only problem I have is that changing those later don't propagate to change the behaviour on the fly. Therefore trying to push all that changes down to their respective element now.

themes/Backend/ExtJs/backend/base/attribute/field/Shopware.form.field.Grid.js
```diff
@@ -94,6 +95,26 @@ Ext.define('Shopware.form.field.Grid', {
...
+
+    setReadOnly: function(readOnly) {
+        var me = this;
+        me.readOnly = readOnly;
+
+        me.setReadOnlyProperties();
+
+
+        var columns = me.grid.columns;
+        console.log(columns);
```
tinect: this shouldn't be here?

tinect (changes requested):

pascalheidmann-bedarf:
@shopwareBot the 5.6 branch isn't up to date is it?

Neither my local repo nor github find any missing commits and this would explain why `shopware/engine/Shopware/Controllers/Backend/Analytics.php` always fails in Travis CI

Pmuckel:
Hey @pascalheidman-bedarf,

thanks for your improvements! I tested them and they look pretty good. 

There need to be some things added, though:
Can you please gray out the readonly attributes e.g. in the article detail view? Otherwise it wouldn't be possible for the customer to detect a readonly attribute. Also, please check the attribute translations (need to be grayed out as well if readonly) and remove the delete-buttons of the multiple selection grid. 

Thanks in advance!

pascalheidmann-bedarf:
@shopwareBot yep but not much time right now.

@Pmuckel will try to add those suggested indicators.

engine/Shopware/Bundle/AttributeBundle/Service/DataPersister.php
```diff
@@ -267,4 +271,29 @@ private function isDateColumn(Column $column)
...
+        $builder = $this->connection->createQueryBuilder();
+        $builder
+            ->select(['config.column_name'])
+            ->from('s_attribute_configuration', 'config')
+            ->where('config.table_name = :tablename')
+            ->andWhere('config.readonly = 1')
+            ->setParameter('tablename', $table)
+        ;
+
+        $results = $builder->execute()->fetchAll();
```
hlohaus: FETCH_COLUMN and you can save yourself the next lines.

pascalheidmann-bedarf:
I rebased to current 5.6 as there were some collisions as the code is quite old now ;)
Seems like there have been a database error @ travis so maybe someone can rerun those tests?

pascalheidmann-bedarf:
@shopwareBot fixed the PR.
The migration wasn't working anymore as there was already another migration with the same class name.

shyim:

PR #1881, opened 2018-11-11
Added code to remove temporary file and directory after media was pro…

SieGeL2k16:
…cessed

If media data is uploaded via API Shopware creates first a tempory file, removes this file immediatly and creates a temporary directory under the same name. The media is then copied into this directory for further processing. However after media data is added to Shopware, these temporary directories are not removed afterwards! This is a massive problem if you bulk-add a lot of articles with images, as the temporary directory gets filled quickly with orphaned files. One of our servers had more than 350.000 temporary directories which occupied nearly the complete hd storage!

It first unlink() the temporary file and afterwards rmdir() the created temporary directory.

Add an article via API with at least one image included and watch your temporary directory. You can see directories being created for every image. But these directories won't disappear after the article was uploaded.

No issue opened (yet), we do require a fast solution to this problem.

No documentation changes required.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Media.php
+++ b/engine/Shopware/Components/Api/Resource/Media.php
@@ -254,6 +254,9 @@ class Media extends Resource
         /** @var Album $album */
         $album = $this->getManager()->find(Album::class, $albumId);
         if (!$album) {
+            // Cleanup temporary file
+            unlink($file);
+            rmdir(dirname($file));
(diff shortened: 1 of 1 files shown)


SieGeL2k16:
Sorry for the circumstances...could you provide a link where all the commit rules are noticed? I understand the permission problem, but i didn't know or read of forcing usage of php-cs-fixer?

PR #1882, opened 2018-11-11
SW-22921 - adds .raw to field name in aggregations for product attrib…

rene-gust:
…ute filters of type text

If you build a product attribute filter with the data type text for the listing view, it will return unexpected values when you work with ElasticSearch. If you deactivate ElasticSearch, the correct values will be returned.

The Change adds '.raw' to the aggregation field name if the data type of the product attribute is text.

The product need to have an attribute of data type text and the values begin with a capital letter . Add a product filter in the backend with that product attribute. In the frontend the Filter Drop Down shows the values with a lowercase letter. So clicking on that value doesnt work, no products are schown.

https://issues.shopware.com/issues/SW-22921

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleES/FacetHandler/ProductAttributeFacetHandler.php
+++ b/engine/Shopware/Bundle/SearchBundleES/FacetHandler/ProductAttributeFacetHandler.php
@@ -32,6 +32,7 @@ use ONGR\ElasticsearchDSL\Search;
 use Shopware\Bundle\AttributeBundle\Service\ConfigurationStruct;
 use Shopware\Bundle\AttributeBundle\Service\CrudService;
 use Shopware\Bundle\AttributeBundle\Service\TypeMapping;
+use Shopware\Bundle\ESIndexingBundle\MappingInterface;
 use Shopware\Bundle\SearchBundle\Condition\ProductAttributeCondition;
 use Shopware\Bundle\SearchBundle\Criteria;
 use Shopware\Bundle\SearchBundle\CriteriaPartInterface;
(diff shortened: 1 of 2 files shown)


engine/Shopware/Bundle/SearchBundleES/FacetHandler/ProductAttributeFacetHandler.php
```diff
@@ -62,14 +63,20 @@ class ProductAttributeFacetHandler implements HandlerInterface, ResultHydratorIn
...
     /**
      * ProductAttributeFacetHandler constructor.
      *
      * @param CrudService $crudService
      */
-    public function __construct(CrudService $crudService)
+    public function __construct(CrudService $crudService, MappingInterface $poductMappingService)
     {
         $this->crudService = $crudService;
+        $this->poductMappingService = $poductMappingService;
```
soebbing: The name of the variable has a typo.

engine/Shopware/Bundle/SearchBundleES/FacetHandler/ProductAttributeFacetHandler.php
```diff
@@ -370,4 +377,45 @@ private function formatDates(array $aggregation)
...
+    /**
+     * @param CriteriaPartInterface $criteriaPart
+     * @param ShopContextInterface  $context
+     *
+     * @return bool
+     */
+    private function isTextAttribute(CriteriaPartInterface $criteriaPart, ShopContextInterface $context)
+    {
+        /** @var ProductAttributeFacet $criteriaPart */
+        $poductMappingService = $this->poductMappingService->get($context->getShop());
```
soebbing: Typo again

soebbing (changes requested):
Great fix! I think the ticket number in the text of the pr is wrong, but the one on the commit message seems correct.
Could you please provide a test case for this change as well?

PR #1883, opened 2018-11-14
Use rsync to delete old cache directory

xf-:
rm -Rf is really slow for small directory (my case 50gb in delete). Fastest easy solution is rsync
https://www.slashroot.in/which-is-the-fastest-method-to-delete-files-in-linux
https://www.dynamsoft.com/help/saw/start%20delta%20transfer.htm used by rsync

Use rsync if available.

Clear a really large cache.

### 6. Checklist

### DIFF
--- a/var/cache/clear_cache.sh
+++ b/var/cache/clear_cache.sh
@@ -17,4 +17,10 @@ rm -f $DIR/../../web/cache/*.txt
 
 $DIR/../../bin/console sw:generate:attributes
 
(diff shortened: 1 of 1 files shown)


xf-:
rsync has also a progress/speed option -maybe interesting to see some interaction. Never tried it on a delete.

xf-:
@soebbing @shyim  why is this change missing in chnagelog?

PR #1886, opened 2018-11-16
Add notify event to seo rebuild cli command

hrombach:

Currently, the only way to react to SEO Index rebuild via CLI is via a hook on a called function inside the command. A dedicated notify-event would be preferable as it is potentially more update-safe.

Also, every time a hook is used, a little kitten dies.

Add a `notify`-event to `\Shopware\Commands\RebuildSeoIndexCommand::execute()` named `Shopware_Command_RebuildSeoIndexCommand_CreateRewriteTable`, triggered once for each subshop like in the CronJob implementation

Rebuild SEO-Index via CLI command. All options described [here](https://developers.shopware.com/blog/2017/07/24/seo-urls-in-plugins/) fail.

-/-

The new event could be added to [this article](https://developers.shopware.com/blog/2017/07/24/seo-urls-in-plugins/).

### 6. Checklist

### DIFF
--- a/engine/Shopware/Commands/RebuildSeoIndexCommand.php
+++ b/engine/Shopware/Commands/RebuildSeoIndexCommand.php
@@ -24,6 +24,7 @@
 
 namespace Shopware\Commands;
 
+use Shopware\Components\ContainerAwareEventManager;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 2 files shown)


engine/Shopware/Commands/RebuildSeoIndexCommand.php
```diff
@@ -141,6 +141,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $this->rewriteTable->sCreateRewriteTableBlog();
             $this->rewriteTable->createManufacturerUrls($context);
             $this->rewriteTable->sCreateRewriteTableStatic();
+
+            Shopware()->Events()->notify(
```
shyim: Don't use Shopware() global, get the eventmanager from the container

hrombach: I was actually in the middle of changing that when you commented 😁

PR #1890, opened 2018-11-20
Made Shopware()->DocPath() compatible with composer installations

jkrzefski:
When used in a composer installation the method `Shopware()->DocPath()` would return `vendor/shopware/shopware/` instead of the actual project root directory.

The project root is now determined by `dirname(dirname(dirname(__DIR__)))` and will be used throughout the config. Also the `Shopware()->DocPath()` method now return the config value of `app.rootDir`.

Run `Shopware()->DocPath()` in a composer installation.

Dunno.

Dunno.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Application.php
+++ b/engine/Shopware/Application.php
@@ -51,11 +51,6 @@ class Shopware extends Enlight_Application
     protected $appPath;
 
     /**
-     * @var string
-     */
-    protected $docPath;
-
-    /**
      * @var Container
      */
     protected $container;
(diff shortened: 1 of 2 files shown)


jkrzefski:
As @shyim pointed out in a discussion about this PR, my changes would break the usage of `$this->DocPath()` inside the config.php file. So now the approach is to decorate the application class inside of the composer project and leaving this implementation as it is. The new PR can be found here https://github.com/shopware/composer-project/pull/28/files

PR #1892, opened 2018-11-21
fix StateManager miscalculation for iOS > 10

webisdead:
- this fixes wrong calculation for viewport-width in statemanager since iOS ignores "user-scalable=no" meta-value

- it returns the correct value if window is zoomed
- it makes the operation more readable

- use Safari on iPad on iOS > 10 (wich is ignoring "user-scalable=no")
- scroll to elements relying on State Manager e.g. .column-headline in footer on https://www.shopwaredemo.de/
-  pinch-zoom a little bit (zoom > 1)
- tap mutiple times on title-element (e.g. "Newsletter")
- see content collapsing

-

-

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.state-manager.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.state-manager.js
@@ -1136,7 +1136,7 @@
                 return width;
             }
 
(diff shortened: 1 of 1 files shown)


sergejreznik:
👍

PR #1893, opened 2018-11-21
Media Garbage Collector: Take account for attribute html fields

buddhaCode:

The Media Garbage Collector moves unused media items to the trash bin. If one created attribute fields as HTML/rich text field and used images in the editor, these usage will not be taken into account for cleaning unused files.

Added a routine to check all html in attributes.

1. Upload new image into a desired album (not trash bin) in media manager.
2. Create attribute field of type "html" on an entity of your choice.
3. Navigate to entity and use the HTML editor to select the uploaded image. Save.
4. Spin on media gargabe collector via cronjob or console command.
5. You image has been moved from the former album to trash bin.

None aware of.

None aware of.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/MediaBundle/GarbageCollectorFactory.php
+++ b/engine/Shopware/Bundle/MediaBundle/GarbageCollectorFactory.php
@@ -26,6 +26,7 @@ namespace Shopware\Bundle\MediaBundle;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\DBAL\Connection;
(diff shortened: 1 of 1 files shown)


buddhaCode:
What do you think about the non-attribute model fields. These are determined in the `GarbageCollectorFactory::getDefaultMediaPositions()` method. Do you think this is nice? Maybe a I could replace it with an annotation directly attached to the property in the model class itself? E.g. like:
```
/**
 * @MediaBundle\MediaPosition(mediaColumn="path", parseType="html")
 */
```

buddhaCode:
@shopwareBot Sure. I see your point.

PR #1896, opened 2018-11-22
Correct spelling error

barbieswimcrew:
Because "devide" doesn't exist in english language it's mandatory to correct this spelling error. Therefore "devide" becomes "divide"

The misspelled "devide" becomes "divide"

Just replace "devide" by "divide" at two different places in the BatchProcess.php class 

-

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/MultiEdit/Resource/Product/BatchProcess.php
+++ b/engine/Shopware/Components/MultiEdit/Resource/Product/BatchProcess.php
@@ -142,7 +142,7 @@ class BatchProcess
                 case 'bigint':
                 case 'decimal':
                 case 'float':
-                    $attributes[$attribute] = ['set', 'add', 'subtract', 'devide', 'multiply'];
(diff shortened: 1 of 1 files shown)


shyim:
Would that not break existing saved Queries in overview module, which used "devide" ?

soebbing:
I think @shyim is right. It would be better to just add the correct spelling instead of replacing the wrong one.

barbieswimcrew:
> I think @shyim is right. It would be better to just add the correct spelling instead of replacing the wrong one.

I see... @shyim @soebbing ... adding another switch case for "divide" would solve that, right?

```
...
case 'divide':
case 'devide':
    $builder->set("{$prefix}.$column", $builder->expr()->quot("{$prefix}.$column", $operationValue));
    break;
case 'multiply':
...
```

barbieswimcrew:
> Hello @barbieswimcrew,
> please squash your commits now. :-)

did it ;-)

PR #1898, opened 2018-11-26
Make Country::$allowShipping default to true to restore backwards compatibility with 5.5.2

cjuner:

Commit https://github.com/shopware/shopware/commit/affcab35459c2a3528b113b0822214a757dd695f#diff-83a08850c6e3a09dfc2638a5828a6fa5R178, released in Shopware v5.5.3, was a breaking change that broke backwards compatibility with the 5.5 series and any previous version of Shopware.

Newly created instances of the `Country` set `allowShipping` to `false` in Shopware v5.5.3 by default.

For example, before this fix, countries programmatically created could no longer be used as part of a new customer's address when creating a customer via the Resource API, as that API would not allow creating a customer with an address located in a country to which shipping is disabled.

To be compatible to Shopware 5.5.3, one had to call `$country->setAllowShipping(true);`, which was not required in previous versions. Hence, `allowShipping` is now set to `true` by default, making any special handling for Shopware 5.5.3 unnecessary.

`Country::$allowShipping` is now `true` by default when creating new instances.

An example case, which shows the broken backwards compatibility:

A country where `allowShipping` is `false` cannot be in the shipping or billing address of a _new_ customer when creating one via the [Customer API Resource](https://github.com/shopware/shopware/blob/affcab35459c2a3528b113b0822214a757dd695f/engine/Shopware/Components/Api/Resource/Customer.php#L46). Attempting to do so causes the following exception:

`Shopware\Components\Api\Exception\CustomValidationException: Country by id 38 is not available for shipping`

> thrown in [shopware/engine/Shopware/Components/Api/Resource/Customer.php:602](https://github.com/shopware/shopware/blob/affcab35459c2a3528b113b0822214a757dd695f/engine/Shopware/Components/Api/Resource/Customer.php#L602)

No known relevant issues.

None, besides the included change entry in `UPGRADE-5.5.md`.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Country/Country.php
+++ b/engine/Shopware/Models/Country/Country.php
@@ -175,7 +175,7 @@ class Country extends ModelEntity
      *
      * @ORM\Column(name="allow_shipping", type="boolean", nullable=false)
      */
-    private $allowShipping = false;
+    private $allowShipping = true;
 
     /**
      * @var \Doctrine\Common\Collections\ArrayCollection
--- a/UPGRADE-5.5.md
+++ b/UPGRADE-5.5.md
@@ -44,6 +44,7 @@ This changelog references changes done in Shopware 5.5 patch versions.
 * Changed default groupKey to get correct subpages of custom pages in mobile menu
 * Changed some mailer options to combo boxes to avoid wrong entries
 * Changed CSV import of snippets to only remove one apostrophe from the beginning of a line 
+* New countries created via programming interfaces default to allowing shipping to restore backwards compatibility to versions before Shopware 5.5.3.
 
 ## 5.5.3
 
--- a/_sql/migrations/1448-set-default-country-allow-shipping-true.php
+++ b/_sql/migrations/1448-set-default-country-allow-shipping-true.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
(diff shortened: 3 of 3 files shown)


engine/Shopware/Models/Country/Country.php
```diff
@@ -175,7 +175,7 @@ class Country extends ModelEntity
      *
      * @ORM\Column(name="allow_shipping", type="boolean", nullable=false)
      */
-    private $allowShipping = false;
+    private $allowShipping = true;
```
shyim, line 178: Please create also a migration, the Doctrine model should match the table definition :blush:

cjuner: Thanks for reviewing!

What should the migration do, in your opinion? Reset all `allow_shipping` values which were set to be `false` to `true`? I would assume users who have set this values themselves would not like the values to be reset again.

As I see it, the original commit already properly migrates previous entries which did not have the field: https://github.com/shopware/shopware/commit/affcab35459c2a3528b113b0822214a757dd695f#diff-00e580f91c830ac6f39d00bc0cf892feR29

Now, what a future commit could add is to introduce a default value for the table schema. But I am not yet aware of any necessity for it. However, I would see no problem adding that, though.

shyim: I would just change the default value of the column, so only new countries are affected

cjuner: > I would just change the default value of the column, so only new countries are affected

As requested: https://github.com/shopware/shopware/pull/1898/commits/24e2296fd5762c26a2d6163a4d2764d2727f5983#diff-600aafb965d84df29a6d9e693ac02f01R28

shyim: And now squash commits :blush:

cjuner: > And now squash commits :blush:

Done: https://github.com/shopware/shopware/pull/1898/commits/eba91c856dfc0fb38e64ccc108e5f1f723fdc4a8

PR #1903, opened 2018-11-29
Fixed InvalidArgumentException in ReflectionHelper for shopware compo…

MrMDeluxe:
When installing Shopware via composer, 
it is not possible to add own CustomerSearchBundleDBAL ConditionHandlers via Plugins.
Since shopware 5.4 is supporting installations via composer, 
the shopware core should consider this when dealing with the docPath.

It adds a check in the verifyClass function of the ReflectionHelper Class, to determine
if shopware is installed via composer (the docPath points to a subdir of vendor) and the
class that should be verified is outside of the vendor dir.
If this is the case, the docPath is changed to the current docRoot.

Install shopware via composer and develope a plugin with a CustomerSearchBundleDBAL ConditionHandler.

-

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/ReflectionHelper.php
+++ b/engine/Shopware/Components/ReflectionHelper.php
@@ -99,6 +99,12 @@ class ReflectionHelper
     private function verifyClass(\ReflectionClass $class, $docPath, array $directories = [])
     {
         $fileName = $class->getFileName();
+
+		// if shopware is installed via composer and $class is not in vendor dir, change $docPath
+        if (strpos($docPath, '/vendor/') !== false && strpos($fileName, '/vendor/') === false) {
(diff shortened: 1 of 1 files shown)


shyim:
We had in Gitter the same discussion recently and i had worked on a solution which fixes multiple scenarios (composer, plugin folder moved to a another location and so on).
Can you test [this](https://dpaste.de/Wvyy) in your setup

MrMDeluxe:
Hey shyim,
i have tested your ReflectionHelper and now it works also for composer installations.
Will the class be part of the next update?

shyim:
Hey @MrMDeluxe .

thank you for your Contribution! We have scheduled that change with the Ticket [SW-23063](https://issues.shopware.com/issues/SW-23063) for Shopware 5.5.5 release.

Thanks!. So i will close this Pull Request :blush:

PR #1904, opened 2018-11-29
Listing Boxes should use Flexbox instead of floating – It behaves smoother and is modern

jonas-pietzsch:

For example on Article listing pages like category pages, articles will behave ugly when their short description texts have different lengths and result in different amounts of line breaks. Due to this, the responsive layout is very unstable and tends to look ugly if short description texts have different length.

Example: [Screenshot with the issue](https://pbs.twimg.com/media/DtMmlh3X4AEG6Bi.jpg)

Flexbox is a modern standard while being compatible with a decent amount of older browsers, too. It makes layouting of grids with columns far more easy and solves the described problem smoothly.

Clearfixes and left floats were thrown out of the listing boxes layout. Flexbox will take care of the layouting by preserving also a width of a "column" in percentage value (just like before).
If there is no bug in my changes, it should do exactly the same as before. In addition in fixes the described ugly behaviour.

Expected result is: [Screenshot with fix](https://pbs.twimg.com/media/DtMnJ3eX4AY2jhf.jpg)

See the change in action here on this shop: https://lifefoto.de/foto-fineart-papier/?p=1
Although implemented and therefor overwritten in a child template, it should perform exactly as in this PR.

* Create a category
* Create some articles (e.g. 4) with short descriptions of different length (e.g. 80, 140 and 200 characters)
* On the shop frontend, visit the category and see the articles of different height form an ugly experience

- none -

I guess none.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/less/_mixins/flexbox.less
+++ b/themes/Frontend/Responsive/frontend/_public/src/less/_mixins/flexbox.less
@@ -1,0 +1,18 @@
+.flexboxContainer() {
+    display: -webkit-box;
+    display: -moz-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+    -webkit-flex-wrap: wrap;
+    flex-wrap: wrap;
+}
+
+.flexboxChild(@width: initial) {
+    -webkit-flex: @width;
+    -ms-flex: @width;
+    flex: @width;
+    -webkit-flex-grow: 0;
+    -ms-flex-positive: 0;
+    flex-grow: 0;
+}
--- a/themes/Frontend/Responsive/frontend/_public/src/less/_modules/listing.less
+++ b/themes/Frontend/Responsive/frontend/_public/src/less/_modules/listing.less
@@ -208,7 +208,7 @@ It contains the viewport specific styles inside media queries.
 }
 
 .listing {
-    .clearfix();
+    .flexboxContainer();
     margin: -2% 0 0 -2%;
     padding: 0 0 2% 0;
 }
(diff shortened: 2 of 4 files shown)


shyim:
Hey,
have you tested that on IE 11 on Windows 7? Caniuse shows "Known Issues" for IE 11 and we have to support it 😊.

PS: Please squash your commits

jonas-pietzsch:
Hey @shyim,
squashed my commits. I didn't test it on IE11 in Windows 7.
I would require help to do that. First, I don't have a working setup for this. Secondly, I only fixed this for a shop I'm maintaining and later thought, it should be open sourced because it's simple.

I would be very happy if someone else could take over the cross-browser-testing.

klarstil:
Hey @jverhoelen 
you can use [https://crossbrowsertesting.com/](https://crossbrowsertesting.com/) for cross browser testing. The free account is all what you need.

larsbo:
@jverhoelen 
Or you can use this free vms from Microsoft especially for IE testing:
https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

jonas-pietzsch:
Hey all,
I never ran Shopware locally. 1,5 hours ago I really wanted to do so, but I had to quit because the lack of documentation. Honestly: how do I start this? I found several scripts about starting something, I found a docker-compose file which complaints about `sh: dev-ops/docker/actions/.copy_mysql-config.sh: No such file or directory`.

More honestly: I can't and don't want to put much time into it. Could somebody help me please?
Where can I find any documentation how to run and access this properly? The README is very thin, also the getting started docs linked from there do not mention any way to run this in Docker. I would prefer the Docker-Compose option if it would work...

hlohaus:
You have to use psh. Example:
$ ./psh.phar docker:start

shyim:
The Docker setup in the project is currently broken and will be fixed next week. 
You could try removing this line https://github.com/shopware/shopware/blob/5.5/dev-ops/docker/actions/start.sh#L7. Or use a local setup and follow the steps in README

Or the fastest way: Download 5.5.3 from the Download Page, replace the themes folder from your Branch :)

PR #1905, opened 2018-11-30
Added command to refresh topseller

FiveDigital:
Topseller can currently only be refreshed via backend or cron. 

Adds a console command to refresh the topseller index.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Commands/RefreshTopsellerCommand.php
+++ b/engine/Shopware/Commands/RefreshTopsellerCommand.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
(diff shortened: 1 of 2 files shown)


FiveDigital:
Please also note that there is some kind of misbehaviour using the cron strategy because of this:

```
        if (!($this->isTopSellerActivated()) || $strategy !== self::AGGREGATE_STRATEGY_CRON_JOB) {
            return true;
        }
```

If the topsellers are not shown in the frontend the index is not generated by the cron. If the index is not generated the `sales` are not put into the elasticsearch index ...

larsbo:
@FiveDigital the wrong frontend dependency has been fixed with Shopware 5.5.3 (https://github.com/shopware/shopware/pull/1262) 😃

FiveDigital:
Ah nice - the PR with the birthday party 😬 👍

shyim:
Code Style check fails 😉

FiveDigital:
It should pass now.

PR #1906, opened 2018-11-30
Set correct regex for xss filter to filter data-* attributes correctly

zapato:
Usage of URL domain names which contain `data-` could not be used. For example when a domain name contains `data-` then Shopware will set the `$_SERVER[HTTP_HOST]` to `NULL`

So we are not able to use Shopware without workarounds for our Domains `*.data-room.de` and `*.data-graphis.de`

Adds the missing `=` to the xss regex to filter a html `data-*` attribute and keeps values like the `HTTP_HOST` with `data-` values.

Install Shopware to an domain which contains `data-` and call `$this->Request()->getHttpHost()` into a controller

none

none

### 6. Checklist

- [x] I have squashed any insignificant commits (only a single commit)


zapato:
Hi @shopwareBot,
I signed the CLA, today

hlohaus:
You cam simply disable the input filter.

> So we are not able to use Shopware without workarounds for our Domains *.data-room.de and *.data-graphis.de

PR #1913, opened 2018-12-07
Fix: DataTime Bug

tw-94:
Fix: application crashed if isnt datatime but has Format like 3070-0408-121221

Plugins with a search function will Response Error: 503 - Service Unavailable

Fixes the problem that a string in **String** format is no longer initialized as a **date**, cause execute further lines like shopware\engine\Shopware\Controllers\Backend\Application.php :
    --> error is not caught that it is a string
    $date = new DateTime($value);
    $value = $date->format('Y-m-d');

When entering a Format like "3070-0408-121221" or "3070-04131308-12111662221" on search bar with a plugin like Shopware example "SwagProductListingExtension."

shopware\engine\Shopware\Controllers\Backend

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Application.php
+++ b/engine/Shopware/Controllers/Backend/Application.php
@@ -1141,7 +1141,7 @@ class Shopware_Controllers_Backend_Application extends Shopware_Controllers_Back
                     break;
                 }
                 //Fix: check if type is 'DateTime', otherwise return
-                if(!is_a($value, 'DateTime')) {
+                if (!$value instanceof \DateTime) {
 					break; 
 				}
                 $date = new DateTime($value);
--- a/engine/Shopware/Controllers/Backend/Application.php
+++ b/engine/Shopware/Controllers/Backend/Application.php
@@ -1140,10 +1140,12 @@ class Shopware_Controllers_Backend_Application extends Shopware_Controllers_Back
                     $value = '%' . $value . '%';
                     break;
                 }
+                
                 //Fix: check if type is 'DateTime', otherwise return
                 if (!$value instanceof \DateTime) {
                     break; 
                 }
+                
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Application.php
```diff
@@ -1140,7 +1140,10 @@ protected function formatSearchValue($value, array $field, $expression = null)
                     $value = '%' . $value . '%';
                     break;
                 }
-
+                //Fix: check if type is 'DateTime', otherwise return
+                if(!is_a($value, 'DateTime')) {
```
bcremer: Please use the `instanceof` operator here:

```
if (!$value instanceof \DateTime)
```

tw-94: changed, lovely regards.

shyim: And now squash your commits 😉

engine/Shopware/Controllers/Backend/Application.php
```diff
@@ -413,7 +413,7 @@ public function save($data)
                 'property' => $violation->getPropertyPath(),
             ];
         }
-
+        /** @var catch wrong date conversion $errors */
```
nlubisch: This docblock makes no sense. Please remove it.

nlubisch (changes requested):

nlubisch:
Please squash your commits.
For a explanation about how to do it look here: https://davidwalsh.name/squash-commits-git

engine/Shopware/Controllers/Backend/Application.php
soebbing: Why not use the `\DateTimeInterface`?

PR #1914, opened 2018-12-07
added event TemplateMail_CreateMail_MailContext to change mail context

uehler:
if you want to add some information to a mail context, you have to override the createMail-method in Shopware_Components_TemplateMail, which isn't really fun and can break stuff

adds the event TemplateMail_CreateMail_MailContext, where you can change the mail context

try to add information to the context of a mail

### 6. Checklist


uehler:
That's a real Quick-Pick ;) Great!

PaddyS:
Quickpick

PR #1917, opened 2018-12-11
Remove platform override in composer.json for allowing newest PHP version

pascalheidmann-bedarf:
When installing additional composer packages which are only available for PHP 7 composer won't install those. Also forcing a completely outdated version of PHP (PHP 5.6 life ends in 19 days) is not a very nice move as you should always use the newest compatible version of 3rd party libs.

Removing `config.platform` key from `composer.json`

https://issues.shopware.com/issues/SW-23133

### 6. Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -127,10 +127,7 @@
     },
     "config": {
         "autoloader-suffix": "Shopware",
-        "optimize-autoloader": true,
-        "platform": {
-            "php": "5.6.4"
-        }
+        "optimize-autoloader": true
     },
     "scripts": {
         "cs-check": "php-cs-fixer fix --dry-run -v",


bcremer:
The platform config is crucial for building the installation zip file.
If the platform config is removed a run of `composer install` might pull in dependencies that are not compatible with PHP 5.6 which is the lowest supportet PHP Version in Shopware.

Also for the Shopware core developers it's important that only dependencies a pulled in via composer that are PHP 5.6 compatible.

Of course you can run Shopware with a higher PHP Version. This restriction is set in the `require` section of the `composer.json`-file, see: https://github.com/shopware/shopware/blob/v5.5.4/composer.json#L18

Alternatively you can use [shopware/composer-project](https://github.com/shopware/composer-project).
With the `composer-project` Shopware itself is pulled as a dependency so you can define the platform however you want.

bcremer:
Interesting bit from the Shopware 5.5 Upgrade Guide:

https://developers.shopware.com/developers-guide/shopware-5-upgrade-guide-for-developers//#system-requirements-changes

> The minimum PHP version still is PHP 5.6.4 or higher, though PHP 7.x is highly encouraged. Please be aware that PHP 5.6 and PHP 7.0 reach their "end of life" (the end of the official support by the PHP Group) by the end of the year. **Shopware will drop support for those versions of PHP with the next minor version.**

So there is a good chance that shopware will drop PHP 5.6 Support for Shopware 5.6.
Maybe @soebbing can chime in here and shed a light.

soebbing:
As @bcremer already pointed out, it is not that easy to simply raise the required PHP version, both from a technical and also from an organizational standpoint, as lot of customers still use PHP 5.6. 

Also, the mentioned `composer-project` is a good alternative if you depend on more recent versions of PHP.

In any case, changing such fundamental requirements in a patch release like 5.5.5 is out of the question. Therefore we'll have to close this PR.

pascalheidmann-bedarf:
@soebbing I saw this more of a PR for 5.6 anyway based on the implications that were mentioned.
Any chance for that change by then?

soebbing:
@pascalheidman-bedarf Oh, I see. I thought the target was 5.5 as that is the target branch of this PR ;)

Shopware 5.6 is supposed to drop the support for PHP 5.6, so feel free to create a new PR to that branch. Please also change the README.md accordingly, do a search for "5.6.4" on the codebase to find other references to the old PHP version and run a `composer update` to receive possibly updated libraries

PR #1918, opened 2018-12-13
Changed itemSize calculation of swProductSlider for correct itemSize …

Venistro:
…and observance of minItemWidth and minItemHeight on product slider with padding

It fixes the calculation of itemSize on swProductSlider with padding for correct itemSize and observance of minItemWidth/minItemHeight. 

It calculates the containerSize of swProductSlider correctly on the container of all slider items and not on the size of the whole slider to avoid display errors in some viewports and that itemSize could fall below minItemWidth/minItemHeight on product slider with padding.   

Adding a larger padding to swProductSlider and resize the viewport. 

https://issues.shopware.com/issues/SW-23145

### 6. Checklist

### DIFF
--- a/UPGRADE-5.5.md
+++ b/UPGRADE-5.5.md
@@ -11,6 +11,11 @@ This changelog references changes done in Shopware 5.5 patch versions.
 * Added new event `TemplateMail_CreateMail_MailContext` to `engine/Shopware/Components/TemplateMail.php`
 * Added internal locking to sitemap generation so the sitemap isn't generated multiple times in parallel
 
+### Changes
+
+* Changed `itemSize` calculation of `swProductSlider` for correct `itemSize` and observance of `minItemWidth`/`minItemHeight` on product slider with padding
+
+
 ## 5.5.4
 
 [View all changes from v5.5.3...v5.5.4](https://github.com/shopware/shopware/compare/v5.5.3...v5.5.4)
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.product-slider.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.product-slider.js
@@ -505,7 +505,7 @@
         setSizes: function (orientation) {
             var me = this,
                 o = orientation || me.opts.orientation,
-                containerSize = (o === 'vertical') ? me.$el.innerHeight() : me.$el.innerWidth(),
+                containerSize = (o === 'vertical') ? me.$container.height() : me.$container.width(),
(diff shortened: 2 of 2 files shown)


PaddyS:
Hey @Xiut,

can you be more precise on how to reproduce the issue? :-)
LIke this: How much padding did you add to which element - and how?
How does it look like then? (Maybe with a screenshot attached)

I've just added several paddings to several elements of the slider and never had any issues.

Venistro:
Hey @PaddyS,

sorry, it was late when I wrote this pull request, so I forgot to put my code for reproducing this issue :( 

Following code adds padding to the right element and shows an error in the console when the width of the slider items falls below itemMinWidth:  

```
$('*[data-product-slider="true"]').css('padding', '70px');

$.subscribe('plugin/swProductSlider/onSetSizes', function(event, plugin){
   
    if( plugin.itemSize < plugin.opts.itemMinWidth )
    {
        console.error(plugin.itemSize + ' < ' + plugin.opts.itemMinWidth); 
    } 

});
```

It depends very much on how the product boxes and sliders and everything else around them is designed or which itemMinWidth/itemMinHeight you use, how much you see the error. 

Edit: And of course, it depends on how much the width of product-slider--container is different to the product-slider Element. It dont have to be only padding on the product-slider element.

PaddyS:
Hey @Xiut,

I am really sorry, but I am still unable to reproduce any issues with this.

1st: Can you maybe attach a screenshot of how it looks when it breaks in your setup without your changes?

2nd: Does it work, if you add the padding via CSS instead of JS?
The size calculation is done only once and manipulating the item-size afterwards via JS may already be the issue in your case.

Venistro:
Hey @PaddyS, 

my code should show that minItemWidth will not be maintained, when you add padding to the .product-slider Element or make other changes where the width of the .product-slider--container element is smaller than the .product-slider Element. 

It doesn't matter whether the padding is added via JavaScript or CSS, for example, because the sizes are recalculated not only once, but with each resize of the window. 

At the moment I cannot show the theme of my client, but I can show you the issue on the responsive theme:
**Without my fix:** 
without_fix

**With my fix:**
with_fix

The slider got a padding of 30px left and right and the minItemWidth is 220px (default value).

With the Responsive theme, the problem may not make much difference, but in some designs it is important that the width of the productboxes is greater than or equal to the minItemWidth. 
I mean, that's what the value is for and with the small fix this is kept until the container itself is smaller than minItemWidth. 

And with bigger difference between .product-slider and .product-slider--container the problem becomes even bigger. 

But if it's not a issue for you to fix, I can just post my little fix here in the form of a code snippet, for those who also have this problem. If that's not a problem.

PR #1925, opened 2018-12-17
Add filter to sBasket

danielwolf1:
Add filter Shopware_Modules_Basket_CheckBasketQuantities_ProductQuantities in order to modify OutOfStock variable for each basket item.

In order to add individual code to check product quantity

This change simply adds a new filter.

### 6. Checklist


shyim (changes requested):
Please squash the commits

PR #1926, opened 2018-12-17
Add filter Shopware_Modules_Basket_CheckBasketQuantities_ProductQuant…

danielwolf1:
Add filter Shopware_Modules_Basket_CheckBasketQuantities_ProductQuantities in order to modify OutOfStock variable for each basket item.

1. Why is this change necessary?
In order to add individual code to check product quantity

2. What does this change do, exactly?
This change simply adds a new filter.

6. Checklist
[x]  I have written tests and verified that they fail without my change
[x]  I have squashed any insignificant commits
[x] This change has comments for package types, values, functions, and non-obvious lines of code
[x]  I have read the contribution requirements and fulfil them.

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -259,6 +259,11 @@ class sBasket
             }
         }
 
+        $products = $this->eventManager->filter('Shopware_Modules_Basket_CheckBasketQuantities_ProductsQuantity', $products, [
+            'subject' => $this,
+            'hideBasket' => $hideBasket,
+        ]);
+
         return ['hideBasket' => $hideBasket, 'articles' => $products];
     }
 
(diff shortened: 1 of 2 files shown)


engine/Shopware/Core/sBasket.php
```diff
@@ -259,6 +259,11 @@ public function sCheckBasketQuantities()
             }
         }
 
+        $products = $this->eventManager->filter('Shopware_Modules_Basket_CheckBasketQuantities_ProductQuantities', $products, [
```
shyim: Please change the event name to the one below, because there are multiple products with one quantity each
```suggestion
        $products = $this->eventManager->filter('Shopware_Modules_Basket_CheckBasketQuantities_ProductsQuantity', $products, [
```

shyim:
Andddd now squash :)

danielwolf1:
> Andddd now squash :)

Doooone! (finally) :-)

PR #1927, opened 2018-12-18
Fix performance leak in sGetSimilarArticles

Ancillius:
The function sGetSimilarArticles generates SQL code that does not perform well on shops with many articles (e.g. 300k), since the left joins unneccessary create multiple lines. In an example case the result with 350k articles, 500  categories (approx 3 categories for 1 article - its a book store - ) this query generates over 2,1 million lines  and runs on a quite big dedecated server for 7-15sek, since the big result tables needs to be written to disc my the database.

To group and sort 2,1 million lines for 4 random lines picket by a simple relevance is not good.

When a crawler comes by a shop and crawls many unavailable products (ignoring the sitemap) the server is flooded by long running queries and begins to lag.

The idea here is, to split those 3 relevances into independent queries und unionize the result. This yields the exact same four randomized results (in terms of relevance, whilst of course not the exact same 4, since the DB does not have to order the result).

Performance leak is severe. Brought a PX91 NVME Hetzner Server to its knees.

The way 4 random products are picked ordered by the same relevance as before.

Crawl non existing products, or non-available products to call the Product-Not-Found-Error page in a "big" shop (read above).

None

### 6. Checklist

- [ - ] I have written tests and verified that they fail without my change -- no its live on www.beam-shop.de and the performance impact on an up to date mariaDB is enourmous (Factor 15000)
- [ - ] This change has comments for package types, values, functions, and non-obvious lines of code - Its "just" SQL. Not more complex than before.
- [ x ] I have read the contribution requirements and fulfil them. - Think so

### DIFF
--- a/engine/Shopware/Core/sMarketing.php
+++ b/engine/Shopware/Core/sMarketing.php
@@ -511,50 +511,119 @@ class sMarketing
     {
         $limit = empty($limit) ? 6 : (int) $limit;
         $productId = empty($articleId) ? (int) $this->sSYSTEM->_GET['sArticle'] : (int) $articleId;
+        $sql = <<<SQL
+SELECT u.articleID, u.articleName, u.Rel 
+  FROM (
+    
+    ( 
+    SELECT DISTINCT
+   
+      a.id as articleID, 
+      a.name as articleName, 
+      3 as Rel
+      
+    FROM s_articles a
+    
+      INNER JOIN s_articles_categories_ro ac 
+        ON ac.articleID = a.id 
+        AND ac.categoryID = {$this->categoryId}
+        
+      INNER JOIN s_categories c 
+        ON c.id = ac.categoryID 
+        AND c.active = 1
+      
+      LEFT JOIN s_articles_avoid_customergroups ag 
+        ON ag.articleID = a.id 
+        AND ag.customergroupID = {$this->customerGroupId}
+      
+      INNER JOIN s_articles_similar s 
+        ON s.articleID = a.id 
+        AND s.relatedarticle = a.id
+        
+      INNER JOIN s_articles_categories_ro s1 
+        ON s1.articleID = a.id
+        
+      INNER JOIN s_articles_categories_ro s2 
+        ON s2.categoryID = s1.categoryID 
+        AND s2.articleID = a.id
+        
+      WHERE a.active = 1
+        AND ag.articleID IS NULL
+        AND a.id != {$productId}
+        
+    LIMIT {$limit}
+  ) 
+  UNION ( SELECT DISTINCT
+   
+      a.id as articleID, 
+      a.name as articleName, 
+      2 as Rel
+      
+    FROM s_articles a
+    
+      INNER JOIN s_articles_categories_ro ac 
+        ON ac.articleID = a.id 
+        AND ac.categoryID = {$this->categoryId}
+        
+      INNER JOIN s_categories c 
+        ON c.id = ac.categoryID 
+        AND c.active = 1
+      
+      LEFT JOIN s_articles_avoid_customergroups ag 
+        ON ag.articleID = a.id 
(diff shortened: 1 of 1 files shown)


engine/Shopware/Core/sMarketing.php
```diff
@@ -494,49 +494,118 @@ public function sGetSimilarArticles($articleId = null, $limit = null)
...
+   
+      a.id as articleID, 
+      a.name as articleName, 
+      3 as Rel
+      
+    FROM s_articles a
+    
+      INNER JOIN s_articles_categories_ro ac 
+        ON ac.articleID = a.id 
+        AND ac.categoryID = {$subject->categoryId}
```
shyim: That looks for me for a Typo. I guess you have copied that code from your replace hook? :D
Maybe you want $this->categoryId instead

shyim:
You have a wrong target branch. Please change it to 5.5

engine/Shopware/Core/sMarketing.php
Ancillius: Yes you are right... i am sorry, will correct....

Ancillius:
> You have a wrong target branch. Please change it to 5.5

Tbh we would need it in 5.4 also... and I tested on 5.4 (seems totally legit for 5.5, but since we run on 5.4...)

shyim:
Anddddddddddddddd squash the commits :sweat_smile:

Ancillius:
> Anddddddddddddddd squash the commits

Yep, just realized there seems to be no option for that in github edit mode (gitlab has that tough). Did it locally now.

Ancillius:
For the 5.5 branch, whats your preferred way? Switch this PR or open another one similar to this for 5.5.

I would love to see those changes (if accepted) in 5.4, too.

shyim:
We are not planning anymore a another 5.4 release, so please change this to 5.5 :blush:

Ancillius:
Changed to 5.5

wbob:
> The idea here is, to split those 3 relevances into independent queries und unionize the result

@Ancillius, thanks for taking it up to renovate this query, it is popular in slowlogs if `RelatedArticlesOnArticleNotFound` is enabled - though did you test it on a filled `s_articles_similar` dataset? With the demoset (`_sql/demo/latest.sql`) for articleid 196 ("Artikel mit ähnlichen Produkten"), the old query returns relevance 3 rows, while the UNION query will not.

I think the UNION approach with 3 relevance tiers is a good idea, but if I'm not in the wrong, some more work is neccessary to fulfil the querys original purpose.

philipreinken:
Hey @wbob ,

I'm getting similar results, the higher relevances (> 1) seemingly aren't being
selected anymore.

I think it might be due to this inner join in both of the higher relevance
queries:

```
INNER JOIN s_articles_similar s 
  ON s.articleID = a.id 
  AND s.relatedarticle = a.id
```

This would lead to nothing being selected, as no product is related to itself in
`s_articles_similar`.

Can you confirm this? If so, we'd be glad if you could address this issue :slightly_smiling_face:

wbob:
the former query did 2 things at the same time - rank by similar article and check for same category. It was a left join to `o.id` (others I presume). Within the UNION approach that query can be simplified.
```
    SELECT s.relatedarticle,a.name
      FROM ...
INNER JOIN s_articles_similar s 
        ON s.relatedarticle = a.id
        ..
     WHERE s.articleid = @aid
  GROUP BY s.relatedarticle;
```
While I thought the idea of UNIONs to be interesting, I came to the conclusion it is better to do individual querys per relevance and only query the next, less focused tier if the result comes back empty. Like: `if (empty($similarArticles)) { if (empty($sameLeafCategoriesArticles)) { return $randomArticles }}`. I don't think many end-users populate `s_articles_similar`, so a performant sameLeafCategoriesArticles query could be handy.

PR #1931, opened 2018-12-21
Resolve one-to-many relations in backend applications

JoshuaBehrens:
Without this fix it was not possible to save extjs models with a one-to-many relation. A specific backend application can extend the changed behaviour to make it work right away but this change had to be done on multiple applications.

It extends the analysis of referenced models by loading models from arrays for a one-to-many relation.

Two levels of nested extjs relations is my understanding.

### 4. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Application.php
+++ b/engine/Shopware/Controllers/Backend/Application.php
@@ -866,7 +866,7 @@ class Shopware_Controllers_Backend_Application extends Shopware_Controllers_Back
                     //remove the foreign key data.
                     unset($data[$field]);
                 }
-            } elseif ($mapping['type'] === ClassMetadataInfo::MANY_TO_MANY) {
+            } elseif (in_array($mapping['type'], [ClassMetadataInfo::MANY_TO_MANY, ClassMetadataInfo::ONE_TO_MANY], true)) {
                 /**
                  * @ORM\ManyToMany associations.
                  *


shyim:
Hey @JoshuaBehrens,

thanks for the Contribution :blue_heart:.
Can you please rebase your Branch to the newest changes in the 5.5 Branch, so we can sure that the tests are passing :blush:

pascalheidmann-bedarf (changes requested):
Pls add your change to the doc block

JoshuaBehrens:
@pascalheidman-bedarf Your comment is justified. I guess I need to simplify my production problem to have a good example for the doc block.

shyim:
Hey @JoshuaBehrens,
could you provide a Code-Example to archive the Issue? I have tested your Changes against the example plugin of https://developers.shopware.com/developers-guide/backend-components/associations/ and could not find any changes :)

JoshuaBehrens:
I have time for it this evening. I wanted to fresh up all the PRs this week :)

JoshuaBehrens:
I guess I worked down the application to the problem. I still have to anonymize it.

shyim:
I have tested your provided Plugin. Your change does not affect anything.

It still continues at owning side condition. https://github.com/shopware/shopware/blob/3d8fe7cfeddc935367eb2550ced7b1eb9e2eb8ce/engine/Shopware/Controllers/Backend/Application.php#L838

Are you sure with your change?

PaddyS:
Hey @JoshuaBehrens,

we're still not a hundred percent sure if we really want that.
Can you maybe provide an example plugin including the actual issue?
Like some case of "Here I'd like to save Foo, but that doesn't work at the moment".

This way we could check if this is an actual issue in Shopware or just a weird scenario, that could and **should** be solved with a different approach.

Doesn't have to be public here, PN in Gitter or anything would be fine as well.

Would that be possible?

JoshuaBehrens:
I will mail you guys the original plugin then.

pascalheidmann-bedarf:
Any chance that this will fixed/merged some day?

JoshuaBehrens:
To clear things up I've been traveling back in time (code-wise) when the original error happened and recreated it:

> Fatal error: Uncaught TypeError: Argument 1 passed to `FooBar\Models\ModelWithManyToOneRelationToArticle::setArticle()` must be an instance of `Shopware\Models\Article\Article`, `array` given, called in `/vendor/shopware/shopware/engine/Shopware/Components/Model/ModelEntity.php` on line 56 and defined in `/custom/plugins/FooBar/Models/ModelWithManyToOneRelationToArticle.php` on line 129

It can be hotfixed when you overwrite your controllers `resolveExtJsData` and unset the passed data at the key of the relation. `unset($data['article']);` so the relation is ignored. Not a real fix if you want to save it but a fix to stop the error from happening.

shyim:
I have discussed this PR with @JoshuaBehrens and it can be fixed with an `setOneToMany` ([example](https://github.com/shopware/shopware/blob/00b097df06ec124ab8399eaec403a06374802f18/engine/Shopware/Models/Shop/TemplateConfig/Element.php#L283)) in the model. This method is already used in many places for the same problem.

Would this fix also your problem @pascalheidmann-bedarf ?

PR #1933, opened 2018-12-28
Autoselect Variant with non default configurator

pascalheidmann-bedarf:
If you change your configurator to not standard (e.g. image or selection) the default variant will not be auto selected (if no `number` parameter is passed)

Allow any configurator be valid. Also remove not initialized variable `$configuration`.

Without PR create a variant article and check autoselection in frontend. Then change the variant article to image and load the article in frontend again without any `number` parameter. This won't select the default selection.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sArticles.php
+++ b/engine/Shopware/Core/sArticles.php
@@ -1104,11 +1104,10 @@ class sArticles
              * Check if a variant should be loaded. And load the configuration for the variant for pre selection.
              *
              * Requires the following scenario:
-             * 1. $number has to be set (without a number we can't load a configuration)
-             * 2. $number is equals to $productNumber (if the order number is invalid or inactive fallback to main variant)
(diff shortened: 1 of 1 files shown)


engine/Shopware/Core/sArticles.php
```diff
@@ -1108,7 +1108,7 @@ public function sGetArticleById($id = 0, $sCategoryID = null, $number = null, ar
              * 2. $number is equals to $productNumber (if the order number is invalid or inactive fallback to main variant)
              * 3. $configuration is empty (Customer hasn't not set an own configuration)
```
shyim: You forgot to update the phpdoc

pascalheidmann-bedarf:
@shopwareBot would implementing this as a feature toggle be acceptable, either as a configurator prop or global?

Considering that all but the default configurator do not autoselect is a major UX flaw as it prevents to easily putting the default variant (see *default*?) into your basket and requiring another click.

PR #1935, opened 2019-01-03
show company in widget if it's not empty

tinect:
current statement doesn't work and shows always first- and lastname

replaces #1860 

add check for empty company

### 6. Checklist

### DIFF
--- a/tests/Functional/Controllers/Backend/WidgetsTest.php
+++ b/tests/Functional/Controllers/Backend/WidgetsTest.php
@@ -21,6 +21,7 @@
  * trademark license. Therefore any rights, title and interest in
  * our trademarks remain entirely with us.
  */
+use Doctrine\DBAL\Connection;
 
 /**
  * @category  Shopware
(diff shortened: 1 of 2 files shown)


tests/Functional/Controllers/Backend/WidgetsTest.php
```diff
@@ -156,4 +168,135 @@ public function testIfNoConversionAfterEightDays()
...
+            'lastname' => 'Test',
+            'street' => 'Teststreet 1',
+            'country_id' => 2,
+        ];
+
+        $this->prepareTestGetVisitors($addressData);
+
+        $response = $this->dispatch('backend/widgets/getVisitors?page=1&start=0&limit=25');
+
+        $jsonBody = json_decode($response->getBody(), true);
```
shyim: $this->View()->getAssign() does the same

tinect: So true 🙇

PR #1936, opened 2019-01-03
Create Alias for shopware.api.customer_stream to fix cronjob

pascalheidmann-bedarf:
The CustomerStream cronjob uses the API resource class as identifier. This causes issues as `Shopware\Components\Api\Manager::getResource()` is using plain `strtolower` instead of converting `CamelCase` to `snake_case` which is used in DI services.

Adds an alias for `shopware.api.customer_stream` to `shopware.api.customerstream` so the services is found. Instead of using `strtolower` a proper transformation would be better but might break things...

- Try to run the customerstream cronjob (`php bin/console sw:cron:run Shopware_CronJob_RefreshCustomerStreams` in your shopware root) 
- Get error message `Uncaught ArgumentCountError: Too few arguments to function Shopware\Components\Api\Resource\CustomerStream::__construct(), 0 passed (...)`
- Apply fix and rerun cron
- Profit 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/services.xml
+++ b/engine/Shopware/Components/DependencyInjection/services.xml
@@ -643,6 +643,8 @@
             <argument id="shopware.customer_stream.criteria_factory" type="service"/>
         </service>
 
+        <service id="shopware.api.customerstream" alias="shopware.api.customer_stream" />
+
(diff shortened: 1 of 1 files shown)


shyim:

PR #1938, opened 2019-01-03
Fix overwrite protection time zone bug when saving an order detail or deleting one

windaishi:
When client and server are in different timezones you cannot save an order detail nor remove one.

Changes the format of the transferred timestamp to timezone-independent ISO format.

Like here: https://github.com/shopware/shopware/commit/d389efdd56a48b68e81791dd3c4fd17aeb927118#diff-3de4436f3f4322b381d75442bd6edbf0L62

Access a Shopware backend with a browser working in a different timezone as the server and try to edit/delete an order position.

N/A

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/order/controller/detail.js
+++ b/themes/Backend/ExtJs/backend/order/controller/detail.js
@@ -304,7 +304,7 @@ Ext.define('Shopware.apps.Order.controller.Detail', {
 
         e.record.save({
             params: {
-                changed: order.get('changed'),
+                changed: order.get('changed') ? order.get('changed').toISOString() : null,
             },
             callback:function (data, operation) {
                 var records = operation.getRecords(),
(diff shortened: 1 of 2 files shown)


themes/Backend/ExtJs/backend/order/controller/detail.js
```diff
@@ -304,7 +304,7 @@ Ext.define('Shopware.apps.Order.controller.Detail', {
 
         e.record.save({
             params: {
-                changed: order.get('changed'),
+                changed: order.get('changed') ? order.get('changed').toISOString() : null,
```
shyim, line 307: There are more places with the same code
e.g Customer, Product overview

windaishi: As far as I can remember, this already has been fixed here:
https://github.com/shopware/shopware/blob/d389efdd56a48b68e81791dd3c4fd17aeb927118/themes/Backend/ExtJs/backend/article/model/article.js#L62
and here: 
https://github.com/shopware/shopware/blob/d389efdd56a48b68e81791dd3c4fd17aeb927118/themes/Backend/ExtJs/backend/customer/model/customer.js#L53

PR #1940, opened 2019-01-07
Adding shopware mail template name to zend mail object

Isengo1989:
There is only one real Event for checking the Mail before sending it Enlight_Components_Mail_Send - other events are only notify-events. Therefore it would be nice to know which Mail will be sent (sORDER / sREGISTERCONFIRMATION etc.) in case you want to send a mail with third party (mailchimp, emarsys etc.)

Right now there is no indicator whatsoever which mail will be sent. You can only differ these mails by their subject, which will eventually change (subshop, some admin etc.)

It sets a template name available in the Zendmail object, which is needed for external services and other catches.

New protected template (setter/getter) which can be set in the TemplateMail Class

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Components/Mail.php
+++ b/engine/Library/Enlight/Components/Mail.php
@@ -71,6 +71,13 @@ class Enlight_Components_Mail extends Zend_Mail
     protected $_plainSubject = null;
 
     /**
+     * Property for the template name. Can be filled by setTemplateName function.
+     *
+     * @var null
+     */
+    protected $templateName = null;
+
+    /**
      * Magic setter method
      *
      * @param string $name
@@ -310,6 +317,22 @@ class Enlight_Components_Mail extends Zend_Mail
         $this->_plainSubject = $subject;
 
         return parent::setSubject($subject);
+    }
+
+    /**
+     * @param string $templateName
+     */
+    public function setTemplateName($templateName)
+    {
+        $this->templateName = $templateName;
+    }
+
+    /**
+     * @return string
+     */
+    public function getTemplateName()
+    {
+        return $this->templateName;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


engine/Library/Enlight/Components/Mail.php
```diff
@@ -312,6 +319,22 @@ public function setSubject($subject)
...
+     */
+    public function setTemplate($template)
+    {
+        $this->template = $template;
+    }
+
+    /**
+     * @return string
+     */
+    public function getTemplate()
```
shyim: Maybe something like getTemplateName 🤔

engine/Shopware/Components/TemplateMail.php
```diff
@@ -21,7 +21,6 @@
  * trademark license. Therefore any rights, title and interest in
  * our trademarks remain entirely with us.
  */
-
```
shyim: This line breaks travis :)

Isengo1989: Weird - I ran php-cs-fixer before commiting. Thx - I will fix it tmrw

engine/Shopware/Components/TemplateMail.php
```diff
@@ -317,6 +317,8 @@ public function loadValues(\Enlight_Components_Mail $mail, Mail $mailModel, $ove
             }
         }
 
+        $mail->setTemplate($mailModel->getName());
```
shyim: setTemplateName :see_no_evil:

Isengo1989: haha, "I will change this really quick, before I go to work. But don't forget to change the call of the setter" :duck: 

Travis is failing all the time - but because of:
"HP Fatal error:  Uncaught Doctrine\ORM\ORMException: The EntityManager is closed. in /home/travis/build/shopware/shopware/vendor/doctrine/orm/lib/Doctrine/ORM/ORMException.php:144" right?

Isengo1989:
@shopwareBot fixed the issue and tests passed

PR #1941, opened 2019-01-07
Correctly use typeof keyword and use strict comparison

DanielRuf:
`typeof` is a keyword, not a function. Also strict comparison is generally better.

Change `typeof($var)` to `typeof $var` and change a few loose comparisons to strict comparisons.

### 6. Checklist


shyim:
Thanks for the contribution :blue_heart:

PR #1944, opened 2019-01-07
Correctly check the plugin search term length if it is 0

DanielRuf:
The check for `length < 0` always return `false` as `length` can never be smaller than `0`.

Fixes the length check, a length property can never be smaller than `0`

### 6. Checklist


shyim:
Thanks for the contribution :blue_heart: !

PR #1946, opened 2019-01-07
Remove accidentally typed character

DanielRuf:
There is an extra `z` in the CSS which breaks the next CSS rule (see https://codepen.io/DanielRuf/pen/JwZKmY)

Remove the wrong `z` character.

### 6. Checklist


shyim:

PR #1947, opened 2019-01-07
Remove unnecessary conditional structure

DanielRuf:
The conditional is not needed here it is uses the same code in both logical branches.

Remove the unneeded conditional.

### 6. Checklist


shyim:
Thanks for the contribution :blue_heart: !

PR #1950, opened 2019-01-08
Enable the testWithNumericArticleNumbers test

DanielRuf:
The `testWithNumericArticleNumbers` is currently not executed. It's unclear why. Let's see if it fails or if it succeeds and what we can do then.

### 6. Checklist

### DIFF
--- a/tests/Functional/Plugins/Core/MarketingAggregate/Components/AlsoBoughtTest.php
+++ b/tests/Functional/Plugins/Core/MarketingAggregate/Components/AlsoBoughtTest.php
@@ -84,7 +84,8 @@ class AlsoBoughtTest extends AbstractMarketing
 
     public function testInitAlsoBought()
     {
-        $this->assertCount(30, $this->getAllAlsoBought());
+        print_r($this->getAllAlsoBought());
+        $this->assertCount(27, $this->getAllAlsoBought());
     }
 
     public function testRefreshBoughtArticles()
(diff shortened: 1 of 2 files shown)


mitelg:
strange, the test itself seems not to be the problem

DanielRuf:
https://travis-ci.org/shopware/shopware/jobs/476856003#L1437

DanielRuf:
Hm, is there some side effect which causes this? Because there are 3 products in the result and now 3 are missing from one assertion result.

DanielRuf:
Seems so. Shall we adjust the assertion or fill up the "also bought" articles to get 30 again?

tests/Functional/Plugins/Core/MarketingAggregate/Components/AlsoBoughtTest.php
```diff
@@ -84,7 +84,8 @@ protected function tearDown()
 
     public function testInitAlsoBought()
     {
-        $this->assertCount(30, $this->getAllAlsoBought());
+        print_r($this->getAllAlsoBought());
```
shopwareBot, line 87: Here is a debug :relaxed:

DanielRuf: Ah sorry, forgot to remove it but yes, the PR was meant as "WIP" currently and I've expected some feedback from @mitelg about the right solution here.

> Seems so. Shall we adjust the assertion or fill up the "also bought" articles to get 30 again?

https://github.com/shopware/shopware/pull/1950#issuecomment-452471240

PR #1952, opened 2019-01-08
Rename CSS property to break-inside as column-break-inside is not a v…

DanielRuf:
…alid CSS property

`column-break-inside` is not a valid CSS selector so this will only work in webkit browsers.

Use the right `break-inside` property.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/_public/src/less/_mixins/column-count.less
+++ b/themes/Frontend/Bare/frontend/_public/src/less/_mixins/column-count.less
@@ -18,5 +18,6 @@ Please refer to <http://caniuse.com/multicolumn> to see the browser support tabl
 
 .column-break-inside(@type: avoid) {
     -webkit-column-break-inside: @type;
-    column-break-inside: @type;
+    page-break-inside: @type;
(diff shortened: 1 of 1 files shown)


themes/Frontend/Bare/frontend/_public/src/less/_mixins/column-count.less
```diff
@@ -18,5 +18,5 @@ Please refer to <http://caniuse.com/multicolumn> to see the browser support tabl
 
 .column-break-inside(@type: avoid) {
     -webkit-column-break-inside: @type;
-    column-break-inside: @type;
-}
\ No newline at end of file
+    break-inside: @type;
```
shopwareBot, line 22: Please add also `page-break-inside` property

DanielRuf: Done. Shall I squash commits or it is ok to have two?

shyim: Squash :)

PR #1953, opened 2019-01-08
Remove the CSS touch-callout property which does not officially exist

DanielRuf:
`touch-callout` is an exclusive property in webkit based browsers and only exists with the prefix.

Remove the wrong `touch-callout` and the Android reference.

### 6. Checklist


DanielRuf:
I'll consider this but I am generally doing atomic commits. Some may consists of multiple fixed issues but it depends on the findings and the reports that I have (multiple pages of found issues).

PR #1955, opened 2019-01-08
Remove the delete operator as it does not remove variables

DanielRuf:
`delete` is meant to delete properties of objects. Variables can not be unset or deleted. `let` is recommended for scoped variables.
https://caniuse.com/#feat=let

Either this or we just remove the `delete` line.

Replace the `delete` with a block scoped `let`.

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/translation/controller/main.js
+++ b/themes/Backend/ExtJs/backend/translation/controller/main.js
@@ -357,7 +357,6 @@ Ext.define('Shopware.apps.Translation.controller.Main',
                 }
 
                 data.push({ valueField: field.emptyText, displayField: label || field.fieldLabel });
-                delete label;
             }
         });
 


shyim:
See IE 11 Support. So can't be used :)

DanielRuf:
I do not directly see a `for` loop here.

> let variables are not bound separately to each iteration of for loops

mitelg:
`Ext.each()` uses a `for` loop internally

see here: https://raw.githubusercontent.com/shopware/shopware/5.5/engine/Library/ExtJs/ext-all-debug.js
search for `Ext.each =` and then for `ExtArray =`

DanielRuf:
> `Ext.each()` uses a `for` loop internally
> 
> see here: [/engine/Library/ExtJs/ext-all-debug.js@`5.5` (raw)](https://raw.githubusercontent.com/shopware/shopware/5.5/engine/Library/ExtJs/ext-all-debug.js)
> search for `Ext.each =` and then for `ExtArray =`

Right.  But this is a different scope and not about iteration variables in the initialization of the loop.
Nonetheless, we could also just remove the line with the `delete` keyword as it does nothing (it is a `noop` and returns `false` as it does nothing here).

For reference (with some workarounds but I would prefer to delete the `delete` operator).

DanielRuf:
> Either this or we just remove the `delete` line.

Please read the whole conversation carefully. `delete` does nothing and returns false. So shall I just remote the line and keep the `var` untouched?

DanielRuf:
I've reverted / changed back the `let` change.

shyim:
And now squash please

DanielRuf:
> And now squash please

Done.

DanielRuf:
I think rewording is better here. I'll do that now.

shyim:
Hey @DanielRuf ,
thanks for the Contribution! :blue_heart: . Your Change has been merged

PR #1957, opened 2019-01-08
Remove unused parameter from removeResizeListener invocation

DanielRuf:
`removeResizeListener` accepts only one parameter so the second parameter (overloading) is not needed.

Remove the unused parameter in the `removeResizeListener` invocation.

### 6. Checklist


DanielRuf:
Mh, it may be better to not track vendor files (if they should not be changed) at all and use composer-patches and patch-package (Yarn, npm, ...) to patch issues in them ("do not edit vendor files").

DanielRuf:
See https://github.com/chartjs/Chart.js/pull/5970

PR #1961, opened 2019-01-10
cache ConfigReader

tinect:
Cache configs of plugins for more than the current stack.

It uses Cache with lifetime in CachedConfigReader.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Plugin/CachedConfigReader.php
+++ b/engine/Shopware/Components/Plugin/CachedConfigReader.php
@@ -25,6 +25,7 @@
 namespace Shopware\Components\Plugin;
 
 use Shopware\Models\Shop\Shop;
+use Zend_Cache_Core as Cache;
 
 /**
  * @category Shopware
@@ -39,16 +40,18 @@ class CachedConfigReader implements ConfigReader
     private $reader;
 
     /**
-     * @var array
+     * @var Cache
      */
-    private $configStorage;
+    private $cache;
 
     /**
      * @param ConfigReader $reader
+     * @param Cache        $cache
      */
-    public function __construct(ConfigReader $reader)
+    public function __construct(ConfigReader $reader, Cache $cache)
     {
         $this->reader = $reader;
+        $this->cache = $cache;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


engine/Shopware/Components/Plugin/CachedConfigReader.php
```diff
@@ -39,20 +40,24 @@ class CachedConfigReader implements ConfigReader
...
+    public function __construct(ConfigReader $reader, Cache $cache)
     {
         $this->reader = $reader;
+        $this->cache = $cache;
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @throws \Zend_Cache_Exception
```
shopwareBot: Please remove the throws

engine/Shopware/Components/Plugin/CachedConfigReader.php
```diff
@@ -62,10 +65,18 @@ public function getByPluginName($pluginName, Shop $shop = null)
...
+            $config = $this->cache->load($cacheKey, true);
+        }
+
         if (!isset($this->configStorage[$cacheKey])) {
-            $this->configStorage[$cacheKey] = $this->reader->getByPluginName($pluginName, $shop);
+            $config = $this->reader->getByPluginName($pluginName, $shop);
         }
 
-        return $this->configStorage[$cacheKey];
+        $this->cache->save($config, $cacheKey, ['Shopware_Config'], 86400);
```
hlohaus, line 74: why you save it, when you loaded it from cache?

hlohaus (changes requested):

engine/Shopware/Components/Plugin/CachedConfigReader.php
```diff
@@ -62,10 +65,16 @@ public function getByPluginName($pluginName, Shop $shop = null)
             $cacheKey = $pluginName;
         }
 
-        if (!isset($this->configStorage[$cacheKey])) {
-            $this->configStorage[$cacheKey] = $this->reader->getByPluginName($pluginName, $shop);
+        $config = [];
+
+        if ($this->cache->test($cacheKey)) {
+            $config = $this->cache->load($cacheKey, true);
```
mitelg: an early `return $config` is missing here

engine/Shopware/Components/Plugin/CachedConfigReader.php
tinect: Cause.... there are no reasons :-) Thank you.

PR #1962, opened 2019-01-10
Added new form element to exclude customergroup key

Isengo1989:
Imagine a shop with a lot of customergroups and a big menu (that is always the same). Every time a user logs in, a new menu will be generated and this will take a long time if you have +150 categories. There should be a way to exclude customergroups to be in the cache key.

auswahl_009

auswahl_242

Espacially the checkout (login/register) is critical, because users will more likely dicard their shopping cart if the login takes longer then a few seconds.

This change adds a new form field **excludeCustomergroup** which will decide if the customergroup will be used in the cache key or not.

The option (boolean) is set to 0 by default (same behavior as before) and even with the plugin not reinstalled it is not true and therefore the customergroups will be considered as usual. I also added the shop scope for caching, cachetime and hoverDelay.

Browse in a big shop (+150 categories) and login with customergroup which is not default.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1453-add-customer-group-key-exemption-option.php
+++ b/_sql/migrations/1453-add-customer-group-key-exemption-option.php
@@ -1,0 +1,76 @@
+<?php
+
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+class Migrations_Migration1453 extends \Shopware\Components\Migrations\AbstractMigration
+{
+    /**
+     * Migration to make new "Exclude Customergroup" option visible in advanced menu without reinstallation
+     *
+     * @param string $modus
+     */
+    public function up($modus)
+    {
+        $sql = <<<'SQL'
+SET @pluginID = (SELECT DISTINCT `id` FROM `s_core_plugins` WHERE `name`='AdvancedMenu' AND `active`=1);
+SQL;
+        $this->addSql($sql);
+
+        $sql = <<<'SQL'
+SET @formID = (SELECT DISTINCT `id` FROM `s_core_config_forms` WHERE `plugin_id`=@pluginID);
+SQL;
+        $this->addSql($sql);
+
+        $sql = <<<'SQL'
+INSERT IGNORE INTO s_core_config_elements (form_id, name, value, label, description, type, required, position, scope, options) VALUES(
+@formID,
+'excludeCustomergroup',
+'i:0;',
+'Kundengruppen ausschließen',
+'Alle Kundengruppen erhalten das gleiche Menü (bessere Perfomance)',
+'boolean',
+0,
+0,
+1,
+0x613A303A7B7D
+);
+SQL;
+        $this->addSql($sql);
+
+        $sql = <<<'SQL'
+SET @elementID = (SELECT id FROM s_core_config_elements WHERE form_id=@formID AND `name`='excludeCustomergroup');
+SQL;
+        $this->addSql($sql);
+
+        $sql = <<<'SQL'
+INSERT IGNORE INTO s_core_config_element_translations (element_id, locale_id, label, description) VALUES (
+@elementID,
+2,
+'Exclude Customergroup',
+'All customergroups will have the same menu (better perfomance)');
+SQL;
+
+        $this->addSql($sql);
+    }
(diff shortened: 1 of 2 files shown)


Isengo1989:
> Hi @Isengo1989, nice PR! Did you check what happens with this change on an update of an existing shop? Your new field is only created when the plugin is installed, which won't happen in that case.

@shopwareBot actually I have no Idea how to do so. Is it "best practice" to write the element in the database with a query on update?

I could call the getVersion() Method in the Bootstrap.php file + enrich the getInfo() Method with the version to show that an update is available

```
    /**
     * @return array
     */
    public function getInfo()
    {
        return [
            'version' => $this->getVersion(),
            'label' => $this->getLabel(),
        ];
    }

    public function getVersion()
    {
        return '1.0.0';
    }
```

I think I can remeber that it is not possible to add a config element without a reinstallation tho, so it would be some INSERT INTO...

The menu works fine, even if the field does not exists. So I don't think it is necessary to force the update.

If you can hint me in the right direction I can find a suitable solution.

Isengo1989:
> Hey @Isengo1989,
> 
> we're afraid there's no way to handle this without using a SQL query.
> 
> You might want to have a look into this example, as we already had to do this in the past:
> https://github.com/shopware/shopware/blob/5.5/_sql/migrations/406-add-advanced-menu-listeners.php#L42

@shopwareBot ok, if the rest is fine, I will look into creating a migration class for this. Is there a naming convention? (**121**-fix-cache-label.php / **Migrations_Migration121** seem to be random)

mitelg:
@Isengo1989 the number is counting up. So you have to look at the latest migration file and add `1` :wink:

Isengo1989:
@mitelg well there are missing a lot of numbers ^^

mitelg:
@Isengo1989 this happens, when we start working on a new minor release. so that we have enough numbers open for bugfix releases :wink:

Isengo1989:
@mitelg and in this case? :stuck_out_tongue_winking_eye:

auswahl_098

Anyways, I guess I scrolled to the wrong place to get the pattern. I know what to do, thanks guys (**edit**: girls and divers!) :)

PaddyS:
@Isengo1989 That's just due to weird sorting in GitHub. :-)
The other migrations are also available.

Isengo1989:
> @Isengo1989 That's just due to weird sorting in GitHub. :-)
> The other migrations are also available.

Class **Migrations_Migration1220** is nowhere to be found in the Repo and in Browser search with "1220-" is empty also :confused: 

1229 -> 1230 is sorted correctly. It's sorted okay.

No I am too deep into this, I have to argument :rofl:

PaddyS:
@Isengo1989 Now I see what you mean.
Thought it was about the jump from "122" to "1221", which is just a sorting issue. :-P

Some migration numbers might be missing, e.g. due to reverted changes.

Isengo1989:
@PaddyS ok, got it. Now I wasted enough time of people for the whole week ^^

Isengo1989:
@shopwareBot I added the migration.

Isengo1989:
@shopwareBot dumb question here - do I always need to squash ALL commits so there is only one commit or can I just squash the last 2 commits into one (since there were other changes to other files and the other commit is way behind in the commit history)

Isengo1989:
Well that was fun :rofl: - rebasing/merging is now my least favorite thing in the world :+1: 

@shopwareBot done

mitelg:
@Isengo1989 Commits should be semantically separated. So e.g. `Changed sBasket core class` and `Fix frontend product detail page`. but this should be avoided: `Fixed sBasket` and `Run cs-fixer` or `Code review changes`

_sql/migrations/1453-add-customer-group-key-exemption-option.php
```diff
@@ -0,0 +1,76 @@
...
+{
+    /**
+     * Migration to make new "Exclude Customergroup" option visible in advanced menu without reinstallation
+     *
+     * @param string $modus
+     */
+    public function up($modus)
+    {
+        $sql = <<<'SQL'
+SET @pluginID = (SELECT DISTINCT `id` FROM `s_core_plugins` WHERE `name`='AdvancedMenu');
```
shopwareBot: Please also add a check if the plugin is installed, otherwise this could crash.

Isengo1989: @shopwareBot done

_sql/migrations/1453-add-customer-group-key-exemption-option.php
```diff
@@ -0,0 +1,76 @@
...
+{
+    /**
+     * Migration to make new "Exclude Customergroup" option visible in advanced menu without reinstallation
+     *
+     * @param string $modus
+     */
+    public function up($modus)
+    {
+        $sql = <<<'SQL'
+SET @pluginID = (SELECT DISTINCT `id` FROM `s_core_plugins` WHERE `name`='AdvancedMenu' AND `active`=1);
```
PaddyS, line 36: Please check if the plugin is **installed** ("installation_date IS NOT NULL").

Also, you still need to check if this query actually returned an ID, and if not, do not proceed with the migration.
Distinct is also not necessary. :-)

_sql/migrations/1453-add-customer-group-key-exemption-option.php
```diff
@@ -0,0 +1,76 @@
...
+        $sql = <<<'SQL'
+SET @formID = (SELECT DISTINCT `id` FROM `s_core_config_forms` WHERE `plugin_id`=@pluginID);
+SQL;
+        $this->addSql($sql);
+
+        $sql = <<<'SQL'
+INSERT IGNORE INTO s_core_config_elements (form_id, name, value, label, description, type, required, position, scope, options) VALUES(
+@formID,
+'excludeCustomergroup',
+'i:0;',
```
PaddyS, line 49: Also, this should be a boolean value (b), not an integer (i). :-)

PR #1964, opened 2019-01-10
Swift mailer

hlohaus:
Replace zend mail with swift mailer.
Features:

Redirect Plugin for redirect all mails to a address
NULL-Transporter: Disable mails in the backend
Global BBC Address
Additionally config for mailer_from, reply_to and return_path
 mail() function is deprecated see swiftmailer/swiftmailer#866 (due to security reasons)
Sendmail is now the default transport
Sendmail is now a system requirement

### 6. Checklist

### DIFF
--- a/_sql/migrations/1606-swift-mailer.php
+++ b/_sql/migrations/1606-swift-mailer.php
@@ -1,0 +1,93 @@
+<?php
+
+class Migrations_Migration1606 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $sql = <<<'EOD'
+SET @formId = (SELECT id FROM `s_core_config_forms` WHERE name='Mail');
+EOD;
+        $this->addSql($sql);
+
+        $sql = <<<'EOD'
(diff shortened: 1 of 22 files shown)


hlohaus:
https://github.com/swiftmailer/swiftmailer/issues/866

shyim:
Hey @hlohaus,
unfortunately we cannot drop the current default option mail. Maybe we could use the 5.x version instead where mail is still available?

What do you think about that?

hlohaus:
Yes, maybe. But swift mailer has as reason for that. I added sendmail as system requirement. All linux servers use sendmail. mail() is only a wrapper for this.

hlohaus:
@shyim Okay, i set the version to 5 and make mail() as fallback for sendmail?

hlohaus:
Okay, i re-added Zend_Mail.

shyim:
Hey,
we don't want to switch to an already deprecated Mailer. If you want to create a Pull Request for Symfony Mailer, we would be very happy.

PR #1968, opened 2019-01-10
Properly check if variables are set and not NULL or undefined

DanielRuf:
Some variables are not correctly checked and in this case an undefined value is used in some code parts and specific cases. This could lead to errors and things which are not correctly working.

Properly check for `'undefined'` and `null`.

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/base/application/Shopware.window.Progress.js
+++ b/themes/Backend/ExtJs/backend/base/application/Shopware.window.Progress.js
@@ -679,12 +679,12 @@ Ext.define('Shopware.window.Progress', {
         var me = this, record;
 
         //no current task passed? Take the first task in the tasks array
-        if (current == undefined && tasks.length > 0) {
+        if (typeof current === 'undefined' && tasks.length > 0) {
             current = tasks.shift();
         }
 
         //contains the current task no more record/data? Then get the next from the array
-        if (current.data && current.data.length <= 0) {
+        if (typeof current !== 'undefined' && current.data && current.data.length <= 0) {
             current = tasks.shift();
         }
 
--- a/themes/Backend/ExtJs/backend/article/controller/detail.js
+++ b/themes/Backend/ExtJs/backend/article/controller/detail.js
@@ -403,7 +403,9 @@ Ext.define('Shopware.apps.Article.controller.Detail', {
             return;
         }
         var baseField = mainWindow.down('article-base-field-set');
-        baseField.numberField.validationRequestParam = article.getMainDetail().first().get('id');
+        if (article) {
+            baseField.numberField.validationRequestParam = article.getMainDetail().first().get('id');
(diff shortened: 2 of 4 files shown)


themes/Backend/ExtJs/backend/base/application/Shopware.window.Progress.js
```diff
@@ -679,12 +679,12 @@ Ext.define('Shopware.window.Progress', {
         var me = this, record;
 
         //no current task passed? Take the first task in the tasks array
-        if (current == undefined && tasks.length > 0) {
+        if (typeof current === 'undefined' && tasks.length > 0) {
```
mitelg, line 682: you could also use https://docs.sencha.com/extjs/4.1.1/#!/api/Ext-method-isDefined

DanielRuf: Which is another layer and adds a few more CPU cycles. Not sure if this is needed and if so we should use it in more places then ;-)

DanielRuf: And afaik we want to drop Ext JS in the (far) future so this makes it easier to refactor.

DanielRuf:
@shyim @larsbo why are you confused?

themes/Backend/ExtJs/backend/base/application/Shopware.window.Progress.js
mitelg: we will drop ExtJs, that is right. but none of this code will be re-used :wink: 
see: https://github.com/shopware/administration

DanielRuf: Sure, but still, why the wrapper when we can directly use the right and only correct code? I know that frameworks like Ext JS and also Foundation Sites (where I contribute too) have their own `typeof` methods but at this line we do not need another Ext JS helper method imo.

soebbing: I'm with @mitelg here, we should use those functions in more places.

These wrapper functions exist to make such checks more readable, consistent and maintainable. As you can see here, doing them yourself each time is error prone and makes the task at hand less obvious.

DanielRuf: > is error prone

For me it is easier to use the native `typeof` operator. Otherwise I would have to check what the wrapper does. And if so I would change other files and lines too where the helper method is not used. And it is actually not used in many places where still `typeof` is used and where we know what it exactly does.

I think it makes no real difference here so it is probably not necessary to use the method here ;-)

DanielRuf:

Done.

DanielRuf:
Oh man =(

PR #1969, opened 2019-01-10
Host Source Sans Pro font locally (data protection)

DanielRuf:
Add generic font name as fallback

To improve the data protection we should host the fonts locally and prevent issues when resources are blocked due to CSP and other solutions.

Host the Source Sans Pro files locally.

### 6. Checklist


bcremer:
Source Sans Pro is licensed under the [Open Font License](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL_web).

From PERMISSION & CONDITIONS (emphasizes by me):

> 2) Original or Modified Versions of the Font Software may be bundled,
> redistributed and/or sold with any software, **provided that each copy
> contains the above copyright notice** and this license. These can be
> included either as stand-alone text files, human-readable headers or
> in the appropriate machine-readable metadata fields within text or
> binary files as long as those fields can be easily viewed by the user.

This should be double checked before merge.

DanielRuf:
> This should be double checked before merge.

It is already contained in the woff files.
You are allowed to host them.
We do not ship the source files of the fonts. Same with the https://google-webfonts-helper.herokuapp.com/fonts/ tool which is recommended by Google.

https://github.com/google/fonts#self-host-fonts-available-from-google-fonts

https://theleagueof.github.io/licenses/ofl-faq.html

> 1.10 Does the full OFL license text always need to accompany the font?
The only situation in which an OFL font can be distributed without the text of the OFL (either in a separate file or in font metadata), is when a font is embedded in a document or bundled within a program. In the case of metadata included within a font, it is legally sufficient to include only a link to the text of the OFL on http://scripts.sil.org/OFL, but we strongly recommend against this. **Most modern font formats include metadata fields that will accept the full OFL text, and full inclusion increases the likelihood that users will understand and properly apply the license.**

Also see https://github.com/Coding-Coach/coding-coach/pull/85

bildschirmfoto 2019-01-11 um 08 29 44

bildschirmfoto 2019-01-11 um 08 31 57

DanielRuf:
> Most modern font formats include metadata fields that will accept the full OFL text, and full inclusion increases the likelihood that users will understand and properly apply the license.

DanielRuf:

Done.

DanielRuf:
Huh, what? Sorry but thus PR is related to GDPR. Seems it makes not much sense to contribute more here =( Even the Magento team takes the time even if there are many PRs.

DanielRuf:
Huh, what? Sorry but this PR is related to GDPR. Seems it makes not much sense to contribute more here =( Even the Magento team takes the time even if there are many PRs.

PR #1970, opened 2019-01-10
Use an RSA algorithm with a OAEP padding to prevent chosen-ciphertext…

DanielRuf:
… attacks

By default `openssl_public_encrypt` does not use an algorithm with OAEP padding and so the generated strings are vulnerable to oracle padding and chosen ciphertext attacks.

Set the padding parameter for `openssl_public_encrypt` to `OPENSSL_PKCS1_OAEP_PADDING` to improve the security.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/OpenSSLEncryption.php
+++ b/engine/Shopware/Components/OpenSSLEncryption.php
@@ -77,7 +77,7 @@ class OpenSSLEncryption
         $encryptedMessage = openssl_encrypt($data, $encryptionMethod, $key, false, $iv);
 
         $encryptedKey = '';
-        if (!true === openssl_public_encrypt($key, $encryptedKey, $publicKey)) {
+        if (!true === openssl_public_encrypt($key, $encryptedKey, $publicKey, OPENSSL_PKCS1_OAEP_PADDING)) {
             $errors = [];
             while ($errors[] = openssl_error_string());
             $errorString = implode("\n", $errors);


DanielRuf:
Looks like there is a flaky test https://travis-ci.org/shopware/shopware/jobs/478060071#L1437

bcremer:
More about this issue here: https://paragonie.com/blog/2016/12/everything-you-know-about-public-key-encryption-in-php-is-wrong#php-openssl-rsa-bad-default

DanielRuf:

Done.

teiling88:
Please change this hint from Php Inspection too: 

image

DanielRuf:
@teiling88 this is a line which is not affected by my change.
This is the PHP inspection and the issue is the interface and return value of the `getBytes` method of the `Random` class. Not sure why it was done with another class as wrapper which does `random_bytes($len)` in the method and not more.

teiling88:
> This is the PHP inspection and the issue is the interface and return value of the getByte
Yes but the function doesn't use openssl_random_pseudo_bytes(). So you please change this line too because it fits in the actual context.

DanielRuf:
`openssl_random_bytes` should not be used as with PHP 7.0+ you should use `random_bytes()` directly wich is a CSPRNG.

And http://php.net/manual/en/function.openssl-random-pseudo-bytes.php is a PRNG, it may not be CS

> It also indicates if a cryptographically strong algorithm was used to produce the pseudo-random bytes, and does this via the optional crypto_strong parameter. It's rare for this to be FALSE, but some systems may be broken or old.


You do not want to weaken the security by relying on openssl at all

https://stackoverflow.com/questions/38716613/generate-a-single-use-token-in-php-random-bytes-or-openssl-random-pseudo-bytes.

https://security.stackexchange.com/a/102082/6090

You will often see Scott as he is very active in the security industry.

> You should be aware of [PHP bug #70014](https://bugs.php.net/bug.php?id=70014) which was pretty recent and affects the reliability of `openssl_random_pseudo_bytes()`.

> I've been working on [paragonie/random_compat](https://github.com/paragonie/random_compat), which backports `random_bytes()` from PHP 7 into PHP 5. One of the fallbacks it supports is `openssl_random_pseudo_bytes()`, but if it can read directly from `/dev/urandom` it will prefer that instead.

DanielRuf:
> Yes but the function doesn't use openssl_random_pseudo_bytes(). So you please change this line too because it fits in the actual context.

There is nothing to be changed in the line that you show me as this is an internal PHP inspection rule which checks what instance the Random class returns.

DanielRuf:
I would say it is a false positive detection imo and it is not related to the padding algorithm.

teiling88:
Bug is reported - @kalessil will release a bugfix soon. Thanks @kalessil  :-)

kalessil:
You are welcome @teiling88 =)

DanielRuf:
Ehm what? Sorry but this is a special PR. More PRs are currently not planned.

Can we reopen this one?
Security is important.

PaddyS:
Hey @DanielRuf,

We investigated that issue and figured out, that there's no longer a need for the class as a whole.
Therefore, we deprecated the class and will remove it in a future release.

DanielRuf:
Hi @PaddyS,

thanks for the feedback and also thanks for letting me know.

Might be good to clarify what you mean with "own implementation" because you should never roll your own crypto and there are several good libraries from paragonie and others =)

https://github.com/shopware/shopware/commit/ba5ea87b4401528936d92d8172690850f4896fe0

PR #1972, opened 2019-01-13
Symfony Session

hlohaus:
Replaces zend session with symfony session.

### 6. Checklist


engine/Shopware/Components/DependencyInjection/Bridge/Session.php
```diff
@@ -40,74 +46,207 @@ class Session
...
      */
-    public function createSession(Container $container, \SessionHandlerInterface $saveHandler = null)
+    public function factoryStorage(Container $container)
     {
-        $sessionOptions = $container->getParameter('shopware.session');
+        $sessionOptions = $this->getOptions($container);
 
         if (!empty($sessionOptions['unitTestEnabled'])) {
-            \Enlight_Components_Session::$_unitTestEnabled = true;
+            $storage = new PhpBridgeSessionStorage();
```
shyim, line 101: Maybe change to NullSessionHandler. The use case seems for unit testing https://github.com/symfony/symfony/blob/4.2/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/NullSessionHandler.php#L15 :)

engine/Shopware/Components/DependencyInjection/Bridge/Session.php
```diff
@@ -40,74 +46,207 @@ class Session
...
+    }
+
+    /**
+     * @param Container $container
+     *
+     * @return string
+     */
+    private function getSessionName($container)
+    {
+        if ($container->has('shop')) {
```
shyim, line 190: The has method shows is the service already there, when not it tries to load it. 
`shop`  is a synthetic service, so it will crash here :) It should be fixed by using initialized instead

engine/Shopware/Components/DependencyInjection/Bridge/Session.php
```diff
@@ -40,74 +46,207 @@ class Session
...
+        }
+    }
 
-        $sessionOptions['name'] = 'session-' . $shop->getId();
+    /**
+     * @param $container
+     * @param NativeSessionStorage $storage
+     * @param $sessionOptions
+     */
+    private function setSessionStorageOptions($container, NativeSessionStorage $storage, $sessionOptions)
```
shyim, line 139: Typehints are missing

engine/Shopware/Components/DependencyInjection/Bridge/Session.php
```diff
@@ -40,74 +46,207 @@ class Session
...
+    {
+        $config = [
+            'db_table' => 's_core_sessions',
+            'db_id_col' => 'id',
+            'db_data_col' => 'data',
+            'db_lifetime_col' => 'expiry',
+            'db_time_col' => 'modified',
+            'lock_mode' => $sessionOptions['locking'] ? PdoSessionHandler::LOCK_TRANSACTIONAL : PdoSessionHandler::LOCK_NONE,
+        ];
+        if (!$container->has('shop')) {
```
shyim, line 243: initialized :)

shyim:
You could also add to all new classes typehints + return types, because 5.6 branch min PHP version is 7.1 :)

engine/Library/Enlight/Components/Test/Controller/TestCase.php
```diff
@@ -153,6 +153,8 @@ public function reset()
     {
         $app = Shopware();
 
+        $container = $app->Container();
```
shopwareBot, line 156: Why would you do that?

hlohaus:
I closed it, because the tests a really broken.

PR #1973, opened 2019-01-14
replace CountryId by Position

tinect:
There seems to be a confusing sorting in backend for countries. This PR adds Column "Position" and tries to the list as they will result in frontend. 

Please see also: https://github.com/shopware/shopware/commit/1efcb59a93370277e61836a8807ed5df744f9782#r31928592

- add position-column

image

### 6. Checklist


soebbing:
HI @tinect, nice idea! But we tried to have the most used country in front and still maintain some general sensible order.
How about removing the `id` column, adding `position` instead and keeping the current default sorting?

tinect:
@soebbing Sounds good :-) PR changed.

shyim:
Hey @tinect,

thanks for the Contribution! :blue_heart: . Your change has been merged.

PR #1975, opened 2019-01-15
Standardized shopId and added more description info

Isengo1989:
Commands and the options/arguments should be standardized. Especially at the beginning this can be annoying and confusing. Sometimes it is an option, then an argument, then it takes multiple IDs then only single and then it is called shop instead of shopId.

Changed 2 arguments to options / added more info how to use multiple IDs / changed optionname from **shop** to **shopId**

**EDIT**: Added 4 deprecation warnings + TODOs for removal in the next major version

-

-

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Commands/WarmUpHttpCacheCommand.php
+++ b/engine/Shopware/Commands/WarmUpHttpCacheCommand.php
@@ -54,7 +54,9 @@ class WarmUpHttpCacheCommand extends ShopwareCommand
         $this
             ->setName('sw:warm:http:cache')
             ->setDescription('Warm up http cache (everything by default)')
-            ->addArgument('shopId', InputArgument::OPTIONAL, 'The Id of the shop')
+            /* @deprecated since 5.6, to be removed in 6.0 */
+            ->addArgument('shopId', InputArgument::OPTIONAL | InputArgument::IS_ARRAY, 'The Id of the shop (deprecated)')
+            ->addOption('shopId', null, InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY, 'The Id of the shop (mulitple IDs -> shopId={1,2})')
             ->addOption('clear-cache', 'c', InputOption::VALUE_NONE, 'Clear complete httpcache before warmup')
             ->addOption('concurrent-requests', 'b', InputOption::VALUE_OPTIONAL, 'Integer representing the maximum number of requests that are allowed to be sent concurrently. To many URLs at a time may cause script timeouts, memory issues or block your HTTP server', 1)
             ->addOption('category', 'k', InputOption::VALUE_NONE, 'Warm up categories')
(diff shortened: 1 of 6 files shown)


mitelg:
hey @Isengo1989 
good idea, but that would break existing setups, e.g. for deployments. Is there a way to do it without BC? Either way this should be done on the 5.6 branch

Isengo1989:
> hey @Isengo1989
> good idea, but that would break existing setups, e.g. for deployments. Is there a way to do it without BC? Either way this should be done on the 5.6 branch

That's true. How about adding the option in **RebuildSeoIndexCommand.php** and **WarmUpHttpCacheCommand.php** (leave the argument) - and check if either one of them is set. If it is the argument it will output

`$output->writeln('Deprecated argument - will be removed in Shopware 6.0...blabla');`

It should be mentioned in the Updateguide for 5.6 as a deprecated command parameter and be removed in 5.x or 6.0.

Could be merged with theses changes already with a //TODO in the two commands. The other 3 changes do not affect a deployment (shopId is the same in ThemeCacheGenerateCommand.php and PluginConfigListCommand.php/PluginConfigSetCommand.php are most likely not used in deployment)

Isengo1989:
@mitelg btw - why did travis fail? Can you give me a hint?

engine/Shopware/Commands/WarmUpHttpCacheCommand.php
```diff
@@ -82,18 +81,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
         /** @var UrlProviderFactoryInterface $urlProviderFactory */
         $urlProviderFactory = $this->container->get('shopware_cache_warmer.url_provider_factory');
 
-        // Get every shop to warm
-        $shopId = $input->getArgument('shopId');
+        $inputShops = $input->getOption('shopId');
+
+        /** @var \Shopware\Models\Shop\Repository $shopRepository */
         $shopRepository = $this->container->get('models')->getRepository(Shop::class);
```
shopwareBot, line 100: Hey @Isengo1989,

just add `$shops = null` here to fix the broken travis build.

mitelg:
@Isengo1989 

> Line   Commands/WarmUpHttpCacheCommand.php    
> ------ --------------------------------------- 
> 138    Variable $shops might not be defined.

Isengo1989:
@mitelg added deprecation warnings to the arguments/options + TODOs for removal in the next major version. This way full BC is given.

Isengo1989:
How come that this https://github.com/shopware/shopware/blob/ac0b66e4c0b4724e9e5ed30f75f9914114671221/engine/Shopware/Controllers/Frontend/Account.php#L25 throws a travis error in my PR when other PRs pass the CI just fine? @shyim

mitelg:
@Isengo1989 this is an unused `use` statement. not related to your PR :wink:

Isengo1989:
@mitelg I know. I was just wondering how a PR with the same base could pass all tests (https://github.com/shopware/shopware/pull/1979) - but I just saw that this file was fixed in the PR ^^. So I guess I will wait till the change is merged in the 5.6 branch.

Is the rest okay?

Isengo1989:
@shopwareBot done - removed TODOs and added "@deprecated" annotation above the arguments/options

PR #1976, opened 2019-01-16
leftJoining article attributes instead of inner

EtienneBruines:
The current INNER JOIN excludes results that don't have an entry within `s_articles_attributes`. That is unnecessary -- many details/variants won't have an entry. Neither will articles by default. 

Replaces the INNER JOIN by a LEFT JOIN. 

```php
$context = Shopware()->Container()->get('shopware_storefront.context_service')
    ->createProductContext($shopId);

$criteria = new Criteria();
$criteria->addBaseCondition(
    $this->getCondition('SW10123.3')
);
$results = Shopware()->Container()->get('shopware_search.product_number_search')
    ->search($criteria, $context);
```

Now look at the `$results` - if the detail did not have an `Attribute`, it will not have found it. 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/QueryBuilderFactory.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/QueryBuilderFactory.php
@@ -173,7 +173,7 @@ class QueryBuilderFactory implements QueryBuilderFactoryInterface
                  AND product.active = 1'
             );
         }
-        $query->innerJoin(
+        $query->leftJoin(
             'variant',
             's_articles_attributes',
             'productAttribute',


mitelg:
@EtienneBruines afaik, all created `article` entities should have an empty attribute row per default. If you create a product via the backend or API. so a product without an attribute seems invalid

EtienneBruines:
@mitelg Sounds reasonable. It's up to you (Shopware) to whether:
1. Close this PR, as it's not required. 
2. Merge this PR, as a "first step" into making the attribute-row optional. For lots of entities, this seems optional already. The `Attribute` object is null when instantiating the object anyway.

PR #1978, opened 2019-01-16
Removes impossible to reach or redundant condition branches

jkrzefski:
It is not. But the redundant condition isn't either.

Removes a redundant condition and an else-branch that could never be reached.

1. Look at code
2. Be confused
3. Done

None

None

### 6. Checklist


shyim:
Hi @jkrzefski ,

PR #1983, opened 2019-01-18
Use RouterInterface / Remove clone router

hlohaus:

Use RouterInterface in Bundles for better decorating.
Remove clone router because less head.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/HrefLangService.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/HrefLangService.php
@@ -31,7 +31,7 @@ use Shopware\Bundle\StoreFrontBundle\Struct\Shop;
 use Shopware\Bundle\StoreFrontBundle\Struct\ShopContextInterface;
 use Shopware\Components\Model\ModelManager;
 use Shopware\Components\Routing\Context;
-use Shopware\Components\Routing\Router;
+use Shopware\Components\Routing\RouterInterface;
 use Shopware\Models\Shop\Shop as ShopModel;
 use Shopware_Components_Config as Config;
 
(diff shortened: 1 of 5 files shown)


engine/Shopware/Bundle/StoreFrontBundle/Service/Core/HrefLangService.php
```diff
@@ -158,20 +158,18 @@ private function getLanguageShops(Shop $shop)
     /**
      * @param int $shopId
      *
-     * @return Router
+     * @return Context
+     *                 e
```
jkrzefski:
```suggestion
```

I think this line should not exist.

engine/Shopware/Bundle/StoreFrontBundle/Service/Core/HrefLangService.php
```diff
@@ -158,20 +158,17 @@ private function getLanguageShops(Shop $shop)
     /**
      * @param int $shopId
      *
-     * @return Router
+     * @return Context
      */
-    private function getRouter($shopId)
+    private function getContext($shopId)
```
mitelg, line 163: why is this change necessary? 
imho it is not related to the goal of the PR to exchange the tyehints and should come in another PR

hlohaus: technically it does the same thing.
I just did not think it was so nice to clone the router.

hlohaus: In addition, it fixes that the config does not fit

mitelg: then please put that in a separate commit and change the description of your PR. This is nowhere mentioned

hlohaus:
Okay.  There are now 2 commits and i updated the description

PR #1985, opened 2019-01-18
Fixed CSRF-token verification loophole

oktupol:

Currently, the CSRF-token validation for POST-Requests can be bypassed entirely by providing the header `X_REQUESTED_WITH: XMLHttpRequest`. This also means that there is no CSRF-token validation for any Ajax-Requests

With this change, this is no longer possible

Send a POST-request (e.g. changing an address / confirming an order) using curl on the command line. Provide the header stated above and no CSRF-Token.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/CSRFTokenValidator.php
+++ b/engine/Shopware/Components/CSRFTokenValidator.php
@@ -136,10 +136,6 @@ class CSRFTokenValidator implements SubscriberInterface
             return;
         }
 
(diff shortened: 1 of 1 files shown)


soebbing:
Hi @oktupol, after we've discussed this changed internally (and with you, offline) we're closing this PR as it doesn't actually fix a security issue.

If anybody stumbles across an issue like this, please create an issue on https://issues.shopware.com/createissue so we can confirm and, if necessary, fix it before making it public.

PR #1986, opened 2019-01-21
Add "laststock" to "Apply Standard Data"

aherking:
This allows the user to apply the laststock flag from the preselection variant with the "Apply Standard Data" function in backend.

Adds the "lastStock" information to the getMappingData function

Actually the last stock flag is not applied to the other variants if using "Apply Standard Data"

https://issues.shopware.com/issues/SW-23131

### 6. Checklist


shyim:

PR #1989, opened 2019-01-22
PHP 7.2+ - Doctrine to 2.6.3

EtienneBruines:
Let's just see if the tests run on the CI server and if any doctrine-related issues occur. 

The current version of doctrine is not compatible with PHP 7.3, due to 
```
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in ./lib/Doctrine/ORM/UnitOfWork.php on line 2718
```
See also https://github.com/doctrine/orm/pull/7325

The same goes for `zendframework/zend-code`. The latest version of `zend-code` is supported by `ocramius/proxy-manager` at 2.2.0, but that one requires PHP 7.2. So we require PHP 7.2 or above. :tada: 

- `php` to `7.2`
- `doctrine/orm` to `2.6.3`
- `ocramius/proxy-manager` to `2.2.2`

### 6. Checklist


bcremer:
I noticed you deleted `/engine/Library/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php`. I am pretty sure this file is required  for Shopware to work. The file overwrites the one in the vendor directory and contains a little hack, see: https://github.com/shopware/shopware/blob/c22a29376ff5a713d70afc310c371e1faa21d77c/engine/Library/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php#L671

EtienneBruines:
@bcremer Agreed. Back then, you apparently already put in a lot of effort; and with the new version, there's still no better solution. 

I'll be updating the files, though. Coming right up.

EtienneBruines:
For future reference:

https://gitter.im/shopware/shopware?at=5c4833abc45b986d11a6e9a6
> @soebbing:
> Policy of Shopware is to support the oldest current actively supported PHP version - that would be PHP 7.1

I will be removing any PHP 7.2 requirement.

soebbing:
Just here to point out: You've sparked a new internal discussion, @EtienneBruines,  if it is really worth a hack to support a PHP version that will only be around for about half a year before it is deprecated. 

So the last word on this hasn't been spoken, I'll get back to you soon.

bcremer:
PHP 7.1 is already in Security Only mode so it would make sense to support PHP 7.3. 

image

mitelg:
my vote for minimum PHP 7.2 :+1:

soebbing:
Our prayers have been heard! Shopware 5.6 will have PHP 7.2 as a minimum requirement, @EtienneBruines!
So please please squash your commits, have a look at the CI-errors and this PR is good to go! 👍 💙

EtienneBruines:
I am unfortunately unable to locate the issue. 

```
1) Shopware\Tests\Functional\Plugins\Core\MarketingAggregate\Components\AlsoBoughtTest::testInitAlsoBought
     [exec] Failed asserting that actual size 27 matches expected size 30.
```

#help-needed

EtienneBruines:
It should be ready now :smile:

PR #1991, opened 2019-01-24
autoselect variant in sArticle, based on setting in backend

pascalheidmann-bedarf:
If you change your configurator to not standard (e.g. image or selection) the default variant will not be auto selected (if no number parameter is passed)

Allow any configurator be valid. Also remove not initialized variable $configuration. This is behind a settings toggle so you can still have the old behaviour.

Without PR create a variant article and check autoselection in frontend. Then change the variant article to image and load the article in frontend again without any number parameter. This won't select the default selection.

- https://github.com/shopware/shopware/pull/1933 (this 

According to https://github.com/shopware/shopware/pull/1933 this PR will be BC breaking if it is enabled. As this is behind a toggle this shouldn't be a problem.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1606-make-configurators-autoselect.php
+++ b/_sql/migrations/1606-make-configurators-autoselect.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+use Shopware\Components\Migrations\AbstractMigration;
+
+class Migrations_Migration1606 extends AbstractMigration
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function up($modus)
+    {
+        $sql = <<<SQL
+        SET @form_id = (SELECT form.id FROM s_core_config_forms form WHERE form.`name` = "Frontend79" LIMIT 1);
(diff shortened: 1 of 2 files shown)


mnaczenski:
Hi @pascalheidman-bedarf 

thank you for your PR. 
Our quality assurance found some issues with the user-experience regarding this PR. We can't currently think of a possible solution that works in "every" case. Perhaps you can suggest a better solution?

The main reason the "selection" and "picture" configurator type won't preselect the main variant is, that you might create "dead ends" for the users, since both configurators are able to disable the selection of unavailable variants. I tried to record a GIF that should clearify this issue.

Peek 2019-03-25 06-58

If the pre-selected variant would be one where the counterpart is disabled, the user won't be able to change the variant (in my example from "weiß" to "blau"). One might say that the button "Reset Selection" is displayed in this case, but since the user did not select anything, it is not clear for him, what this button does. The standard configurator is able to handle this issue, because it displays every variant, even if it is disabled or deleted. 

@shyim suggested that you create a PR that will add extensibility, so you can create a simple plugin for it, if we don't find a solution that would work in every case (since we are a standared software product).


Moritz

This PR had the same issue: https://github.com/shopware/shopware/pull/1828

pascalheidmann-bedarf:
@mnaczenski I haven't dived in the code yet that handles both configurator templates but my suggestion would be to ignore disabled products and allow all selection at that point - basically as it would be without a selection. That way the _reset_ button isn't needed and we still have to improved usability of a pre selection.
Is this a reasonable solution for shopware?

EveryDayISeeMyDream:
Hey @pascalheidman-bedarf 
your proposed solution is unfortunately somewhat vaguely formulated. How exactly do you imagine the solution to be?
My wish would be that the preselection is loaded and the "Reset selection" button disappears completely. At this point, the selection fields for the individual options must then behave in such a way that they are grayed out and indicate that this article is not available in this variant. If the option can be made clickable, the incompatible option would have to be reset so that the customer can see exactly that something has changed. You can see this with the example here: https://www.amazon.de/dp/B07GLL9X23/ > If you select size "39.5" here, almost all colors will be grayed out, if you then click on one of these grayed out options, the size option will reset so that you know that you have to make a different selection of size here.
This should happen with the Image Configurator and the Selection Configurator.

Greetings from the QA :)

pascalheidmann-bedarf:
@EveryDayISeeMyDream The image selector on amazon was kind of my inspiration and should be doable :)

For the selection configurator I'm not that sure as right now it relies on the `disabled` prop of the `<select>` dropdown. AFAIK those can't be styled via css.
This means that it would have to go through a 3rd party lib like https://developer.snapappointments.com/bootstrap-select/examples/

EveryDayISeeMyDream:
@shyim Could you give an advice here how the select-configurator could style entries as disabled (and show a notice) but make them still clickable to match the desired solution?

Generally I'm not a friend of using different libraries for such small things, this would blow up our package size and we had to take care about updating another library in case of security issues.

PR #1992, opened 2019-01-24
Update rewrites and cms links

Isengo1989:
There are still plenty of shops online with ticket pages indexed - since there was no cleanup of the rewrite table and cms pages (where many shops links forms).

auswahl_103

It updates the rewrite and cms table and replaces **ticket** by **forms** so even without sw:rebuild:seo:index they will be redirected correctly and linked forms on cms pages work.

Update from Shopware < 5.5.0 with forms links in cms pages

-

-

### 6. Checklist

### DIFF
--- a/_sql/migrations/1453-update-cms-links.php
+++ b/_sql/migrations/1453-update-cms-links.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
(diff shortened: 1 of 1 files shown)


shyim:
Your migration breaks SwagTicketSystem, cause it uses the same controller name

Isengo1989:
> Your migration breaks SwagTicketSystem, cause it uses the same controller name

@shyim Never installed it - can you give me an example of the pattern? Maybe I can write a different LIKE or just use the cms fix - I guess this is the most urgent fix - since **sw:rebuild:seo:index**  will not fix it.

mnaczenski:
Your screenshot shows a list of urls with a debug configuration. Without this configuration those liks will throw a normal 404 as far as I see.

shyim:
They use the excact same controller (ticket) and same paramter sFid. So there is no way in the rewrite table to find the right one 🙈.

Maybe the better way would be, that the rebuild seo index removes those entries.

Isengo1989:
> Your screenshot shows a list of urls with a debug configuration. Without this configuration those liks will throw a normal 404 as far as I see.

True that - but this should not happen anyway - if it gets catched fine. But some users will just not find this issue by themselfes - so lets fix it :) 

**Suggestions**
1. Remove old entries in seo index command
2. Clean cms table with the first update statement (I doubt that anyone will link it in the s_cms_static/CMS page context)

soebbing:
Hi @Isengo1989!

We've discussed your change internally and have to decline it.

There are multiple reasons for it: First, as already has been pointed out, this is only an issue on systems that have the debug-config enabled. Second: We really don't want to mess with the urls-table of our customers, there be dragons. The controller name might be reused or something.

Third, and most importantly: The plugin for the ticket system is still being sold and used by customers (https://store.shopware.com/swagticketsystem/ticketsystem.html). Changing those URLs would break all those installations!

Maybe a more appropriate solution would be something to remind shop owners that they are running their shop in a debug-configuration and tell them what implications this has? Something like a notification in the backend?

Isengo1989:
@soebbing ok, I understand. For the future it would be nice if there was an info in the Upgrade Guide (i didn't find any) when a controller/action is changed in the Rewrite table. I don't think backend notify is necessary.

soebbing:
@Isengo1989 I think this change was made a long time ago (2014-2015), that might be a reason (though not an excuse).

But I take your comment as a strong reminder for us to document our changes well. I know we still have a lot to do in that regard.

PR #1993, opened 2019-01-25
Update email regex

pontoffeltier:
The old regex allowed for invalid email-addresses to be entered

The new regex validates emails like the `<input type="email">` does. The regex is actually proposed by the w3c as an implementation for this (https://www.w3.org/TR/html5/sec-forms.html#valid-e-mail-address)

Enter "test@-.com" as your e-mail-address in the register form. The address will be accepted by shopware even though it's not a valid address. Use a text-input, not type="email" for this or the browser throws an error because the input actually validates this correctly (which does not mean, shopware should simply accept it)

-

-

### 6. Checklist

- [-] I have written tests and verified that they fail without my change


aragon999:
Since this is a much more complicated RegEx I would suggest to not only add this one additional Test Case, but a more comprehensive list, see for example the Chromium tests: 
https://cs.chromium.org/chromium/src/third_party/blink/web_tests/fast/forms/resources/ValidityState-typeMismatch-email.js

PR #1994, opened 2019-01-25
SW-22603 allow category filter in addition to sCategory base condition

jnugh:
This allows the user to filter for categories in product streams which uses categoryFilter to for filtering and sCategory for product stream selection

This fixes the issue described in https://issues.shopware.com/issues/SW-22603 where category filters are not applied in product streams.

Criteria can now contain a base condition with a category and additional category conditions. 
A product stream request with category filters contains `sCategory` to identify the stream to be used and `categoryFilter` to do the actual filtering.
The category in `sCategory` is later replaced be the shops root category id and can therefor not be used to narrow the filter.

Create a stream, enable filters and render the stream with the cf parameter like this: http://shopware.ddev.local/cat/index/sCategory/76?p=1&o=1&n=12&cf=38

https://issues.shopware.com/issues/SW-22603

- 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
+++ b/engine/Shopware/Bundle/SearchBundle/CriteriaRequestHandler/CoreCriteriaRequestHandler.php
@@ -109,7 +109,8 @@ class CoreCriteriaRequestHandler implements CriteriaRequestHandlerInterface
             $criteria->addBaseCondition(
                 new CategoryCondition($ids)
             );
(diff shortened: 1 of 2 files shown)


jnugh:
Does that mean https://issues.shopware.com/issues/SW-22603 cannot be fixed?
The issue is still marked open and is marked as high priority. Are there different solutions you would accept here? Maybe a second code path so that filtering is technically not possible on standard listing pages without product streams?

Actually the product stream has no selected category, sCategory is the category the stream is embedded into and will be replaced by the shops root category before the actual query is being executed.

PR #1995, opened 2019-01-25
surround stylesheet and javascript with blocks

tinect:
There should be a shorter way to change the content of the iteration

surround css and javascript-adding with blocks

### 6. Checklist


shyim:
Thanks for the Contribution. :blue_heart:

PR #1996, opened 2019-01-25
Backend: Allow opening customers from orders (again?)

pascalheidmann-bedarf:
Often you have an order open but you need to see more things about that customer like other orders or comments etc. To open that specific customer you have to search for them or if you are lucky you still have searched for the order in the order list and have it still open as there is a customer button in action column.
It seems like this functionality has been there some time or at least was planned but not executed as the event listener is already inside the order detail controller.

Add a customer button to open a new ExtJS window with the customer to order detail overview panel. 

### 6. Checklist


themes/Backend/ExtJs/backend/order/view/detail/overview.js
```diff
@@ -639,6 +648,27 @@ Ext.define('Shopware.apps.Order.view.detail.Overview', {
         ];
     },
 
+    /**
+     * Adds a link to the customer that placed the order
+     *
+     * @returns Ext.Button|null
+     */
+    createCustomerLink: function() {
```
shopwareBot: Maybe rename to "createCustomerButton"?

Also, to keep the interface consistent, just disable the button if the customer doesn't exist.

pascalheidmann-bedarf: Good points.
I changed it to be a "tool" in `Ext.form.Panel` so it is now placed at the right side and not pushed in front of the title as every other solution I found to solve that problem wouldn't work...

PR #1998, opened 2019-01-28
Also consider links when collecting media which should not be removed

aragon999:
When adding for example a PDF via the media manager in a shop page, not an image is created but a link to that file. 
When now running the media garbage collector, this file gets removed even though it is used in a shop page.

Also consider links such that the media garbage collector does not remove these media.

See **1**

\-

\-

### 6. Checklist


aragon999:
Done :-)

aragon999:
Done :-)

PR #1999, opened 2019-01-29
Always fallback to DefaultFrom in Enlight_Components_Mail 

pascalheidmann-bedarf:
If you send some mail with `Enlight_Components_Mail`, for example when changing the order state and sending out an update to your customer and you don't provide an sender, setting the `From` header to an empty string won't set the fallback default mail address.

If the sender email address provided to `$mail->setFrom()` is an empty string fall back to `$address = null` to allow `Zend_Mail` to use `setFromToDefaultFrom()`.

1.  Go to an order
2. Change order state to cancelled
3. Send state change mail to customer
4. Customer gets mail but without a sender as it is an empty string (AFAIK there is no input for sender mail)
5. Now use this PR and go to step 1 and retry. Now shop's default mail address will be used.

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Components/Mail.php
+++ b/engine/Library/Enlight/Components/Mail.php
@@ -205,6 +205,11 @@ class Enlight_Components_Mail extends Zend_Mail
             throw new \RuntimeException('Potential code injection in From header');
         }
 
+        // explicit nulling so fallback to parent::setFromToDefaultFrom() works
+        if (empty($email)) {
+            $email = null;
+        }
+
         $this->_fromName = $name;
 
         return parent::setFrom($email, $name);


pascalheidmann-bedarf:
Hey @shopwareBot what exactly do you mean with this? Changing the setting in `Shopware_Controllers_Backend_Order::sendMailAction()`?

If yes I strongly disagree as this is a problem that can happen at a lot of places just by not correctly checking and you should never send a mail without any sender address - so enforcing to at least set the default address might not the finest but still the best way to make this safe.

bneumann97:
Hello @pascalheidman-bedarf,

I wasn't able to reproduce this error on any system. I changed the state of an order to cancelled and sent the email, and there was always a sender.

pascalheidmann-bedarf:
@shopwareBot while this was a case that problem happened to us this should be fixed anyway... Will take a look at it soon anyway

PR #2000, opened 2019-01-30
Added Shopware_Command_Before_Run

pixelwerk:

As of today there exists a "Shopware_Command_After_Run" event but no ""Shopware_Command_Before_Run" event. As a consequence there is no way to hook into commands before they are run. This is a problem if for instance you need to obtain access to the input or output objects passed to the console command.

Adds a new event "Shopware_Command_Before_Run" which is emitted just before a console command is run

At present, there is no event emitted before a console command is run.

-

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Console/Application.php
+++ b/engine/Shopware/Components/Console/Application.php
@@ -119,10 +119,20 @@ class Application extends BaseApplication
             $command->setContainer($this->kernel->getContainer());
         }
 
-        $exitCode = parent::doRunCommand($command, $input, $output);
-
         /** @var \Enlight_Event_EventManager $eventManager */
         $eventManager = $this->kernel->getContainer()->get('events');
+
+        $event = $eventManager->notifyUntil('Shopware_Command_Before_Run', [
+            'command' => $command,
+            'input' => $input,
+            'output' => $output,
+        ]);
+
+        if ($event) {
+            return (int) $event->getReturn();
+        }
+
+        $exitCode = parent::doRunCommand($command, $input, $output);
 
         $eventManager->notify('Shopware_Command_After_Run', [
            'exitCode' => $exitCode,


mitelg:
congrats to PR #2000 :smile: :+1: :blue_heart:

aragon999:
Maybe it would be good to have a mechanism to stop the execution of the command, in the before event, so maybe a `notifyUntil`-event?

shyim:
Why should a plugin have the power to stop a cli command :thinking:

aragon999:
One use case I just thought of, when I wrote the original post was that when some API update for the products and media is running, that you might not want to run `sw:thumbnail:cleanup` for example (when you run this via "console command crontab").

aragon999:
But I think the question should be thought of different, "Why should a plugin not be allowed to stop a console command from executing?".
I mean there are some downsides, i.e. when you have many plugins installed and don't know what exactly each plugin is doing, then you might not know why a command is not properly executing, but I guess that plugin has a reason to stop the execution at that point.

engine/Shopware/Components/Console/Application.php
```diff
@@ -119,11 +119,17 @@ protected function doRunCommand(Command $command, InputInterface $input, OutputI
             $command->setContainer($this->kernel->getContainer());
         }
 
-        $exitCode = parent::doRunCommand($command, $input, $output);
-
         /** @var \Enlight_Event_EventManager $eventManager */
         $eventManager = $this->kernel->getContainer()->get('events');
 
+        $eventManager->notify('Shopware_Command_Before_Run', [
```
shopwareBot: Can you please change this event to notifyUntil?

shyim:
Can you please squash also the commits. Thanks :blue_heart:

engine/Shopware/Components/Console/Application.php
twtgrowth: I updated the pull request

shyim:
Thanks for the Contribution :blue_heart:.

PR #2002, opened 2019-01-31
Use correct autoloader and config file when shopware is a composer dependency

JoshuaBehrens:
When you write a plugin test in a composer-project environment like it is written in the [dev docs](https://developers.shopware.com/developers-guide/plugin-testing/) you can refer to the correct testing bootstrap file but it does not load the correct configuration file.

If the special case of a composer project environment is detected it uses a different `autoload.php` and `config.php` file.

1. Have a plugin in a composer project in `custom/project/FooBar`
2. Bootstrap tests
    ```php
    <?php
    // resides in custom/project/FooBar/Tests/Functional/Bootstrap.php
    ini_set('error_reporting', E_ALL);
    ini_set('display_errors', '1');
    ini_set('display_startup_errors', '1');

    require __DIR__ . '/../../../../../tests/Functional/bootstrap.php';
    ```
3. phpunit plugin folder
4. Cannot test as a wrong `autoload.php` is loaded
    ```
    Warning: require(/var/www/html/foobar/vendor/shopware/shopware/tests/Functional/autoload.php): failed to open stream: No such file or directory in /var/www/html/foobar/vendor/shopware/shopware/tests/Functional/bootstrap.php on line 30
    Fatal error: require(): Failed opening required '/var/www/html/foobar/vendor/shopware/shopware/tests/Functional/autoload.php' (include_path='/var/www/html/foobar/engine/Library:/var/www/html/foobar/vendor/shopware/shopware/engine/Library:.:/usr/share/php') in /var/www/html/foobar/vendor/shopware/shopware/tests/Functional/bootstrap.php on line 30
    ```
5. Amend bootstrap require path
6. phpunit plugin folder
7. Cannot test as a wrong `config.php` is loaded
    ```
    Fatal error:  Uncaught RuntimeException: Could not connect to database. Message from SQL Server: SQLSTATE[HY000] [2002] No such file or directory in /var/www/html/foobar/vendor/shopware/shopware/engine/Shopware/Components/DependencyInjection/Bridge/Db.php:78
    Stack trace:
    #0 /var/www/html/dynfoobaratics/vendor/shopware/shopware/engine/Shopware/Kernel.php(300): Shopware\Components\DependencyInjection\Bridge\Db::createPDO(Array)
    #1 /var/www/html/foobar/vendor/shopware/shopware/tests/Functional/bootstrap.php(41): Shopware\Kernel->boot()
    #2 /var/www/html/foobar/vendor/shopware/shopware/tests/Functional/bootstrap.php(67): TestKernel::start()
    #3 /var/www/html/foobar/custom/project/FooBar/Tests/Functional/Bootstrap.php(7): require('/var/www/html/f...')
    #4 /var/www/html/foobar/vendor/phpunit/phpunit/src/Util/Fileloader.php(64): include_once('/var/www/html/f...')
    #5 /var/www/html/foobar/vendor/phpunit/phpunit/src/Util/Fileloader.php(48): PHPUnit\Util\Fileloader::load('/var/www/html/f...')
    #6 /var/www/htm in /var/www/html/foobar/vendor/shopware/shopware/engine/Shopware/Components/DependencyInjection/Bridge/Db.php on line 78
    ```
8. Play some tetris on arcade machine to regain saneness

The dev docs should just have some paths to :clipboard: :ramen: from, depending on their test bootstrap location.

### 6. Checklist

> :ballot_box_with_check: I have written tests and verified that they fail without my change 
> :grin: it is THE test

### DIFF
--- a/tests/Functional/bootstrap.php
+++ b/tests/Functional/bootstrap.php
@@ -22,7 +22,27 @@
  * our trademarks remain entirely with us.
  */
 
-require __DIR__ . '/../../autoload.php';
+$searchDirectory = __DIR__;
+
+while (true) {
+    $newSearchDirectory = realpath($searchDirectory . DIRECTORY_SEPARATOR . '..');
+
+    if ($searchDirectory === false || strlen($searchDirectory) < 3 || $newSearchDirectory === $searchDirectory) {
+        throw new RuntimeException('No autoloader found');
+    }
+
+    $searchDirectory = $newSearchDirectory;
+
+    if (file_exists($autoloadFile = implode(DIRECTORY_SEPARATOR, [$searchDirectory, 'app', 'autoload.php']))) {
+        require $autoloadFile;
+        break;
+    }
+
+    if (file_exists($autoloadFile = implode(DIRECTORY_SEPARATOR, [$searchDirectory, 'vendor', 'autoload.php']))) {
+        require $autoloadFile;
+        break;
+    }
+}
 
 class TestKernel extends \Shopware\Kernel
 {
@@ -48,6 +68,12 @@ class TestKernel extends \Shopware\Kernel
 
     protected function getConfigPath()
     {
+        if (file_exists(__DIR__ . '/../../../../../app/config/config.php')) {
+            // shopware as composer dependency
+            return __DIR__ . '/../../../../../app/config/config.php';
+        }
+
+        // shopware standalone
         return __DIR__ . '/config.php';
     }
 }


JoshuaBehrens:
Will there be a unified way to normalize a path to the tests bootstrap/kernel? I can't have plugins installed in a shopware standalone setup refer to a different bootstrap file than it would refer to in a composer project setup.
During plugin development I should just have to choose between 5.2 structure and pre-5.2 structure. But in a composer project this simply wont work. Although the `tests` folder is symlinked, `__DIR__` uses the absolute resolved path to the file.
Even worse for the pre-5.2 structured plugins as they moved up in the hierarchy. Although pre-5.2 is a legacy one should think about it.
It is ok for me if you decline that the multiple `..` might go to a completely different application and might break the tests as well. I am fine with a custom test kernel for composer projects as well. But this ends up in a complete mess in the test bootstrap per plugin where to get the correct test kernel from.

tests/Functional/bootstrap.php
```diff
@@ -22,7 +22,13 @@
  * our trademarks remain entirely with us.
  */
 
-require __DIR__ . '/../../autoload.php';
+if (file_exists(__DIR__ . '/../../../../../app/autoload.php')) {
+    // shopware as composer dependency
+    require __DIR__ . '/../../../../../app/autoload.php';
+} else {
+    // shopware standalone
+    require __DIR__ . '/autoload.php';
```
jkrzefski: I believe you wanted to require the same file as before.

```suggestion
    require __DIR__ . '/../../autoload.php';
```

JoshuaBehrens: Good hint. I had a chat with @soebbing about it and follow a different solution to the autoload file. There is still good solution for the config file missing.

JoshuaBehrens:
Regarding the configuration path. What about doing the same as done for the autoloader? I think of a loop that does the same like the autoloader loop but does not go any directories further then the autoloader parent directory.

soebbing:
Sounds like it might work, the only problem I see is that there is no single file path that might contain the `config.php`, but multiple. That might result in a lot of file checks

JoshuaBehrens:
Can we somehow use the autoloader to find the configuration file? Like a configuration finder that just knows where it is?

shyim:
Hey @JoshuaBehrens,
I don't see really the need to change this in shopware repository. What about placing in composer-project itself bootstrap files for Plugins, in the same place?

I tested it quickly and it works. I have created in composer-project template also the tests' folder with the same folder structure, but with the configuration for composer. For me it looks like your pain has been resolved with doing that.

**Composer-Project**: https://github.com/shopware/composer-project/commit/4bcc0cc6e93af356b76828f9ec842cc86fef95d6
**FroshProfiler**: https://github.com/FriendsOfShopware/FroshProfiler/commit/e7be5d4362a4f53f58fcd9fe3f0766e5fa663579

What do you think about that?

JoshuaBehrens:
I tested it locally on a composer installation and a normal installation and it worked out as @shyim stated. But doesn't this mean, that every plugin has to have this testing setup like shyim? If I am right the current cli tools do not generate something similar to this.

JoshuaBehrens:
So this will result in a PR for the cli tools instead to create suitable plugin skeletons in the first place as well as some changes in the documentation. Thanks for the thoughts about it.

PR #2006, opened 2019-02-05
SW-17109 do not use uppercase values for product filter

jnugh:
With the current implementation it is impossible to filter certain products by name.
Using `toUpperCase()` is not bijective 
```js
str.toUpperCase().toLowerCase() === str.toLowerCase() //might be false
```

This is an issue for chars like `ß` where `'ß'.toUppwerCase() == 'SS'` in most modern browsers.

For that reason the raw user input should be used as a filter.

The Lexer will store the original input in the AST if the token is of type `values`.

Search for `ß` in the demo data. Articles like "Kicker Figuren Set, große Auswahl" will not be in the result set.

https://issues.shopware.com/issues/SW-17109

None

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/article_list/resource/lexer.js
+++ b/themes/Backend/ExtJs/backend/article_list/resource/lexer.js
@@ -99,7 +99,7 @@ Ext.define('Lexer', {
             expectation,
             valid,
             finalCheck,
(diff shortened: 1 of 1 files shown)


shyim:

PR #2007, opened 2019-02-05
Improved performance of privacy-cleanup

fivetide:
On high traffic instances, cleanup of guest users and basket/orders might fetch many ids. Doing these cleanups right in the database might see a significant performance increase.

instead of fetching ids manually, postprocessing them for user/order-status and then passing them into delete-where-in, we perform the deletion directly in the database

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Privacy/PrivacyService.php
+++ b/engine/Shopware/Components/Privacy/PrivacyService.php
@@ -34,7 +34,7 @@ class PrivacyService implements PrivacyServiceInterface
     private $connection;
 
     /**
-     * MaintenanceService constructor
+     * PrivacyService constructor
      *
      * @param Connection $connection
      */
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Privacy/PrivacyService.php
```diff
@@ -24,113 +24,78 @@
...
-            }
-            if ($validOrder) {
-                unset($users[$id]);
-            }
-        }
-
-        $this->deleteFromTable('s_user', array_keys($users));
+        $threshold = (new \DateTime())->modify(sprintf('-%d months', $months));
+
+        $idsQuery = $this->entityManager->createQueryBuilder()
```
shopwareBot: Can you please use the DBAL connection for this?
Or is there any specific reason for using the ORM here?

Also, remember to revert your changes in the constructor then.

fivetide: In my opinion, there is no benefit in writing queries to the database directly. Using the ORM to construct these improves readability, enables autocompletion and makes these queries future-proof in case the database-fields ever change. The only caveat is that doctrine does not support delete with join or dependant subqueries, so i had to write a raw query for the deletion (line 67 to 73).

teiling88: DBAL has a QueryBuilder too. :-) 
Auto completion is available if you use PHPStorm with an configured Database Connection as well.

fivetide: @teiling88 while you are technically correct, i still fail to see any benefit in exposing the implementation details (database column names) here. Using the ORMs query builder nicely takes care of the join condition and those nasty inconsitencies like basket.date / datum and the mixed case on foreign keys (order.userID).

jkrzefski:
IIRC Shopware prefers DBAL over Doctrine for all storefront related queries, because of performance reasons. With DBAL there is no ORM layer which has to be resolved and you can optimize the query by limiting the select to the required information. I did not test your PR for performance but I find it hard to believe that this implementation should be the best possibility in terms of performance.

fivetide:
@jkrzefski, neither is this storefront-related - but cronjobs on the backend - nor does it do any select or hydration. ORM is only used to create clean sql for delete-queries...

engine/Shopware/Components/Privacy/PrivacyService.php
bcremer: The use of a QueryBuilder (DBAL or ORM) is not necessary nor helpful here as the query contains no dynamic parts but the actual query params.

```php
// query not tested
$query = <<<SQL
DELETE user 
FROM s_user user
LEFT JOIN s_order order ON user.id = order.userID AND order.status != -1
WHERE user.accountmode = 1 
AND user.firstLogin < :threshold
AND order.id is null
SQL;

$connection->executeUpdate(
    $query,
    ['threshold' => $threshold->format('Y-m-d H:i:s')]
);
```

shopwareBot: Hey @fivetide,

while you're right about the performance, we'd still like to have some consistency in this.
Building an SQL with an ORM, only to execute it in DBAL then, might be confusing.
That's why we'd rather have both QueryBuilder instances to be from DBAL, just for consistency reasons.

shyim:

PR #2009, opened 2019-02-06
Make annotations reader public

hlohaus:

Make the reader public so you can decorate them.
For example you can remove the Assert\NotBlank or Assert\Regex in models.

### 6. Checklist


shyim:
thanks for the Contribution! :blue_heart:

PR #2010, opened 2019-02-07
Fix conversion of Resource to service id with CamelCase-Name

fivetide:
Shopware\Components\Api\Manager::getResource('CustomerStream'); would fail to find the Service shopware.api.customer_stream

Fix conversion of CustomerStream to customer_stream

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Manager.php
+++ b/engine/Shopware/Components/Api/Manager.php
@@ -25,6 +25,7 @@
 namespace Shopware\Components\Api;
 
 use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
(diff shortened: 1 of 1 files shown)


mitelg:
hey @fivetide 
there is already a fix for that https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Components/DependencyInjection/services.xml#L647

fivetide:
@mitelg , right, didn't see that one. Though it looks more like a workaround. This fix would actually make the behaviour consistent in case any future APIs are added with CamelCase names... But If you don't expect that to happen, feel free to close this PR.

PR #2012, opened 2019-02-07
Pulled method up into the interface

JoshuaBehrens:
You cannot simply decorate VariantHelper as the [ImmediateDeliveryFacetHandler](https://github.com/shopware/shopware/blob/24affc52190ed03dd1cf5b8f90d14a8945b5c5c1/engine/Shopware/Bundle/SearchBundleDBAL/FacetHandler/ImmediateDeliveryFacetHandler.php#L105) expects a specific implementation of it.

To decorate VariantHelper

Extend the interface to contain a public function that is already in the current implementation

1. Try to decorate VariantHelper
2. Open a listing
3. :bomb: 

This should be part of the UPGRADE-XX.md as this changes an interface.

### 5. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/VariantHelper.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/VariantHelper.php
@@ -150,7 +150,7 @@ class VariantHelper implements VariantHelperInterface
     }
 
     /**
-     * @param QueryBuilder $query
+     * {@inheritdoc}
      */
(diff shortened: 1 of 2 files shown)


JoshuaBehrens:
There is still an error lunging around that makes travis fail that I saw in multiple other PRs. I do not expect it to be me.

JoshuaBehrens:
Yeah sure. Makes sense.

PR #2016, opened 2019-02-09
Added new option to AdvancedMenu

Isengo1989:
See also -> https://github.com/shopware/shopware/pull/1962

Imagine a shop with a lot of customergroups and a big menu (that is always the same). Every time a user logs in, a new menu will be generated and this will take a long time if you have +150 categories. There should be a way to exclude customergroups to be in the cache key.

auswahl_009

auswahl_242

Espacially the checkout (login/register) is critical, because users will more likely dicard their shopping cart if the login takes longer then a few seconds.

This change adds a new form field **excludeCustomergroup** which will decide if the customergroup will be used in the cache key or not.

The option (boolean) is set to 0 by default (same behavior as before) and even with the plugin not reinstalled it is not true and therefore the customergroups will be considered as usual. I also added the shop scope for caching, cachetime and hoverDelay.

Browse in a big shop (+150 categories) and login with customergroup which is not default.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1455-add-customer-group-key-exemption-option.php
+++ b/_sql/migrations/1455-add-customer-group-key-exemption-option.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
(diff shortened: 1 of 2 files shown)


Isengo1989:
@PaddyS created a new PR to have a clean start and weekend laziness :smile:

shyim:
The migration is still wrong. See https://github.com/shopware/shopware/blob/5.5/_sql/migrations/406-add-advanced-menu-listeners.php#L13 as example

Isengo1989:
@shyim so the check if the plugin is active or not should not be in the query? Just saw I copy and pasted the wrong class name anyways - so i will fix that too.

shyim:
Remove active from condition AND execute the query like in the example to check that the plugin is installed.

shyim:
Thanks for the Contribution! :blue_heart:

PR #2018, opened 2019-02-12
Added event to modify query in ListProductGateway

oktupol:

With this change, it's now possible to modify the query used to display products in the store front.

It adds an event to which plugin developers may subscribe

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/ListProductGateway.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/ListProductGateway.php
@@ -66,21 +66,29 @@ class ListProductGateway implements Gateway\ListProductGatewayInterface
     private $connection;
 
     /**
+     * @var \Enlight_Event_EventManager
+     */
+    private $eventManager;
+
+    /**
      * @param Connection                  $connection
      * @param FieldHelper                 $fieldHelper
      * @param Hydrator\ProductHydrator    $hydrator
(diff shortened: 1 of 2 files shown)


oktupol:
While this works, I haven't found any way to decorate `ListProductGateway` without breaking other decorators. The only way I found to modify the query in a decorator was following:

```php
<?php
class ListProductGatewayDecorator extends ListProductGateway
{
    public function __construct($connection, $fieldHelper, $hydrator, $config) {
        parent::__construct($connection, $fieldHelper, $hydrator, $config);
    }

    protected function getQuery($numbers, $context) {
        $query = parent::getQuery($numbers, $context);

        // apply magic

        return $query;
    }
}
```

If I have another decorator `OtherDecorator`, which comes first in the decoration chain, `OtherDecorator`'s changes would be discarded, or, to be more precise, `OtherDecorator` would never be loaded.

oktupol:
Would it be ok to have a `ListProductQueryService` into which I source the getQuery-method out?

soebbing:
What you are doing in the code above is not a decoration, that's why other decorators are breaking.

But I can see what you're trying to achieve, I am not entirely sure what the best way to implement this would be.

My first thought would be something like the `FieldHelper`-class in the the `ListProductGateway`: a simple Helper-class with an interface that plugins can decorate, which would allow for a modification of the `QueryBuilder`.

What do you think about that?

oktupol:
That sounds good. I'll try that

PR #2021, opened 2019-02-14
redirect Note-Actions instead of forwarding

tinect:
Looks better :-)

change forward to redirect

- add product to note
- switch to note
- remove product from note
- see address in browser

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Note.php
+++ b/engine/Shopware/Controllers/Frontend/Note.php
@@ -51,7 +51,8 @@ class Shopware_Controllers_Frontend_Note extends Enlight_Controller_Action
         if (!empty($this->Request()->sDelete)) {
             Shopware()->Modules()->Basket()->sDeleteNote($this->Request()->sDelete);
         }
-        $this->forward('index');
+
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Frontend/Note.php
```diff
@@ -62,7 +63,7 @@ public function addAction()
             $this->View()->sArticleName = Shopware()->Modules()->Articles()->sGetArticleNameByOrderNumber($orderNumber);
         }
 
-        $this->forward('index');
+        $this->redirect(['action' => 'index']);
```
uehler, line 66: you will loose the view variable with the redirect, won't you?

shyim: You lose in forward and redirect the view variables. I guess the important change is here, that the GET/POST parameters are away.

shyim:
Please change the target to 5.6. It's a BC break

shyim:
Hey @tinect,
your change has been merged in 5.6. Thanks for your Contribution! :blue_heart:

PR #2022, opened 2019-02-15
add filter-event Shopware_Controllers_Frontend_Account_Login_Redirect

tinect:
Give us the possibility to manipulate the redirection after successfull Login

add an filter-event `Shopware_Controllers_Frontend_Account_Login_Redirect`

### 6. Checklist


hlohaus:
You can not do that with a postdispatch event?

tinect:
using this event I don't need to check the state of login and errors again. I just can start at the place I want.

PR #2023, opened 2019-02-17
Fix Logger mode

foxlady:
The backend shows a notification about warmup error and says it has been logged. But 404 errors won't be logged in default, cause of Monolog min level (default error).

Changed logger level from notice to error

Warmup 404 pages, there is currently no entry in log

### 6. Checklist


tinect:
This will result in reporting errors for hidden categories again, won't it?

shyim:
@tinect Created a second Ticket to solve the category issue: https://issues.shopware.com/issues/SW-23530. We will fix that together

shyim:
Hey @foxlady,

PR #2024, opened 2019-02-19
Fix conversion of Resource to service id with CamelCase-Name

fivetide:
Shopware\Components\Api\Manager::getResource('CustomerStream'); would fail to find the Service shopware.api.customer_stream

Fix conversion of CustomerStream to customer_stream

### 6. Checklist

This PR is an updated version of https://github.com/shopware/shopware/pull/2010 for 5.5-branch

### DIFF
--- a/engine/Shopware/Components/Api/Manager.php
+++ b/engine/Shopware/Components/Api/Manager.php
@@ -26,6 +26,7 @@ namespace Shopware\Components\Api;
 
 use Shopware\Components\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
+use Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter;
 
 /**
  * API Manger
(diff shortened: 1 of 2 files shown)


fivetide:
I took the opportunity to add aliases and deprecate inconsistent service-ids. Hope this is fine.

fivetide:
> Hey @fivetide,
> 
> can you please add a condition to check for the old name first?
> 
> ```
> if (!$this->container->has('shopware.api.' . strtolower($name))) {
>     $resource = $container->get('shopware.api.' . (new CamelCaseToSnakeCaseNameConverter())->normalize($name));
> }
> ```
> The reason for this are plugins, that also created a new API resource with a concatenated lowercase service-name. Those would break with your changes.

My change includes service-aliases for those, including deprecated-tags that will inform the developer.

shyim:
We think about custom api resources from plugins

soebbing:
> My change includes service-aliases for those, including deprecated-tags that will inform the developer.

If I understand correctly, this only applies for the currently existing resources.

Consider a custom plugin, creating a resource `shopware.api.storefront` with class `Somecompany/StoreFront`. With your change, this resource won't be found anymore by `\Shopware\Components\Api\Manager::getResource`, since there is no deprecated alias for this custom service.

shyim:

PR #2025, opened 2019-02-19
Change s_order_basket.quantity and s_order_details.quantity to unsigned int

aragon999:
It could happen, for example from a plugin, that the quantity is inserted is negative. Afterwards the calculation of the shipping costs fail, with the following error:
image
It is now very hard to reconstruct, which plugin or at what point the negative quantity was inserted. 

Changes the database quantity fields to unsigned int, such that the error occurs when a plugin tries to set a negative quantity.

Easiest way would be to manually change the quantity in the database.

\-

\-

### 6. Checklist

### DIFF
--- a/_sql/migrations/1457-change-quantity-to-unsinged-int.php
+++ b/_sql/migrations/1457-change-quantity-to-unsinged-int.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * Shopware 5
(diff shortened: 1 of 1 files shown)


aragon999:
Will look into the tests later, strange that they succeeded on my local machine. Maybe I forgot to apply the migration :see_no_evil:

PR #2029, opened 2019-02-20
show update-version in plugin-manager

tinect:
The word "Update" is too little :-)

show available version instead of text "Update"
image

### 6. Checklist


mitelg:
how about "Update to v2.0.0" as text?

tinect:
> how about "Update to v2.0.0" as text?

Yes, could be a solution. But looks quite ugly cause the label above will be shorter and looks different. I think the label are clear as they are, aren't they? :-)

tinect:
image
You mean like this? I like! :-) Hopefully it's correct,

Tell me, if correct, so I will squash.

PaddyS:
Hey @tinect,

As we previously discussed, this minor change comes with a lot of work to be done.
There's several edge cases to be considered, multiple styles to be changed, etc., so the overall benefit doesn't really fit well to the work that needs to be done.

Therefore we're closing this PR.

PR #2030, opened 2019-02-20
Update header.less

Julianfleischmann:
Missing Semicolon could be an error in less

Adding a missing semicolon

No issues until. But the code is now a bit more clean ;)

No issue.

### 6. Checklist


tinect:
there are multiple locations with this. I think, there is no need, unless there is a decision to add it everywhere.
Interesting discussion: https://stackoverflow.com/questions/11939595/leaving-out-the-last-semicolon-of-a-css-block

PR #2031, opened 2019-02-21
Make customergroup keys longer

pascalheidmann-bedarf:
The customergroup model allows up to 5 chars while the customer model allows 15 characters for the customer group. The lower limit makes naming customer groups hard where you have several similar group names, eg. if you have companies as your customers where each company has their own customergroup. Having more space to use improves it a lot.

Upping the customergroup model's limit to match the limit set in the customer model to 15 characters.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1607-allow-longer-customergroup-keys.php
+++ b/_sql/migrations/1607-allow-longer-customergroup-keys.php
@@ -1,0 +1,38 @@
+<?php
+
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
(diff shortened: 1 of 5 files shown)


_sql/migrations/1607-allow-longer-customergroup-keys.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function up($modus): void
+    {
+        $this->addSql("ALTER TABLE `s_core_customergroups` CHANGE `groupkey` `groupkey` VARCHAR(15);");
+    }
+
+    /**
+     * @param string $modus
+     */
+    public function down($modus): void
```
shyim: Plugin Migrations need only down

shyim (changes requested):

PaddyS:
Whoops. Wrong order.
Anyway, please still fix this issue. :-)

pascalheidmann-bedarf:
@shopwareBot @PaddyS good point.

But: searching through all models I only found references for `$customerGroupKey` in `\Shopware\Models\Article\Configurator\Template\Price::$customerGroupKey` and `\Shopware\Models\Article\Price::$customerGroupKey` - both where with a space of 30 characters. Reducing this to match the others limit of 15 characters feels bad as this means potential (but unrealistic) data loss.

Searching further with broader `$customerGroup` revealed an even longer version: `Shopware\Models\Newsletter\Newsletter::$customerGroup` with 255 characters.
Gladly other instances only had a reference for the customergroup id (typical `int(11)`).

What is your suggestion regarding this discrepancy of those fields? As we don't use the newsletter module I don't have data for that usecase anyway but I would feel okay going up to 30 characters but 255 seem overkill.

pascalheidmann-bedarf:
@shopwareBot all right.
I changed all Doctrine annotations to 15 characters but didn't include a migration for those models that had space for more characters (to prevent potential data loss if other changed those fields by hand)

shyim:
Users can't loose data. The groupkey is defined in s_core_customergroups and the users can't have longer text as there, cause its joined in multiple places

shyim:
Thanks for your Contribution :blue_heart:

PR #2033, opened 2019-02-22
Fix translation of product attributes via RestAPI

mnaczenski:
Currently the product attributes won't be translated correctly, since they are not saved in the s_articles_translation table. The issue is caused by a mix of camelCase and snake_case. The s_core_translations uses camelCase and the s_article_translations uses snake_case to identify the attribute.

The foreach will add every attribute as snake_case to the array as well, otherwhise this check will fail: https://github.com/shopware/shopware/blob/cfcbcc820efefc3943280c2f9356da23c556f1b4/engine/Shopware/Components/Translation.php#L540

aasaasasasas vs aasa_asas_asas

Add a new attribute column like "aasa_asas_asas" to your product attributes as translatable and try to translate this attribute via RestAPI.

` $minimalTestArticle = [
        'name' => 'Test',
        'translations' => [
            3 => [
                'name' => 'Test-EN',
                '__attribute_aasaAsasAsas' => "sdgfdghdfghfgdhfghfgd",
                '__attribute_attr1' => 'testetst',
                'shopId' => 3,
            ]
        ]
    ];.`

attr1 will be saved correctly, the new attribute not.

https://issues.shopware.com/issues/SW-23584

Since the attributes are only added to the array as snake_case and not overwritten, this should not break anything. 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Translation.php
+++ b/engine/Shopware/Components/Translation.php
@@ -24,6 +24,7 @@
 
 use Doctrine\DBAL\Connection;
 use Shopware\Bundle\AttributeBundle\Service\CrudService;
+use Symfony\Component\Serializer\NameConverter\CamelCaseToSnakeCaseNameConverter;
 
 /**
  * Shopware Translation Component
@@ -532,7 +533,15 @@ class Shopware_Components_Translation
         $columns = $schemaManager->listTableColumns('s_articles_translations');
         $columns = array_keys($columns);
 
+        $converter = new CamelCaseToSnakeCaseNameConverter();
+
         foreach ($data as $key => $value) {
+            if (strpos($key, '__attribute_') !== false) {
+                $columnname = str_replace('__attribute_', '', $key);
+                $columnname = $converter->normalize($columnname);
+                $data[$columnname] = $value;
+            }
+
             $column = strtolower($key);
             $column = str_replace(CrudService::EXT_JS_PREFIX, '', $column);
 


engine/Shopware/Components/Translation.php
```diff
@@ -483,6 +483,18 @@ protected function fixArticleTranslation($languageId, $articleId, $data)
         );
         $fallbacks = array_column($fallbacks, 'id');
 
+        $unserializedData = unserialize($data);
+
+        //Adds the attribute as snake_case to the array
+        foreach ($unserializedData as $columnname => $value) {
+            if (strpos($columnname,'__attribute_') !== false) {
+                $columnname = strtolower(preg_replace('/(?<!^)[A-Z]/', '_$0', $columnname));
```
FiveDigital: Perhaps it's better to do this with the `CamelCaseToSnakeCaseNameConverter` class?

mnaczenski: I will have a look at it, thank you!

hlohaus:
see my fix https://github.com/shopware/shopware/pull/1591

mnaczenski:
Done

hlohaus:
Maybe this a better fix:
https://github.com/shopware/shopware/pull/1591
We should write it right, so it works everywhere.
Additionally s_article_translations doesn't contain any attributes

engine/Shopware/Components/Translation.php
```diff
@@ -483,6 +484,19 @@ protected function fixArticleTranslation($languageId, $articleId, $data)
         );
         $fallbacks = array_column($fallbacks, 'id');
 
+        $unserializedData = unserialize($data);
+        $converter = new CamelCaseToSnakeCaseNameConverter();
+
+        //Adds the attribute as snake_case to the array
+        foreach ($unserializedData as $columnname => $value) {
+            if (strpos($columnname, '__attribute_') !== false) {
+                $columnname = $converter->normalize($columnname);
```
shopwareBot: Can you move the normalize to prepareArticleData method

hlohaus:
? Please read my comment. Your fix only work in search. Not in the listing, detail page etc.

shyim:
Hey @hlohaus,
we will import both Pull Requests and decide when we work on it which one we merge

hlohaus:
Okay, but this pull request doesn't make sense. 
The product (search) table has only 5 attributes.
That the reason that the unit tests fail.

mnaczenski:
when creating an attribute in the backend, the s_articles_translations is also extended with the new attribute column. 
auswahl_068

hlohaus:
Okay, but attr5 to attr20 are missing in default.
So the unit tests doesn't work with your changes.
Maybe you can add them?
Additionally you should re add the check:
            if (in_array($column, $columns)) {
                $data[$column] = $value;
            }
So only exist columns will be added.

soebbing:
Hi @mnaczenski, thank you for your contribution! 💙 

We're closing your PR since the fix of @hlohaus seems more promising. So your issue will be fixed nevertheless!

PR #2034, opened 2019-02-25
Fix import attribute translation with underscore

hlohaus:
see title.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Category.php
+++ b/engine/Shopware/Components/Api/Resource/Category.php
@@ -27,7 +27,6 @@ namespace Shopware\Components\Api\Resource;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use RuntimeException;
 use Shopware\Components\Api\Exception as ApiException;
-use Shopware\Models\Attribute\Category as CategoryAttribute;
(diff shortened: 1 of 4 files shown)


engine/Shopware/Components/Api/Resource/Category.php
```diff
@@ -436,31 +435,12 @@ private function prepareMediaData(array $data, CategoryModel $categoryModel)
...
-            unset($properties[$field]);
-        }
-
-        foreach ($metaData->getIdentifierFieldNames() as $identifierFieldName) {
-            unset($properties[$identifierFieldName]);
-        }
-
+        /** @var \Shopware\Bundle\AttributeBundle\Service\CrudService $crud */
+        $crud = $this->getContainer()->get('shopware_attribute.crud_service');
+        $list = $crud->getList('s_articles_attributes');
```
shopwareBot: Did you test your changes? You are in Category Resource and reading Product Attributes.
Could you please also add some test-cases?

hlohaus: Oh, you are right. There is the issue too. I will update my pull request.

shyim:
Thanks for your contribution! :blue_heart:

PR #2035, opened 2019-02-26
SW-23605 | HttpCache beim ersten Aufruf fehlerhaft / Could not write …

arosenhagen:
Fatal error: Uncaught RuntimeException: Could not write cacheKey in engine/Shopware/Components/HttpCache/Store.php:206

fix empty return value to be 'true' instead of empty

On initial visit of any URL when HttpCache is enabled.

[SW-23605 | HttpCache beim ersten Aufruf fehlerhaft / Could not write cacheKey](https://issues.shopware.com/issues/SW-23605)

none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/HttpCache/Store.php
+++ b/engine/Shopware/Components/HttpCache/Store.php
@@ -391,5 +391,7 @@ class Store extends BaseStore
         }
 
         @chmod($path, 0666 & ~umask());
+
+        return true;
     }
 }


soebbing:
Hi @arosenhagen, 

thank you for your fix, but the build will fail due to the static code analyzer. I think a `return true;` at the end of the `save()`-method being called there should do the same. Can you verify that?

soebbing:
For everybody who searches for a quick fix for the issue above: This error only occurs when the debug-setting for the HTTP-Cache is active. Removing the following lines from your `config.php` fixes the issue :
```php
…
'httpcache' => [
    'debug' => true
]
```

PR #2036, opened 2019-02-26
allow custom attributes in document templates

juliusbentz:
This change is necessary to use custom attributes in any document templates

With this change the function getPositions selects all columns of s_articles_attributes including custom shop-specific attributes instead of only at.attr1 - at.attr20. 

If custom article attributes are added, they cannot be used in document templates.
With this change all attributes that worked before still work and custom ones can be used too.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Document/Order.php
+++ b/engine/Shopware/Models/Document/Order.php
@@ -386,11 +386,21 @@ class Shopware_Models_Document_Order extends Enlight_Class implements Enlight_Ho
      */
     public function getPositions()
     {
+        // Workaround to fetch all attribute columns without shadowing id columns from other tables
+        $attributeColumns = 'at.attr1, at.attr2, at.attr3, at.attr4, at.attr5,
+                             at.attr6, at.attr7, at.attr8, at.attr9, at.attr10,
+                             at.attr11, at.attr12, at.attr13, at.attr14,
+                             at.attr15, at.attr16, at.attr17, at.attr18,
+                             at.attr19, at.attr20';
+        $attributeColumnNames = Shopware()->Db()->fetchRow('SELECT * FROM s_articles_attributes LIMIT 1');
+        if (!empty($attributeColumnNames)) {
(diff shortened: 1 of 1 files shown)


shyim:
Hey, 
your implementation would break the articleID.
Maybe implement something similar to this https://github.com/shopware/shopware/blob/308789ab660c1d2bf5776333cd13b757a935646a/engine/Shopware/Core/sExport.php#L827-L835

juliusbentz:
In regard to your comment the workaround has been used to avoid breaking the ArticleID.

engine/Shopware/Models/Document/Order.php
```diff
@@ -386,11 +386,21 @@ public function processOrder()
      */
     public function getPositions()
     {
+        // Workaround to fetch all attribute columns without shadowing id columns from other tables
+        $attributeColumns = 'at.attr1, at.attr2, at.attr3, at.attr4, at.attr5,
+                             at.attr6, at.attr7, at.attr8, at.attr9, at.attr10,
+                             at.attr11, at.attr12, at.attr13, at.attr14,
+                             at.attr15, at.attr16, at.attr17, at.attr18,
+                             at.attr19, at.attr20';
+        $attributeColumnNames = Shopware()->Db()->fetchRow('SELECT * FROM s_articles_attributes LIMIT 1');
```
shopwareBot, line 395: Can you change this to a DBAL Query Builder?

engine/Shopware/Models/Document/Order.php
```diff
@@ -386,11 +386,21 @@ public function processOrder()
...
+        if (!empty($attributeColumnNames)) {
+            unset($attributeColumnNames['id'], $attributeColumnNames['articleID'], $attributeColumnNames['articledetailsID']);
+            $attributeColumns = 'at.' . implode(', at.', array_keys($attributeColumnNames));
```
shyim, line 398: And it would be cool, if you could add after that fetching from database a filter event. So plugin developers could resolve Attribute fields with Single or Multiselection (due only id saved)

PR #2037, opened 2019-02-26
Adds configuration to have the basket cleared on logout

jkrzefski:
you might not wish to clear the basket when a customer logs out.

Adds a new configuration option under "Basic settings > Shopping cart / item details" to decide whether the basket should be cleared on logout or not. It is set to true per default to not break the existing behaviour.

1. Login as customer.
2. Put a product into your basket.
3. Logout.
4. Now the basket is cleared. The new configuration makes it possible to not have the basket cleared.

https://issues.shopware.com/issues/SW-13110

-

### 6. Checklist

### DIFF
--- a/_sql/migrations/1457-new-config-clear-basket-on-logout.php
+++ b/_sql/migrations/1457-new-config-clear-basket-on-logout.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


_sql/migrations/1457-new-config-clear-basket-on-logout.php
```diff
@@ -0,0 +1,32 @@
...
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+class Migrations_Migration1457 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $this->addSql('INSERT INTO s_core_config_elements (form_id, name, value, label, description, type, required, position, scope, options) VALUES (147, "clearBasketAfterLogout", "b:1;", "Warenkorb beim Abmelden leeren", null, "boolean", 0, 0, 0, null);');
```
shopwareBot, line 29: Can you change the default from true to false. So that after an Update nothing changes.

jkrzefski: If you look closely, you will notice that I made the default true for this exact reason. If the value is true, the basket will be cleared. For now it will always be cleared, so true would be the correct default value here.

jkrzefski:
@shopwareBot I rebased my commit and changed the target branch. Do you need anything else from me?

shyim:

PR #2039, opened 2019-02-28
Fix attribute valid function

hlohaus:
### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Struct/Attribute.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Struct/Attribute.php
@@ -41,12 +41,15 @@ class Attribute extends Struct implements \JsonSerializable
     /**
      * @param array $data
      *
-     * @throws \Exception
+     * @throws \InvalidArgumentException
      */
     public function __construct($data = [])
     {
+        if (!is_array($data)) {
+            throw new \InvalidArgumentException('Injected data should be an array');
+        }
         if (!$this->isValid($data)) {
-            throw new \Exception('Class values should be serializable');
+            throw new \InvalidArgumentException('Class values should be serializable');
         }
         $this->storage = $data;
     }
@@ -71,12 +74,12 @@ class Attribute extends Struct implements \JsonSerializable
      * @param string $name
      * @param mixed  $value
      *
-     * @throws \Exception
+     * @throws \InvalidArgumentException
      */
     public function set($name, $value)
     {
         if (!$this->isValid($value)) {
-            throw new \Exception('Class values should be serializable');
+            throw new \InvalidArgumentException('Class values should be serializable');
         }
 
         $this->storage[$name] = $value;
@@ -119,22 +122,24 @@ class Attribute extends Struct implements \JsonSerializable
      */
     private function isValid($value)
     {
-        if (is_scalar($value)) {
-            return true;
-        }
         if ($value instanceof \JsonSerializable) {
             return true;
         }
-        if (!is_array($value)) {
+
+        if (is_scalar($value) || $value === null) {
             return true;
         }
 
-        foreach ($value as $val) {
-            if (!$this->isValid($val)) {
-                return false;
+        if (is_array($value)) {
+            foreach ($value as $val) {
+                if (!$this->isValid($val)) {
+                    return false;
+                }
             }
+
+            return true;
         }
 
-        return true;
+        return false;
     }
 }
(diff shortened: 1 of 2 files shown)


engine/Shopware/Bundle/StoreFrontBundle/Struct/Attribute.php
```diff
@@ -75,10 +75,6 @@ public function exists($key)
      */
     public function set($name, $value)
     {
-        if (!$this->isValid($value)) {
-            throw new \Exception('Class values should be serializable');
```
bcremer, line 79: Why do you think this guard is unnecessary?
This class implements `JsonSerializable` so all attributes must be serializable too.
It should not be allowed to pass in a file handle (`resource`) or anything else that can not be represented as a string.

engine/Shopware/Bundle/StoreFrontBundle/Struct/Attribute.php
```diff
@@ -111,30 +107,4 @@ public function jsonSerialize()
...
-            return true;
-        }
-
-        foreach ($value as $val) {
-            if (!$this->isValid($val)) {
-                return false;
-            }
-        }
-
-        return true;
```
shyim: This should be false

engine/Shopware/Bundle/StoreFrontBundle/Struct/Attribute.php
```diff
@@ -111,30 +107,4 @@ public function jsonSerialize()
...
-    private function isValid($value)
-    {
-        if (is_scalar($value)) {
-            return true;
-        }
-        if ($value instanceof \JsonSerializable) {
-            return true;
-        }
-        if (!is_array($value)) {
-            return true;
```
bcremer: Should also return false.
Can be any complex Object that is not `JsonSerializable`.

This class needs a test.

aragon999:
Maybe something like the following:

```php
private function isValid($value)
{
    if ($value instanceof \JsonSerializable) {
        return true;
    }

    if (is_scalar($value) || $value === null) {
        return true;
    } elseif (is_array($value)) {
        foreach ($value as $val) {
            if (!$this->isValid($val)) {
                return false;
            }
        }
    }

    return false;
}
``` 

and yes, this should probably be tested, do you refactor your PR, or should I create a new one?

tests/Unit/Bundle/StoreFrontBundle/Struct/AttributeStructTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+class AttributeStructTest extends TestCase
+{
+    /**
+     * @dataProvider testDataProvider
+     *
+     * @param mixed $value
+     * @param bool  $result
+     */
+    public function testIsValid($value, $result)
```
bcremer: Please separate into two data providers for valid and invalid data:

```
testWithValidValues($value)
{
    $attr = new Attribute(['attr' => $value]);
    $this->assertSame($value, $attr->get($value));
}

testWithInvalidValues($value)
{
    $this->expectException(\Exception::class);
     new Attribute(['attr' => $value]);
}

File: engine/Shopware/Bundle/StoreFrontBundle/Struct/Attribute.php [hunk 5]
```diff
@@ -45,6 +45,9 @@ class Attribute extends Struct implements \JsonSerializable
      */
     public function __construct($data = [])
     {
+        if (!is_array($data)) {
+            throw new \Exception('Injected data should be a array');
```
PaddyS: *an array :-)

shyim: and InvalidArgumentException

engine/Shopware/Bundle/StoreFrontBundle/Struct/Attribute.php
```diff
@@ -45,6 +45,9 @@ class Attribute extends Struct implements \JsonSerializable
      */
     public function __construct($data = [])
     {
+        if (!is_array($data)) {
+            throw new \InvalidArgumentException('Injected data should be an array');
+        }
         if (!$this->isValid($data)) {
             throw new \Exception('Class values should be serializable');
```
shopwareBot: This should be InvalidArgumentException

shopwareBot (changes requested):

engine/Shopware/Bundle/StoreFrontBundle/Struct/Attribute.php
```diff
@@ -76,7 +79,7 @@ public function exists($key)
     public function set($name, $value)
     {
         if (!$this->isValid($value)) {
-            throw new \Exception('Class values should be serializable');
+            throw new \RuntimeException('Class values should be serializable');
```
shopwareBot: Here also InvalidArgumentException

shyim:
We have merged it into 5.6 Branch. Thanks for your contribution! :blue_heart:

PR #2040, opened 2019-03-01
add pseudoPercent to backend + adjust price-columns

tinect:
The management of "Prozentrabatt" isn't clear for users:
- who didn't read the documentarion attentively
- the position between price and pseudoPrice is confusing as it will indicate to be the discount for pseudoPrice=>price

image

- move `percent` before `price`
- make `percent` empty, if it's 0. so first field will be empty for better reading and unsterstanding, too.
- add new column `pseudoPercent`. this field won't be stored into database, just live calculated. There is no need to store it.

https://docs.shopware.com/de/shopware-5-de/produkte-und-kategorien/produkte-anlegen#preise
I would adjust, if desired :-)

### 6. Checklist


shyim:

PR #2046, opened 2019-03-05
Add snippets everywhere

hlohaus:
Adds support for snippets in emails, product exports and eg.

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Components/Snippet/DummyResource.php
+++ b/engine/Library/Enlight/Components/Snippet/DummyResource.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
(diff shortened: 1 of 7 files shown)


shyim:
Hey @hlohaus,
can you add please some tests?

PaddyS:
Hey @hlohaus,

I've tested your Pull Request and unfortunately found an issue with it.

When using plugins like "AboCommerce", some snippets in the Article module are not properly loaded.
It seems that's just an issue in files like these:
```

// {namespace name="backend/abo_commerce/article/view/main"}
// {block name="backend/article/view/detail/window"}
// {$smarty.block.parent}
Ext.define('Shopware.apps.Article.view.detail.AboCommerceWindow', {
    override: 'Shopware.apps.Article.view.detail.Window',
```

In those cases, the snippets are not loaded from the given namespace. In fact, the given namespace is fully ignored here and instead the snippet's default value is used and saved as a new snippet.

Can you please have a look into this?

PR #2050, opened 2019-03-06
Fixes unknown `has` for HttpFoundation Request

EtienneBruines:
The method `has` exists for individual ParameterBags, but not the Request as a whole.

Looks for the argument in all parameter bags individually. Basically checking if the `get` method was able to provide *any* value. 

screenshot-2019-03-06_16 13 16

### 6. Checklist


engine/Shopware/Bundle/ControllerBundle/ArgumentResolver/RequestParameterResolver.php
```diff
@@ -35,7 +35,9 @@ class RequestParameterResolver implements ArgumentValueResolverInterface
      */
     public function supports(Request $request, ArgumentMetadata $argument)
     {
-        return $request->has($argument->getName());
+        return $request->attributes->has($argument->getName())
```
shyim: Please do $request->get with a null check instead

shyim:

PR #2052, opened 2019-03-07
Show wrong opcache settings in system info

niklasbuechner:
If two Shopware shops are running on the same server, they might override each others model definition, if the OPCache module is not configured correctly. Depending on the definition, OPCache does not include the directory path in its caching which can lead to multiple files with the same name overriding each other. As a result, the shop will experience sporadic model definition errors which are hard to track down.

This pr will add the critical OPCache settings into the requirements tab if the OPCache module is enabled.

This change adds the `opcache.use_cwd = 1` and the `opcache.validate_root = 1` settings into the requirements tab in case OPCache is used.

Configure OPCache incorrectly with `opcache.use_cwd = 0` and `opcache.validate_root = 0`. The requirements tab will now show if the correct settings are defined.

None

Add the options into the sysadmins guide: https://developers.shopware.com/sysadmins-guide/shopware-5-performance-for-sysadmins/#opcode-cache

### 6. Checklist

### DIFF
--- a/recovery/install/src/Requirements.php
+++ b/recovery/install/src/Requirements.php
@@ -71,6 +71,7 @@ class Requirements
             'phpVersionNotSupported' => false,
         ];
 
+        $checks = [];
         foreach ($this->runChecks() as $requirement) {
             $check = [];
 
@@ -95,6 +96,12 @@ class Requirements
                 }
             }
 
+            $checks[] = $check;
+        }
+
+        $checks = array_merge($checks, $this->checkOpcache());
+
+        foreach ($checks as $check) {
             if (!$check['check'] && $check['error']) {
                 $check['status'] = 'error';
                 $result['hasErrors'] = true;
@@ -215,6 +222,44 @@ class Requirements
     private function checkModRewrite()
     {
         return isset($_SERVER['MOD_REWRITE']);
+    }
+
+    /**
+     * Checks the opcache configuration if the opcache exists.
+     */
+    private function checkOpcache()
+    {
+        if (!extension_loaded('Zend OPcache')) {
+            return [];
+        }
+
+        $useCwdOption = $this->compare('opcache.use_cwd', ini_get('opcache.use_cwd'), 1);
+        $opcacheRequirements = [[
+            'name' => 'opcache.use_cwd',
+            'group' => 'core',
+            'required' => 1,
+            'version' => ini_get('opcache.use_cwd'),
+            'result' => ini_get('opcache.use_cwd'),
+            'notice' => '',
+            'check' => $this->compare('opcache.use_cwd', ini_get('opcache.use_cwd'), 1),
+            'error' => '',
+        ]];
+
+        if (fileinode('/') > 2) {
+            $validateRootOption = $this->compare('opcache.validate_root', ini_get('opcache.validate_root'), 1);
+            $opcacheRequirements[] = [
+                'name' => 'opcache.validate_root',
+                'group' => 'core',
+                'required' => 1,
+                'version' => ini_get('opcache.validate_root'),
+                'result' => ini_get('opcache.validate_root'),
+                'notice' => '',
+                'check' => $this->compare('opcache.validate_root', ini_get('opcache.validate_root'), 1),
+                'error' => '',
+            ];
+        }
+
+        return $opcacheRequirements;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


wbob:
Sanity checks are good - `opcache.use_cwd` is true by default in both development and production php.ini for 6 years now and a warning is helpful if it is disabled.

`opcache.validate_root` will check the inode number of the root to distinguish chroot environments, prevent compromise and avoid key collisions - but for non-chrooted environments it shouldn't be marked as a requirement. Can you wrap the check in a `if (fileinode('/') > 2)` ?

I wondered about the origin of validate_root / validate_permission - [bugs.php.net#69090](https://bugs.php.net/bug.php?id=69090) is an interesting read.

niklasbuechner:
Unfortunately, the check `if (fileinode('/') > 2)` does not work on all systems. Depending on which system you run, the check would result in a false positive. (https://unix.stackexchange.com/questions/14345/how-do-i-tell-im-running-in-a-chroot/24248#24248)

In addition the benefit of always including this setting in the `requirements` tab far outweighs the consequences of not including it.
- Not showing the user these settings leads to many incorrectly configured OPCaches which result in weird 500 errors for shop administrators and their customers.
- All checks that are execute by the opcache are cached by the kernel so there should be almost no performance overhead.

Therefore it is preferable to show the warning too often, then to not show it when needed.

wbob:
Yes, that inode check is unsophisticated (can it be improved?) - but so is the requirements check. Without "some" chroot awareness the check will generate more unneccesarry warnings (for non-chrooted envs) than false positives (chroot + on separate mountpoint) or false negatives (non-ext4 fs).

Maybe an addition to the opcache section in the dev/admin docs can be more elaborate. Though I understand less people will read it. Your change is in good faith and to warn others, I'm not sure what is best - in the end, validate_root is not a default setting for regular php envs.

Reading through that php bug, its recent 2018 comment and the opcache [code](https://github.com/php/php-src/blob/php-7.3.0/ext/opcache/ZendAccelerator.c#L2338), I think with multiple chroots to separate mountpoints, one ought to run one master process per chrooted fpm pool and not be dependant on the validate-settings, as inode is not a sufficient separator and some dysfunctionality remains.

wbob:
I wouldn't put convenience over correctness, if somebody chroots php envs, one has to read the docs and enable validate_root independently, other users are not affected. The `use_cwd` check though is a hard requirement and a default, if @niklasbuechner is up for it and agrees, merging that part would be helpful for those truly misconfigured environments.

niklasbuechner:
@shopwareBot I implemented the changes

shyim:

soebbing:
The call of `filenode('/')` throws an exception on systems that have a configured `open_basedir`,`/` might not be accessible in that setup.

While it is easy to simply catch the exception and move on, I am unsure on what such a setup would mean with regards to the `opcache.validate_root` check. Is the check obsolete in such cases?

niklasbuechner:
No, this is one of the scenarios in which the check is relevant. If we assume that a shared hoster uses `open_basedir` to restrict file access then `opcache.validate_root` will need to be checked in order to verify that the OPCache module is configured correctly.

soebbing:
Shopware 5.5.10 currently silently catches the exception that get's created to not crash the installation on such systems. But the `opcache.validate_root` check get's omitted currently in such cases.

Maybe you could create a new PR, fixing the check in such cases?

recovery/install/src/Requirements.php
```diff
@@ -217,6 +224,44 @@ private function checkModRewrite()
...
+            'group' => 'core',
+            'required' => 1,
+            'version' => ini_get('opcache.use_cwd'),
+            'result' => ini_get('opcache.use_cwd'),
+            'notice' => '',
+            'check' => $this->compare('opcache.use_cwd', ini_get('opcache.use_cwd'), 1),
+            'error' => '',
+        ]];
+
+        if (fileinode('/') > 2) {
```
amenk, line 248: This is very magic, what does this check do ? This seems to cause problems with openbasedir restrictions on some systems. is this necessary? see also https://issues.shopware.com/issues/NEXT-10564

shyim: You should consider using chrooted php-fpm. open_basedir introduces many performance issues cause of disabled realpath cache

amenk: Remark: I don't use open basedir, just tried try help a bit in the linked issue.

Still trying to understand this if statement out of curiosity.

And if open basedir is officially not supported it should go to the readme

shyim: We did modified the patch after import and had the same problem. But it should run into this empty catch https://github.com/shopware/shopware/blob/406925fbb6a8e98beb94e71ceca1fd473a6edc31/recovery/install/src/Requirements.php#L256

shyim: This has been wrong copied to Shopware 6. So I don't see an issue here :)

PR #2053, opened 2019-03-08
Added missing type cast for ordernumbers

JoshuaBehrens:
Hooking into `sOrder::sSaveOrder` and reading the return value gives you different types depending on the fact whether there are ordernumber filters present or not.

Adds a cast to match the return type declared in the phpdocs.

1. Have a subscriber:
```php
<?php declare (strict_types=1);
/// ...
    {
        return [
            'sOrder::sSaveOrder::after' => 'onSaveOrder',
        ];
    }

    public function onSaveOrder(\Enlight_Hook_HookArgs $args)
    {
        /** @var string $ordernumber */
        $ordernumber = $args->getReturn();
        $this->processOrdernumber($ordernumber);
    }

    public function processOrdernumber(string $ordernumber)
    {
    }
```
2. This will break at calling `processOrdernumber` as the variable was not a string as the number incrementor returns an int
3. Have any `Shopware_Modules_Order_GetOrdernumber_FilterOrdernumber` subscriber that returns the number as string
4. Now the call `processOrdernumber` does not break

### 4. Checklist

### 5. Bonus

The quickfix is the following subscriber:
```php
<?php
// ...
    {
        return [
            'Shopware_Modules_Order_GetOrdernumber_FilterOrdernumber' => 'makeOrdernumberOfTypeString',
        ];
    }

    public function makeOrdernumberOfTypeString(Enlight_Event_EventArgs $args)
    {
        return (string) $args->getReturn();
    }
```

### DIFF
--- a/engine/Shopware/Core/sOrder.php
+++ b/engine/Shopware/Core/sOrder.php
@@ -253,7 +253,7 @@ class sOrder
             ['subject' => $this]
         );
 
-        return $number;
+        return (string) $number;
     }
 
     /**


soebbing:
Hi @JoshuaBehrens,

since this might actually be a small break, we'll move the PR to the 5.6 branch internally.

JoshuaBehrens:
I didn't expect it to be breaking but I understand that I change types here. Thanks for taking this intense rebase on your side :grin:

shyim:

PR #2055, opened 2019-03-09
Normalized null type hints for ORM nullables

JoshuaBehrens:
I used the ORM via API resources for exporting orders. Reading data from the billing and shipping address properties failed because the phpdoc typehints said: you will get strings. I expected empty strings on empty inputs. I got null. It is ok to be null if you have a look at the Doctrine ORM hints and on the database, but the accessor methods do barely know.

Added `|null` to a lot of phpdoc typehints. AND one completely wrong phpdoc typehint in the voucher-model at the customer property/accessors.

Use the ORM and have expectations given by (phpdoc) typehints.

### 4. Checklist

### DIFF
--- a/engine/Shopware/Models/Article/Article.php
+++ b/engine/Shopware/Models/Article/Article.php
@@ -47,12 +47,10 @@ class Article extends ModelEntity
     /**
      * OWNING SIDE
      *
-     * @var \Shopware\Models\Tax\Tax
+     * @var \Shopware\Models\Tax\Tax|null
      *
      * @Assert\NotBlank()
      * @Assert\Valid()
-     *
-     * @var \Shopware\Models\Tax\Tax
      *
      * @ORM\OneToOne(targetEntity="Shopware\Models\Tax\Tax")
      * @ORM\JoinColumn(name="taxID", referencedColumnName="id")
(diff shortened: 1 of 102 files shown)


shyim:
Hey @JoshuaBehrens,
the Phpstan Check is failing. Can you fix it?

JoshuaBehrens:
I fixed all phpstan errors that I could find in my files.

JoshuaBehrens:
I thought I had run it. But yeah sure.

JoshuaBehrens:
I guess I am not aware of every business logic to apply this change you suggest. The examples you listed up make sense to me but I need to know all of these rules.

My latest thoughts about it are modifying the getters but keeping the setters. So one is not fooled regarding the return values which represent the possible states on the database and the setters instead match the business logic so you cannot set needed data to null. How about that?

What about making these columns non-nullable? Too much BC?

JoshuaBehrens:
Rebase is out

JoshuaBehrens:
I additionally changed the type of `Emotion\Emotion::device` from int to string.

JoshuaBehrens:
I am currently on it

shyim:

PR #2056, opened 2019-03-11
Allow any attribute name in Risk Management

Gugiman:
Since 5.2 attributes can be added to database easily with any name.
Since today the risk management module only allows attributes with the old static naming like "attr{number}"

It allows any kind of attribute name

Create an article attribute with any name and use it in the old format on attribute rules in risk management.

none

none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -1910,36 +1910,43 @@ class sAdmin
     {
         if (!empty($order['content'])) {
             $value = explode('|', $value);
-            if (!empty($value[0]) && isset($value[1])) {
-                $number = (int) str_ireplace('attr', '', $value[0]);
 
+            $crudService = Shopware()->Container()->get('shopware_attribute.crud_service');
+            $columnData = $crudService->get('s_articles_attributes', $value[0]);
+
+            if ($columnData === null && is_numeric($value[0])) {
+                $columnData = $crudService->get('s_articles_attributes', 'attr' . $value[0]);
+                $value[0] = 'attr' . $value[0];
+            }
+
+            if ($columnData !== null && !empty($value[0]) && isset($value[1])) {
                 $sqlProductOrderNumber = $this->connection->createQueryBuilder()
-                   ->select(['s_articles_attributes.id'])
-                   ->from('s_order_basket, s_articles_attributes, s_articles_details')
-                   ->where('s_order_basket.sessionID = :sessionID')
-                   ->andWhere('s_order_basket.modus = 0')
-                   ->andWhere('s_order_basket.ordernumber = s_articles_details.ordernumber')
-                   ->andWhere('s_articles_details.id = s_articles_attributes.articledetailsID')
-                   ->andWhere('s_articles_attributes.attr' . $number . ' = :attrValue')
-                   ->setParameters([
-                       'attrValue' => $value[1],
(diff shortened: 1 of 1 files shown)


engine/Shopware/Core/sAdmin.php
```diff
@@ -1910,8 +1910,11 @@ public function sRiskATTRIS($user, $order, $value)
     {
         if (!empty($order['content'])) {
             $value = explode('|', $value);
-            if (!empty($value[0]) && isset($value[1])) {
-                $number = (int) str_ireplace('attr', '', $value[0]);
+            
+            $entityManager = Shopware()->Container()->get('models');
+            $attributesMeta = $entityManager->getClassMetadata("\Shopware\Models\Attribute\Article");
```
hlohaus: use curd servcie. In orm the attribute fields CamelCase.

shyim: @Gugiman can you use CrudService instead ORM Models?

Gugiman: Done

engine/Shopware/Core/sAdmin.php
```diff
@@ -1910,8 +1910,11 @@ public function sRiskATTRIS($user, $order, $value)
     {
         if (!empty($order['content'])) {
             $value = explode('|', $value);
-            if (!empty($value[0]) && isset($value[1])) {
-                $number = (int) str_ireplace('attr', '', $value[0]);
+            
+            $crudService = Shopware()->Container()->get('shopware_attribute.crud_service');
+            $columnData = $crudService->get('s_articles_attributes',$value[0]);
```
soebbing: The Codestyle checker will fail here (and on the same line below) due to the missing space after the comma.

soebbing (dismissed):

engine/Shopware/Core/sAdmin.php
Gugiman: fixed
thanks for noticing

engine/Shopware/Core/sAdmin.php
```diff
@@ -1910,8 +1910,11 @@ public function sRiskATTRIS($user, $order, $value)
     {
         if (!empty($order['content'])) {
             $value = explode('|', $value);
-            if (!empty($value[0]) && isset($value[1])) {
-                $number = (int) str_ireplace('attr', '', $value[0]);
+            
+            $crudService = Shopware()->Container()->get('shopware_attribute.crud_service');
+            $columnData = $crudService->get('s_articles_attributes', $value[0]);
```
hlohaus, line 1915: getList please

Gugiman: Why not using the get() method directly when it is already using the getList() method to iterate its result for searching for a specific column?

When using the getList() method here, I would copy&paste most of the get() method to check if the columname really exists.

hlohaus: getList returns the columns, see https://github.com/shopware/shopware/pull/2034/files#diff-4d3ec247b97b8f7d17712b7b36537992R2535

Gugiman:
Does the squash look right?
Tbh, I'm not sure.

Gugiman:
This looks better now :)

Gugiman:
Maybe anykind of progress so far?

shyim:

PR #2057, opened 2019-03-11
Refreshes plugin manager before install and update plugin commands

jkrzefski:
Because I always forget to do it when I want to use `sw:plugin:update`.

Refreshes the plugins before `sw:plugin:install`, `sw:plugin:reinstall` and `sw:plugin:update` commands.

1. Run `sw:plugin:update` without running `sw:plugin:refresh` first.
2. Wonder why nothing happens.
3. Remember that you must first run `sw:plugin:refresh`.

None.

The commands `sw:plugin:install`, `sw:plugin:reinstall` and `sw:plugin:update` now feature a new option called `no-refresh`. If it exists, the plugin list will not be refreshed.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Commands/PluginInstallCommand.php
+++ b/engine/Shopware/Commands/PluginInstallCommand.php
@@ -53,6 +53,12 @@ class PluginInstallCommand extends PluginCommand
                 InputOption::VALUE_NONE,
                 'Activate plugin after intallation.'
             )
+            ->addOption(
+                'no-refresh',
+                null,
+                InputOption::VALUE_NONE,
+                'Do not refresh plugin list.'
+            )
             ->setHelp(<<<'EOF'
 The <info>%command.name%</info> installs a plugin.
 EOF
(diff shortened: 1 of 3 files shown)


engine/Shopware/Commands/PluginInstallCommand.php
```diff
@@ -66,6 +72,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         /** @var InstallerService $pluginManager */
         $pluginManager = $this->container->get('shopware_plugininstaller.plugin_manager');
+        if (!$input->getOption('no-refresh')) {
+            $pluginManager->refreshPluginList();
+            $output->writeln(sprintf('Successfully refreshed'));
```
mitelg: sprintf for a normal string? 🤔

jkrzefski: @mitelg Oh, I'm sorry. I took it from here: https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Commands/PluginRefreshCommand.php#L58

jkrzefski: I removed the sprintf.

PR #1628, opened 2018-04-10
Added clearing of old note entries

shyim:
### 1. Why is this change necessary?
There are currently many old expired notes, which cant be used anymore

### 2. What does this change do, exactly?
Removes expired notes

### 4. Please link to the relevant issues (if any).
https://issues.shopware.com/issues/SW-21514

### 6. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have squashed any insignificant commits
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I have read the contribution requirements and fulfil them.

PR #2060, opened 2019-03-12
update userId in notes on login and NotePostDispatch

tinect:
With https://github.com/shopware/shopware/commit/bf254f2b1a080bc8014860efd413d0092d069a66 notes without userId will be deleted. Unless the userId will not be updated, the Clearing will unwanted delete to many notes. We have to update userID, so used notes won't be deleted.

add updateUserID ~~in NotePostDispatch and~~ on Login

- don't login in Shop (on default you'll check you're not loggedin in checkout latest :-) )
- add product to note
- login
- you will your note, but...
- (without this change) userId will stay 0, so it will be deleted
- (with this change) userId was updated to yours, so it won't be deleted

https://github.com/shopware/shopware/pull/1628

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -3216,9 +3216,10 @@ class sAdmin implements \Enlight_Hook
             $newHash = $this->passwordEncoder->reencodePassword($plaintext, $hash, $encoderName);
         }
 
+        $userId = (int) $getUser['id'];
+
         if (!empty($newHash) && $newHash !== $hash) {
             $hash = $newHash;
-            $userId = (int) $getUser['id'];
             $this->db->update(
                 's_user',
                 [
@@ -3229,9 +3230,23 @@ class sAdmin implements \Enlight_Hook
             );
         }
 
+        // Update note userID
+        $uniqueId = $this->front->Request()->getCookie('sUniqueID');
+        if (!empty($uniqueId)) {
+            $this->connection->executeQuery(
+                'UPDATE s_order_notes SET userID = :userId, sUniqueID = NULL WHERE sUniqueID = :uniqueId AND userID = 0',
+                [
+                    'userId' => $userId,
+                    'uniqueId' => $uniqueId,
+                ]
+            );
+            //destroy cookie
+            $this->front->Response()->setCookie('sUniqueID');
+        }
+
         $this->session->offsetSet('sUserMail', $email);
         $this->session->offsetSet('sUserPassword', $hash);
-        $this->session->offsetSet('sUserId', $getUser['id']);
+        $this->session->offsetSet('sUserId', $userId);
 
         if (!$this->sCheckUser()) {
             return;


engine/Shopware/Core/sBasket.php
```diff
@@ -1426,24 +1426,37 @@ public function sAddNote($articleID, $articleName, $articleOrderNumber)
...
 
         if (!empty($cookieData) && empty($uniqueId)) {
-            $uniqueId = Random::getAlphanumericString(32);
-            $this->front->Response()->setCookie('sUniqueID', $uniqueId, time() + (86400 * 360), '/');
+            /*
+             * if there is a user, we don't need UniqueId
+             */
+            if ($userId) {
+                $uniqueId = '';
+                $this->front->Response()->setCookie('sUniqueID', $uniqueId, strtotime('-1 year'), '/');
```
shopwareBot: Can you inline the variable $uniqueId and replace strtotime with 0

engine/Shopware/Core/sAdmin.php
```diff
@@ -676,7 +676,8 @@ public function logout()
     {
         $this->moduleManager->Basket()->clearBasket();
 
-        Shopware()->Session()->unsetAll();
+        $this->session->offsetUnset('sUserId');
```
shopwareBot: Why explicit unsetting sUserId when in the next line everything is unsetted

engine/Shopware/Controllers/Frontend/Note.php
```diff
@@ -35,7 +35,21 @@ public function preDispatch()
 
     public function postDispatch()
     {
-        Shopware()->Session()->sNotesQuantity = Shopware()->Modules()->Basket()->sCountNotes();
+        $session = $this->get('session');
+
+        $session->offsetSet('sNotesQuantity', Shopware()->Modules()->Basket()->sCountNotes());
```
shopwareBot: Don't use global Shopware() function

tinect:
All the boss wants :-)

shyim:
Hey,

Sorry for the late response :see_no_evil: .
I have discussed your Change with our internal QA. We decided that only on login the notes should be migrated and removed from cookie. So its **only** associated to the user account.

We want to prevent cases like

* Customer fills product in note
* Customer logins (getting migrated)
* Logout
* Login in another account, but you see notes from customer 1 and **can manage them**

engine/Shopware/Core/sAdmin.php
```diff
@@ -3231,9 +3232,20 @@ protected function loginUser($getUser, $email, $password, $isPreHashed, $encoder
             );
         }
 
+        // Update note userID
+        $uniqueId = $this->front->Request()->getCookie('sUniqueID');
+        if (!empty($uniqueId)) {
+            $this->connection->executeQuery(
+                'UPDATE s_order_notes SET userID = :userId WHERE sUniqueID = :uniqueId AND userID = 0',
```
PaddyS: Set `UniqueID` on `NULL` in this case.
Also, remove the `sUniqueID` cookie afterwards.

shyim:

PR #2063, opened 2019-03-19
assign filesize in product-downloads

tinect:
currently the size of downloads is 0. it won't be set by shopware. Maybe we could remove the field from database?!

get filesize from mediaService

https://issues.shopware.com/issues/SW-23682

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/DownloadGateway.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/DownloadGateway.php
@@ -97,6 +97,7 @@ class DownloadGateway implements Gateway\DownloadGatewayInterface
 
         $query->from('s_articles_downloads', 'download')
             ->leftJoin('download', 's_articles_downloads_attributes', 'downloadAttribute', 'downloadAttribute.downloadID = download.id')
+            ->innerJoin('download', 's_media', 'media', 'media.path = download.filename')
             ->where('download.articleID IN (:ids)')
             ->setParameter(':ids', $ids, Connection::PARAM_INT_ARRAY);
 
--- a/engine/Shopware/Models/Article/Download.php
+++ b/engine/Shopware/Models/Article/Download.php
@@ -86,6 +86,9 @@ class Download extends ModelEntity
     /**
      * @var float
      *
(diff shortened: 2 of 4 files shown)


engine/Shopware/Components/Compatibility/LegacyStructConverter.php
```diff
@@ -476,7 +476,7 @@ public function convertProductStruct(StoreFrontBundle\Struct\Product $product)
                 'id' => $download->getId(),
                 'description' => $download->getDescription(),
                 'filename' => $this->mediaService->getUrl($download->getFile()),
-                'size' => $download->getSize(),
+                'size' => $this->mediaService->getSize($download->getFile()),
```
shyim: Please fill the right size in $download instead. Reading file sizes on the fly makes the frontend slower.

engine/Shopware/Models/Article/Download.php
```diff
@@ -178,6 +179,8 @@ public function getFile()
      * @param float $size
      *
      * @return Download
+     *
+     * @deprecated will be removed in 5.7
```
shyim: Please remove the size usages also in extjs files

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/DownloadGateway.php
```diff
@@ -96,6 +96,7 @@ public function getList($products, Struct\ShopContextInterface $context)
         $query->select($this->fieldHelper->getDownloadFields());
 
         $query->from('s_articles_downloads', 'download')
+            ->leftJoin('download', 's_media', 'media', 'media.path = download.filename')
```
shyim: Is here any reason why a leftJoin should be used instead of an innerJoin?

engine/Shopware/Models/Article/Download.php
```diff
@@ -86,6 +86,7 @@ class Download extends ModelEntity
     /**
      * @var float
      *
+     * @deprecated will be removed in 5.7
```
shyim: Please add to the deprecation since when its deprecated and an alternative

mitelg:
thanks for your contribution @tinect 👍🎉💙

PR #2068, opened 2019-03-25
Fixed SearchIndexer to not delete keywords unintentionally

csauthoff:
For shops with few articles, but many variants, keywords would be deleted unintentionally. Had a case with a client, where there were only a few articles, but highly configurable. Search for special keywords always failed, until I found this issue.

While populating the search index, keywords are deleted if they match on more than 90% of the number of articles (SQL COUNT s_articles). But when you do heavy use of variants, s_articles_details count would be preferable over s_articles, as potentially it could be much higher.

See 2.

### 6. Checklist
* [ ]  I have written tests and verified that they fail without my change
* [x]  I have squashed any insignificant commits
* [ ]  This change has comments for package types, values, functions, and non-obvious lines of code
* [x]  I have read the contribution requirements and fulfil them.

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/SearchTerm/SearchIndexer.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/SearchTerm/SearchIndexer.php
@@ -284,7 +284,7 @@ class SearchIndexer implements SearchIndexerInterface
             $sql_join
 
             GROUP BY keywordID, fieldID
-            HAVING COUNT(*) > (SELECT COUNT(*)*0.9 FROM `s_articles`)
+            HAVING COUNT(*) > (SELECT COUNT(*)*0.9 FROM `s_articles_details`)
         ";
 
         $collectToDelete = $this->connection->fetchAll($sql);
--- a/engine/Shopware/Bundle/SearchBundleDBAL/SearchTerm/SearchIndexer.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/SearchTerm/SearchIndexer.php
@@ -37,7 +37,7 @@ class SearchIndexer implements SearchIndexerInterface
     /**
      * Percent of matches to delete search index keywords
      *
-     * @type float
+     * @var float
      */
     const INDEX_DELETE_THRESHOLD = 0.9;
 
@@ -88,7 +88,7 @@ class SearchIndexer implements SearchIndexerInterface
             return;
         }
 
-        $interval = (int)$this->config->get('cacheSearch');
+        $interval = (int) $this->config->get('cacheSearch');
 
         if (empty($interval) || $interval < 360) {
             $interval = 86400;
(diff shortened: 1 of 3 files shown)


soebbing:
Hi @csauthoff, thank you very much for your contribution!

The basic idea of that check is: If some keyword matches nearly EVERY product in the shop, then there is no value of said keyword to describe a specific product, so the keyword doesn't need to be indexed.

I am not sure that changing the threshold from 90% of all main products to 90% of all variants is reasonable: the keywords themselves come from the main variants. As far as I am aware there are no variant specific keywords (other than the ID) being indexed. Therefore the only effect of this change would be to increase the threshold of 90% and the keywords currently being dropped would again match (nearly) all products - which leads to a bad searching experience and a reduction in performance.

But if an increase in the percentage (from the default 90% to something above that) helps in your use case, maybe it would be a good idea to make this 90% threshold configurable? What do you think?

csauthoff:
Hi @soebbing,

I totally get the point but let me further describe the issue I had:
As said, in my case, there were only a few articles with many variants. We added an index on s_articles_details.suppliernumber (via backend configuration), which always stayed the same for all variants of a single article. 

Let's suppose the following article configuration:

```
Article A:
	- Variant 1 (suppliernumber: foo)
	- Variant 2 (suppliernumber: foo)
	- Variant 3 (suppliernumber: foo)
	- Variant 4 (suppliernumber: foo)
	- Variant 5 (suppliernumber: foo)
	- Variant 6 (suppliernumber: foo)

Article B:
	- Variant 1 (suppliernumber: bar)
	- Variant 2 (suppliernumber: bar)

Article C:
	- Variant 1 (suppliernumber: hello)
	- Variant 2 (suppliernumber: hello)	
```

If you now search for "foo", it'll match 6 times (for all variants of article "A"), which is more than 90% of the number of articles (3*.9=2.7). Thus, it's never possible to find these articles.
If you'd search for "bar" or "hello" though, it's fine, as both would match two times, which is less than 90% of 3 articles. 
This should clarify the issue, please correct me if i'm wrong.

So probably the most accurate solution would be to make the COUNT condition dependent of the referenced table. Please see my last commit for a first proposal (not thoroughly tested though).

soebbing:
I think your idea is great! If you add some test cases for your change, then we can merge it very soon!

Bonus points if you manage to change the implementation to a `QueryBuilder` :)

csauthoff:
Please see my last two commits for an implementation of QueryBuilder and a simple test case for the SearchIndexer :)

csauthoff:
Si claro, checks pass now!

shyim:
Hey,

csauthoff:
CLA signed!

pascalheidmann:
@csauthoff looks like this branch is a bit out of date as it is targeted at 5.5 while Shopware is at 5.6.
You might want to rebase onto 5.6 and force push the changes.

shyim:
Sorry for the late response.
Thanks for your contribution! :blue_heart:

PR #2069, opened 2019-03-27
sw:plugin:install Activate plugin although it is already installed

JoshuaBehrens:
When you run `sw:plugin:install --activate ...` and the plugin is already installed it won't get activated.

Don't stop after checking the installation state of the plugin and continue with optional activation.

1. `sw:plugin:install ...`
2. Wait and forget you installed it (optional)
3. `sw:plugin:install --activate ...`
4. Use things from the freshly installed and activated plugin
5. Wait wat? :confused: 
6. Why is my service not loaded?
7. Random coworker: did you activate your plugin and clear the cache?
    Me: Sure I did, see :point_right: `sw:plugin:list | grep ...`
    Random coworker: no you did not lawl

### 4. Checklist

### DIFF
--- a/engine/Shopware/Commands/PluginInstallCommand.php
+++ b/engine/Shopware/Commands/PluginInstallCommand.php
@@ -81,14 +81,14 @@ EOF
             return 1;
         }
 
+        $installationContext = null;
+
         if ($plugin->getInstalled()) {
             $output->writeln(sprintf('The plugin %s is already installed.', $pluginName));
-
-            return 1;
+        } else {
+            $installationContext = $pluginManager->installPlugin($plugin);
+            $output->writeln(sprintf('Plugin %s has been installed successfully.', $pluginName));
         }
-
-        $installationContext = $pluginManager->installPlugin($plugin);
-        $output->writeln(sprintf('Plugin %s has been installed successfully.', $pluginName));
 
         $activationContext = null;
 


engine/Shopware/Commands/PluginInstallCommand.php
```diff
@@ -83,13 +83,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         if ($plugin->getInstalled()) {
             $output->writeln(sprintf('The plugin %s is already installed.', $pluginName));
-
-            return 1;
+        } else {
+            $installationContext = $pluginManager->installPlugin($plugin);
```
mitelg, line 89: why do you want to execute the install process again? I think this could be critical. instead of removing the return. you could add the activation logic here

JoshuaBehrens: What do you mean with again? I moved the installation part into an else block. There should be no duplicate process

mitelg: lol, my fault :smile: 
you are right. everything totally fine :blush:

JoshuaBehrens:
Uhm, why is it tagged as incomplete?

soebbing:
Uh, sorry, there should have been a comment.

It is incomplete since the tests fail due to a problem with the `$installContext` - it is only available when the plugin was installed and is needed in the clearCache function below.

shyim:
Hey @JoshuaBehrens,

PR #2070, opened 2019-03-27
Improved variant listing (Pagesize, selection delete & copy data)

pascalheidmann-bedarf:
The variant listing leave quite a few things desired like changing the variant count as well as an option to apply changes to a selection.

This PR contains 3 commits:
The 1st one is adding a combobox for selecting the `pageSize` to control the visible variant count in the grid.
The 2nd one improves on the selection model to add checkboxes in a separate column as well as adding a dedicated `Delete selection` button. Both the logic for the delete button as well as the button for copying article data are already present and capable of using the SelectionModel to limit their actions to only the selected variants (but haven't used it yet).
And the 3rd finally uses hooks and crosses instead of "false" and "true" string values for both bool columns while also changing their width to a more appropriate width.

### 6. Checklist


shyim:

pascalheidmann-bedarf:
@shyim @shopwareBot @soebbing while this PR is officially part of 5.6.0 at least the changes to the config forms aren't in it 😕 Maybe something hasn't worked when picking the changes?

shyim:
I will lookup Friday

shyim:
Did you mean the changes with the pagesize?

pascalheidmann-bedarf:
Yes @shyim

shyim:
It has been removed with https://github.com/shopware/shopware/commit/af2b6771ac90315569e6d6b67f753ab1acc0c0d4
https://issues.shopware.com/issues/SW-24255
QA was against this, cause the UI has been broken

PR #2071, opened 2019-03-28
Check plugin validation before activation

pascalheidmann-bedarf:
There is no check on plugin activation if the conditions, especially `requiredPlugins`, are still met. 

Assert plugin conditions.

1. Have 2 Plugins where one is dependent on the other one.
2. Install both but don't activate yet.
3. Uninstall the base plugin
4. Activate dependent plugin

If the dependent plugin decorates a symfony service of the parent plugin you are completely f**k'd as now all calls try to instantiate the non existent service and therefore crash.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
@@ -247,9 +247,10 @@ class PluginInstaller
      */
     public function activatePlugin(Plugin $plugin)
     {
-        $context = new ActivateContext($plugin, $this->release->getVersion(), $plugin->getVersion());
-
         $bootstrap = $this->getPluginByName($plugin->getName());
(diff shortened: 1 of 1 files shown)


engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginInstaller.php
```diff
@@ -247,6 +247,9 @@ public function updatePlugin(Plugin $plugin)
      */
     public function activatePlugin(Plugin $plugin)
     {
+        $pluginBootstrap = $this->getPluginByName($plugin->getName());
+        $this->requirementValidator->validate($pluginBootstrap->getPath() . '/plugin.xml', $this->release->getVersion());
+
         $context = new ActivateContext($plugin, $this->release->getVersion(), $plugin->getVersion());
 
         $bootstrap = $this->getPluginByName($plugin->getName());
```
shopwareBot: Can you move this line to start of this function and remove your $pluginBootstrap?

shyim:

PR #2072, opened 2019-03-28
Order backend property list by property position

tophcodes:
In the frontend properties are displayed in the order that is specified within Items -> Properties in the backend. I would expect the properties to be ordered the same way in the item's "Properties" tab, but instead those are ordered by creation date (or ID) of the original properties within Items -> Properties.

It adds an `orderBy` statement to the query that loads an item's properties from the database. That way instead of the database infering the insert order as the result order, the `position` column is used.

1. Create a handful of property options for a group in Items -> Properties
2. Reorder these properties
3. Edit the properties of an item and select multiple options of the previously reordered ones
4. Save your changes
5. Observe that the order of properties within the preview changes based on the original creation order of the properties

None

None

### 6. Checklist


shyim:

PR #2073, opened 2019-03-29
hide non-sense informations in ConfigCache while Redis

tinect:
When Redis is set, information of cacheDir will be shown.

Show informations of redis 
image

Set Redis, see Caches/Performance showing sizes of default CacheDir

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/CacheManager.php
+++ b/engine/Shopware/Components/CacheManager.php
@@ -276,17 +276,29 @@ class CacheManager
      */
     public function getConfigCacheInfo()
     {
-        $cacheConfig = $this->container->getParameter('shopware.cache');
-        $dir = null;
+        $backendCache = $this->cache->getBackend();
 
-        if ($this->cache->getBackend() instanceof \Zend_Cache_Backend_Apcu) {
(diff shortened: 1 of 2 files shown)


engine/Shopware/Components/CacheManager.php
```diff
@@ -286,6 +286,13 @@ public function getConfigCacheInfo()
             $info['size'] = $this->encodeSize($apcInfo['mem_size']);
             $apcInfo = apcu_sma_info();
             $info['freeSpace'] = $this->encodeSize($apcInfo['avail_mem']);
+        } elseif ($this->cache->getBackend() instanceof \Zend_Cache_Backend_Redis) {
+            $info = [];
+            /** @var \Doctrine\Common\Cache\RedisCache $cache */
+            $cache = $this->container->get('shopware.model_config')->getMetadataCacheImpl();
```
shyim: You have mixed up Zend_Cache and Doctrine Cache
$this->cache is not the same as Doctrine Cache.
You can configure Zend_Cache to use Redis, but Doctrine could use Apcu. These caches can be configured separate
See more here https://developers.shopware.com/developers-guide/shopware-config/#redis-configuration

shyim:

PR #2074, opened 2019-04-01
Fix sMinimumCharge with currency Factor fixes SW-16586

aherking:
- to display the current minimum charge in snippet when using a currency factor

https://issues.shopware.com/issues/SW-16586
### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -835,7 +835,11 @@ SQL;
             );
             $smarty = Shopware()->Container()->get('template');
             $template = $smarty->createTemplate(sprintf('string:%s', $snippet));
-            $template->assign('sMinimumCharge', $voucherDetails['minimumcharge']);
+            if ($factor !== 0) {	    
+            	$template->assign('sMinimumCharge', $voucherDetails['minimumcharge'] * $factor);
+            } else {
+            	$template->assign('sMinimumCharge', $voucherDetails['minimumcharge']);
+            }
 
             $sErrorMessages[] = $template->fetch();
 


engine/Shopware/Core/sBasket.php
```diff
@@ -835,7 +835,11 @@ public function sAddVoucher($voucherCode, $basket = '')
             );
             $smarty = Shopware()->Container()->get('template');
             $template = $smarty->createTemplate(sprintf('string:%s', $snippet));
-            $template->assign('sMinimumCharge', $voucherDetails['minimumcharge']);
+            if ($factor !== 0) {
```
jkrzefski, line 838: I think `$factor` will never be `0` because of the [condition up here](https://github.com/aherking/shopware/blob/233c27757cb9ab36927231323185c8d1d89e9ed1/engine/Shopware/Core/sBasket.php#L819) so you do not have to check for `0` and can always multiply by `$factor`. I also think, [this condition](https://github.com/aherking/shopware/blob/233c27757cb9ab36927231323185c8d1d89e9ed1/engine/Shopware/Core/sBasket.php#L827) is unnecessary, but this could be a separate pr.

mitelg: @jkrzefski you see that the PR was closed? :smile:

jkrzefski: @mitelg Yes, I had this open in a new tab for a while and had wanted to send this comment. Then I had something else to do and came back to this tab afterwards. So I only saw that it was closed after I sent my comment. I was too lazy to delete it again. Anyway: Do you agree with me that [this condition](https://github.com/aherking/shopware/blob/233c27757cb9ab36927231323185c8d1d89e9ed1/engine/Shopware/Core/sBasket.php#L827) is obsolete?

mitelg: @jkrzefski yes sir, indeed

PR #2075, opened 2019-04-01
Removes obsolete condition for currency factor

jkrzefski:
It isn't. Just cleans up some obsolete code.

Removes a condition that checks the variable `$factor` for `0`. But in [this condition](https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Core/sBasket.php#L819) it is already checked and if it is `0` it falls back to `1`. Therefore it can never be `0` here.

None.

None.

None.

### 6. Checklist

- [x] I have written tests ~~and verified that they fail without my change~~

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -816,17 +816,13 @@ SQL;
         }
 
         // Including currency factor
+        $factor = 1;
         if ($this->sSYSTEM->sCurrency['factor'] && empty($voucherDetails['percental'])) {
             $factor = $this->sSYSTEM->sCurrency['factor'];
             $voucherDetails['value'] *= $factor;
-        } else {
-            $factor = 1;
         }
 
-        $basketValue = 0;
-        if ($factor !== 0) {
-            $basketValue = $amount['totalAmount'] / $factor;
-        }
+        $basketValue = $amount['totalAmount'] / $factor;
         // Check if the basket's value is above the voucher's
         if ($basketValue < $voucherDetails['minimumcharge']) {
             $snippet = $this->snippetManager->getNamespace('frontend/basket/internalMessages')->get(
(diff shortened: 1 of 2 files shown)


engine/Shopware/Core/sBasket.php
```diff
@@ -824,9 +824,7 @@ public function sAddVoucher($voucherCode, $basket = '')
         }
```
mitelg, line 823: `$factor = 1` should be initialised before the if :wink:

jkrzefski: Why should it? In both branches of the condition `$factor` is set. So it is safe to assume it was initialized.

uehler: I think @mitelg means
$factor=1
if(condition){
$factor = xy
}

so you don't need the else statement

engine/Shopware/Core/sBasket.php
```diff
@@ -824,9 +824,7 @@ public function sAddVoucher($voucherCode, $basket = '')
...
 
         $basketValue = 0;
```
uehler: you can remove this line because you override it in the next line

jkrzefski: Good catch. Will do.

engine/Shopware/Core/sBasket.php
jkrzefski: Ok, I can understand that. I'll do that.

jkrzefski:
@shopwareBot I will take a look at it soon.

shyim:
Hey @jkrzefski,
are you still on it?

jkrzefski:
@shyim Yeah, I was a bit busy the last week. But I'll look into this over the weekend.

jkrzefski:
I have added a test for `sBasket::sAddVoucher()` with different currency factors. I test it with `0`, `.5`, `1`, `1.5` and `2`. Let me know, if this is a sufficient implementation.

shyim:
Thanks for your Contribution!

PR #2076, opened 2019-04-03
changed initial type according to other class properties

benboit:
Under some circumstances index with bin/console sw:es:index:populate was not possible anymore. 

Uncaught TypeError: Argument 1 passed to Shopware\Bundle\ESIndexingBundle\Product\ProductListingVariationLoader::createSplitting() must be of the type array, null given

Full product configurations being null should not lead to the whole process aborting.

It changes the initial type of the class property to the expected type. In that way not populated properties wont make the process crash.

There seemed to be articles having no valid variant-configuration. This may happen.

none

none

### 6. Checklist


shyim:

PR #2077, opened 2019-04-04
Adds support for third party authentication adapters

jkrzefski:
If a plugin wants to add an authentication adapter, it would have no effect. This is because when iterating over all given adapters and trying to authenticate the user with them, the routine will always use the first adapter in the set of adapters.

This change will use the actual adapter of the current iteration instead of always using the first one.

1. Try to add an authentication adapter.
2. Wonder why it does not work.
3. Find out, you never had a chance to begin with.
4. Rage for a few minutes. (optional)
5. Replace the default adapter entirely.

None.

Maybe one could add a tutorial in the documentation on how to add a custom authentication adapter.

### 6. Checklist

- [x] I have written tests ~~and verified that they fail without my change~~

### DIFF
--- a/tests/Functional/Components/AuthTest.php
+++ b/tests/Functional/Components/AuthTest.php
@@ -1,0 +1,137 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+use Shopware\Components\Password\Encoder\PasswordEncoderInterface;
+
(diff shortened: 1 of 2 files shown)


JoshuaBehrens:
Great now that you find it I can finally add our new login.

jkrzefski:
@shopwareBot You're right. The logic is for some reason duplicated here. I will put some effort into researching why this is the case and hopefully I will come up with a simplification and tests.

buddhaCode:
Damn it. Wrote a SSO integration for sw4 in those days. What a pain!

jkrzefski:
@buddhaCode Maybe you can provide some more knowledge on this topic then (and save me some research time). Every bit of information about the authentication component could be helpful.

jkrzefski:
@shopwareBot Yeah, I was a bit busy the last week. But I'll look into this over the weekend.

jkrzefski:
@shopwareBot I have simplified the `authenticate` method and added some tests for it. Let me know, if this looks good to you.

tests/Functional/Components/AuthTest.php
```diff
@@ -0,0 +1,130 @@
...
+        // Authenticate with correct password; should succeed
+        $adapter->setIdentity($username);
+        $adapter->setCredential($password);
+        static::assertTrue($this->auth->authenticate($adapter)->isValid());
+        $user = $this->auth->getStorage()->read();
+        static::assertInstanceOf('stdClass', $user);
+        static::assertEquals($username, $user->username);
+
+        // Housekeeping
+        $this->db->delete(
```
shyim: Maybe use a transaction instead

jkrzefski: @shyim You mean I should begin a transaction, create the admin user and then roll it back? Will that still work with the `sw:admin:create` command or will I have to implement that myself?

shyim: Should work also there :)
Maybe you could also insert a new row to s_core_auth easily

jkrzefski: @shyim I added transactions and a rollback. I also changed the command execution to a simple insert statement.

jkrzefski:

shyim:

PR #2078, opened 2019-04-05
Fixes wrong Display/Preselection of Shop in Customer Detail View (Backend)

aherking:

When creating/updating customers with REST-API you are only able to set shopID, so the "language" will stay default. In customer list the correct Shop is shown, but when opening the detail of a customer there will be the languageId preselected.
customer1
customer3
customer2

Show Shop instead of language

see Screenshots

### 6. Checklist


hlohaus:
The language depends on subshop and subshopID an the main shop...
It will also not work for frontend user.

mnaczenski:
The field description might be misleading, but the selectbox shows the correct values. You can change the language in the customer details, not the shop. By changing this to the shopID the language would be changed on save, since the "shopID" in the details maps to "language" in the database.

Current behavior:
Auswahl_024

The correct way to fix this would be

- Rename the snippet to "Language" instead of "Shop", because this is what this field does
- Add a new field for the selection of the shop

aherking:
Done

shyim:

PR #2079, opened 2019-04-05
Remove deletion of Mpdf Fonts

Linusten:

Removing fonts is not necessary

It preserves the Fonts of Mpdf

Simply DO NOT DELETE!

Ticket: SW-23844
https://issues.shopware.com/issues/SW-23844

### 6. Checklist

### DIFF
--- a/build/composer-post-install-cmd.sh
+++ b/build/composer-post-install-cmd.sh
@@ -7,7 +7,6 @@ root_dir="$script_dir"/..
 
 set -x
 
-rm -rf -- "$root_dir"/vendor/mpdf/mpdf/ttfonts
 rm -rf -- "$root_dir"/vendor/google/protobuf/src
 rm -rf -- "$root_dir"/vendor/google/protobuf/java
 rm -rf -- "$root_dir"/vendor/google/protobuf/objectivec


soebbing:
Hi @Linusten, thank you for your PR!

These fonts are deleted since they aren't used in Shopware and with around 87MB nearly double the size of the vendor folder (~113MB without these fonts).

Why would you want to keep them?

Linusten:
Hi @soebbing,
we need these Fonts because we want to publish a Plugin that creates Individual Coupons. 
Anyway many Plugins like the Kilb Configurator and other Coupon Plugins need these Fonts. 
Isnt it stupid to manipulate a Package which comes as it is, everybody depends on its state and you manipulate it? Should a developer check every Package to make sure that it is not manipulated?

soebbing:
So why don't you provide these fonts in your plugin if your plugin depends on it? 

I think it is more stupid to double the size of an installation for everybody when only very few actually use these fonts. That is the tradeoff we made here.

Linusten:
The Problem is that you have an Upload limit of 8MB for Plugins, we tried to Provide it but you do not allow it.

Linusten:
That is the tradeoff we made here. @soebbing

soebbing:
I see your problem, @Linusten and we'll discuss a solution internally.

I still don't think keeping the fonts is the way to go but at least we need to document the deletion better. And we need a workaround for those that really make use of these fonts.

jkrzefski:
@Linusten Maybe you can download the fonts during the installation of your plugin. Not an ideal solution, but it is a way to bypass the filesize limit for plugins.

Linusten:
@jkrzefski  Why do i have to create a bypass? Mpdf provides the fonts... It makes no sense for me to find a different way, just because mpdf was manipulated. It is a solution for sure, but it is always a hack. Lets find a good solution, i dont want to create more and more mess... The next one comes and downloads symfony just because shopware changed something in it.

soebbing:
Hi again @Linusten!

Here's what we'll do to mitigate your problem:
- We'll create a plugin containing the missing fonts on which you can depend in your own plugin. That way even older Shopware shops can make use of it. Relevant [Ticket](https://issues.shopware.com/issues/SW-23846).
- We'll document the deletion of fonts in the devdocs, explaining the reasoning and the workaround, incl. a link to the plugin.
- We'll modify the composer project for Shopware to also delete these fonts so that it behaves similar to a classic Shopware installation by default. Of course this can be reverted easily.
- In the long run we try to move away from mpdf to something lighter like dompdf so that we do not have to delete the fonts anymore

Thanks again for pointing your issue out to us. I'll close your PR since this fix is not viable for us at the moment.

PR #2080, opened 2019-04-07
Changed LIKE search for ordernumber to use the index on the table.

SieGeL2k16:
This dramatically decreases the search time and also corrects the placeholder
search of ordernumbers.

It improves the search time for ordernumbers with LIKE operator

Rewrites the SQL to improve search operation with LIKE. Also it fixes the search for parts of ordernumbers.

In very large shops (more than 2 Million articles) it took more than 6 seconds with original query. With rewritten SQL the search time drops down to 0.5 seconds.

No issue yet opened, the fix was need ASAP as our customer was not very happy with the search time.

### 6. Checklist


shyim:

PR #2083, opened 2019-04-10
Enable defining pdo options in config

mschop:

Definiting connection options for pdo is needed, to enable tls encrypted connection to the mysql database. In many cloud environments (like Microsoft Azure) it is recommended to encrypt the traffic between the app server and the database if the Database is managed by the hoster / cloud admins.

Otherwise the traffic is not protected again MITM-Attacks.

Additionally, shop owners can use this options to add additional pdo options that are not related to this.

Passing options from config to the PDO connection.

https://developers.shopware.com/developers-guide/shopware-config should be updated with the new config option.

### 6. Checklist

- [ ] I have written tests and verified that they fail without my change -> I don't know, how to test this. Any ideas? If I modify the db settings in the tests, other tests could fail because of this changes. I don't think, that this change is risky in any way, because it just passes some parameters to the PDO object initialization

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/Bridge/Db.php
+++ b/engine/Shopware/Components/DependencyInjection/Bridge/Db.php
@@ -49,7 +49,8 @@ class Db
             $conn = new \PDO(
                 'mysql:' . $connectionString,
                 $dbConfig['username'],
-                $password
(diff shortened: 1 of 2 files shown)


shyim:

PR #2087, opened 2019-04-16
[user-scalable="no"] wird nicht mehr benötigt

christopherholst:
Nach den aktuellen Audits von Lighthouse wird [user-scalable="no"] nicht mehr bewertet und soll entfernt werden, wir haben dies einmal getestet und es funktioniert alles noch wie gehabt und google meckert nicht mehr.

According to Lighthouse, Google no longer needs this parameter.

It removes a meanwhile unnecessary parameter for the viewport.

Carry out an audit in the Lighthouse, then look at the error under use.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/index/header.tpl
+++ b/themes/Frontend/Bare/frontend/index/header.tpl
@@ -40,7 +40,7 @@
     {/block}
 
     {block name='frontend_index_header_meta_tags_mobile'}
-        <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <meta name="mobile-web-app-capable" content="yes">
         <meta name="apple-mobile-web-app-title" content="{if $theme.appleWebAppTitle != ""}{$theme.appleWebAppTitle|escapeHtml}{else}{{config name=sShopname}|escapeHtml}{/if}">
         <meta name="apple-mobile-web-app-capable" content="yes">


tinect:
Hey @snHolst,
`es funktioniert alles noch wie gehabt` is wrong. With this change mobile-user can now scale/zoom page - as it says. Tested with different devices.

christopherholst:
Hey @tinect,
even if the parameter is set you can scale. So he is superfluous and causes nothing more.

tinect:
This would destroy behaviour on non Safari.
:apple: removed user-scalable in Safari

https://twitter.com/thomasfuchs/status/742531231007559680/photo/1

pascalheidmann-bedarf:
As this stupid tag is bad UX anyway by disabling the user's ability to use zoom it should be removed anyways.
There is a reason you can also disable this enforcement in mobile chrome in accessability settings (and I would recommend doing so)

tinect:
Why should there be a possibilty to zoom, if the content is correct? I see no need to remove this in shopware.

I see very bad ux by removing it, due to unneeded and unwanted zoomings while visiting and working with sites. 

Dobble tab in ipad or iphone in Detail page and see the shop looking weird.

pascalheidmann-bedarf:
@tinect http://adrianroselli.com/2015/10/dont-disable-zoom.html

I don't know what apple does when zooming (but as you can't disable it anyway anymore I don't care that much) but at least on Android it just zooms without reformatting (-> canvas size stays intact).
Not sure what you do that you accidentally trigger those as long as you don't use your device like my grandfather with 3 fingers at a time to swipe around ;)

tinect:
Hey @pascalheidman-bedarf 
I cannot find any productive and issue-related content in your reply in relation to my objections 🤔 

However, we need to get shopware's reply.

shyim:

PR #2088, opened 2019-04-16
Adds new command sw:tinker based on psy/psysh

jkrzefski:
We used to have an interactive shell while Shopware included Symfony 2. Now this interactive shell is gone because Symfony does no longer have one. With PsySH we can have a new interactive shell that is even more powerful than what we used to have. PsySH is a REPL shell and integrates great with Symfony based applications. For more information about it visit https://psysh.org/

1. Adds a new dependency: `psy/psysh`
2. Adds a new command: `sw:tinker`

Execute `sw:tinker` and have fun in an interactive REPL shell that can execute every present php console command while also being able to evaluate php code live and have full access to the dependency injection container and everything else you only have at runtime.

None.

Maybe the developer documentation should mention this as a way to tinker around in a running environment. Maybe not.

### 6. Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,7 +79,8 @@
         "fig/link-util": "1.0.0",
         "league/flysystem": "1.0.46",
         "league/flysystem-aws-s3-v3": "1.0.19",
-        "superbalist/flysystem-google-storage": "6.0.0"
+        "superbalist/flysystem-google-storage": "6.0.0",
+        "psy/psysh": "^0.9.9"
     },
     "suggest": {
         "ext-apcu": "*",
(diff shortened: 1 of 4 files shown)


JoshuaBehrens:
I want to point out the best command psysh has. `wtf`

jkrzefski:
Looks like the composer.lock is messed up. I'll take a look at it tomorrow.

jkrzefski:
I fixed the composer.lock file.

mitelg:
psy/psysh uses a library from good ol' @dnoegel :open_mouth: :+1:

PR #2089, opened 2019-04-18
Show inactive payments in risk management

JoshuaBehrens:
There is no easy way to configure risk management for inactive payment means. There are workarounds like enabling the payment mean but disabling the payment for specific for the shop countries or don't connect it to shipping costs but these workarounds can be easily forgotten and it will be a hassle to find out what the source of the problem is.

I changed a query to allow also inactive payment means and sort them by activity state. To see in what kind of state they are I added the same icons like in the payment window to see whether a payment mean is active or inactive.

1. Install a new payment mean for a specific type of customers
2. Try to limit the usage of this payment mean before activation
3. Cannot select the disabled payment mean
4. Be annoyed
5. Use a workaround
6. Be distracted
    be distracted
    Original comic: https://heeris.id.au/2013/this-is-why-you-shouldnt-interrupt-a-programmer/
7. Configure the active payment mean
8. Wonder why you cannot select the active payment mean in the storefront
9. Backtrace yourself, talk to the :duck: and undo the workaround

Maybe somewhere should be stated the meaning of the icons.

### 5. Checklist


shyim:

PR #2091, opened 2019-04-19
Prevents broken event listeners from being registered

jkrzefski:
When a plugin registers an event listener, it creates a database entry with a callback for that listener. If this callback will later be not callable anymore and the listened event is triggered, it will result in a fatal error. For some reason this only seems to affect the cli. A listener could become not callable if e.g. the plugin is removed from the system without using the plugin manager to uninstall it.

Before the event listeners are registered, it will now be checked if they are callable. If they are not, they are not registered in the event manager. So now, the error will not be triggered anymore.

1. Install e.g. [SwagImportExport](https://github.com/shopwareLabs/SwagImportExport) and activate it.
2. Go to the plugin directory and remove SwagImportExport.
3. Try to execute a command in the cli, e.g. `sw:cache:clear`.

None.

None.

### 6. Checklist

PS: While working on this, I discovered, that there are several event listeners in the default Shopware database, that are not callable. But since the listened events seem to not exist anymore, they do not actually matter.

### DIFF
--- a/_sql/migrations/1463-remove-uncallable-event-listeners.php
+++ b/_sql/migrations/1463-remove-uncallable-event-listeners.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+class Migrations_Migration1463 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $uncallableListeners = [
+            'Shopware_Plugins_Backend_Menu_Bootstrap::onInitResourceMenu',
+            'Shopware_Plugins_Core_Router_Bootstrap::onFilterAssemble',
+            'Shopware_Plugins_Core_Router_Bootstrap::onFilterUrl',
+            'Shopware_Plugins_Core_Router_Bootstrap::onAssemble',
+            'Shopware_Plugins_Backend_Check_Bootstrap::onGetControllerPathBackend',
+            'Shopware_Plugins_Core_MarketingAggregate_Bootstrap::initTopSeller',
+        ];
+
+        $sql = 'DELETE FROM s_core_subscribes WHERE listener = "%s";';
+
+        foreach ($uncallableListeners as $uncallableListener) {
+            $this->addSql(sprintf($sql, $uncallableListener));
+        }
+    }
+}
--- a/engine/Library/Enlight/Event/EventManager.php
+++ b/engine/Library/Enlight/Event/EventManager.php
@@ -24,6 +24,8 @@
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Enlight\Event\SubscriberInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 
 /**
  * The Enlight_Event_EventManager stores all event listeners.
@@ -50,6 +52,26 @@ class Enlight_Event_EventManager extends Enlight_Class
      *                              registerListener(Enlight_Event_Handler $handler) function.
      */
     protected $listeners = [];
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
+     * Enlight_Event_EventManager constructor.
+     * @param LoggerInterface $logger
+     */
+    public function __construct()
+    {
+        $this->logger = new NullLogger();
+        parent::__construct();
+    }
+
+    public function setLogger(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
+    }
 
     /**
      * Returns all event listeners of the Enlight_Event_EventManager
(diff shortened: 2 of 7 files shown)


shyim:
Why should we fix that?
You have "deleted" the Plugin in the wrong way.  If you uninstall and delete it, this problem would not happen.

jkrzefski:
@shyim I believe, this adds to the stability of your system. It seems rather unstable if one faulty entry in `s_core_subscribes` is all it takes to completely break the cli and at the same time not showing an error message. The command just terminates with code 255 and the user does not even receive a hint to the problem. If a callback is provided to be called on a specific occasion, that callback should at least be callable.

shyim:
When you catch those issues. The developer will never get feedback if he does something wrong.
Thats the main issue what i have with your Fix.

jkrzefski:
I will add error logging for faulty listeners. To prevent Shopware from blowing up its logfiles I will also add a migration to remove the faulty listeners that come with Shopware by default. Anything else?

jkrzefski:
I have added error logging and a migration to remove said faulty listeners. But now travis is giving me a hard time. I ran the unit tests locally and I cannot reproduce the result that the travis gets. There it says:

```
1) Shopware\Tests\Unit\Components\Event\EventManagerTest::testAddSubscriber
RuntimeException: Shopware Kernel not booted
```

I can only assume this is related to the way I handle error logging in the EventManager. I use the `Shopware()` monolith function to get the core logger. That at least would be my best guess. But I have no clue why this does not happen in my local test environment. A little hint or help here would be much appreciated.

shyim:
The unit and functional tests has to been separately executed. The easiest way to execute it right is ant build-continuous

jkrzefski:
@shyim Thank you. I am now able to reproduce the error in my local environment. But when it comes to testing the event manager as a unit, I run into this problem where I cannot have it be an independent unit but instead the event manager now depends on the logger service. I know there is a `ContainerAwareEventManager` that probably will not encounter this problem, but there are two issues that prevent me from moving my changes into that class.

1. The unit test in question only spawns an instance of `Enlight_Event_EventManager`, so my checks would not run if they were only located in the `ContainerAwareEventManager`.
2. The method I chose for this check (`registerListener`) is callable both internally and externally. So if I were to override this method, only the external calls would be affected (if they even choose to use the `ContainerAwareEventManager` at all). The internal calls would still use the unchanged version of the `registerListener` method and bypass my check. So that would require a lot more changes to both classes.

And that does not even cover the possibility that an event manager is not necessarily an instance of the `ContainerAwareEventManager`. So in short: How do I get an independent unit and at the same time have access to a logger service? Or do I have to alter the unit test to have the kernel booted up before the test runs?

shyim:
I had fixed it with https://dpaste.de/axMh . Maybe try that out

jkrzefski:
@shyim Thank you for your help. I applied your patch and the build is now passing.

engine/Library/Enlight/Event/EventManager.php
```diff
@@ -98,6 +120,28 @@ public function addListener($eventName, $listener, $priority = 0)
      */
     public function registerListener(Enlight_Event_Handler $handler)
     {
+        if (($handler instanceof Enlight_Event_Handler_Default && !is_callable($handler->getListener())) ||
```
hlohaus: in the events a isCallable method would be nice.
less code here.

jkrzefski: @hlohaus I have added a method `isCallable` to the handler and implemented them for the default and plugin handler. Does this look good to you?

Btw: I have also done a slight change to the log statement. Instead of building the message with `sprintf` I now use placeholders and supply the values as context. Not much but I thought, I should mention it.

shyim: This introduces a BC and this pr have then to target 5.6 :thinking:

jkrzefski: @shyim What if I make the method in `Enlight_Event_Handler` not abstract but instead return `true` by default so it is not mandatory to override it. That would not be a BC, right?

hlohaus: Nice 👯‍♂ And now a getListenerName() :D

jkrzefski:
@shopwareBot As much as I like to see this PR scheduled, the problem @shyim pointed out should at least be discussed. I understand that adding a new abstract method to the `Enlight_Event_Handler` class introduces a BC. So I would like some feedback on my idea to not make the `isCallable` method abstract but return `true` by default and override it in `Enlight_Event_Handler_Default` and `Enlight_Event_Handler_Plugin`.

shyim:
A method which is called "isCallable" which returns everytime "true" feels wrong to me. I have searched the Community Store mirror and none plugin would have impact of that. We will discuss this again internaly when we someone working on this Pull Request. So its currently complete from your side :blush:

jkrzefski:
@shyim Alright, thank you. 💙

jkrzefski:
I noticed something about the way I compose the log messages: The are displayed with the placeholders instead of the values from the context. A short google search shows that I would need a new monolog processor to replace the placeholders. See https://github.com/symfony/symfony-standard/issues/981#issuecomment-223898528 for more info.

So my question is: Is this worth enabling a new monolog processor or should I go back to composing the messages with `sprintf`?

jkrzefski:
I have changed the logging back to sprintf now.

jkrzefski:

jkrzefski:
Would this not be a great opportunity to merge this PR into 5.6 right before its release? ;)

_sql/migrations/1463-remove-uncallable-event-listeners.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    public function up($modus)
+    {
+        $uncallableListeners = [
+            'Shopware_Plugins_Backend_Menu_Bootstrap::onInitResourceMenu',
+            'Shopware_Plugins_Core_Router_Bootstrap::onFilterAssemble',
+            'Shopware_Plugins_Core_Router_Bootstrap::onFilterUrl',
+            'Shopware_Plugins_Core_Router_Bootstrap::onAssemble',
+            'Shopware_Plugins_Backend_Check_Bootstrap::onGetControllerPathBackend',
+            'Shopware_Plugins_Core_MarketingAggregate_Bootstrap::initTopSeller',
```
shyim: This method exists. Are you sure here?

jkrzefski: @shyim I could not find it. Can you provide a link to this method, please?

PaddyS:
Hello @jkrzefski,

We've merged your PR internally and it will be available here soon. :-)

shyim:
Unfortunately this change broke multiple things Old Plugin system plugins (like Pickware) not working anymore, Lazy Loading of Plugins / Subscribers are broken, cause isCallable triggers instantiating. We have reverted this change 
https://github.com/shopware/shopware/commit/e66c306949de736257f5e8921be560278354053e
Feel free to create a new PR with a different solution and with tests

jkrzefski:
@shyim Can you please provide more details about the problem with this solution? I don't understand the cause of these issues.

jkrzefski:
@shyim I would like to fix the issues with this PR. Can you provide me with some more information?

shyim:
Create a plugin in the old way. Overwrite the afterInit method. Just subscribe to an Unknown event.
The afterInit method will be executed. So lazy is broken

jkrzefski:
Ok, thank you. I will take a look at this.

PR #2093, opened 2019-04-24
Variant support for custom orderNumberRegex

hlohaus:

@soebbing has forget the regex in the variant listing in the backend.
So ordernumber still doesn't match, if the user has changed the regex in the config.

Fix it

Change the regex in the config / open variant listing in the backend.

Nope

Nope

### 6. Checklist


shyim:

PR #2095, opened 2019-04-24
Refactored shopContext creation out of ContextService

twtgrowth:

It is not possible to elegantly decorate, overwrite or in any otherway change the creation of the ShopContext.  

By moving the creation of the ShopContext to a factory which is injected to the ContextService it is possible to change the process of ShopContext creation by exchanging the factory implementation. 

Before it was necessary to overwrite the whole ContextService in order to get control and change how the ShopContext is created.

### 6. Checklist


shyim:
And squash :)

PR #2096, opened 2019-04-26
invalidate http cache on entity removal

buddhaCode:

Page is still available in HTTP cache when entity got removed.

Change removes affected pages from HTTP cache when an entity got deleted.

1. Activate HTTP cache
2. Open product page (copy link or leave tab open)
3. Enter backend in another tab
4. Delete corresponding produkt via backend
5. Refresh product page (maybe flush browser cache before)
6. Product page is still active and alive but assumend to be dead 💀 

### 6. Checklist

### DIFF
--- a/_sql/migrations/1463-invalidate-http-cache-on-entity-removal-subscriber.php
+++ b/_sql/migrations/1463-invalidate-http-cache-on-entity-removal-subscriber.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * Shopware 5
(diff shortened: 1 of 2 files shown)


shyim:
This affects currently only new installations. You will need here also a migration.

buddhaCode:
@shyim Ups. Sure thing. Didn't notice, that this is a legacy plugin.
Is there any naming convention for the migrations? Guess the merger have to change the file name, anyway.

shyim:

PR #1488, opened 2018-01-29
Console bash completion feature

JoshuaBehrens:
### 1. Why is this change necessary?
For easier usability of console commands.

Did you ever cloned a category tree and did not know what category ids you have?
You wanted to configure a plugin via commandline and ~~rage~~ be upset when you had to copy/rewrite every single config key?

Here are some examples using improved commands:

#### Basic completion
[basic](https://asciinema.org/a/M1WTwW7PadDycB4Iq14UatqWs)

#### Nice things
[nice](https://asciinema.org/a/48k4jWeCKpUTCvqNLMYD4HJJi)

#### Plugin reinstall workflow
[plugin](https://asciinema.org/a/6bZITFV3cOJVETUP2PazBksqD)

Still a downside. You have to register it in your `.bashrc` using `source <(cd /home/vagrant/www/shopware && bin/console _completion --generate-hook)`. Maybe the big Shopware hosters will add this by default to make their customers happy.

### 2. What does this change do, exactly?
Implements for all core/shipped bundle commands the CompletionAwareInterface

### 3. Describe each step to reproduce the issue or behaviour.
Press tab to complete your shopware console commands. It just does not work.

### 4. Please link to the relevant issues (if any).
Life with shopware 💙 

### 5. Which documentation changes (if any) need to be made because of this PR?
You can add this feature to the documentation.

### 6. Checklist

- [ ] I have written tests and verified that they fail without my change
- [x] I have squashed any insignificant commits
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfil them.

PR #2097, opened 2019-04-28
Console bash completion feature is back for shopware 5.6

JoshuaBehrens:
For easier usability of console commands.

Did you ever cloned a category tree and did not know what category ids you have?
You wanted to configure a plugin via commandline and rage be upset when you had to copy/rewrite every single config key?

As this is a reworked pull request that timed out previously I refer to that one https://github.com/shopware/shopware/pull/1488 . I suggest you reading the original pull request to get a deeper understanding.

Implements for all core/shipped bundle commands the CompletionAwareInterface

Press tab to complete your shopware console commands. It just does not work.

You can add this feature to the documentation.

### 5. Checklist

### DIFF
--- a/engine/Shopware/Commands/PluginInstallCommand.php
+++ b/engine/Shopware/Commands/PluginInstallCommand.php
@@ -25,13 +25,46 @@
 namespace Shopware\Commands;
 
 use Shopware\Bundle\PluginInstallerBundle\Service\InstallerService;
+use Shopware\Components\Model\ModelRepository;
+use Shopware\Models\Plugin\Plugin;
+use Stecman\Component\Symfony\Console\BashCompletion\Completion\CompletionAwareInterface;
+use Stecman\Component\Symfony\Console\BashCompletion\CompletionContext;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
-class PluginInstallCommand extends PluginCommand
+class PluginInstallCommand extends ShopwareCommand implements CompletionAwareInterface
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function completeOptionValues($optionName, CompletionContext $context)
+    {
+        return [];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function completeArgumentValues($argumentName, CompletionContext $context)
+    {
+        if ($argumentName === 'plugin') {
+            /** @var ModelRepository $repository */
+            $repository = $this->getContainer()->get('models')->getRepository(Plugin::class);
+            $queryBuilder = $repository->createQueryBuilder('plugin');
+            $result = $queryBuilder->andWhere($queryBuilder->expr()->eq('plugin.capabilityInstall', 'true'))
+                ->andWhere($queryBuilder->expr()->isNull('plugin.installed'))
+                ->select(['plugin.name'])
+                ->getQuery()
+                ->getArrayResult();
+
+            return array_column($result, 'name');
+        }
+
+        return [];
+    }
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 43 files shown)


bneumann97:
Hey @JoshuaBehrens,
thanks for your contribution 💙 

I just check your PR and I get some errors while checking all the commands.

Those give me all the same error:
./bin/console sw:plugin:activate
./bin/console sw:plugin:deactivate
./bin/console sw:plugin:install
./bin/console sw:plugin:reinstall
./bin/console sw:plugin:uninstall

And this one gives me a different error: 
./bin/console sw:plugin:delete 

Please check and fix them.

JoshuaBehrens:
I will check that

JoshuaBehrens:
The problem is that somehow `setContainer` is not called correctly anymore. The commands do not work standalone directly from the container. That is why I made them now "standalone" working like the api resources are standalone working from the DI.

engine/Shopware/Components/DependencyInjection/Compiler/ConfigureContainerAwareCommands.php
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Shopware\Components\DependencyInjection\Compiler;
+
+use Shopware\Components\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
+use Symfony\Component\DependencyInjection\Reference;
+
+class ConfigureContainerAwareCommands implements CompilerPassInterface
```
shyim, line 33: This is not a good solution to fix inject container for your completion. Commands of old plugin system will still explode, cause they don't have the container. I would prefer to override the find method of the Console application and set the container regardless where the command come from. (https://github.com/shopware/shopware/blob/5.6/engine/Shopware/Components/Console/Application.php#L40)

engine/Shopware/Commands/PluginInstallCommand.php
```diff
@@ -25,13 +25,46 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function completeArgumentValues($argumentName, CompletionContext $context)
+    {
+        if ($argumentName === 'plugin') {
+            /** @var ModelRepository $repository */
+            $repository = $this->getContainer()->get('models')->getRepository(Plugin::class);
+            $queryBuilder = $repository->createQueryBuilder('plugin');
+            $result = $queryBuilder->andWhere($queryBuilder->expr()->eq('plugin.capabilityEnable', 'true'))
```
shyim: plugin.capabilityInstall. Currently it completes HttpCache and this throws a exception

engine/Shopware/Commands/PluginListCommand.php
```diff
@@ -26,13 +26,48 @@
...
+                'activate',
+                'inactive',
+            ];
+        }
+
+        if ($optionName === 'namespace') {
+            $namespaces = [
+                'core',
+                'frontend',
+                'backend',
```
shyim, line 54: ShopwarePlugins and ProjectPlugins are missing

engine/Shopware/Components/DependencyInjection/Compiler/ConfigureContainerAwareCommands.php
JoshuaBehrens: This is in my POV a correct solution. This only breaks because these commands implement the completion interface where I rely on the container. Other commands don't have this interface implemented with container usage and therefore don't rely on the container in an other moment than command execute. The original setContainer-call is still there and thus nothing is broken.

In comparison to the impact on the performance I'd say the current solution is heavier in execution once (on container build) and only if needed. The suggest solution impacts every time any command is request disregarding whether it needs it. But admittedly this is a minimal difference.

I test your suggested change and it works out. So it is a valid solution.

engine/Shopware/Commands/PluginInstallCommand.php
JoshuaBehrens: Good one. Fixed.

engine/Shopware/Commands/PluginListCommand.php
JoshuaBehrens: I can add them and see their usage for the first time. I didn't know they existed. Good one.

shyim:
Hey @JoshuaBehrens,

PR #2101, opened 2019-05-02
add attributes table for pdf receipt template

buddhaCode:

For fun ;) The more attributes, the better.

This PR adds an attribute table for s_core_documents. Also, the snippets for the s_order_documents attributes have been renamed from "PDF Vorlage" to "Bestellung - Dokument", because it's heavily misleading.

None.

None.

Not aware of.

### 6. Checklist


shyim:
You missed the doctrine entity and maybe you should also add attributes at places where documents fetched otherwise this feature would be not very useful :)

buddhaCode:
@shopwareBot Okay, here some use cases:

- Imagine one has several different document types, all based on the same template file, and you want to set some config options which will change the behavior in the template file.
- Say u wanna enable specific processing options for document types, e.g. allow multiple documents per type and order or allow part document creation for some, but not all, document types.

@shyim Your right. Thought that was to easy ;) I will take care. Maybe next week.

buddhaCode:
I am. But I'm quite busy right now. But I'll deliver the missing parts.

PR #2103, opened 2019-05-03
make queries in order repo more extendable

buddhaCode:

Make some queries in the order repository more extendable.

I tried to extend the `getDocuments()` method by it's attributes. Unfortunately, that's only possible using a replace or after hook.

- Splits the documents, payments and details query methods into two separate methods to make the query builder extendable
- Fix some doc blocks

Not aware of.

Not aware of.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Order/Document/Repository.php
+++ b/engine/Shopware/Models/Order/Document/Repository.php
@@ -43,9 +43,11 @@ class Repository extends ModelRepository
      * To determine the total number of records, use the following syntax:
      * Shopware()->Models()->getQueryCount($query);
      *
-     * @param int      $orderId
-     * @param int|null $limit
-     * @param int|null $offset
+     * @param int        $orderId
+     * @param array|null $filter
+     * @param array|null $orderBy
+     * @param int|null   $limit
+     * @param int|null   $offset
      *
      * @return \Doctrine\ORM\Query
      */
--- a/engine/Shopware/Models/Order/Repository.php
+++ b/engine/Shopware/Models/Order/Repository.php
@@ -159,8 +159,6 @@ class Repository extends ModelRepository
      * @param int|null     $offset
      * @param int|null     $limit
      *
-     * @internal param $ids
-     *
      * @return \Doctrine\ORM\Query
      */
     public function getOrdersQuery($filters = null, $orderBy = null, $offset = null, $limit = null)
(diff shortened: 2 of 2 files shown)


engine/Shopware/Models/Order/Document/Repository.php
```diff
@@ -44,6 +44,8 @@ class Repository extends ModelRepository
      * Shopware()->Models()->getQueryCount($query);
      *
      * @param int      $orderId
+     * @param null     $filter
```
shyim: array|null

engine/Shopware/Models/Order/Document/Repository.php
```diff
@@ -44,6 +44,8 @@ class Repository extends ModelRepository
...
+     * @param null     $orderBy
```
shyim: array|null

engine/Shopware/Models/Order/Repository.php
```diff
@@ -159,9 +159,11 @@ public function getOrderStatusQueryBuilder($filter = null, $order = null)
      * @param int|null     $offset
      * @param int|null     $limit
      *
-     * @internal param $ids
+     * @throws \Exception
```
shyim: Please add throws only, if the exception is thrown in the method

engine/Shopware/Models/Order/Repository.php
```diff
@@ -159,9 +159,11 @@ public function getOrderStatusQueryBuilder($filter = null, $order = null)
...
      *
      * @return \Doctrine\ORM\Query
+     *
+     * @internal param $ids
```
shyim: Looks like fail, can you remove it?

buddhaCode: Removed

engine/Shopware/Models/Order/Repository.php
buddhaCode: Do you know any way to tell PHPStrom what ;)

engine/Shopware/Models/Order/Document/Repository.php
buddhaCode: fixed

engine/Shopware/Models/Order/Document/Repository.php
buddhaCode: fixed

engine/Shopware/Models/Order/Repository.php
shyim: I have disabled that inspection :laughing:

shyim:

PR #2104, opened 2019-05-03
Cleanup ACLs

buddhaCode:

The ACL resources and privileges are not up to date. The `performance` ACL should be renamed to `cache`, as the `performance` backend controller doesn't use any ACLs, but the `cache` controller does.

API endpoints are missing ACLs at all.

This PR adds missing ACLs resources and privileges as described above. Existing ACLs will be taken into account.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1463-cleanup-acls.php
+++ b/_sql/migrations/1463-cleanup-acls.php
@@ -1,0 +1,125 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+class Migrations_Migration1463 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $sql = <<<'SQL'
+-- backend controller renamed
+UPDATE IGNORE s_core_acl_resources SET `name` = 'cache' WHERE `name` = 'performance';
+
+-- add missing acls for api endpoints
+INSERT INTO s_core_acl_resources (`name`) SELECT 'address' FROM dual WHERE NOT EXISTS (SELECT * FROM s_core_acl_resources WHERE `name` = 'address');
+SET @addressResourceId = (SELECT id FROM s_core_acl_resources WHERE `name` = 'address' LIMIT 1);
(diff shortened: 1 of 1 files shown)


_sql/migrations/1463-cleanup-acls.php
```diff
@@ -0,0 +1,125 @@
...
+ * our trademarks remain entirely with us.
+ */
+
+class Migrations_Migration1463 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $sql = <<<'SQL'
+-- backend controller renamed
+UPDATE IGNORE s_core_acl_resources SET `name` = 'cache' WHERE `name` = 'performance';
```
shyim, line 31: Nope.........
Every user can open now the performance settings. Please create a new resource 🙈

buddhaCode: Okay I can see your point. Is it (bc/update) save to move the existing clear, ... privileges to the new cache resource? These privileges aren't working accoirdingly, but maybe linked to existing roles (like in my case)

shopwareBot: Renaming the resource would be a bc. Please create a new resource with the permissions

_sql/migrations/1463-cleanup-acls.php
```diff
@@ -0,0 +1,125 @@
...
+
+-- add missing acls for api endpoints
+INSERT INTO s_core_acl_resources (`name`) SELECT 'address' FROM dual WHERE NOT EXISTS (SELECT * FROM s_core_acl_resources WHERE `name` = 'address');
```
shyim, line 34: Can you simplify it to INSERT INTO IGNORE?

_sql/migrations/1463-cleanup-acls.php
buddhaCode: @shopwareBot Thanks for your reply. I understand the issue with the resource. But what about the privileges performance.read, performance.update, performance.clear. These are dead. Remove them or rename/move them to the cache resource, while keeping the privilege database id?!?

shyim: Controller does not have to implement "read". Its checked automaticly by menu and loading of controller. Also the acls are used in extjs, so they are not useless. Maybe add these privileges also to the php controller

_sql/migrations/1463-cleanup-acls.php
buddhaCode: There is a chance that a simple insert (and ignore) will result in two entries with the `name` "address". The `name` column has no unique key constraint.

_sql/migrations/1463-cleanup-acls.php
buddhaCode: Okay. Now I got it ;) The "performance" resource is queried in the ExtJS views/controllers via "acl_is_allowed" smarty helper. But the AJAX request is fired against the "cache" controller. Not the "performance" controller. It is okay to clean this up? Change the ExtJS views/controllers to query the "cache" resource?

e.g. https://github.com/shopware/shopware/blob/5.5/themes/Backend/ExtJs/backend/performance/controller/direct.js#L59

shyim: Its ok :)

_sql/migrations/1463-cleanup-acls.php
shyim: Ahh okay

_sql/migrations/1463-cleanup-acls.php
buddhaCode: Puhh. This BC think is making some headache. I renamed ACL privilege `performance.clear` to `cache.clear`. Guess this is already a BC, isn't it? I would also consolidate `cache.clear` and `cache.delete`, because it's basically the same. `clear` is used in backend GUI, `delete` in api resource. But this would break the privilege check in cache API resource.

soebbing: Breaking a privilege check somewhere is a no-go. You might standardize the name and deprecate the other one, making sure e.g. that `cache.delete` still works for now and `cache.clear` works as well, so we can remove `cache.delete` in a later version

shopwareBot: Are you still on it?

buddhaCode:
I am. Was just sick the last days.

PR #2106, opened 2019-05-07
add missing GTINs

tinect:

GTIN13 shouldn't be forced.
https://www.barcode.graphics/education-gtin-data-structures/

Check length of string and select matching GTIN. GTIN13 at first, cause it should be common.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/detail/content/header.tpl
+++ b/themes/Frontend/Bare/frontend/detail/content/header.tpl
@@ -19,7 +19,16 @@
 
                     {block name="frontend_detail_index_data_ean"}
                         {if $sArticle.ean}
-                            <meta itemprop="gtin13" content="{$sArticle.ean}"/>
+                            {$sArticle.ean = $sArticle.ean|trim}
+                            {if $sArticle.ean|strlen == 13}
+                                <meta itemprop="gtin13" content="{$sArticle.ean}"/>
+                            {elseif $sArticle.ean|strlen == 8}
+                                <meta itemprop="gtin8" content="{$sArticle.ean}"/>
+                            {elseif $sArticle.ean|strlen == 12}
+                                <meta itemprop="gtin12" content="{$sArticle.ean}"/>
+                            {elseif $sArticle.ean|strlen == 14}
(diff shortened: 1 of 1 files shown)


themes/Frontend/Bare/frontend/detail/content/header.tpl
```diff
@@ -19,7 +19,15 @@
 
                     {block name="frontend_detail_index_data_ean"}
                         {if $sArticle.ean}
-                            <meta itemprop="gtin13" content="{$sArticle.ean}"/>
+                            {if $sArticle.ean|strlen == 13}
```
shyim, line 23: Can you add trim here? If the user copies an ean with whitespace this construct would not work

tinect: Good catch, thank you! Would suggest to trim before line 21.
`{$sArticle.ean = $sArticle.ean|trim} `
Okay?

themes/Frontend/Bare/frontend/detail/content/header.tpl
```diff
@@ -18,8 +18,17 @@
                     {/block}
 
                     {block name="frontend_detail_index_data_ean"}
+                        {$sArticle.ean = $sArticle.ean|trim}
```
mitelg: this should be done after the if check. 
I don't what happens if you try to trim something which is not there :wink:

themes/Frontend/Bare/frontend/detail/content/header.tpl
```diff
@@ -19,7 +19,16 @@
 
                     {block name="frontend_detail_index_data_ean"}
                         {if $sArticle.ean}
-                            <meta itemprop="gtin13" content="{$sArticle.ean}"/>
+                            {$sArticle.ean = $sArticle.ean|trim}
```
hlohaus, line 22: btw. {$eanLenght = $sArticle.ean|trim|strlen} would be faster + less code

tinect: Thanks. Sure, but have to define it additionally. Otherwise the ean won't be displayed trimmed.

shyim:

PR #2109, opened 2019-05-08
Fix wrong permissions for cache folders/files

ingowalther:
The default permission configuration for cache folders/files will result in wrong file permissions.

"0777 & ~umask()" will return an decimal permission value. (e.g. with umask 0002 this will return 509).

This values will be injected into the "Zend_Cache_Backend_File" Component which will call "octdec" on "hashed_directory_perm" and "cache_file_perm" in the constructor. Because these values are already decimals this will return wrong permission values and these will be set during cache creation. 

This change will convert the decimal results back to octal numbers.

/

/

/

### 6. Checklist


ingowalther:
This solution wont work correctly

PR #2110, opened 2019-05-08
Fix wrong permissions for cache folders/files

ingowalther:
The default permission configuration for cache folders/files will result in wrong file permissions.

"0777 & ~umask()" will return an decimal permission value. (e.g. with umask 0002 this will return 509).

This values will be injected into the "Zend_Cache_Backend_File" Component which will call "octdec" on "hashed_directory_perm" and "cache_file_perm" in the constructor. Because these values are already decimals this will return wrong permission values and these will be set during cache creation.

This change will convert the decimal results back to octal numbers before injecting them into the "Zend_Cache_Backend_File" Component.

/

/

/

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/DependencyInjection/Bridge/Cache.php
+++ b/engine/Shopware/Components/DependencyInjection/Bridge/Cache.php
@@ -51,6 +51,14 @@ class Cache
     {
         $backendOptions['release'] = $release;
 
+        if(isset($backendOptions['hashed_directory_perm'])) {
+            $backendOptions['hashed_directory_perm'] = decoct($backendOptions['hashed_directory_perm']);
+        }
+
+        if(isset($backendOptions['cache_file_perm'])) {
+            $backendOptions['cache_file_perm'] = decoct($backendOptions['cache_file_perm']);
+        }
+
         $backend = $this->createBackend($backend, $backendOptions);
         $cacheCore = $this->createCacheCore($frontendOptions);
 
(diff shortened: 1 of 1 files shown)


ingowalther:
I'm not sure that's a good solution. If someone overwrites the permissions in the config, they will probably specify this as an octal number. If this happens, my customization leads to incorrect permissions. Unfortunately, it's not so easy to tell whether the value is octal or decimal. Maybe it would be a solution to remove the "octdec" calls in the ""Zend_Cache_Backend_File" component.

soebbing:
Hi @ingowalther!

I checked your change, but it looks like `octdec` is only used on the permission-settings if the value is a string (see [here](https://github.com/shopware/shopware/blob/01d604dba3ff13761e063499f83508c5501fdbcd/engine/Library/Zend/Cache/Backend/File.php#L147)).

While testing around, I found that the same setting is being used in the [TemplateManager](https://github.com/shopware/shopware/blob/7543603bb9d1c467b6c06f3b088da5980d371d7e/engine/Library/Enlight/Template/Manager.php#L76), but there the same conversion is missing.

Maybe this is the source of your problems? I am currently trying to implement a similar conversion case as in the `Backend/File.php` there, to streamline both behaviours. Might this fix your issue?

ingowalther:
Hello @soebbing,

I seem to have missed the "is_string" check. I have to take a closer look again at why it leads to different results with the adjustments than without.

shyim:

PR #2111, opened 2019-05-09
Update advanced menu click

hlohaus:

On a tablet device the menu does't open on first click.
If you have a sub menu, it make sense.
But without it should open directly.

It fix it.

Step 1: Install Advanced Menu.
Step 2: Create some categories with sub entries or not.
Step 3: Open the shop with a tablet or if you don't have a tablet you can use the browser console with device toolbar.

You must click twice to open an upper navigation link.
With this change a navigation entry without a submenu will be open directly.

### 6. Checklist


soebbing:
Could you please fill out step 3. so we are able to reproduce the problem more easily? Also, what cases/devices did you check your change with?

PR #2112, opened 2019-05-09
SW-14204 Add trailing slash support

hlohaus:

See https://issues.shopware.com/issues/SW-14204

It makes a second sql request without trailing slash.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1646-ignore-trailing-slash.php
+++ b/_sql/migrations/1646-ignore-trailing-slash.php
@@ -1,0 +1,21 @@
+<?php
+class Migrations_Migration1646 Extends Shopware\Components\Migrations\AbstractMigration
+{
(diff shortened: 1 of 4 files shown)


soebbing:
Wouldn't that change lead to possibly two different contents presented for one URL with and without a slash?

hlohaus:
Because of main=0 it will be redirected.

soebbing:
Could you please add a few test cases for this change?

shyim:
Hey,
Sorry for the late Response. I have discussed it with our SEO guru, and we need here a new option in SEO router to enable / disable this feature. It should be also deactivated on update.

Can you add this to this PR? Thanks! :blue_heart:

shyim:

PR #2113, opened 2019-05-10
Prevent breaking change in snippet manager constructor

felixbrucker:

This change is required to prevent breaking plugin classes which `extends \Shopware_Components_Snippet_Manager`. We do this for example to provide localized API error messages (useful for app clients) or to provide localized user information messages during the initial installation of a plugin (i.e. when the plugin's snippets have not been imported yet).

This change makes the `themeDir` constructor parameter optional.

Create a class which extends the snippet manager and call the parent constructor with only 3 parameters.

n/a

n/a

### 6. Checklist


shyim:
Thank you for your Contribution! :blue_heart:

PR #2114, opened 2019-05-11
Corrected translation of order status processing string

budda:
Improved English language used in the order list based on end user customer feedback.

Changes one of the order status strings.

"In Process" doesn't make sense, suspect it should be "In progress" for orders being processed?

Check the order status drop down options


soebbing:
Hi @budda,

I checked for more occurrences of 'in process', found some more and changed them as well. Also, I changed the target branch of your PR to 5.5 (which will be merged upstream to 5.6 later on), I hope that is in your interest.

Thanks for your PR! 💙

PR #2115, opened 2019-05-13
Refactor url provider

hlohaus:

The url providers are not extendable.
Additionally the categories query isn't anymore recursive.
It makes now one call for all categories.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SitemapBundle/Provider/ManufacturerUrlProvider.php
+++ b/engine/Shopware/Bundle/SitemapBundle/Provider/ManufacturerUrlProvider.php
@@ -24,8 +24,7 @@
 
 namespace Shopware\Bundle\SitemapBundle\Provider;
 
-use Doctrine\DBAL\Driver\Connection as ConnectionInterface;
-use Doctrine\DBAL\Query\QueryBuilder;
+use Shopware\Bundle\SitemapBundle\Repository\ManufacturerRepositoryInterface;
 use Shopware\Bundle\SitemapBundle\Struct\Url;
 use Shopware\Bundle\SitemapBundle\UrlProviderInterface;
 use Shopware\Bundle\StoreFrontBundle\Struct\ShopContextInterface;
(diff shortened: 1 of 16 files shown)


soebbing:
@hlohaus I am not exactly why you want to do do. These services weren't build with extendability in mind! You should rather implement the interface yourself with your custom service and just add the appropriate tag to have your URLs exported.

How does your use-case look like?

hlohaus:
As user i want to change the getCriteria or getQueryBuilder to filter urls.
But without this change you have to completely copy the class...
See: https://github.com/hlohaus/SheSitemapFilter

shyim:
When you really want to filter urls, you should use https://github.com/shopware/shopware/blob/5.5/engine/Shopware/Bundle/SitemapBundle/UrlFilterInterface.php

hlohaus:
I doesn't like the filters. I can only register one and there is already one?
Additionally i don't want to make a second sql statement.

shyim:
Okay. But your changes does not improve to decorate this service. You make it better to replace the service 😉.
Maybe we need a QueryBuilder service instead

engine/Shopware/Bundle/SitemapBundle/Provider/ManufacturerUrlProvider.php
```diff
@@ -24,7 +24,7 @@
 
 namespace Shopware\Bundle\SitemapBundle\Provider;
 
-use Doctrine\DBAL\Driver\Connection as ConnectionInterface;
```
shyim, line 27: Why replace the interface usage by an explict implementation?

hlohaus: Only this class implements createQueryBuilder.

hlohaus:
I added a repository service for all resources.

shyim:
Hey @hlohaus,
unfortunately the CI build is failing. Can you fix it?

PR #2116, opened 2019-05-14
Support single order detail update via API

JoshuaBehrens:
If you want to update a single order detail's status via API you have to know all order details and send them over otherwise you will delete all those which you do not update in your API call.

It always loads the existing details and only overrides them if given in the payload.

1. Have an order with at least two order details.
2. Update this order via API with the following payload
```json
{
  "details": [{
    "id": 42,
    "status": 2
  }]
}
```
3. Every detail that has not the id 42 is now deleted as the foreign key to the order is lost on doctrine level and it deletes the orphans.

A hint to the order api documentation should be added.

### 5. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Order.php
+++ b/engine/Shopware/Components/Api/Resource/Order.php
@@ -24,6 +24,7 @@
 
 namespace Shopware\Components\Api\Resource;
 
+use Doctrine\Common\Collections\ArrayCollection;
 use Shopware\Components\Api\Exception as ApiException;
 use Shopware\Models\Country\Country as CountryModel;
 use Shopware\Models\Country\State;
@@ -266,7 +267,7 @@ class Order extends Resource
             throw new ApiException\NotFoundException(sprintf('Order by id %d not found', $id));
         }
 
-        $params = $this->prepareOrderData($params);
+        $params = $this->prepareOrderData($params, $order);
 
         $order->fromArray($params);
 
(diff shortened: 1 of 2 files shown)


shyim:
And how do i overwrite all details now?
We need here something like "__options_details" like in other api resources which the default is replace

JoshuaBehrens:
@shyim Overwrite in the common API sense is also adding new items. This is not possible due to the whitelist here. See my comment: https://github.com/shopware/shopware/blob/4a20e600e880530c3d99b251b2beab68c9d2f3bc/engine/Shopware/Components/Api/Resource/Order.php#L605

Is it ok if I allow deleting on the option "replace"?

JoshuaBehrens:
Added test and delete support. Review the tests how it is done. It will use the common replace parameter.

JoshuaBehrens:
Checked and cleared

shyim:
I don't see any difference.

PUT /orders/57
```
{
	"details": [
		{
			"id": 201,
			"status": 2
		}
	]
}
```

With your changes and without all details are gone.

JoshuaBehrens:
Did you add the `__options_details`?

shyim:
:man_facepalming:  Long day..  Is now in code-review thanks :blush:

JoshuaBehrens:
I'd like to set my fix as default but you are very protective about the API (plausible) :blue_heart:

shyim:

PR #2121, opened 2019-05-26
make config parameters modifiable  while order document creation

buddhaCode:

I need to modify the config parameters prior to the creation of order documents. Besides, it would be nice to have them available in the model.

* Added event `Shopware_Models_Order_Document_Filter_Config` to modify config settings for document creation
* Added `_config` property to class `Shopware_Models_Document_Order` to make it usable while model creation

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -173,6 +173,16 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
         /** @var Shopware_Components_Document $document */
         $document = Enlight_Class::Instance('Shopware_Components_Document');
 
+        $config = Shopware()->Container()->get('events')->filter(
+            'Shopware_Models_Order_Document_Filter_Config',
+            $config,
+            [
+                'subject' => $document,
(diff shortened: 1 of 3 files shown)


engine/Shopware/Components/Document.php
```diff
@@ -173,6 +173,15 @@ public static function initDocument($orderID, $documentID, array $config = [])
         /** @var Shopware_Components_Document $document */
         $document = Enlight_Class::Instance('Shopware_Components_Document');
 
+        $config = Shopware()->Container()->get('events')->filter(
+            'Shopware_Models_Order_Document_Filter_Config',
+            $config,
+            [
+                'orderID' => $orderID,
```
shyim, line 181: Please pass subject

buddhaCode: This is static context. Isn't it pretty much senseless?

shyim: $document would be the context 🤔

buddhaCode: Ur right.

PR #2130, opened 2019-06-05
Considers minimum purchase quantity for graduated prices

jkrzefski:
Say you have a product with graduated prices for quantities from `1` to `10` and from `11` to `beliebig`. Now say you set the minimum purchase quantity to `15`. The table will still show the prices up to `10` but you cannot buy the product in such low quantities.

It removes every entry from the displayed price table that is impossible to purchase. It also raises the smallest displayed quantity to the maximum of `prices.to` or `variant.minpurchase`. So the case above would then display one row starting from `15`.

See 1.

https://git.shopware.com/plugins/SwagCustomProducts/merge_requests/2

None.

### 6. Checklist


shyim:
Can you add some test cases?

jkrzefski:
@shyim I added a test case. Is this sufficient or should I test something else, too?

jkrzefski:

shyim:

PR #2132, opened 2019-06-11
Better paging for several backend list views

pascalheidmann-bedarf:
Usability

- Adding more page size options to
  - article list (up to 150)
  - order list (up to 250)
- adding page size combo to
  - supplier list
  - config lists (shops, customer groups, cron, ...)

### 6. Checklist


engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/store/local_plugin.js
```diff
@@ -31,7 +31,7 @@
 Ext.define('Shopware.apps.PluginManager.store.LocalPlugin', {
     extend:'Shopware.store.Listing',
 
-    pageSize: 20000,
```
shyim: Did you tested the extjs grouping feature with 200 plugins, when you have pagination?

shyim:
Please change the target branch to 5.6

engine/Shopware/Plugins/Default/Backend/PluginManager/Views/backend/plugin_manager/store/local_plugin.js
pascalheidmann-bedarf: Actually nope as I don't have that much plugins. But the 20k limit feels also very much like a hack 😁 
I'll remove it for now from PR as this might break the pagination as you suggested.

PR #2137, opened 2019-06-14
create event classes for shopware events

uehler:
The Symfony Event System and therefore shopware 6 have those nice named events. For example when the shopware 6 product detail page was loaded, it will emit the event ProductPageLoadedEvent. It has getters for all the data it provides and your IDE helps you with code completion.
This pull request would create similar events for shopware 5.x, but I wanted to make sure this feature is wanted before I implement it. I also don't really know where to put those event classes in the shopware core, just tell me a good place.

You also won't have to copy the event names into your subscriber

```
use Enlight\Event\SubscriberInterface;

final class MySubscriber implements SubscriberInterface
{
    public static function getSubscribedEvents(): array
    {
        return [
            AddVoucherStartEvent::EVENT_NAME => 'mySubscriberCallback',
        ];
    }


    public function mySubscriberCallback(AddVoucherStartEvent $event)
    {
        echo $event->getCode();
    }
}
```

it is not, but might be helpful

replaces all those generic Enlight_Event_EventArgs with named events, which provide getters and setters for the event data.

If you subscribe to any event in Shopware, you will get an Enlight_Event_EventArgs as parameter of your subscriber, but either you know which data is provided or you have to check where the event is emitted.
With the named events your IDE will provide you with code completion.

the event documentation should be extended, I would provide those changes too, if you like the idea of this PR

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/Events/Articles/GetArticleChartsEvent.php
+++ b/engine/Shopware/Core/Events/Articles/GetArticleChartsEvent.php
@@ -1,0 +1,64 @@
+<?php declare(strict_types=1);
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Core\Events\Articles;
+
+use Enlight_Event_EventArgs;
+use sArticles;
+
+class GetArticleChartsEvent extends Enlight_Event_EventArgs
+{
+    public const EVENT_NAME = 'Shopware_Modules_Articles_GetArticleCharts';
+
+    public function getSubject(): sArticles
+    {
+        return $this->get('subject');
+    }
+
+    public function getCategoryId(): int
+    {
+        return $this->get('categoryId');
+    }
+
+    public function getProducts(): array
+    {
+        return $this->get('products');
+    }
+
+    /**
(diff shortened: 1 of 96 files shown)


shyim:
Maybe that could be a Task that can be done with https://github.com/rectorphp/rector
Just search for a Expr\_New and then create a class based on event name and pass the arguments there :sweat_smile:

uehler:
I will have a look into it
where would be a good place for all those events?

mitelg:
really nice idea! Also because it's not a break :blush: :+1: 
Maybe the events should be placed, where they are dispatched?

uehler:
good idea
so for example I would place the AddVoucherStartEvent in engine/Shopware/Core/Events?

mitelg:
yes, for example :+1: and also for each core class a subdirectory. but that is only my idea :blush: 
what do you think @shyim @soebbing ?

uehler:
@shopwareBot yes, I will do it this week

engine/Shopware/Core/Events/Articles/GetArticleChartsEvent.php
```diff
@@ -0,0 +1,64 @@
...
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Core\Events\Articles;
+
+use Enlight_Event_EventArgs;
+use sArticles;
+
+class GetArticleChartsEvent extends Enlight_Event_EventArgs
```
mitelg, line 30: I'm wondering if we should use the correct translation "product" in the class name? :thinking: at some day devs might only use `Event::class` to subscribe to events, so we could get rid of another layer of false translations :wink:

mitelg:
really nice work! :+1:

engine/Shopware/Core/Events/Articles/GetArticleChartsEvent.php
uehler: good idea, I'll change that

mitelg: and also for the method names :wink:

engine/Shopware/Plugins/Default/Frontend/Seo/Bootstrap.php
```diff
@@ -208,7 +208,8 @@ private function canBeMinified(Enlight_Controller_Action $controller): bool
         /** @var Enlight_Event_EventManager $events */
         $events = $this->get('events');
 
-        if ($event = $events->notifyUntil(__CLASS__ . '_canBeMinified', [
+        // TODO add event class and register namespace!
+        if ($event = $events->notifyUntil('Shopware_Plugins_Frontend_Seo_Bootstrap_canBeMinified', [
```
PaddyS, line 212: Hey @uehler,
will you fix this one too? :-)

uehler: i will look into this pr this week again
completely forgot about it 😅

shyim: Any new news?

PR #2138, opened 2019-06-17
Fix padding on hidden cookie-permission banner breakpoint update

xf-:
## 1. Why is this change necessary?
On breakpoint update the height of banner gets recalculated without checking the cookie

Check if cookie isset and no banner displayed (same as on init)

If breakpoints is triggered the update sets a new height. If the element has a height in css it will be used. (display none is active). -> Depends on your theme

### 6. Checklist


xf-:
Fixed @shopwareBot

PR #2139, opened 2019-06-17
Change escaping for snippets on csv export

mnaczenski:
The exported file contains wrong escaping for Excel. The End-Separator is missing and the singe quote is also wrong.

Bildschirmfoto 2019-06-17 um 13 49 53

Change the escaping to a double quote and complete the quoting by adding another one after the string.

Bildschirmfoto 2019-06-17 um 13 53 09

Export snippets as "CSV (Excel)" in the snippets manager of the backend

### 6. Checklist


shyim:

PR #2141, opened 2019-06-21
Prevent DateTime from given errors on valid (numeric) formats

ArjanLankhaar:
If you parse a unix timestamp in the constructer of a DateTime object, there is a chance that the DateTime object is complaining about an "Unexpected character", altough the value is a valid timestamp (you can see the error in the screenshot). This is a pretty common problem, if you search for the error on google. 
DateTimeStringTypeError

If a value has to be converted to a database value, I've added a check if the value is numeric. If this is the case, it is a unix timestamp, so I us the `setTimestamp()` method to prevent this error from happening. If the value is not numeric, it is probably a string like "now", which I do pass into the constructor.

Have a model with a DATETIME/TIMESTAMP column and try to write this to the database.

There are some relevant issues, but I'm not sure whether there are shopware specific issues to mention. At least, here is someone who had the same problem in his application:
https://stackoverflow.com/questions/17427503/php-datetime-construct-failed-to-parse-time-string-xxxxxxxx-at-position-x

None

### 6. Checklist


shyim:
Your Builds are failing

PascalThesing:
Dear ArjanLankhaar, 

Thank you for your first pull request and for the time which you spend on it. 

At the moment, we are trying to get a more “stock” doctrine implementation. 

Therefore, we are going to remove our own internal doctrine types in the future. 

Your mentioned error is not a doctrine type problem. 

The reason for the error is the way how the corresponding property is set in the model. 

Please introduce your change to the setter in the model itself. 

Yours faithfully  

Pascal

ArjanLankhaar:
Hi @shyim 
> Your Builds are failing

I cannot solve this because this is out of my hand since the error message is:
"The program 'ant' is currently not installed. To run 'ant' please ask your administrator to install the package 'ant'."

shyim:
You have to change the target to 5.6. We currently don't plan to release an another 5.5.
Please also consider @PascalThesing comment

PascalThesing:
Hey @ArjanLankhaar, 

Please rebase your branche with the current Shopware 5.6 branch. 
There are changes which should affect your pull request.

ArjanLankhaar:
Hey @PascalThesing,
I've rebased my bugfix branch to 5.6

> At the moment, we are trying to get a more “stock” doctrine implementation.

Does this mean that this change is no more needed in 5.6 or in one of the later updates?

Kind regards,

Arjan

PascalThesing:
Dear @ArjanLankhaar, 

you error should have changed with 5.6. The reason for your problem is, that you aren't writing a daytime into the model. This is not a doctrine error. It is a programming error. Please provide us a stack trace or change the corresponding setter of the attribute. 

This is an example of a working setter of an attribute:

`

    public function setDate($date)
    {
        if (!$date instanceof \DateTimeInterface) {
            $date = new \DateTime($date);
        }

        $this->date = $date;

        return $date;
    }

`

It changes the Value to a DateTime if it is not.

If not, we have to close your Pull request. 

Yours faithfully 

Pascal

PR #2142, opened 2019-06-21
Added getQuery function to CategoryGateway

lx-wnk:
Added possibility to overwrite only the categoryQuery instead of overwriting the whole getList function or work with arrays after
Because it should be possible to edit only the query instead of the getList method

Puts the Query in an own method and executes it in the getList method
Removed deprecated int|array from get method

Try to edit the query instead of overwrite the getList method.

I don't know any

I don't know any

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
@@ -24,7 +24,8 @@
 
 namespace Shopware\Bundle\StoreFrontBundle\Gateway\DBAL;
 
-use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\ResultStatement;
+use PDO;
 use Shopware\Bundle\StoreFrontBundle\Gateway;
 use Shopware\Bundle\StoreFrontBundle\Service\MediaServiceInterface;
 use Shopware\Bundle\StoreFrontBundle\Struct;
@@ -32,29 +33,14 @@ use Shopware\Bundle\StoreFrontBundle\Struct;
 class CategoryGateway implements Gateway\CategoryGatewayInterface
 {
     /**
+     * @var Gateway\CategoryQueryHelperInterface
+     */
+    private $queryHelper;
+
+    /**
      * @var Hydrator\CategoryHydrator
      */
     private $categoryHydrator;
-
-    /**
-     * The FieldHelper class is used for the
-     * different table column definitions.
-     *
-     * This class helps to select each time all required
-     * table data for the store front.
-     *
-     * Additionally the field helper reduce the work, to
-     * select in a second step the different required
-     * attribute tables for a parent table.
-     *
-     * @var FieldHelper
-     */
-    private $fieldHelper;
-
-    /**
-     * @var Connection
-     */
-    private $connection;
 
     /**
      * @var MediaServiceInterface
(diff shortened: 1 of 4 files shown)


shyim:
Hey, your change does not allow decorating getQuery. Please create a new service which builds the query and inject the service to get the query. You can checkout the listproduct gateway as example.

Also, please change the target to 5.6

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
```diff
@@ -24,71 +24,45 @@
...
+use PDO;
 use Shopware\Bundle\StoreFrontBundle\Gateway;
 use Shopware\Bundle\StoreFrontBundle\Service\MediaServiceInterface;
 use Shopware\Bundle\StoreFrontBundle\Struct;
 
 class CategoryGateway implements Gateway\CategoryGatewayInterface
 {
     /**
-     * @var Hydrator\CategoryHydrator
+     * @var CategoryQueryHelper
```
shyim: CategoryQueryHelperInterface

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
```diff
@@ -187,23 +135,24 @@ public function getList(array $ids, Struct\ShopContextInterface $context)
         return $categories;
     }
 
+    /**
+     * @param Struct\ShopContextInterface $context $
```
shyim: Remove $ at end

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
```diff
@@ -187,23 +135,24 @@ public function getList(array $ids, Struct\ShopContextInterface $context)
...
+     *
+     * @return QueryBuilder
+     */
+    protected function getQuery(array $numbers, Struct\ShopContextInterface $context)
```
shyim, line 137: add return type

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
```diff
@@ -187,23 +135,24 @@ public function getList(array $ids, Struct\ShopContextInterface $context)
         return $categories;
     }
 
+    /**
+     * @param Struct\ShopContextInterface $context $
+     *
+     * @return QueryBuilder
```
shyim: remove phpdoc

engine/Shopware/Bundle/StoreFrontBundle/Gateway/CategoryGatewayInterface.php
```diff
@@ -45,11 +45,10 @@ public function getList(array $ids, Struct\ShopContextInterface $context);
      * - Assigned media object
      * - Core attribute of the media object
      *
-     * @param int $id
      *
      * @return Struct\Category
      */
-    public function get($id, Struct\ShopContextInterface $context);
+    public function get(int $id, Struct\ShopContextInterface $context);
```
uehler: this is a breaking change, isn't it?

lx-wnk: Sure but if you take a look at [this](https://github.com/shopware/shopware/blob/ad5ea26116917637ffac6d71a99a74180b69887f/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php#L89) it has been marked to be removed in 5.6. 
So on my opinion it should be handled like this and marked in the CHANGELOG.

uehler: the array handling is deprecated but if you add a type to $id you will afaik break every plugin which decorates that service because you changed the interface

lx-wnk: You are absolutely right. I changed that!

shyim:

PR #2143, opened 2019-06-28
Fix priority in event manager

hlohaus:

If you have many plugins the priority isn't right... 

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Event/EventManager.php
+++ b/engine/Library/Enlight/Event/EventManager.php
@@ -41,15 +41,31 @@ use Enlight\Event\SubscriberInterface;
 class Enlight_Event_EventManager extends Enlight_Class
 {
     /**
-     * @var Enlight_Event_Handler[] Contains all registered event listeners. A listener can be registered by the
+     * @var bool
+     */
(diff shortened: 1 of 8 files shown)


hlohaus:
i added a test, wich fail without my change.

soebbing:
Since this is a very important and basic functionality of Shopware, we'll need some more tests, eg. that adding events without any priority are kept in order, that multiple events WITH priorities are correctly sorted and a combination of both.

Also, to be able to add this functionality without breaking existing installations, it would be great if this new behavior could be activated by a config switch for the moment. That way users could test it in a save way before we activate it by default in a future version, similar to what we're currently doing with the order plugins are activated.

hlohaus:
I added a config switch and improved the tests.

hlohaus:
I added deprecations for #2151.
Also i fixed the removing of plugin handler.

shyim:
Hey @hlohaus,
when we do something on our Event Manager, we want to switch then to Symfony Events (#2151). So I will close this Pull Request then
Anyway thanks for your Contribution! :blue_heart:

PR #2144, opened 2019-07-03
Use ES prefix for backend index name

FiveDigital:
If the ES prefix is missing it's not possible to have multiple ES backend indices in the same ES Server.

It adds the ES prefix used for the frontend indices to the backend indices, too.

Build the ES backend index with and without this change.

https://issues.shopware.com/issues/SW-23540

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/EsBackendBundle/EsBackendIndexer.php
+++ b/engine/Shopware/Bundle/EsBackendBundle/EsBackendIndexer.php
@@ -33,6 +33,11 @@ class EsBackendIndexer
     const INDEX_NAME = 'backend_index';
 
     /**
+     * @var string
+     */
+    private $prefix;
+
+    /**
      * @var Client
      */
     private $client;
@@ -53,11 +58,13 @@ class EsBackendIndexer
     private $esVersion;
 
     public function __construct(
+        string $prefix,
         Client $client,
         \IteratorAggregate $repositories,
         EvaluationHelperInterface $evaluation,
         string $esVersion
     ) {
+        $this->prefix = $prefix;
         $this->client = $client;
         $this->repositories = $repositories;
         $this->evaluation = $evaluation;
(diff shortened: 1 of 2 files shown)


engine/Shopware/Bundle/EsBackendBundle/EsBackendIndexer.php
```diff
@@ -53,11 +58,13 @@ class EsBackendIndexer
     private $esVersion;
 
     public function __construct(
+        $prefix
```
shyim: string

engine/Shopware/Bundle/EsBackendBundle/services.xml
```diff
@@ -8,6 +8,7 @@
         <defaults public="true" />
 
         <service class="Shopware\Bundle\EsBackendBundle\EsBackendIndexer" id="shopware_es_backend.indexer">
+        	<argument>%shopware.es.prefix%</argument>
```
shyim, line 11: Maybe create a new prefix option in shopware.es.backend

FiveDigital: Maybe, let's see if the PR fits to you at all.

FiveDigital:
I won't find the time in the next 2 weeks because of holidays.

shyim:
Hey,
I am currently working on https://issues.shopware.com/issues/SW-23540. I will import your commit and do the last tasks myself. Thanks for your Contribution! :blue_heart: 
And Happy Holiday :blush:

FiveDigital:
Please also consider that the backend index does not use the shard settings here: https://github.com/shopware/shopware/blob/5.6/engine/Shopware/Configs/Default.php#L152-L153

Due to this the default ES sharding config will be used. That's not the way it's expected. Perhaps you can also fix that, too?

shyim:
@FiveDigital fyi: https://github.com/shopware/shopware/commit/728e29c0c6061c9429996ec42f54be38c3e8e041

FiveDigital:
Thank you! Looks fine 👍

PR #2145, opened 2019-07-08
Prefer core config values over plugins

pascalheidmann-bedarf:
When having a plugin with the same config key shopware can't distinguish those values from core configs as all are organised in the same table. The main differentiator is the plugin_id field on forms which is used for cleanup on plugin deinstallation.
As the config SQL query returns newer entries after older ones, plugins with the same config keys can override core fields.

EDIT: This is superseded by https://github.com/shopware/shopware/pull/2147 which replaces Zend_Db with Doctrine ModelManager and DBAL builder for improved readability.
You might close this PR and continue over at the newer PR.

Prefer core config (=> plugin_id = null) over plugins while still allowing custom config values be in the global config object.
As an alternative you might add a condition that prohibits all plugin values in the SQL query.

- Create a plugin with a config field called "mail" without default value
- install & activate plugin
- have lots of problems when sending mails which won't explicitly set the sender and therefore use the default one

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Config.php
+++ b/engine/Shopware/Components/Config.php
@@ -262,6 +262,7 @@ class Shopware_Components_Config implements ArrayAccess
      */
     protected function readData()
     {
+        // use order by on fetch query to prefer core configs over config values from plugins with the same name
         $sql = "
             SELECT
               LOWER(REPLACE(e.name, '_', '')) as name,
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Config.php
```diff
@@ -287,6 +287,8 @@ protected function readData()
 
             LEFT JOIN s_core_config_forms forms
               ON forms.id = e.form_id
+            
+            ORDER BY forms.plugin_id DESC
```
shopwareBot, line 292: Can you add a comment here to describe why this is important

mitelg:
@pascalheidman-bedarf is this PR obsolete now?

PR #2146, opened 2019-07-08
Fix fallback for backend order mail to use default sender

pascalheidmann-bedarf:
The backend order mail dialog passes both a `fromMail` and `fromName` to the backend order module when sending PDFs. If those values are broken, for example via https://github.com/shopware/shopware/pull/2145 those will be empty strings.
When then passing those empty strings to `\Enlight_Components_Mail` respective its parent `\Zend_Mail` there is no check for a valid mail address but a check for `null`. When passing an empty string this in turn sets the sender to this empty string and the `\Zend_Mail::$_defaultFrom` won't be used but also you even aren't able to set FROM header later anyway.
Some mail servers won't accept mails without a FROM header and therefore the mail will be lost.

The bug itself should be quite rare as with a default shopware installation the function should at least provide the default field data (`info@example.com`) but combined with the error described in https://github.com/shopware/shopware/pull/2145 this is a real pain ;)

With this PR shopware won't blindly pass empty strings to the mail object.

1. Set the `from` either via shopware backend or directly via `config.php` to an empty string
Alternative way: create a plugin that messes with config values by colliding with config key for `mail`
3. try sending document mails from backend order

- https://github.com/shopware/shopware/pull/1999
- https://github.com/shopware/shopware/pull/2145
- https://issues.shopware.com/issues/SW-23411

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Order.php
+++ b/engine/Shopware/Controllers/Backend/Order.php
@@ -998,9 +998,11 @@ class Shopware_Controllers_Backend_Order extends Shopware_Controllers_Backend_Ex
         } else {
             $mail->setBodyText($this->Request()->getParam('content', ''));
         }
+        $fromMail = $this->Request()->getParam('fromMail');
+        $fromName = $this->Request()->getParam('fromName');
 
-        $mail->setFrom($this->Request()->getParam('fromMail', ''), $this->Request()->getParam('fromName', ''));
-        $mail->addTo($this->Request()->getParam('to', ''));
+        $mail->setFrom($fromMail ?: null, $fromName ?: null);
+        $mail->addTo($this->Request()->getParam('to'));
 
         $mail->setAssociation(LogEntryBuilder::ORDER_ID_ASSOCIATION, $orderId);
 


soebbing:
Hi @pascalheidman-bedarf!

I tried to reproduce your issue on the current 5.6-branch by setting the following values in the `config.php`:
```
    'mail' => [
        'type' => 'smtp',
        'from' => [
            'email' => '',
            'name' => ''
        ]
    ],
```
Which didn't work, the values from the backend config still got pulled.

So I went do the database and changed the values to an empty string there, I still got the email. I changed the database value to a `null` and still got the mail.

I tried your code, got the mail and afterwards changed the code to check the params for empty strings, passing `null` into `setFrom` in that case: still I got the email.

If I use an empty string as the `to` Parameter, I get an error that the mail could not be send.

Basically, I cannot reproduce your issue. What was the behavior like after your change? It looks like it might be depended on your mail server.

pascalheidmann-bedarf:
@soebbing yep, this might be dependent on the mailserver configuration. For example our production mailserver will send mails with default sender (but not shopware's) or none if it is set to an empty string.
Will take another look at it, nonetheless this is basically hardening against bad input.

soebbing:
Since your other PR #2155 already removes the empty string defaults (see https://github.com/shopware/shopware/commit/d364250bae166e9c0768db75783efa5babc0e6b3#diff-eb6fedb01d116450d2e537647a1ca6c0R997), this PR has become unnecessary and can be closed. 

Thank you anyway! 💙

PR #2148, opened 2019-07-10
Consider additional text for variants in basket update

aragon999:
Currently when a variant product is added to the basket it is first inserted with the correct product name, but after the update, the name is updated without the additional information regarding variants. 
This change should be also reviewed in terms of performance. If this fix properly scales to large shops and so on, since the additonal text service always loads the complete variant configuration: https://github.com/shopware/shopware/blob/406925fbb6a8e98beb94e71ceca1fd473a6edc31/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/AdditionalTextService.php#L58-L88

Add the needed information.

Add a variant product to the cart.

\-

\-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2868,6 +2868,10 @@ SQL;
             array_column($additionalInformation, 'ordernumber'),
             $this->contextService->getShopContext()
         );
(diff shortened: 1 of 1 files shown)


shyim:

PR #2150, opened 2019-07-12
make active flag of categories translatable

dasistweb-gmbh:
hide the categories from menu and make not reachable via category controller

because there's no ability to hide categories (that are translated and used for more than one sub- or language shop) in one or more specific shops

make active flag of categories translatable and check the activity flag in the Core Category Service

-

-

category documentation

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/CategoryService.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/CategoryService.php
@@ -84,8 +84,14 @@ class CategoryService implements Service\CategoryServiceInterface
     {
         $customerGroup = $context->getCurrentCustomerGroup();
 
-        return array_filter($categories, function (Struct\Category $category) use ($customerGroup) {
+        $categories = array_filter($categories, function (Struct\Category $category) use ($customerGroup) {
             return !(in_array($customerGroup->getId(), $category->getBlockedCustomerGroupIds()));
         });
+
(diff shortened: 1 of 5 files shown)


engine/Shopware/Bundle/StoreFrontBundle/Service/Core/CategoryService.php
```diff
@@ -84,8 +84,14 @@ private function filterValidCategories($categories, Struct\ShopContextInterface
     {
         $customerGroup = $context->getCurrentCustomerGroup();
 
-        return array_filter($categories, function (Struct\Category $category) use ($customerGroup) {
+        $categories = array_filter($categories, function (Struct\Category $category) use ($customerGroup) {
             return !(in_array($customerGroup->getId(), $category->getBlockedCustomerGroupIds()));
         });
+
+        $categories = array_filter($categories, function (Struct\Category $category) use ($customerGroup) {
```
uehler, line 91: you have to remove the category.active = 1 condition in the category gateway, don't you? Otherwise you can only check for translated active flags on categories which are active

shyim:
Hey,
it's not so easy. When the active state is saved in s_core_translations serialized, you cannot filter the categories in SQL. This introduces then issues with limiting and offset.

So we need something like `s_articles_translations` for categories, todo the active check in SQL directly.

And did you checked following things:
* Inactive categories in sitemap
* Inactive categories in sitemap.xml
* Inactive categories in http cache warmup

We added recently a new option to blog categories to limit them to language shops. Maybe this topic could be extended and would solve your problem?

PR #2151, opened 2019-07-12
Symfony events

hlohaus:
###   1. Why is this change necessary?

Alternative to #2143. We use symfony as event listener list.

- You can now use symfony listener and subscribers.
- Adds symfony console events
- You can use enlight events in symfony and symfony events in enlight.
- HttpKernel added. So KernelEvents and smyfony controller are useable.

Breaks:
- Return values of the methods getListeners and getAllListeners in event manger changed from Enlight_Event_Handler to callable.
- Renamed getListener to getMethod in Enlight_Event_Handler_Plugin. getListener returns now a callable.

Example with symfony event subscriber:
 https://github.com/hlohaus/SheTest/blob/master/Subscribers/ConsoleEvents.php#L30

Example controller:

https://github.com/hlohaus/SheTest/blob/master/SheTest.php#L49

2x faster... 💃 

### 6. Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,6 +51,7 @@
         "symfony/options-resolver": "3.4.29",
         "symfony/expression-language": "3.4.29",
         "symfony/web-link": "3.4.29",
+        "symfony/event-dispatcher": "3.4.29",
         "monolog/monolog": "1.23.0",
         "doctrine/common": "2.10.0",
         "doctrine/collections": "1.5.0",
--- a/engine/Shopware/Kernel.php
+++ b/engine/Shopware/Kernel.php
@@ -24,7 +24,6 @@
 
 namespace Shopware;
 
-use Enlight_Controller_Request_RequestHttp as EnlightRequest;
 use Shopware\Bundle\AttributeBundle\DependencyInjection\Compiler\StaticResourcesCompilerPass;
 use Shopware\Bundle\BenchmarkBundle\DependencyInjection\Compiler\MatcherCompilerPass;
 use Shopware\Bundle\ContentTypeBundle\DependencyInjection\RegisterDynamicController;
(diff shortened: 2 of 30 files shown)


engine/Shopware/Kernel.php
```diff
@@ -54,6 +54,7 @@
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\ParameterBag\ParameterBag;
+use Symfony\Component\EventDispatcher\DependencyInjection\RegisterListenersPass;
```
shyim, line 56: Please add a new require to the composer.json, this works currently cause an other dependency require event-dispatcher

composer.json
```diff
@@ -81,7 +81,8 @@
         "league/flysystem-aws-s3-v3": "1.0.19",
         "superbalist/flysystem-google-storage": "6.0.0",
         "stecman/symfony-console-completion": "0.10.1",
-        "beberlei/doctrineextensions": "1.2.1"
+        "beberlei/doctrineextensions": "1.2.1",
+        "symfony/event-dispatcher": "^4.3"
```
shyim: sort it to the other symfony packages 😅

engine/Shopware/Components/Controller.php
```diff
@@ -0,0 +1,78 @@
...
+    private $container;
+
+    public function __construct(ContainerInterface $container)
+    {
+        $this->container = $container;
+    }
+
+    public function dispatch(Request $request): Response
+    {
+        if (!$request instanceof \Enlight_Controller_Request_RequestHttp) {
```
hlohaus: Maybe we cab change the request object in shopware.php. Then it will not be converted. @shyim

build/build.xml
```diff
@@ -238,11 +238,15 @@
         <exec executable="${script.phpunit}" failonerror="true" dir="${test.dir.shopware}">
             <arg value="--config=phpunit_unit.xml.dist"/>
             <arg value="--exclude-group=elasticSearch"/>
+            <arg value="--stop-on-failure"/>
```
shyim: Can you revert this changes

composer.json
```diff
@@ -51,6 +51,7 @@
         "symfony/options-resolver": "3.4.29",
         "symfony/expression-language": "3.4.29",
         "symfony/web-link": "3.4.29",
+        "symfony/event-dispatcher": "^4.3",
```
shyim: Please use 3.4.29 like the other packages from symfony

PascalThesing:
@hlohaus have you tested the different hooks?

hlohaus:
I have linked in description a Plugins with Symfony Events. Or wich hooks do you krank?

PascalThesing:
I am thinking about Shopware hooks. They use the same syntax e.g. 'Shopware_Controllers_Frontend_Checkout::confirmAction::after'

hlohaus:
Yes, they should work. There are tests

shyim:
Hey,
we discussed your PR. Can you split the Events part and Httpkernel part into own Pull Requests?
Currently we are not sure how many Symfony we want to integrate into Shopware 5, so we can discuss that part later on.
We currently don't want two dispatch systems and two ways to do the same thing.

PR #2152, opened 2019-07-12
Add getter and setter to Config\Form model

hrombach:
For field \Shopware\Models\Config\Form::$plugin

Currently, if you have an instance of `\Shopware\Models\Config\Form` and you want the associated `\Shopware\Models\Plugin\Plugin`, you have to get the id and fetch the Plugin model manually

It adds `\Shopware\Models\Config\Form::getPlugin` and `\Shopware\Models\Config\Form::setPlugin`

Fetch an instance of `\Shopware\Models\Config\Form` and try to get the associated `\Shopware\Models\Plugin\Plugin`?

### DIFF
--- a/engine/Shopware/Models/Config/Form.php
+++ b/engine/Shopware/Models/Config/Form.php
@@ -27,6 +27,7 @@ namespace Shopware\Models\Config;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\ORM\Mapping as ORM;
 use Shopware\Components\Model\ModelEntity;
+use Shopware\Models\Plugin\Plugin;
 
 /**
  * @ORM\Table(name="s_core_config_forms")
@@ -35,7 +36,7 @@ use Shopware\Components\Model\ModelEntity;
 class Form extends ModelEntity
 {
     /**
-     * @var \Shopware\Models\Plugin\Plugin
+     * @var Plugin|null
      *
      * @ORM\ManyToOne(targetEntity="Shopware\Models\Plugin\Plugin", inversedBy="configForms")
      * @ORM\JoinColumn(name="plugin_id", referencedColumnName="id")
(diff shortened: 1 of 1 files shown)


engine/Shopware/Models/Config/Form.php
```diff
@@ -387,4 +388,19 @@ public function hasTranslations()
     {
         return $this->translations->count() > 0;
     }
+
+    public function getPlugin(): Plugin
```
shyim: this can be nullable

engine/Shopware/Models/Config/Form.php
```diff
@@ -387,4 +388,19 @@ public function hasTranslations()
...
+    {
+        return $this->plugin;
+    }
+
+    /**
+     * @return static
+     */
+    public function setPlugin(Plugin $plugin): self
```
shyim: nullable setter

engine/Shopware/Models/Config/Form.php
```diff
@@ -387,4 +388,19 @@ public function hasTranslations()
...
         return $this->translations->count() > 0;
     }
+
+    public function getPlugin(): Plugin
+    {
+        return $this->plugin;
+    }
+
+    /**
+     * @return static
```
shyim, line 398: remove phpdoc

hrombach: `@return static` will improve code completion and static analysis in case the class is extended though.

hrombach:
@shyim I went ahead and added the nullable type hints and also edited the phpdoc to keep things consistent.

engine/Shopware/Models/Config/Form.php
shyim: We don't add phpdocs anymore when the types already says everything.

hrombach: But the type declarations do not say everything. 

`@return static` does not trigger php-cs-fixer's `no_superfluous_phpdoc_tags` check because it instructs IDEs and static analyzers to respect late static binding when resolving return types.

shyim:
Hey,
can you squash your commits?

hrombach:
Done :)

shyim:

PR #2147, opened 2019-07-10
Use Doctrine DBAL instead of Zend_Db in Core Config

pascalheidmann-bedarf:
### 1. Why is this change necessary?
Dependency reduction and improved readability by using Doctrine DBAL builder instead of plain sql.
Plus this PR contains https://github.com/shopware/shopware/pull/2145 (preferring core config over plugin config values)

### 2. What does this change do, exactly?
- Replacing db reference in core config with reference of Doctrine's DBAL Connection
- Using DBAL Builder instead of plain sql
- preferring core config over plugin config values by sorting by plugin_id (core configs are `null`)

### 3. Describe each step to reproduce the issue or behaviour.
For core config over plugin config see https://github.com/shopware/shopware/pull/2145 

### 4. Please link to the relevant issues (if any).
https://github.com/shopware/shopware/pull/2145

### 6. Checklist

- [x] I have written tests and verified that they fail without my change
- [x] I have squashed any insignificant commits
- [x] This change has comments for package types, values, functions, and non-obvious lines of code
- [x] I have read the contribution requirements and fulfil them.

PR #2154, opened 2019-07-15
Prefer Core config over Plugin config values

pascalheidmann-bedarf:
Plugin config values pollute the core config but even worse can override those values as those are typical newer in the database.

Based on #2147 this will ~remove all plugin config completely from Core config~ prefer Core config values over plugin values.

See #2145 

- #2145 
- #2147

Devs should be aware that they shouldn't use the Core config for accessing plugin config. For now only on key collision changes will occur (but those were already faulty before).
Next step is completely remove as originally proposed but as this will be definitely BC this will be another change for 5.7. (will make another PR for 5.7 when this is accepted)

~While this shouldn't be done in the first place plugin devs can't use the core config (`Shopware()->Config()`) anymore for accessing their config values.
As this is a potential BC but 5.6 isn't released yet this shouldn't be a problem.~

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Config.php
+++ b/engine/Shopware/Components/Config.php
@@ -280,7 +280,10 @@ class Shopware_Components_Config implements ArrayAccess
                 'parentShop.element_id = e.id AND parentShop.shop_id = :parentShopId')
             ->leftJoin('e', 's_core_config_values', 'fallbackShop',
                 'fallbackShop.element_id = e.id AND fallbackShop.shop_id = :fallbackShopId')
-            ->leftJoin('e', 's_core_config_forms', 'forms', 'forms.id = e.form_id');
+            ->leftJoin('e', 's_core_config_forms', 'forms', 'forms.id = e.form_id')
+            // prefering core config over plugin
+            // going for @deprecated for plugins with SW 5.7 by removing plugin data completely
+            ->orderBy('forms.plugin_id', 'DESC');
 
         $builder->setParameters([
             'fallbackShopId' => 1, // Shop parent id
--- a/tests/Functional/Components/Config/ShopwareConfigTest.php
+++ b/tests/Functional/Components/Config/ShopwareConfigTest.php
@@ -1,0 +1,117 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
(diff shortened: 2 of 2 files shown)


pascalheidmann-bedarf:
As it turns out the plugin Statistics (`/Default/Frontend/Statistics/`) is using the core config.

I fixed those in #2156

hlohaus:
The unit tests fails, because it' breaking... :/

engine/Shopware/Components/Config.php
```diff
@@ -262,39 +263,32 @@ protected function load()
      */
     protected function readData()
     {
-        $sql = "
-            SELECT
-              LOWER(REPLACE(e.name, '_', '')) as name,
+        $builder = $this->_db->createQueryBuilder();
+        $builder
+            ->select('LOWER(REPLACE(e.name, "_", "")) as name,
```
shyim: Please add multiple addSelect instead one big select

soebbing:
As the RC1 of Shopware 5.6 has already been released, a BC for that release isn't an option anymore. The earliest we can merge this would be the start of the 5.7 branch.

And since this change breaks the build, please modify the tests accordingly. Please make sure the Core-Config works as intended (only Core-configs should be considered, no plugin config), as well as the other way around in the relevant Configreader service.

pascalheidmann-bedarf:
@soebbing the test itself is correct but the PR breaks a legacy plugin (see #2156).

Would be the alternative - preferring non plugin values over plugin values instead of removing those completely - be an acceptable way to get this bug fixed in 5.6 with removing those completely in 5.7?

As shown creating a plugin with config key "mail" breaks the core config "mail" value which should classify this as a really annoying bug and therefore be patched asap.

soebbing:
@pascalheidman-bedarf I think simply preferring the core config values would be a better and more compatible solution. It might be a step towards ONLY returning core configs, but I do not see a way to throw deprecation messages automatically if the service is used for plugin values.

pascalheidmann-bedarf:
@soebbing can you remove the `Missing Tests` label as the fix as the PR now only prefers core config over plugin config and therefore the tests aren't broken anymore?

soebbing:
You're right, I'll remove the tag.

soebbing:
Sorry, I forgot something: Could you please add a test that makes sure that the intended behaviour now works? What I mean is: Test that a plugin config with the same name as a core config still returns the value of the core. Also: A plugin config that doesn't exist in the core still get's returned.

pascalheidmann-bedarf:
I'm irritated by the existing functional test at `/tests/Function/Components/Config/ConfigTest.php` as it isn't testing the Component but the `Enlight_Config` class and therefore should be under libs I guess?
That's the reason I created a new test class with maybe not the best fitting name.

tests/Functional/Components/Config/ShopwareConfigTest.php
```diff
@@ -0,0 +1,125 @@
...
+        Shopware()->Container()->reset('config');
+    }
+
+    /**
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    protected function tearDown()
+    {
+        $sql = <<<SQL
+DELETE s_core_config_forms, s_core_config_form_translations, s_core_config_elements, s_core_config_element_translations, s_core_config_values
```
shyim: Please use Shopware\Tests\Functional\Traits\DatabaseTransactionBehaviour instead

pascalheidmann-bedarf: Do I just have to use the trait and am good to go?

soebbing: Adding this trait wraps each test case in a transaction, so normal database actions are reverted after each test case.

shyim:
It has been merged into 5.7. We don't want such change in a patch version

PR #2155, opened 2019-07-15
Adding event to Shopware_Controllers_Backend_Order::sendMailAction to make it extendable

pascalheidmann-bedarf:
Right now you can't modify mails send via backend to the customer without great pain.
For example:
- always add your additional ToS documents to the mail if an invoice is attached
- add additional receivers for specific documents
- ...

All passed request data are now published via an event so devs can use it to modify those parameters and therefore adding/removing recipients, attachments etc.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Order.php
+++ b/engine/Shopware/Controllers/Backend/Order.php
@@ -989,18 +989,42 @@ class Shopware_Controllers_Backend_Order extends Shopware_Controllers_Backend_Ex
         }
 
         $mail = clone $this->container->get('mail');
-        $mail = $this->addAttachments($mail, $orderId, $attachments);
         $mail->clearRecipients();
-        $mail->setSubject($this->Request()->getParam('subject', ''));
 
-        if ($this->Request()->getParam('isHtml')) {
-            $mail->setBodyHtml($this->Request()->getParam('contentHtml', ''));
+        $mailData = [
+            'attachments' => $attachments,
+            'subject' => $this->Request()->getParam('subject', ''),
+            'fromMail' => $this->Request()->getParam('fromMail'),
+            'fromName' => $this->Request()->getParam('fromName'),
+            'to' => [$this->Request()->getParam('to')],
+            'isHtml' => $this->Request()->getParam('isHtml'),
+            'bodyHtml' => $this->Request()->getParam('contentHtml', ''),
+            'bodyText' => $this->Request()->getParam('content', ''),
+        ];
+
+        /** @var Enlight_Event_EventManager $events */
+        $events = $this->get('events');
+        $mailData = $events->filter(
+            'Shopware_Controllers_Order_SendMail_Prepare',
+            $mailData,
+            [
+                'subject' => $this,
+                'orderId' => $orderId,
+                'mail' => $mail,
+            ]
+        );
+
+        $mail->setSubject($mailData['subject']);
+
+        $mail->setFrom($mailData['fromMail'], $mailData['fromName']);
+        $mail->addTo($mailData['to']);
+
+        if ($mailData['isHtml']) {
+            $mail->setBodyHtml($mailData['bodyHtml']);
         } else {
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Order.php
```diff
@@ -989,18 +989,42 @@ public function sendMailAction()
...
+        $mail->setSubject($mailData['subject']);
+
+        $mail->setFrom($mailData['fromMail'], $mailData['fromName']);
+        $mail->addTo($mailData['to']);
+
+        if ($mailData['isHtml']) {
+            $mail->setBodyHtml($mailData['bodyHtml']);
         } else {
-            $mail->setBodyText($this->Request()->getParam('content', ''));
+            $mail->setBodyText($mailData['content']);
```
shyim: $mailData['content'] is not defined :see_no_evil:

mitelg: should be `bodyText`

PascalThesing:
Dear Pascal,

it is already possible to extend this action in Shopware 5.6.
You could use a decorator as described here: https://github.com/shopware/devdocs/commit/89715dc2d3dd7f8134d97040615610a868a35c2d#diff-c86cf8140bd62b8392b3db269fa0f8dcR220

> Since Shopware 5.6 it is also possible to extend or exchange Shopware controllers over the DI Container. You can find an example of decorating a service here.

Information about decorating a services can be found here: https://developers.shopware.com/developers-guide/plugin-quick-start/#decorate-a-service

mitelg:
@PascalThesing but with a decoration you have to copy-paste nearly the whole method to just change a bit of the logic. A new event fits better here, in my opinion :blush:

PascalThesing:
@mitelg Then the logic should be split up into different methods and moved to a service.

mitelg:
@PascalThesing that is probably another topic :wink:

hlohaus:
you can modify the request or hook the sendStatusMail method. A event is not really necessary

pascalheidmann-bedarf:
@hlohaus I strongly disagree.
Hooking would mean copying the whole function and if you have several plugins that modify, those will ~overwrite each other~ send several mails with the same content.
Also I'm not sure how you would get any other attachments besides order documents attached (see my first example) by modifying the request object.

In the end only with the event approach it is possible to have all these use cases covered as long as you don't go the full route to encapsulate the whole mail preparation logic into its own decoratable service (which feels a bit overkill for just one single mail creation).

soebbing:
Just fyi: Replace Hooks work just like decorations, all of them are called

uehler:
all those years everybody was crying for more events, now somebody creates a new one and everybody loses their mind :D

mitelg:
I really don't see any problem here. only because there is already a (hacky) way to achieve what @pascalheidman-bedarf wants, we should not implement a new event??

For me, this is a normal quick-pick :blush:

engine/Shopware/Controllers/Backend/Order.php
```diff
@@ -989,18 +989,42 @@ public function sendMailAction()
...
+        if ($mailData['isHtml']) {
+            $mail->setBodyHtml($mailData['bodyHtml']);
         } else {
-            $mail->setBodyText($this->Request()->getParam('content', ''));
+            $mail->setBodyText($mailData['bodyText']);
         }
-
-        $mail->setFrom($this->Request()->getParam('fromMail', ''), $this->Request()->getParam('fromName', ''));
-        $mail->addTo($this->Request()->getParam('to', ''));
+        $mail = $this->addAttachments($mail, $orderId, $mailData['attachments']);
```
pascalheidmann-bedarf, line 1027: Just realised that in my original PR I simply passed `$attachments` instead of the potential modified array.

shyim: Thanks for the hint, I will update the internal pr

PR #2156, opened 2019-07-15
Old default plugins: Use plugin config instead of relying in global core config object

pascalheidmann-bedarf:
The default `Statistics` and `Cron` plugin both use the core Config object instead of relying on its plugin config values. As the core config is vulnerable to be overridden by plugin values (see #2154 for a fix) this should be fixed.

This PR blocks #2154 as the related unit tests fail.

Replaces calls to the global core config object to the plugin calls. As those plugins use the legacy file structure they call the `Config()` method in their `bootstrap.php` files instead of a more modern `shopware.plugin.config_reader`.

#2154 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/Statistics/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Frontend/Statistics/Bootstrap.php
@@ -108,7 +108,7 @@ ShopWiki;Bot;WebAlta;;abachobot;architext;ask jeeves;frooglebot;googlebot;lycos;
             return false;
         }
         $result = false;
-        $bots = preg_replace('/[^a-z;]/', '', strtolower(Shopware()->Config()->botBlackList));
+        $bots = preg_replace('/[^a-z;]/', '', strtolower($this->Config()->get('botBlackList')));
         $bots = explode(';', $bots);
         if (!empty($userAgent) && str_replace($bots, '', $userAgent) != $userAgent) {
             $result = true;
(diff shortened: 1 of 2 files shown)


engine/Shopware/Plugins/Default/Frontend/Statistics/Bootstrap.php
```diff
@@ -108,7 +108,7 @@ public function checkIsBot($userAgent)
             return false;
         }
         $result = false;
-        $bots = preg_replace('/[^a-z;]/', '', strtolower(Shopware()->Config()->botBlackList));
```
shyim, line 111: $this->Config()->get(...)

pascalheidmann-bedarf: Those old Bootstrap files... :D

engine/Shopware/Plugins/Default/Frontend/Statistics/Bootstrap.php
```diff
@@ -173,8 +173,8 @@ public function shouldRefreshLog(Enlight_Controller_Request_Request $request)
         ) {
             return false;
         }
-        if (!empty(Shopware()->Config()->blockIp)
-            && strpos(Shopware()->Config()->blockIp, $request->getClientIp()) !== false
+        if (!empty($this->getConfig()['blockIp'])
```
shyim: $this->getConfig() does not exist :see_no_evil:

shyim:

PR #2157, opened 2019-07-18
Add router cache

hlohaus:

https://gitter.im/shopware/offtopic?at=5d2edce97dbfff3242c2552f

It's cache the result of rewriteList.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Routing/RewriteGenerator/CachedRepository.php
+++ b/engine/Shopware/Components/Routing/RewriteGenerator/CachedRepository.php
@@ -1,0 +1,115 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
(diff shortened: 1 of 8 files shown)


mitelg:
hey @hlohaus 
the tests are failing with your change

engine/Shopware/Components/Routing/RewriteGenerator/CachedRepository.php
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+namespace Shopware\Components\Routing\RewriteGenerator;
+
+use Doctrine\DBAL\Query\QueryBuilder;
+use Zend_Cache_Core as Cache;
+
+class CachedRepository implements RepositoryInterface
+{
+    private $innerService;
```
shyim, line 35: phpdoc

tests/Functional/Bundle/StoreFrontBundle/HrefLangTest.php
```diff
@@ -43,6 +43,8 @@ protected function setUp()
         // Easier to see english link
         Shopware()->Db()->executeQuery('UPDATE s_core_shops SET base_url = "/en", category_id = 3 WHERE id != 1');
 
+        Shopware()->Container()->get('cache')->clean();
```
shyim: Maybe add a config.php parameter to enable or disable the cache and disable it in cache

engine/Shopware/Components/Routing/RewriteGenerator/Repository.php
```diff
@@ -0,0 +1,74 @@
...
+     * @var Connection
+     */
+    protected $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    public function getQueryBuilder(): QueryBuilder
```
shopwareBot: This has to be in a own service, otherwise decoration does not work here

hlohaus: It should work. you have to implement two methods.

soebbing: Point is: These methods don't help you in that case. Even worse: It's unnecessary! Let's say I want to implement a Redis Repository, that will have to provide a `getQueryBuilder()` method! Better to implement something like a `RewriteRepositoryHelper`-service that is responsible for providing this `QueryBuilder` instance. You can inject that service into your decoration as well.

PR #2159, opened 2019-07-22
Proper attribute entity selection for \Shopware\Models\Order\Detail

pascalheidmann-bedarf:
If you reference `\Shopware\Models\Order\Detail` as an entity for an attribute you will get the default selectors that will fall back to the model field `number` - which is the ordernumber but nothing identifying the detail position itself. 

Added new attribute field handler with the most important data fields.

1. Create new attribute with entity set to `\Shopware\Models\Order\Detail` and `displayInBackend=true`
2. Open the parent entity and try to edit the association
3. Association just contains order number

### 6. Checklist


shyim:

PR #2161, opened 2019-07-23
fixed sAddAccessoriesQuantity

Stapelchips:
When adding accessories to the cart, the value of the "sAddAccessoriesQuantity" parameter is not treated as an array, despite the note in Line 511 explicitly saying:
 
"* @param sAddAccessoriesQuantity = List of article quantities separated by ;"

So this is a function that is expected here, but is never executed because it is not implemented.

The value of "sAddAccessoriesQuantity" gets exploded, just like the first parameter "sAddAccessories".

None - in fact, the documentation is incorrect until this change is done.

### 6. Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,10 +5,14 @@ Please make sure to fulfil our contribution guideline (https://developers.shopwa
 -->
 
 ### 1. Why is this change necessary?
+When adding accessories to the cart, the value of the "sAddAccessoriesQuantity" parameter is not treated as an array, despite the note in Line 511 explicitly saying:
+ 
+"* @param sAddAccessoriesQuantity = List of article quantities separated by ;"
 
+So this is a function that is expected here, but is never executed because it is not implemented.
 
 ### 2. What does this change do, exactly?
-
+The value of "sAddAccessoriesQuantity" gets exploded, just like the first parameter "sAddAccessories".
 
 ### 3. Describe each step to reproduce the issue or behaviour.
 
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,10 +5,14 @@ Please make sure to fulfil our contribution guideline (https://developers.shopwa
 -->
 
 ### 1. Why is this change necessary?
+When adding accessories to the cart, the value of the "sAddAccessoriesQuantity" parameter is not treated as an array, despite the note in Line 511 explicitly saying:
```
shyim, line 8: Can you revert this?

engine/Shopware/Controllers/Frontend/Checkout.php
```diff
@@ -1608,6 +1608,10 @@ private function addAccessories($accessories, $quantities)
             $accessories = explode(';', $accessories);
         }
 
+        if (is_string($quantities)) {
```
mitelg, line 1611: @soebbing @shyim what do think about deprecating one of the two possible input parameter types? only allowing arrays or strings

.github/PULL_REQUEST_TEMPLATE.md
Stapelchips: I'm sorry, I don't know how. :/

shyim: Maybe checkout https://www.atlassian.com/git/tutorials/resetting-checking-out-and-reverting
or easier create a new Pull Request :)

engine/Shopware/Controllers/Frontend/Checkout.php
shyim: I would prefer then strings, because we use it currently already for $accessories

PR #2162, opened 2019-07-23
Remove double aria label

xf-:
Don't use aria-label attribute twice in on node

removed one

### 6. Checklist


shyim:

PR #2163, opened 2019-07-23
Replace tabs with spaces

xf-:
It is annoying to compare sources while shopware stats switching in cart_footer.tpl & confirm_footer.tpl to tabs as only files in project. Also removed last spaces in mixed non vendor files. jamine lib has also **one** single tab in complete file :rage:

Stops insanity! Respect the .editorconfig!

### 6. Checklist

PS: Please check commits for regex \t :)


shyim:
Thanks for your Contribtion! :blue_heart:

PR #2164, opened 2019-07-23
Allow resize on emotion detail window

xf-:
You you open the Backend on a large screen and size is not changeable. Not via close ExtJS window and open it again. Also trick of resizing the OS window won't work.

**Now**
- You detach from your dock and use the 13/15" Notebook screen with lower resultion.
- Same for 720p beamer or 1080 if you use a QHD or 4k display to develop.
- Screen sharing (resenting a backend or discussion an issue). You adjust the window size to Resolution on screen or bandwidth limitation.
- You zoom into the browser window

**Result**
 No way to reach the save button.

Allows resize of the ExtJs window

See 1.

### 6. Checklist


shyim:

PR #2167, opened 2019-07-30
[AJAX-Search] prevent session blocking in slower requests

tinect:
In special environments and specified search-terms the search is slow. The input will fire the ajax search, but submit won't abort the ajax-search.
before

- will abort ajax-search on submit
- won't fire ajax search while submit is in progress.
image

see 1

### 6. Checklist


shyim:

PR #2169, opened 2019-08-02
Fix setOneToMany on model entities

felixbrucker:
Currently some model entities using `setOneToMany` to set a `Collection` property do not initialize these properties and thus `setOneToMany` can throw an exception. Furthermore `setOneToMany` currently does not handle `Collection` as an argument, which breaks the defined typehint.

This change adds a constructor to model entities with uninitialized `Collection` properties which use  `setOneToMany` in their setter and allows `Collection`'s to be used as an argument for `setOneToMany`.

https://github.com/shopware/shopware/blob/c8e95ecd1ba645303d1b17322fe701862e96600e/tests/Unit/Components/Model/ModelEntityTest.php#L485-L492

https://github.com/shopware/shopware/blob/c8e95ecd1ba645303d1b17322fe701862e96600e/tests/Unit/Components/Model/ModelEntityTest.php#L500-L507

n/a

n/a

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Model/ModelEntity.php
+++ b/engine/Shopware/Components/Model/ModelEntity.php
@@ -169,10 +169,10 @@ abstract class ModelEntity
      * <li>So the parameter expect <b>"customer"</b></li>
      * </ul>
      *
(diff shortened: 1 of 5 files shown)


shyim:

PR #2172, opened 2019-08-08
Article detail page - ajax variants - use replaceState instead of pushState for better UX

mkreusch:
Currently selecting variants on the article detail page produces browser history entries that no user will want because they expect to be taken back to the listing page when they use the corresponding browser functionality

Just replace the history.pushState() method by history.replaceState() so the URL still changes for bookmarks but the browser history stays clean.

Open an article with variants and play around with the configurator options. Then try to go back to the listing. You will notice that you have to step through each single selection you have tried.

### 6. Checklist


tinect:
Could be interesting for listing-actions, too. 🤔

PR #2173, opened 2019-08-08
Allow Customer & Address API to use country ISO 3166 code instead of ID

pascalheidmann-bedarf:
Interoperability & flexibility. ISO codes are a standard that nearly every system "speaks".

Check if the parameter _country_ is potential an iso string and use it for lookup.
If we have a valid country we can also use the state ISO 3166-2 code instead of an ID.

Mention the possibility to use a `string` instead of `int` for field `country` in those APIs. 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Address.php
+++ b/engine/Shopware/Components/Api/Resource/Address.php
@@ -26,8 +26,6 @@ namespace Shopware\Components\Api\Resource;
 
 use Shopware\Bundle\AccountBundle\Service\AddressServiceInterface;
 use Shopware\Components\Api\Exception as ApiException;
-use Shopware\Models\Country\Country;
-use Shopware\Models\Country\State;
 use Shopware\Models\Customer\Customer as CustomerModel;
 use Shopware\Models\Shop\Repository;
 use Shopware\Models\Shop\Shop as ShopModel;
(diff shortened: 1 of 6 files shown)


engine/Shopware/Components/Api/Resource/Address.php
```diff
@@ -271,8 +271,23 @@ private function prepareAddressData(array $data, $customerId = null, $filter = f
...
+                ->select('country')
+                ->from(Country::class, 'country')
+                ->where('country.iso3 = :iso or country.iso = :iso')
+                ->setParameter('iso', $data['country']);
+            $data['country'] = $builder->getQuery()->getSingleResult();
+        } else {
+            $data['country'] = null;
+        }
+
+        $data['state'] = !empty($data['state']) ? $em->find(State::class, $data['state']) : null;
```
mitelg: is there something similar for states?

pascalheidmann-bedarf: Kind of.
There is ISO 3166-2 which defines more than 4k different subnational administrative zones aka states. Inside SW core there are just 16 states of Germany as well as 50 states of the USA defined, with the potential of collisions of their shortcode.

When adding support for states (good point by the way) this has to go in tandem with the prior set country as only the combination is unique.
This would mean setting just a state via id can also provide a country but setting a state via shortcode and no country has to result in `null` for both country and state.

pascalheidmann-bedarf:
Added state lookup as suggested by @mitelg

engine/Shopware/Components/Api/Resource/Address.php
mitelg: nice solution :+1:

pascalheidmann-bedarf:
@shopwareBot adding a new field like you proposed seems like a good idea to avoid that problem with the union data type. As this requires defining a priority I would suggest `ID` > ISO country code > state ID.

The additionally proposed new API seems also useful but would be more a new PR.

pascalheidmann-bedarf:
@shopwareBot would adding new fields `countryId` and `countryIso` and possibly deprecating the `country` field be an option, with country behaving like my first version being a viable solution?
That way the fields meaning is clearer going forward.

Adding additional validation (passing both `countryId` and `countryIso` results in an `Exception` if not matching) should do the rest (with documentation of course)

PR #2175, opened 2019-08-12
Make plugin DBALConfigReader readable

pascalheidmann-bedarf:
Readability

Replace plain SQL with DBALQueryBuilder in plugin's `DBALConfigReader`.

Same as https://github.com/shopware/shopware/pull/2147 but this time for plugins instead of core config

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Plugin/DBALConfigReader.php
+++ b/engine/Shopware/Components/Plugin/DBALConfigReader.php
@@ -46,43 +46,27 @@ class DBALConfigReader implements ConfigReader
      */
     public function getByPluginName($pluginName, Shop $shop = null)
     {
-        $sql = <<<'SQL'
-SELECT
-  ce.name,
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Plugin/DBALConfigReader.php
```diff
@@ -46,43 +46,27 @@ public function __construct(Connection $connection)
      */
     public function getByPluginName($pluginName, Shop $shop = null)
     {
-        $sql = <<<'SQL'
```
bcremer, line 49: How exactly is a DBAL query builder more readable for a query that does not contain any dynamic  or conditional parts?
In contrast to https://github.com/shopware/shopware/pull/2147 this code is already using the DBAL connection instead of `Zend_Db`.

pascalheidmann-bedarf: It might be more a personal taste but heredocs look bad with their indentation.

PR #2176, opened 2019-08-12
Use fallback for widget label translation

pascalheidmann-bedarf:
Not having a translation for a given language results in _NULL_ for the label making it hard to identify the widget.

Adds a fallback chain to resolve to the internal widget name instead returning _NULL_ if no translation exists.

1. Create a plugin which adds an widget:
```
$plugin = $context->getPlugin();
$widget = new Widget();
$widget->setName('my-widget');
$widget->setPlugin($plugin);
$plugin->getWidgets()->add($widget);
```
As you already see the label prop is phased out and replaced by the better fitting snippet management.

2. Now install the new plugin, reload your backend and try to add the widget. Instead of a name _null_ will be provided. 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Widgets.php
+++ b/engine/Shopware/Controllers/Backend/Widgets.php
@@ -59,14 +59,16 @@ class Shopware_Controllers_Backend_Widgets extends Shopware_Controllers_Backend_
 
         $data = $builder->getQuery()->getArrayResult();
 
+        $snippets = $this->get('snippets')->getNamespace('backend/widget/labels');
         $widgets = [];
+
         foreach ($data as &$widgetData) {
             if (!$this->_isAllowed($widgetData['name'], 'widgets')) {
                 continue;
             }
 
-            $widgetData['label'] = Shopware()->Container()->get('snippets')->getNamespace('backend/widget/labels')
-                ->get($widgetData['name'], $widgetData['label']);
+            // fallback: translation -> name
+            $widgetData['label'] = $snippets->get($widgetData['name'], $widgetData['name']);
 
             $widgets[] = $widgetData;
         }


engine/Shopware/Controllers/Backend/Widgets.php
```diff
@@ -59,14 +59,16 @@ public function getListAction()
...
+
         foreach ($data as &$widgetData) {
             if (!$this->_isAllowed($widgetData['name'], 'widgets')) {
                 continue;
             }
 
-            $widgetData['label'] = Shopware()->Container()->get('snippets')->getNamespace('backend/widget/labels')
-                ->get($widgetData['name'], $widgetData['label']);
+            // fallback: translation -> label -> name
+            $widgetData['label'] = $snippets->get($widgetData['name'], $widgetData['label']) ?: ($widgetData['label'] ?: $widgetData['name']);
```
shyim: the name should be only "name". $widgetData['label'] does not exist

pascalheidmann-bedarf: Oh, yeah, you are right. Thought just the setter was removed and label used in legacy mode as with `s_core_states`...
But this means I can simplify the code a lot more - see force push'd update

engine/Shopware/Controllers/Backend/Widgets.php
```diff
@@ -59,14 +59,16 @@ public function getListAction()
 
         $data = $builder->getQuery()->getArrayResult();
 
+        $snippets = Shopware()->Container()->get('snippets')->getNamespace('backend/widget/labels');
```
mitelg:
```suggestion
        $snippets = $this->get('snippets')->getNamespace('backend/widget/labels');
```

shyim:

PR #2177, opened 2019-08-14
Better i18n support for emotion elements

pascalheidmann-bedarf:
The use of direct naming emotion elements is a mess. Several elements don't have a correct translation and adding own elements require to use either the name directly or hack snippets into the extjs model.
Adding another problem with the names: even the author of the docs for creating custom emotion elements don't know how they are handled as the [example](https://developers.shopware.com/developers-guide/custom-shopping-world-elements/#registering-a-new-element) will always recreate the same element.

- Move snippets management to the data layer -> snippets will be fetched inside the controller action instead of template. Fetching logic is backwards compatible so plugins still can use the old code.
- renaming default shopping world elements to a more useful internal name (the already used key for snippets)
- making the `name` field of emotion elements unique
- Adding fallback solution for `$componentName` in the function `Shopware\Components\Emotion\ComponentInstaller::createOrUpdate()` so you can leave out the `name` parameter and avoid having inconsistent data and therefore problems with new elements

Use the code [from the custom shopping world example](https://developers.shopware.com/developers-guide/custom-shopping-world-elements/#registering-a-new-element) several times to have the element multiple times.

https://issues.shopware.com/issues/SW-24441

- https://developers.shopware.com/developers-guide/custom-shopping-world-elements/#registering-a-new-element should be modified (even without this PR, see issue/this PR)
- snippet namespace should be mentioned: `backend/emotion/view/detail`
- (internal) shopping world element name has to be unique (technically per plugin but should be unique in general)

### 6. Checklist


_sql/migrations/1635-rename-default-emotion-elements.php
```diff
@@ -0,0 +1,53 @@
...
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+class Migrations_Migration1635 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $updates = [
+            ['Artikel', 'article'],
```
shyim: article :cry:

_sql/migrations/1635-rename-default-emotion-elements.php
```diff
@@ -0,0 +1,53 @@
...
+            ['Kategorie-Teaser', 'category_teaser'],
+            ['Blog-Artikel', 'blog_article'],
+            ['Banner', 'banner'],
+            ['Banner-Slider', 'banner_slider'],
+            ['Youtube-Video', 'youtube'],
+            ['Hersteller-Slider', 'manufacturer'],
```
shyim: manufacturer_slider

_sql/migrations/1635-rename-default-emotion-elements.php
```diff
@@ -0,0 +1,53 @@
...
+            ['Artikel-Slider', 'article_slider'],
```
shyim: product_slider

engine/Shopware/Controllers/Backend/Emotion.php
```diff
@@ -446,12 +447,20 @@ public function libraryAction()
...
 
         $components = $builder->getQuery()->getArrayResult();
+
+        $snippets = $this->get('snippets')->getNamespace('backend/emotion/view/detail');
+        foreach ($components as &$component) {
+            $name = str_replace(' ', '_', strtolower($component['name']));
+
+            $component['fieldLabel'] = $snippets->get($name, $component['name']);
+        }
+
```
shyim, line 472: unset $component

pascalheidmann-bedarf: As I'm using a reference instead of a copy wouldn't `unset($component);` remove that element?

shyim: https://github.com/kalessil/phpinspectionsea/blob/master/docs/performance.md#foreach-variables-reference-usage-correctness :)
Maybe install their Plugin. Its awesome good :blush:

pascalheidmann-bedarf: Ah now I get what you mean. I added it even if the variable is scoped within the function and therefore shouldn't be problematic (as long as nobody is messing with the variables later)

engine/Shopware/Controllers/Backend/Emotion.php
```diff
@@ -446,12 +447,21 @@ public function libraryAction()
...
 
         $components = $builder->getQuery()->getArrayResult();
+
+        $snippets = $this->get('snippets')->getNamespace('backend/emotion/view/detail');
+        foreach ($components as &$component) {
+            $name = str_replace(' ', '_', strtolower($component['name']));
+
+            $component['fieldLabel'] = $snippets->get($name, $component['name']);
+        }
+        unset($component); // to make shyim happy and prevent potential issues (scoped anyway)
```
tinect, line 471: 😆 best comment

shyim:
Hey,
thanks for your Pull Request :blush: 
It looks good, but the element names in the blocks are now gone. Can you fix that?

In 5.6:
old

In your Branch:
new

pascalheidmann-bedarf:
@shyim I guess the missing labels where the result of the migration collision when in the meantime a wild Migration1635 appeared ;)
Sadly I can't yet test it myself as my local setup is broken right now :(

PaddyS:
Hey @pascalheidmann-bedarf,

can you please ping us once you've fixed your local setup and tested your new changes?

pascalheidmann-bedarf:
@PaddyS @shyim found the bug that prevented the correct translation from popping up in the shopping world designer. Turned out Shopware is pushing the elements name once again in the `detailAction`

shyim:
We have merged it into 5.7, cause the renaming could be a breaking change

PR #2178, opened 2019-08-18
only show shipping countries

stefanpoensgen:
When calculating the shipping costs in cart, only those countries should be displayed which are also marked as shipping country

Check if allow_shipping is marked

In cart open shipping costs calculation and open country drop down. All active countries will be displayed.

### 6. Checklist


shyim:

PR #2179, opened 2019-08-20
fix 404 when loading category of another subshop

sensus-media:
Because it is possible to reach categories from other subshops, which leads to confusion for customers and maybe duplicate content.

It fixes an if condition to check if a requested category is empty or is in another subshop. Before the conditions where linked with AND but it should be OR to match both cases.

You need two subshops with different category trees: mainshop.de, subshop.de
Let's say category 123 is the main category of subshop.de
https://mainshop.de/cat/index/sCategory/123 will open the main category of subshop and load the emotions but within mainshop template.

### 6. Checklist


shyim:

PR #2180, opened 2019-08-21
Move greedy array_merge out of loops

pascalheidmann-bedarf:
Several array-operations are quite expensive and `array_merge` is no exception. Executing the same op over and over with the same input is therefore unnecessary.

Improving code performance by moving `array_merge` batching within foreach-loops outside of the loop. This results in a slightly higher temporary memory consumptions but should be beneficial enough in most cases.

I haven't checked every file yet, skipping all those horrible sClasses completely ;)

see https://github.com/kalessil/phpinspectionsea/blob/master/docs/performance.md#slow-array-function-used-in-loop and https://github.com/dseguy/clearPHP/blob/master/rules/no-array_merge-in-loop.md for more related array functions

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/AttributeBundle/Repository/Reader/CategoryReader.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Repository/Reader/CategoryReader.php
@@ -71,8 +71,9 @@ class CategoryReader extends GenericReader
     {
         $parents = [];
         foreach ($data as $id => $row) {
-            $parents = array_merge($parents, explode('|', $row['path']));
+            $parents[] = explode('|', $row['path']);
         }
+        $parents = array_merge([], ...$parents);
         $parents = array_values(array_unique(array_filter($parents)));
 
         $query = $this->entityManager->getConnection()->createQueryBuilder();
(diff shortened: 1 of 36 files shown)


engine/Shopware/Bundle/AttributeBundle/Repository/Reader/CategoryReader.php
```diff
@@ -69,10 +69,11 @@ private function getPathParents($parents, $path)
...
     private function getParents($data)
     {
-        $parents = [];
+        $parents = [[]];
         foreach ($data as $id => $row) {
-            $parents = array_merge($parents, explode('|', $row['path']));
+            $parents[] = explode('|', $row['path']);
         }
+        $parents = array_merge(...$parents);
         $parents = array_values(array_unique(array_filter($parents)));
```
bcremer, line 77: `$parents = [[]];` looks confusing and it's not obvious why the empty array is necessary.

I'd move the empty array down just before the splat operator:

```php
$parents = [];
// loop
$parents = array_merge([], ...$parents);
```

pascalheidmann-bedarf: Good point, looks way cleaner

pascalheidmann-bedarf:
@shopwareBot I fixed some minor possible bugs I found. Looking forward adding tests later.
As all those instances but the [`ContentTypeBundle/Services/Repository`](https://github.com/shopware/shopware/pull/2180/files#diff-5a92ee2f907333d13f50c18231b65917L250-L254) aren't using any references there shouldn't be any side-effects possible at all and can be verified statically.

PaddyS:
Hey @pascalheidmann-bedarf,

can you ping us once more, once you think you're done? :-)
Thank you very much for your effort though! 💙

PR #2181, opened 2019-08-21
Fix RegExp expression issue in filters

maximchik:
The order of the REGEXP parameters in beberlei/DoctrineExtensions is different.
https://github.com/beberlei/DoctrineExtensions/blob/a8747a2aef34f602c365cf2efd1abc94c5ae7e0a/src/Query/Mysql/Regexp.php#L18

Solves the issue in RegExp expressions in filters

Reorder parameters in RegExp to fix work with beberlei/DoctrineExtensions

Using any RegExp expression in filters produces the wrong result.

### 6. Checklist


shyim:
Hey,

PR #2182, opened 2019-08-23
Trim voucher codes for less errors

Isengo1989:
We get a lot of customers who are copying vouchers from mails, websites etc. with a leading space, which will result in an error that the voucher is no longer valid or not available.

It simple trims the voucher before sending the query to the DB.

Create a voucher named "Test02" and try to use it in the checkout form with " Test02".

### 6. Checklist


shyim:

PR #2184, opened 2019-08-26
Add sorting for backend order filter entities dispatch, payment and shop

larsbo:
At the moment the filter values for dispatches and payments are sorted by id which produces random acting orders. The shops are at least sorted by their name but it makes more sence to use the `position` column here as well as in many other places.

Change the sorting of the backend order filter dropdown entries for the entities `dispatch`, `payment` and `shop`.

Click on of the mentioned filters and be like
WTF

nope

nope

### 6. Checklist


shyim:

PR #2186, opened 2019-09-03
Only apply refreshOrderedVariant() for basket items that are products

jkrzefski:

Let the situation be as follows: Your shop has enabled proportional tax calculation. You have a voucher that is restricted to a set of products. A customer has a cart with two products and said voucher. One of the products matches the voucher restriction and the other product does not. The products have different tax rates.

The voucher now is split into two different vouchers; one for each tax rate. The voucher with the tax rate of the matching product has a value below zero. The voucher with a tax rate of the non-matching product has a value of exactly zero, as it does not reduce the value of any line items in the cart.

Shopware would normally not notify the event `product_stock_was_changed` for line items that have a value below zero. So since vouchers typically have a value below zero, this event is not normally notified for vouchers. But in this specific case, we have a voucher in the cart that has a value of exactly zero, so this event does get notified.

Trying to change `sales` and `instock` of a voucher does not break anything, but it is wrong. The actual problematic part is the event notification for `product_stock_was_changed`. Third party plugin developers rely on this event to only be notified when an actual product stock was changed. If no data is found by the provided identifier, this could result in unexpected errors, since the scenario pictured above is kind of an edge case.

Only execute the method `refreshOrderedVariant` when the basket item in question is actually a product. (`modus = 0`)

See 1.

None.

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sOrder.php
+++ b/engine/Shopware/Core/sOrder.php
@@ -762,7 +762,7 @@ class sOrder implements \Enlight_Hook
             $this->sBasketData['content'][$key]['attributes'] = $detailAttributes;
 
             // Update sales and stock
-            if ($basketRow['priceNumeric'] >= 0) {
+            if ($basketRow['priceNumeric'] >= 0 && $basketRow['modus'] == '0') {
                 $this->refreshOrderedVariant(
                     $basketRow['ordernumber'],
                     $basketRow['quantity']


uehler:
What about for example bundles? Those are products, value > 0 and modus 10
or premium products with modus 1?

Could that be a break?

jkrzefski:
@uehler Yes, that would break. The actual condition should be: `is there an item in s_articles_details for the current $basketRow`? Only then it would make sense to actually change `sales` and `instock`. Any idea on how to implement this? We cannot check, if `$basketRow['ordernumber']` matches an item in `s_articles_details` because e.g. vouchers can have ordernumbers that already exist as variants. So how do we check it correctly?

uehler:
Maybe you check, if modus is not one of voucher (2), rebate (3) and surcharge/discount (4)

jkrzefski:
@shopwareBot  Yes, i will change the condition and add a test case. Would it be sufficient to check if the event `product_stock_was_changed` is triggered or should I rather check if the method `refreshOrderedVariant` is executed?

aragon999:
To be compatible with plugins, like the bundle plugin (and others), I would suggest to check for `s_articles_details` with the condition of the `articleID` and `ordernumber` column. Then it should be sure that the correct row is updated. 
And since this is a private method this would be no breaking change.

PR #2187, opened 2019-09-03
Replace invalid timestamps with null

pascalheidmann-bedarf:
Ever tried to load a SQL dump of your database to another host? This will create a lot of problems as the tables create invalid `date`/`datetime` column defaults.

Replace all invalid `0000-00-00` dates with `null` as default. This won't change invalid data in the database but at least new installations won't get corrupted!

`0000-00-00` and `0000-00-00 00:00:00` are invalid dates and should never be used. With this PR applied those fields will be `null` instead of an `\DateTime` object (at least for new rows).

### 6. Checklist


_sql/demo/latest.sql
```diff
@@ -5045,7 +5045,6 @@ INSERT INTO `s_campaigns_groups` (`id`, `name`) VALUES
 
 TRUNCATE TABLE `s_campaigns_html`;
 TRUNCATE TABLE `s_campaigns_links`;
-TRUNCATE TABLE `s_campaigns_logs`;
```
shopwareBot: Why did you do this?

pascalheidmann-bedarf:
@shopwareBot the suggested `1970-01-01` date is valid but it won't be obvious why that date is invalid for Shopware - that would be even worse. As it is **technically** a valid date this brings way more trouble!

I left those code parts like the one you mentioned because of backwards compatibility as I won't update already existing `0000-00-00` dates.

_sql/demo/latest.sql
pascalheidmann-bedarf: This table isn't used anywhere in core, at least I didn't found any evidence. But yeah this is more a thing for a separate PR.

_sql/install/latest.sql
```diff
@@ -909,7 +909,7 @@ CREATE TABLE IF NOT EXISTS `s_campaigns_links` (
 DROP TABLE IF EXISTS `s_campaigns_logs`;
 CREATE TABLE IF NOT EXISTS `s_campaigns_logs` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
-  `datum` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
```
shyim: Please revert the changes in latest.sql and move it to a new migration

pascalheidmann-bedarf: @shyim no changes to `latest.sql` (both demo and install)?
I added it to the existing migration but will commit later depending on your answer regarding the sql files ;)

mitelg: to change the demo is ok, since that are the demo data :wink: 
the install should not be changed

shyim:

PR #2189, opened 2019-09-04
dump beberlei/assert from 2.9.2 to 2.9.9

resslinger:
Assertion::url has errors with PHP 7.3. This issue was fixed with the latest version.

This change dump the version of beberlei/assert from 2.9.2 to 2.9.9.

Test a url with https on Assertion::url

https://github.com/beberlei/assert/commits/v2

### 6. Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         "doctrine/dbal": "2.6.3",
         "doctrine/orm": "2.6.3",
         "doctrine/cache": "1.8.0",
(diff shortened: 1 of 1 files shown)


jkrzefski:
Don't you also have to commit the `composer.lock` file to actually update the dependency? 🤔

mitelg:
@jkrzefski I already mentioned that in the internal PR. the colleagues will take care of that 🙂👍

PR #2190, opened 2019-09-04
Improve service instantiation & use Doctrine DBAL instead of Zend DB in CacheManager

pascalheidmann-bedarf:
Several services are instantiated with parameter "service_container" which can basically be any `Container` that implements the `ContainerInterface` - therefore lots of service classes typehint to narrow with just `Container`.
Then there are several instances where the container is passed to the constructor to load other services. These cases can be simplified a lot.

- Replace all typehints of `Container` with the interface `ContainerInterface` if possible.
- Instantiate services with their required services directly instead of going the route of loading them manually via container
  - if possible remove container completely from services 
- As a bonus I updated the CacheManager to use the `dbal_connection` service instead of `db`.

The PR changes the factory method signature of `Shopware\Components\DependencyInjection\Bridge\Plugin::factory`

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Cache.php
+++ b/engine/Shopware/Components/Api/Resource/Cache.php
@@ -24,10 +24,16 @@
 
 namespace Shopware\Components\Api\Resource;
 
+use Enlight_Controller_Request_Request;
+use Exception;
 use Shopware\Components\Api\BatchInterface;
-use Shopware\Components\Api\Exception as ApiException;
+use Shopware\Components\Api\Exception\NotFoundException;
+use Shopware\Components\Api\Exception\ParameterMissingException;
 use Shopware\Components\CacheManager;
 use Shopware\Components\DependencyInjection\Container;
+use Zend_Cache;
+use Zend_Cache_Core;
+use Zend_Cache_Exception;
 
 /**
  * Cache API Resource
@@ -38,7 +44,7 @@ use Shopware\Components\DependencyInjection\Container;
 class Cache extends Resource implements BatchInterface
 {
     /**
-     * @var \Enlight_Controller_Request_Request
+     * @var Enlight_Controller_Request_Request
      */
     private $request;
 
@@ -70,8 +76,7 @@ class Cache extends Resource implements BatchInterface
     /**
      * @param string $id
      *
-     * @throws \Shopware\Components\Api\Exception\ParameterMissingException
-     * @throws \Shopware\Components\Api\Exception\NotFoundException
+     * @throws ParameterMissingException
      *
      * @return array
      */
@@ -80,7 +85,7 @@ class Cache extends Resource implements BatchInterface
         $this->checkPrivilege('read');
 
         if (empty($id)) {
-            throw new ApiException\ParameterMissingException('id');
+            throw new ParameterMissingException('id');
         }
 
         return $this->getCacheInfo($id);
@@ -108,8 +113,7 @@ class Cache extends Resource implements BatchInterface
     /**
      * @param string $id
      *
-     * @throws \Shopware\Components\Api\Exception\ParameterMissingException
-     * @throws \Shopware\Components\Api\Exception\NotFoundException
+     * @throws ParameterMissingException
      *
      * @return true
      */
@@ -118,7 +122,7 @@ class Cache extends Resource implements BatchInterface
         $this->checkPrivilege('delete');
 
         if (empty($id)) {
-            throw new ApiException\ParameterMissingException('id');
+            throw new ParameterMissingException('id');
         }
 
         $this->clearCache($id);
@@ -131,11 +135,7 @@ class Cache extends Resource implements BatchInterface
      */
     public function getIdByData($data)
     {
-        if (isset($data['id'])) {
-            return $data['id'];
-        }
-
-        return false;
+        return $data['id'] ?? false;
     }
 
     /**
@@ -168,7 +168,7 @@ class Cache extends Resource implements BatchInterface
                     'operation' => 'delete',
                     'data' => $this->delete((string) $id),
                 ];
-            } catch (\Exception $e) {
+            } catch (Exception $e) {
                 $message = $e->getMessage();
 
                 $results[$key] = [
@@ -185,7 +185,7 @@ class Cache extends Resource implements BatchInterface
     /**
      * @deprecated in 5.6, will be removed without a replacement
      *
-     * @return \Enlight_Controller_Request_Request
+     * @return Enlight_Controller_Request_Request
      */
     protected function getRequest()
     {
@@ -199,7 +199,8 @@ class Cache extends Resource implements BatchInterface
      *
      * @param string $cache
      *
-     * @throws \Shopware\Components\Api\Exception\NotFoundException
+     * @throws NotFoundException
+     * @throws Zend_Cache_Exception
      */
     protected function clearCache($cache)
     {
@@ -253,7 +254,7 @@ class Cache extends Resource implements BatchInterface
                 $this->cacheManager->clearOpCache();
                 break;
             default:
-                throw new ApiException\NotFoundException(sprintf('Cache "%s" is not a valid cache id.', $cache));
+                throw new NotFoundException(sprintf('Cache "%s" is not a valid cache id.', $cache));
         }
 
         if (!empty($capabilities['tags'])) {
@@ -270,11 +271,9 @@ class Cache extends Resource implements BatchInterface
      *
      * @param string $cache
      *
-     * @throws \Shopware\Components\Api\Exception\NotFoundException
-     *
-     * @return array
+     * @throws NotFoundException
      */
-    private function getCacheInfo($cache)
+    private function getCacheInfo($cache): array
     {
         switch ($cache) {
             case 'http':
@@ -296,7 +295,7 @@ class Cache extends Resource implements BatchInterface
                 $cacheInfo = $this->cacheManager->getOpCacheCacheInfo();
                 break;
             default:
-                throw new ApiException\NotFoundException(sprintf('Cache "%s" is not a valid cache id.', $cache));
+                throw new NotFoundException(sprintf('Cache "%s" is not a valid cache id.', $cache));
         }
 
         $cacheInfo['id'] = $cache;
--- a/engine/Shopware/Components/Api/Resource/Cache.php
+++ b/engine/Shopware/Components/Api/Resource/Cache.php
@@ -48,6 +48,11 @@ class Cache extends Resource implements BatchInterface
     private $cacheManager;
 
     /**
+     * @var Zend_Cache_Core
+     */
+    private $cache;
+
+    /**
      * Sets the Container.
      *
      * @param Container $container
@@ -57,6 +62,7 @@ class Cache extends Resource implements BatchInterface
         if ($container) {
             $this->request = $container->get('front')->Request();
             $this->cacheManager = $container->get('shopware.cache_manager');
+            $this->cache = $container->get('cache');
         }
         parent::setContainer($container);
     }
@@ -197,10 +203,10 @@ class Cache extends Resource implements BatchInterface
      */
     protected function clearCache($cache)
     {
-        $capabilities = $this->cacheManager->getCoreCache()->getBackend()->getCapabilities();
+        $capabilities = $this->cache->getBackend()->getCapabilities();
 
         if ($cache === 'all') {
-            $this->cacheManager->getCoreCache()->clean();
+            $this->cache->clean();
 
             $this->cacheManager->clearHttpCache();
             $this->cacheManager->clearConfigCache();
@@ -252,9 +258,9 @@ class Cache extends Resource implements BatchInterface
 
         if (!empty($capabilities['tags'])) {
             if (!empty($tags)) {
-                $this->cacheManager->getCoreCache()->clean(\Zend_Cache::CLEANING_MODE_MATCHING_ANY_TAG, $tags);
+                $this->cache->clean(Zend_Cache::CLEANING_MODE_MATCHING_ANY_TAG, $tags);
             } else {
-                $this->cacheManager->getCoreCache()->clean();
+                $this->cache->clean();
             }
         }
     }
--- a/engine/Shopware/Components/CacheManager.php
+++ b/engine/Shopware/Components/CacheManager.php
@@ -123,7 +123,6 @@ class CacheManager
         Shopware_Components_Config $config,
         ContainerAwareEventManager $events,
         PathResolver $themePathResolver,
-
         array $httpCache,
         array $cacheConfig,
         array $templateConfig,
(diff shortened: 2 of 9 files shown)


engine/Shopware/Components/CacheManager.php
```diff
@@ -78,16 +80,23 @@ class CacheManager
...
+    public function __construct(
+        ContainerInterface $container,
+        \Zend_Cache_Core $cache,
+        Configuration $emConfig,
+        Connection $db,
+        \Shopware_Components_Config $config,
+        ContainerAwareEventManager $events,
+        PathResolver $themePathResolver
+    ) {
         $this->container = $container;
```
uehler: You can pass the parameters like `shopware.httpCache.enabled` via constructor too and can get rid of the container injection

pascalheidmann-bedarf: Yeah I know but wasn't sure as there already too many parameters 😅

mitelg: I think, better many explicit parameters as one container parameter :wink:

mitelg: you could try to inject `shopware.httpCache` afaik you will get an array, so you don't need `shopware.httpCache.enabled`, `shopware.httpCache.cache_dir`, etc

pascalheidmann-bedarf:
As suggested by @uehler and @mitelg I now inject all parameters (as arrays if possible) into CacheManager and removed the container completely.

engine/Shopware/Components/CacheManager.php
```diff
@@ -78,35 +86,78 @@ class CacheManager
...
+    public function getCoreCache(): Zend_Cache_Core
     {
         return $this->cache;
     }
 
     /**
      * Clear HTTP-Cache
      */
-    public function clearHttpCache()
+    public function clearHttpCache(): void
```
mitelg: you should not change the method signature of public methods by adding typehints and return types. this will be a breaking change

engine/Shopware/Components/CacheManager.php
```diff
@@ -311,26 +352,21 @@ public function getConfigCacheInfo()
 
     /**
      * Returns template cache information
-     *
-     * @return array
```
mitelg: do not delete return doc

engine/Shopware/Components/CacheManager.php
```diff
@@ -311,26 +352,21 @@ public function getConfigCacheInfo()
...
+        $info = $this->getDirectoryInfo($this->templateConfig['compileDir']);
         $info['name'] = 'Shopware templates';
 
         return $info;
     }
 
     /**
      * Returns template cache information
-     *
-     * @return array
```
mitelg: do not delete return doc

engine/Shopware/Components/CacheManager.php
```diff
@@ -339,37 +375,28 @@ public function getThemeCacheInfo()
 
     /**
      * Returns cache information
-     *
-     * @return array
```
mitelg: do not delete return doc

engine/Shopware/Components/CacheManager.php
```diff
@@ -339,37 +375,28 @@ public function getThemeCacheInfo()
...
+        $info = $this->getDirectoryInfo($this->modelProxyDir);
         $info['name'] = 'Doctrine Proxies';
 
         return $info;
     }
 
     /**
      * Returns cache information
-     *
-     * @return array
```
mitelg: do not delete return doc

engine/Shopware/Components/CacheManager.php
```diff
@@ -339,37 +375,28 @@ public function getThemeCacheInfo()
...
+        $info = $this->getDirectoryInfo($this->hookProxyDir);
         $info['name'] = 'Shopware Proxies';
 
         return $info;
     }
 
     /**
      * Returns cache information
-     *
-     * @return array
```
mitelg: do not delete return doc

engine/Shopware/Components/CacheManager.php
```diff
@@ -391,15 +418,11 @@ public function getOpCacheCacheInfo()
      * Returns cache information
      *
      * @param string $dir
-     *
-     * @return array
```
mitelg, line 395: do not delete return doc

engine/Shopware/Components/CacheManager.php
```diff
@@ -452,13 +475,13 @@ public function getDirectoryInfo($dir)
      * Format size method
      *
      * @param float $bytes
-     *
-     * @return string
```
mitelg, line 456: do not delete return doc

engine/Shopware/Components/CacheManager.php
```diff
@@ -467,8 +490,6 @@ public function encodeSize($bytes)
      * Clear caches by list of CacheManager::CACHE_TAG_* items
      *
      * @param string[] $tags
-     *
-     * @return int
```
mitelg: do not delete return doc

engine/Shopware/Components/CacheManager.php
```diff
@@ -485,8 +506,6 @@ public function clearByTags(array $tags)
      * Clear cache by its tag of CacheManager::CACHE_TAG_*
      *
      * @param string $tag
-     *
-     * @return bool
```
mitelg: do not delete return doc

engine/Shopware/Components/CacheManager.php
```diff
@@ -523,30 +564,30 @@ public function clearByTag($tag)
      *
      * @param string $dir
      */
-    private function clearDirectory($dir)
+    private function clearDirectory($dir): void
```
soebbing: You can add a string type here.

engine/Shopware/Components/CacheManager.php
```diff
@@ -239,16 +285,16 @@ public function clearOpCache()
...
      */
     public function getHttpCacheInfo($request = null)
     {
-        if ($this->container->getParameter('shopware.httpCache.enabled')) {
-            $info = $this->getDirectoryInfo(
-                $this->container->getParameter('shopware.httpCache.cache_dir')
-            );
+        if ($this->httpCache['enabled']) {
+            $info = $this->getDirectoryInfo($this->httpCache['cache_dir']);
         } else {
```
soebbing: You can get rid of the `else` block here by simply initializing the variable `$info = [];` above the `if` condition. The same for the `backend` below.

engine/Shopware/Components/CacheManager.php
soebbing: Please a string type here.

soebbing (changes requested):
Nice PR!

pascalheidmann-bedarf:
@soebbing thx!
Because I was already changing more than signatures I added some deprecations for private methods and also removed indirect calls for `cache` service via `cacheManager` by replacing those with explicit service calls in the `Shopware\Components\Api\Resource\Cache` and cache backend controller.

Also I had to fix two return annotations in `\Shopware_Bootstrap` as those seem to be broken/detected now by PHPStan...

engine/Library/Enlight/Components/Snippet/Manager.php
```diff
@@ -75,6 +75,7 @@ public function __construct($options = null)
         if (isset($options['ignore_namespace'])) {
             $this->ignoreNamespace = (bool) $options['ignore_namespace'];
         }
+        parent::__construct();
```
mitelg: are you sure, that this will not break anything?

engine/Shopware/Components/CacheManager.php
```diff
@@ -78,20 +87,70 @@ class CacheManager
...
+        $this->cacheConfig = $cacheConfig;
+        $this->templateConfig = $templateConfig;
+        $this->docRoot = $docRoot;
+        $this->hookProxyDir = $hookProxyDir;
+        $this->modelProxyDir = $modelProxyDir;
     }
 
     /**
-     * @return \Zend_Cache_Core
+     * @deprecated in 5.7, will be removed in 5.8
```
mitelg: you should describe an alternative for this call

engine/Shopware/Components/CacheManager.php
```diff
@@ -451,102 +548,49 @@ public function getDirectoryInfo($dir)
     /**
      * Format size method
      *
+     * @deprecated in 5.7, will be private in 5.8
```
mitelg: alternative

engine/Shopware/Components/CacheManager.php
```diff
@@ -387,19 +429,74 @@ public function getOpCacheCacheInfo()
...
+                return false;
+        }
+
+        return true;
+    }
+
     /**
      * Returns cache information
      *
+     * @deprecated in 5.7, will be private in 5.8
```
mitelg: alternative

pascalheidmann-bedarf: This is a helper that is used only inside the class itself. I don't see any reason for an alternative, especially as this little formatter is just shy 4 lines of code. If a dev needs it somewhere they should better implement it themselves.
Only possible alternative would be moving it into a completely new helper service that wouldn't really fit anywhere.

engine/Shopware/Components/CacheManager.php
pascalheidmann-bedarf: I can't have any useful alternative but also see no reason for having this public. Right now the function is used just inside the class itself.

engine/Library/Enlight/Components/Snippet/Manager.php
pascalheidmann-bedarf: `\Enlight_Components_Snippet_Manager` has only the abstract class `\Enlight_Class` as its ancestor - therefore only the branch `if (method_exists($this, 'init'))` might ever get executed.
So, yes, as long as no dev tries to extend `\Shopware_Components_Snippet_Manager` or `\Enlight_Components_Snippet_Manager` and adds an `init` method this won't have unwanted side-effects.

engine/Shopware/Components/CacheManager.php
mitelg: ah ok. then you should add this as info :wink: because it is not used in Shopware itself somewhere else, does not mean, some plugins don't use it :blush:

mitelg: totally have overseen the private hint :smile: forget what i've said

engine/Shopware/Components/Api/Resource/Cache.php
```diff
@@ -74,13 +87,16 @@ public function getOne($id)
...
+            throw new ParameterMissingException('id');
         }
 
         return $this->getCacheInfo($id);
     }
 
     /**
+     * @throws ApiException\PrivilegeException
+     * @throws NotFoundException
+     *
```
soebbing: Please remove these Exception annotations.

The rule Shopware follows with regards to annotating Exceptions is: We only declare them when they are thrown in this method explicitly.
The reasoning for this is as follows: If I modify a function and want to declare all implicit Exceptions as well, it would be necessary to check all functions I am using in my modified function for possible exceptions (your IDE helps you there), but I would also need to change all method annotations for every method calling the modified one, since they would have to get the new exception annotation as well. And that is nearly impossible.

engine/Shopware/Components/Api/Resource/Cache.php
```diff
@@ -102,8 +118,10 @@ public function getList()
     /**
      * @param string $id
      *
-     * @throws \Shopware\Components\Api\Exception\ParameterMissingException
-     * @throws \Shopware\Components\Api\Exception\NotFoundException
+     * @throws ApiException\PrivilegeException
+     * @throws NotFoundException
+     * @throws ParameterMissingException
+     * @throws \Zend_Cache_Exception
```
soebbing: Please see the comment above

engine/Shopware/Components/Snippet/Manager.php
```diff
@@ -90,12 +90,12 @@ public function __construct(ModelManager $modelManager, array $pluginDirectories
             ]);
         }
 
-        $this->adapter = new DbAdapter([
+        parent::__construct(new DbAdapter([
```
soebbing: My Spidy-Developer senses are tingeling. Why did you change this? I don't see `$this->adapter` being defined anywhere, but it is used later on.

engine/Shopware/Components/Api/Resource/Cache.php
```diff
@@ -57,15 +68,17 @@ public function setContainer(Container $container = null)
...
     }
 
     /**
      * @param string $id
      *
-     * @throws \Shopware\Components\Api\Exception\ParameterMissingException
-     * @throws \Shopware\Components\Api\Exception\NotFoundException
+     * @throws ApiException\PrivilegeException
+     * @throws NotFoundException               *@throws ApiException\PrivilegeException
+     * @throws ParameterMissingException
```
soebbing, line 79: See comment below

soebbing (changes requested):
This PR by now isn't about Zend DB alone, but a whole lot more generic. I think you should split your PR into multiple commits, so we can test and modify these changes more atomically.

engine/Shopware/Components/Snippet/Manager.php
mitelg: @soebbing see here. the parent has some "nice" possibilities of parameters in its constructor :smile: https://github.com/shopware/shopware/blob/5.6/engine/Library/Enlight/Components/Snippet/Manager.php#L63

shopwareBot: Ok, I wish I wouldn't have seen that. 🤢  But I guess the change is ok then, technically!

soebbing: Ok, I wish I wouldn't have seen that. 🤢 But I guess the change is ok then, technically!

pascalheidmann-bedarf: @soebbing me neither. The main reason I stumbled upon this in the first was my IDE freaking out of missing parent constructor calls. I might add a comment to clarify this.

pascalheidmann-bedarf:
While splitting all changes into their own commits I realised that the changes regarding SnippetManager aren't really part of the rest and therefore I moved them into #2195

shyim:
Wrong pull request number :man_facepalming:

shyim:
When I clear the cache I get following error message
"Fatal error: Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'routerlastupdate' in 'where clause' in /var/www/html/shopware/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:912 Stack trace: #0 /var/www/html/shopware/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(912): PDO->query('SELECT v.shop_i...') #1"

Please test your change before submitting your Pull Request :blush:

engine/Shopware/Components/DependencyInjection/Bridge/Plugins.php
```diff
@@ -24,24 +24,18 @@
...
     /**
      * @return \Enlight_Plugin_PluginManager
      */
     public function factory(
-        Container $container,
-        \Enlight_Loader $loader,
-        \Enlight_Event_EventManager $eventManager,
-        \Shopware $application,
-        array $pluginDirectories
+        \Enlight_Loader $loader, \Enlight_Event_EventManager $eventManager, \Shopware $application, array $pluginDirectories, ConfigReader $configReader
```
mitelg: please use multi lines here

shyim:
@pascalheidmann-bedarf Can i lookup? Maybe add next time new commits instead squashing, i can't see the difference

pascalheidmann-bedarf:
@shopwareBot I rebased to latest version and merged the conflicting changes in `\Shopware_Components_Translation`.
@shyim the mentioned problem happened cause I thought `Expr\Like` would escape... Added the explicit escaping. Next time I will definitely use more commits ;)

shyim:
Next time please test your PR before submitting it. It was just clearing cache :see_no_evil:

pascalheidmann-bedarf:

shyim:

PR #2196, opened 2019-09-12
[WIP] Add support for passing attributes via API

pascalheidmann-bedarf:
While some API endpoints supply attributes not all do. But AFAIK none allows to pass the attributes for create or update calls. 

- adds support for attributes on `create` & `update` calls
- adds attributes to `getOne` and `getList` resource calls that won't make these available yet

1. make sure you have attributes for `media` (`s_media`/`s_media_attributes`)
2. call api endpoint for media with valid post payload containing an `attribute` field
3. check for created media without those passed attributes without PR, with PR they will be there 

### 6. Checklist


pascalheidmann-bedarf:
@shopwareBot As you might see this PR is marked as WIP, both in title as well as PR type, and I opened this ticket because - well its work in progress and I wanted to get some feedback before I add more endpoints... 👎

PR #2199, opened 2019-09-13
Add smarty blocks to delivery_information.tpl to make it extensible

pascalheidmann-bedarf:
You can't really replace/add logic for displaying the delivery time.

Add several smarty blocks so you can extend those in a plugin or child theme

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/plugins/index/delivery_informations.tpl
+++ b/themes/Frontend/Bare/frontend/plugins/index/delivery_informations.tpl
@@ -1,70 +1,88 @@
 {* Delivery informations *}
 {block name='frontend_widgets_delivery_infos'}
     <div class="product--delivery">
-        {if $sArticle.shippingfree}
-            <p class="delivery--information">
-                <span class="delivery--text delivery--text-shipping-free">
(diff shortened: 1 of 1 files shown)


themes/Frontend/Bare/frontend/plugins/index/delivery_informations.tpl
```diff
@@ -1,70 +1,88 @@
...
-            <p class="delivery--information">
-                <span class="delivery--text delivery--text-not-available">
-                    <i class="delivery--status-icon delivery--status-not-available"></i>
-                    {s name="DetailDataNotAvailable"}{config name=notavailable}{/s}
-                </span>
-            </p>
-        {/if}
+        {block name='frontend_widgets_delivery_infos_inner'}
+            {if $sArticle.shippingfree}
+                {block name='frontend_widgets_delivery_infos_shippingfree'}
```
shopwareBot, line 6: Can you fix the indentation in this file?
```
{block name='frontend_widgets_delivery_infos_shippingfree'}
	    <p class="delivery--information">
```

PR #2200, opened 2019-09-16
StateManager.getScrollBarSize() fetches scrollbar-size lazily

screeny05:
Calculating the Scrollbar-Size right at the pageload forces a reflow by the browser. To prevent this ~60ms js-block, we can fetch the scrollbar-size lazily.

image

Change the private property `StateManager._scrollBarSize` to a private method `StateManager._getScrollBarSize` which uses a cache internally, so only one calculation is made.

Also:
* `StateManager.getScrollBarSize()`
* `StateManager.getScrollBarWidth()`
* `StateManager.getScrollBarHeight()`

were changed to call this new function.

Use the Chrome DevTools to profile a shopware-shop. On pageload you will see a flamegraph like in the screenshot above.

-

None, as the removed property was private.

### 6. Checklist

### DIFF
--- a/UPGRADE-5.6.md
+++ b/UPGRADE-5.6.md
@@ -55,6 +55,7 @@ This changelog references changes done in Shopware 5.6 patch versions.
         * `city`
         * `country`
         * `state`
+* Changed `StateManager.getScrollBarSize` to lazily get the size of the scrollbar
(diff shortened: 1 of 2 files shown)


shyim:

PR #2201, opened 2019-09-16
Windows-Fixes for generating theme cache

psychorex:
These changes will fix the problem, that css and js files are not correctly generated on Windows OS Servers when generating the theme caches.

1. It sets the file objects of the generated tmp files to null after releasing the lock, else Windows would still lock the file (open file handler).
2. It replaces the '/' character with DIRECTORY_SEPARATOR in the PathResolver, so it does generate correct file paths.

Generate the theme cache via CLI or web-backend. On Windows OS the css and js files under "web/cache/" are still named with the ".tmp" file extension. After that the shop-frontend is a mess (no css or js can be loaded).

N/A

none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Theme/Compiler.php
+++ b/engine/Shopware/Components/Theme/Compiler.php
@@ -259,6 +259,8 @@ class Compiler
         }
         $file->flock(LOCK_UN);   // release the lock
 
+        $file = null; // release file handles, else Windows still locks the file
+
         rename($this->pathResolver->getTmpCssFilePath($shop, $timestamp), $this->pathResolver->getCssFilePath($shop, $timestamp));
     }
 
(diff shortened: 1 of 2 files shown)


engine/Shopware/Components/Theme/Compiler.php
```diff
@@ -298,6 +300,8 @@ public function compileJavascript($timestamp, Shop\Template $template, Shop\Shop
         $file->fwrite($content);
         $file->flock(LOCK_UN);   // release the lock
 
+        $file = null; // release file handles, else Windows still locks the file
```
shopwareBot, line 303: Wouldn't `unset($file)` work here as well?

psychorex: No it doesn't, I cannot explain why, but it only works when you set it explicitly to another value.

pascalheidmann-bedarf: I guess it comes down to this: https://stackoverflow.com/a/50931077/12025824
`null` frees the memory directly while `unset` will mark the variable as obsolete for the garbage collection. And windows might trigger GC only if low memory.

shyim:
Hey, 

psychorex:
@shyim Done ;-)

shyim:
Thanks! :), FYI: The github webhook hangs, just ignore it for this PR

shyim:
It has been merged with https://github.com/shopware/shopware/commit/7140ad44730c4acca3d336047f7ae636a795c20e

PR #2203, opened 2019-09-19
Address Controller - Handle sessionKey correctly in isShippingRequest()

screeny05:
The correct separator for sessionKey is `,` not `;`

Change the separator in the explode-call

1. In the shop you should have two countries active. One of which should not be available for shipping.
2. Create a new account (no alternative shipping-address)
3. Add a new address with the non-shipping-country selected
3. Go to the checkout
4. Open the modal for editing the current address (which is both shipping & billing)
5. The non-shipping-country is selectable, even though according to the template file it shouldn't be, as this should be a `shippingRequest` but isn't.

This happens because in the template-files and the `Shopware_Controller_Frontend_Address:handleExtraData()`-function the sessionKey is separated by a `,`. Only the `isShippingRequest()`-function handles this wrong.

None

None

### 6. Checklist


shyim:

PR #2205, opened 2019-09-22
Increase actionColumn width for correct rendering in Chrome Wayland

aragon999:
Since I switched to Wayland (sway) in ArchLinux the rendering of the local update button in the plugin manager listing, is not shown anymore in Chrome.
Firefox works fine, but since Chrome is the recommend browser I guess this should be fixed. Note, you can get around the issue if you resize the columns such that you can extend the action column (resize all other columns).
screenshot-22 09 2019-18 33 41
*Without this patch*

screenshot-22 09 2019-18 38 13
*With this patch*
 
Increase the width of the action column in the local plugin listing.

Install sway (or I would suspect other wayland window manager would have the same problems), and look at the plugin manager listing.

\-

\-

### 6. Checklist


PaddyS:
Hey @aragon999,

I've closed the linked ticket, because I was working on a similar ticket already.
The new ticket can be found here: https://issues.shopware.com/issues/SW-24650

Don't worry, I applied your commit to my Pull Request, so your changes will still be applied. :-)

shyim:
Thanks for your Contribution! :blue_heart: . The PR has been internally merged and will later review on GitHub

PR #2207, opened 2019-09-24
fixed issue for old order document types

christopherholst:
fixed issue for old order with document types there are actualy not more exist

if the document ID is no longer available in the system but known at the time of ordering, it can no longer be filtered or sorted by order.

If we do not receive any documents id from the system, the foreach query is not filled with a null

simply put an id in an order in the s_order_documents table that no longer exists in the s_core_documents

-

nothing

### 6. Checklist


shyim:

PR #2209, opened 2019-09-24
Vbidw 70,71

virtua-slash-admin:
I added low stock plugins


virtua-slash-admin:
sorry, missclicked repo :(

PR #2210, opened 2019-09-24
Remove unnecessary break in switch statement of the link filter plugin

aragon999:
I was currently debugging some problem in the link plugin, it took me some (short) time to figure out if there is some difference between `img` and `a` URLs, or not. Spoiler, it's the same :-)

This change removes this confusion, as the `switch` statement is simplified.

\-

\-

\-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Core/PostFilter/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Core/PostFilter/Bootstrap.php
@@ -182,44 +182,29 @@ class Shopware_Plugins_Core_PostFilter_Bootstrap extends Shopware_Components_Plu
         }
 
         $link = $src[3];
-
-        $anchorPart = '';
-        if (strpos($link, '#') !== false) {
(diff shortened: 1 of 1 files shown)


mitelg (commented):
why still using the switch statement at all? 🤔

aragon999:
Not sure actually, as there are even some cases which will never match, I will look into that more closely, but I would suggest to completely remove that `switch` statement

aragon999:
So I looked into it, does anyone know why the links for `iframe`s are not rewritten?

aragon999:
Hey @shopwareBot 

thanks for the reply, so you mean that the link for the `iframe` should not be rewritten? Why should that link not be rewritten (especially since there is a check that only internal links will be rewritten).

aragon999:
I have updated the pull request, and also restructured the handling for links in a more reasonable way. Do you see any modification of the behavior of the code for real applications? Or is something relevant and should be correctly handled:
``

shyim:

PR #2211, opened 2019-09-25
Fix issue with context on console

hlohaus:
You can't use the api resource on console anymore.
You will get the error: 
Zend_Session_Exception: You must call Zend_Session::setId() before any output has been sent to the browser;

It use now only the bundle context and not the old.

### 6. Checklist


shyim:
Hey @hlohaus,
I checked your PR and I don't really like your solution to fix this, cause you add a new method which is not in the Interface. When somebody decorate it, this code will fail.

The main issue are here the warnings. Just fixing the PHP warnings should fix it :wink:

PR #2214, opened 2019-09-26
Prefix MySQL query columns for the checkIfProductIsInBasket QueryBuilder

aragon999:
If one wants to join for example `s_articles_attributes` where also an `articleID` column exists, it is hard to work with this query builder, since the column names are not prefixed and therefore not the `where` condition is not unique.

Prefix the columns correctly.

Use a plugin to extend the DBAL querybuilder in the event `Shopware_Modules_Basket_AddArticle_CheckBasketForArticle` and join the table `s_articles_attributes`.

\-

\-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2184,12 +2184,12 @@ SQL;
     {
         $builder = Shopware()->Models()->getConnection()->createQueryBuilder();
 
-        $builder->select('id', 'quantity')
+        $builder->select('basket.id', 'basket.quantity')
             ->from('s_order_basket', 'basket')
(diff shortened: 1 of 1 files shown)


mitelg (commented):
could this be a breaking change? 🤔

shyim:
It's the same query, just with a better specification of which table. I don't see any breaking change :thinking:

aragon999:
I thought about it, and I also came to the conclusion, that it cannot be a breaking change, since it would break the query if so :-)

mitelg:
yeah, the fetch result should be the same, you are right :blush:

shyim:

PR #2216, opened 2019-09-27
Viewport content snippet

swagTKA:
Sometimes it's useful to set an individual value for the viewport. Actually it would cause overwriting the complete meta_tag_mobile block, which will make updates more difficult.

Add's an snippet for the viewport content

There was a pull request before which changed the value ... this may not the best configuration for everybody:

https://issues.shopware.com/issues/SW-23913

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/index/header.tpl
+++ b/themes/Frontend/Bare/frontend/index/header.tpl
@@ -40,7 +40,7 @@
     {/block}
 
     {block name='frontend_index_header_meta_tags_mobile'}
-        <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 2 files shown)


themes/Frontend/Bare/frontend/index/header.tpl
```diff
@@ -40,7 +40,7 @@
     {/block}
 
     {block name='frontend_index_header_meta_tags_mobile'}
-        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <meta name="viewport" content="{s name='IndexMetaViewportContent'}{/s}">
```
PaddyS, line 43: Hey @swagtoby,

can you implement a Smarty Block here instead?

PR #2217, opened 2019-09-30
Add smarty blocks to delivery_information.tpl to make it extensible

pascalheidmann-bedarf:
You can't really replace/add logic for displaying the delivery time.

Add several smarty blocks so you can extend those in a plugin or child theme

#2199 (was closed cause I was on vacation and couldn't answer in time...)

### 6. Checklist


shyim:

PR #2218, opened 2019-09-30
Fix possible doctrine exception in affiliate marketing (map customer to affiliate)  with setting maxresults

aherking:
If customer.number and customer.id are equal, there will be two results which causes exception because one or NULL result are expected. So we should set the maxresults to 1 .

If you have an equal customer number and customer id (id of s_user), there will be an doctrine exception because "getOneOrNull" is expected in Partner.php Controller.

Set the limit of results to 1

Make customernumber editable and put a number in it which is equal to one userID.

https://issues.shopware.com/issues/SW-23091
### 6. Checklist


mitelg:
hey @aherking,
thanks for your PR :blush: 

I think the query is not correct then. How do you know, that it is the same customer, if ID and number are identical? Seems very random. Setting the limit will fix the exception, but I think we have to fix the where clause.

PR #2220, opened 2019-09-30
Fixed openPdfAction in Order Backend Controller

flopreyser:
Download of order related files like the invoice is not working correctly since the response headers like "Content-Disposition" are set three times in the code. Some browsers (like Microsoft Edge) can't handle these headers correctly which will result in wrong filenames with a comma and/or the word "attachment" appended to the file-extension (like "xxxx.pdf," or "xxxx.pdf, attachment").

Removes the two calls "sendHeaders" and "sendResponse" in the openPdfAction of the Order-Controller. These are unnecessary since the response is sent at the end of the execution with the "send" method of the response object.

1. Login into Shopware 5.6 Backend with Microsoft Edge
2. Open Menu "Customers"->"Orders"
3. Click the "Show Details" Icon at an order
4. Switch to the "Documents"-Tab
5. Create an document and try to download it by clicking on the name of the document

There is an open thread in the german Shopware forum:
https://forum.shopware.com/discussion/63039/shopware-5-6-1-dokumente-enden-mit-komma

-

### 6. Checklist


shyim:

PR #2221, opened 2019-09-30
Fix line calculation with quantity=0

pascalheidmann-bedarf:
With the release of Shopware 5.6 there had been changes how each order position is calculated. Those are handled now inside a separate service (yeay) but those introduce a big problem:
If you set the quantity to `0` the line is still calculated as the base price, eg. quantity=1.

This happens for example if you use Shopware ERP which sets the position's quantity to 0 if you cancel the position or the complete order.
As this breaks lots of things like the official Shopware/Pickware ERP this should be fixed ASAP.

Skip all calculations and return `0.0` if the passed quantity isn't valid (`0` or `null`)

1. Create an order
2. Change the quantity of a position to 0
3. Check order's invoice amount. If you have just one position and no shipping costs it will be the price of the position with qty=1 instead of 0.0

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Cart/NetRounding/RoundLineAfterQuantity.php
+++ b/engine/Shopware/Components/Cart/NetRounding/RoundLineAfterQuantity.php
@@ -28,11 +28,12 @@ class RoundLineAfterQuantity implements RoundingInterface
 {
     public function round(float $netPrice, float $tax, int $quantity = null): float
     {
-        $netPrice = round($netPrice, 2) / 100 * (100 + $tax);
-
-        if ($quantity) {
-            $netPrice = round($netPrice * $quantity, 2);
+        if ($quantity === 0) {
+            return 0.0;
         }
+
+        $netPrice = round($netPrice, 2) / 100 * (100 + $tax);
+        $netPrice = (is_int($quantity) && $quantity !== 1) ? round($netPrice * $quantity, 2) : $netPrice;
 
         return $netPrice;
     }
--- a/engine/Shopware/Components/Cart/NetRounding/RoundLineAfterTax.php
+++ b/engine/Shopware/Components/Cart/NetRounding/RoundLineAfterTax.php
@@ -28,11 +28,12 @@ class RoundLineAfterTax implements RoundingInterface
 {
     public function round(float $netPrice, float $tax, int $quantity = null): float
     {
-        $netPrice = round(round($netPrice, 2) / 100 * (100 + $tax), 2);
-
-        if ($quantity) {
-            $netPrice = round($netPrice * $quantity, 2);
+        if ($quantity === 0) {
+            return 0.0;
         }
+
+        $netPrice = round(round($netPrice, 2) / 100 * (100 + $tax), 2);
+        $netPrice = (is_int($quantity) && $quantity !== 1) ? round($netPrice * $quantity, 2) : $netPrice;
 
         return $netPrice;
     }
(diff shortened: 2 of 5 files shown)


shyim:
Can you write also tests?

pascalheidmann-bedarf:
@shyim I'm on it but will have to continue tomorrow.

engine/Shopware/Components/Cart/NetRounding/RoundLineAfterQuantity.php
```diff
@@ -28,12 +28,13 @@ class RoundLineAfterQuantity implements RoundingInterface
 {
     public function round(float $netPrice, float $tax, int $quantity = null): float
     {
-        $netPrice = round($netPrice, 2) / 100 * (100 + $tax);
-
-        if ($quantity) {
-            $netPrice = round($netPrice * $quantity, 2);
+        if ($quantity === 0) {
```
mitelg, line 31: you should cast `$quantity` to int or also check for `null` 
otherwise, this will always be false if `$quantity` is `null`

engine/Shopware/Components/Cart/NetRounding/RoundLineAfterQuantity.php
```diff
@@ -28,12 +28,13 @@ class RoundLineAfterQuantity implements RoundingInterface
 {
     public function round(float $netPrice, float $tax, int $quantity = null): float
```
mitelg, line 29: why is the default of `$quantity` `null`? :see_no_evil:

engine/Shopware/Components/Cart/NetRounding/RoundLineAfterQuantity.php
pascalheidmann-bedarf: I originally had a check for `!$quantity` and hell broke loose in tests - that's why I thought someone, maybe @shyim as the author, thought it would be good to have some kind of fallback with not passing anything.
If this won't break I will also prefer changing the parameter to `int $quantity = 0` instead of `null` to prevent this thing... 🙈

pascalheidmann-bedarf: After digging into the source code I found 3 instances where the rounding function is used:
- sBasket with no `mode` -> default value `null`
- `OrderModel::processPositions()` with mode 0, 3, 4, 10 or 12 -> default, with quantity added later...
- `OrderModel::calculateInvoiceAmount()` with position's quantity

I modified those instances to explicitly pass `1` so maybe later the default value can be changed...

tests/Functional/Components/Cart/NetRoundingTest.php
```diff
@@ -54,7 +54,7 @@ public function testOldRounding(): void
         $this->setConfig('roundNetAfterTax', false);
         $this->dispatch('/checkout/cart');
 
-        static::assertEquals(23.189999999999998, $this->View()->getAssign('sBasket')['AmountWithTaxNumeric']);
+        static::assertEquals(23.2, $this->View()->getAssign('sBasket')['AmountWithTaxNumeric']);
```
shyim: Why did you change this test?

pascalheidmann-bedarf: Because without it ~will fail~ failed. I guess it has something to do with the way `sBasket` used the service in `sBasket::getBasketProducts()` by getting first the amount for a single item and then multiplying itself afterwards. As now it explicitly passes qty=1 the result will be rounded once again before returning.

pascalheidmann-bedarf:
After @shyim correctly mentioned the changed behaviour in `RoundLineAfterQuantity` I modified it to behave like the old implementation. + added more testcases

shyim:

PR #2222, opened 2019-09-30
Add interfaces for services in AttributeBundle

jkrzefski:
Every service in the dependency injection container deserves to have an interface. It enables third party developers to decorate these services with ease.

It adds new interfaces for the services in the AttributeBundle with all the public methods signatures those services already implement.

None.

The page about the [attribute system](https://developers.shopware.com/developers-guide/attribute-system/) should receive an update to let people know about these new interfaces.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/AttributeBundle/Service/CrudService.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Service/CrudService.php
@@ -27,36 +27,33 @@ namespace Shopware\Bundle\AttributeBundle\Service;
 use Shopware\Components\Model\ModelManager;
 use Shopware\Models\Attribute\Configuration;
 
-class CrudService
+class CrudService implements CrudServiceInterface
 {
-    const EXT_JS_PREFIX = '__attribute_';
-    const NULL_STRING = 'NULL';
-
     /**
      * @var ModelManager
      */
     private $entityManager;
 
     /**
-     * @var SchemaOperator
+     * @var SchemaOperatorInterface
      */
     private $schemaOperator;
 
     /**
-     * @var TableMapping
+     * @var TableMappingInterface
      */
     private $tableMapping;
 
     /**
-     * @var TypeMapping
+     * @var TypeMappingInterface
      */
     private $typeMapping;
 
     public function __construct(
         ModelManager $entityManager,
-        SchemaOperator $schemaOperator,
-        TableMapping $tableMapping,
-        TypeMapping $typeMapping
+        SchemaOperatorInterface $schemaOperator,
+        TableMappingInterface $tableMapping,
+        TypeMappingInterface $typeMapping
     ) {
         $this->entityManager = $entityManager;
         $this->schemaOperator = $schemaOperator;
@@ -65,9 +62,7 @@ class CrudService
     }
 
     /**
-     * @param string $table
-     * @param string $column
-     * @param bool   $updateDependingTables
+     * {@inheritdoc}
      *
      * @throws \Exception
      */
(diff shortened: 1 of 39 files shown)


engine/Shopware/Bundle/AttributeBundle/Service/CrudServiceInterface.php
```diff
@@ -0,0 +1,74 @@
...
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Bundle\AttributeBundle\Service;
+
+interface CrudServiceInterface
+{
+    /**
+     * @param string $table
+     * @param string $column
```
shyim, line 34: please move also the phpdocs

shyim (changes requested):

mitelg:
thanks for your PR @jkrzefski 
you also have to change the typehints in classes, where these classes are used. otherwise the interfaces will have no effect :wink:

jkrzefski:
@mitelg Yes, I will do that.

engine/Shopware/Bundle/AttributeBundle/Service/CrudServiceInterface.php
jkrzefski: Can you explain, what exactly do you mean?

jkrzefski:
Btw: I left out the `@throws` statements of the phpdocs on purpose in the interfaces. The reason for me to not have those in the interfaces is, that I believe they are specific to the implementation. Do you approve this?

mitelg:
@jkrzefski I would say it depends :wink: 
Some interfaces could describe certain throw tags, as the caller of the service will handle them for example. But I don't think that this is the case here

jkrzefski:
@mitelg Ok, then I'll keep it this way. Should I change the phpdocs of the implementations to `{@inheritdoc}`?

engine/Shopware/Bundle/AttributeBundle/Service/CrudService.php
```diff
@@ -27,36 +27,36 @@
 use Shopware\Components\Model\ModelManager;
 use Shopware\Models\Attribute\Configuration;
 
-class CrudService
+class CrudService implements CrudServiceInterface
 {
-    const EXT_JS_PREFIX = '__attribute_';
-    const NULL_STRING = 'NULL';
+    const EXT_JS_PREFIX = CrudServiceInterface::EXT_JS_PREFIX;
```
shyim: Is this really necessary when the implemented interface contains that?
https://3v4l.org/rotPu

jkrzefski: Thank you. I did not think about this. I'll remove it.

engine/Shopware/Bundle/AttributeBundle/Service/TypeMapping.php
```diff
@@ -26,95 +26,95 @@
...
-    const TYPE_HTML = 'html';
-    const TYPE_INTEGER = 'integer';
-    const TYPE_FLOAT = 'float';
-    const TYPE_BOOLEAN = 'boolean';
-    const TYPE_DATE = 'date';
-    const TYPE_DATETIME = 'datetime';
-    const TYPE_COMBOBOX = 'combobox';
-    const TYPE_SINGLE_SELECTION = 'single_selection';
-    const TYPE_MULTI_SELECTION = 'multi_selection';
+    const TYPE_STRING = TypeMappingInterface::TYPE_STRING;
```
mitelg: why the duplication?

shyim: https://github.com/shopware/shopware/pull/2222#discussion_r331024163

shyim:

PR #2223, opened 2019-10-02
Return StateManager from destroyPlugin method to allow chaining

aragon999:
To give the `destroyPlugin` method a similar function signature as most of the other `StateManager` methods.

Returns the `StateManager` instance from the `destroyPlugin` method.

```js
StateManager
    .destroyPlugin('.foo--class', 'baz--plugin')
    .destroyPlugin('.bar--class', 'qux--plugin')
;
```
\-

\-

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.state-manager.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.state-manager.js
@@ -1099,6 +1099,7 @@
          * @method destroyPlugin
          * @param {String|jQuery} selector
          * @param {String} pluginName
+         * @returns {StateManager}
          */
         destroyPlugin: function (selector, pluginName) {
             var $el = (typeof selector === 'string') ? $(selector) : selector,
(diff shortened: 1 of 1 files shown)


themes/Frontend/Responsive/frontend/_public/src/js/jquery.state-manager.js
```diff
@@ -1120,6 +1122,8 @@
                     $currentEl.removeData(name);
                 }
             }
+
+            return me;
```
shyim: Why not just this?

aragon999: Because I feel, uncomfortable to operate on `this` at the end of a function call, if there are some lines of code in between. Since one always needs to check if there was a modification of `this` in between (should not happen in general).
For better code readability I choose this format. I can change it to `this` though and remove the declaration of `me`.

shyim: We use me only, when we need the parent this in a function. This does not match to this.

shyim:

PR #2224, opened 2019-10-03
Changed missing german translation

entensee403:
### 1. Why is this change necessary?
Because it's possible that a german shopware translation displays english messages which should not happen.

### 2. What does this change do, exactly?
Adds missing german translation for 3 text snippets

### 3. Describe each step to reproduce the issue or behaviour.
Go to the article list, activate the split-view mode and try to activate the split-view mode again. You will get an english message stating that it's not possible to activate split view mode again.
not_translated_text

### 4. Please link to the relevant issues (if any).
-

### 5. Which documentation changes (if any) need to be made because of this PR?
-

### 6. Checklist

- [ ] I have written tests and verified that they fail without my change
- [X] I have squashed any insignificant commits
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [X] I have read the contribution requirements and fulfil them.

PR #2225, opened 2019-10-04
Changed missing german translation

entensee403:
This is a copy from https://github.com/shopware/shopware/pull/2224 which was for the wrong branch.

Because it's possible that a german shopware translation displays english messages which should not happen.

Adds missing german translation for 3 text snippets

Go to the article list, activate the split-view mode and try to activate the split-view mode again. You will get an english message stating that it's not possible to activate split view mode again.
not_translated_text

### 6. Checklist


shyim:

PR #2227, opened 2019-10-07
fix wrong value on shipping comment column

resslinger:
In the shipping cost backend widget, the column "Interner Kommentar" has the value of the description instead of the value of the comment column.
That's confusing.

This PR changed the view from the description to the comment field.

### 6. Checklist


shyim:

PR #2229, opened 2019-10-11
Add information about out of stock products and disable checkout proceed button in shopping cart

HH-SW:
Add information about out of stock products and disable checkout proceed button in shopping cart

customers with sold out products in shopping cart should not go in the checkout process

It checks the laststock and instock value like in frontend_checkout_cart_item_quantity_selection

Add product to basket and set instock to 0 and laststock to 1 and compare it with active 0

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Checkout.php
+++ b/engine/Shopware/Controllers/Frontend/Checkout.php
@@ -2141,7 +2141,9 @@ class Shopware_Controllers_Frontend_Checkout extends Enlight_Controller_Action i
             }
 
             if (!empty($item['additional_details'])) {
-                continue;
+                if (!$item['additional_details']['laststock'] || $item['additional_details']['instock'] > 0) {
+                    continue;
+                }
             }
 
             $products[] = $item['articlename'];


HH-SW:
grafik

without this change:
grafik

pascalheidmann-bedarf:
@HH-SW you can simplify the condition to just one single `if`. Also you aren't handling the case of having less qty than the customer wants to order.

HH-SW:
@pascalheidmann-bedarf i know, but for short code lines i used this style.
`                 if (!$item['additional_details']['laststock'] || $item['additional_details']['instock'] > 0) {`
its simpler than in sw-template:
` {if !$sBasketItem.additional_details.laststock || ($sBasketItem.additional_details.laststock && $sBasketItem.additional_details.instock > 0)}`

> Also you aren't handling the case of having less qty than the customer wants to order.
But than the cart-information "following products are not available" is not valid.

There are more conditions like stockmin and minpurchase, but i think it's less important and not compatible at this point with getInvalidProducts.

With my cart-protection and this core-modification we have no more oversells. <3

declined *lol*

shyim:
Please also do not forgot to sign the cla

PR #2231, opened 2019-10-14
Allow deleting prices from non default pricegroups

pascalheidmann-bedarf:
As soon as you enter a price for a price group that isn't the first (default pricegroup) you can delete one entry but you never are able to use the fallback again (via backend extjs GUI)

Allow deleting the last (first) price row if the pricegroup isn't the default price group (`EK`).

Prerequisite: have at least 2 customer groups
1. create or update existing product by adding a price for the second price group
2. The second price group should show a minus icon for deleting each price row
3. Modify a price in the second customer group and change it back
4. Modify the price in the primary customer group - the change won't be reflected in the second customergroup (won't fix)
5. Delete every row you can for that second customer group -> you can't delete the last row -> you are stuck with the prices

### 6. Checklist


shyim:
I have discussed your Change with our internal QA and Design team.
We want that the last entry of non default group has a new icon in the actioncolumn with class sprite-chain where the user can reset it to fallback to EK. Before setting the fallback, it should show a modal with a warning.

Your current changes works only for scale prices. It should also work for normal prices.

pascalheidmann-bedarf:
@shyim good points! I looked into it again and implemented a confirmation request in a completely new event - this made the hassles I hadn't found a solution for way easier to tackle.
As a bonus I added the missing translation for the `delete` tooltip ;)

pascalheidmann-bedarf:

shyim:
Hey @pascalheidmann @pascalheidmann-bedarf,

I found two issues with your PR:

* When you have 2 rows and delete one the icon is not shown until you switch the tabs. Maybe a rerender is missing here.
* When you delete / link the row to the default group go to first tab and then back. You can do it again

PR #2232, opened 2019-10-15
Add laststock to sExport.php

JonidBendo:

`laststock` is needed to correctly define product availability for product feeds. By adding it to this query it will allow users to use it in the item export for feeds.

Adds an existing column to the query so it will be available for use in the template for item exports (item feeds).

Go to shopware backend -> item export 
Create a feed and try to use `laststock` in the body, it does not exist.

`laststock` needs to be added to the available variables here: https://docs.shopware.com/en/shopware-5-en/marketing-and-shopping-worlds/item-exports

### 6. Checklist


shyim:

PR #2233, opened 2019-10-15
Add attribute fields to API + added API testcases

pascalheidmann-bedarf:
While some API endpoints supply attributes not all do. But AFAIK none allows to pass the attributes for create or update calls.
Also the API isn't part of the typical test cases right now and has low coverage.

- adds support for attributes on `create` & `update` calls
- adds attributes to `getOne` and `getList` resource calls that won't make these available yet
- add & extend api tests

1. make sure you have attributes for `media` (`s_media`/`s_media_attributes`)
2. call api endpoint for media with valid post payload containing an `attribute` field
3. check for created media without those passed attributes without PR, with PR they will be there 

- #2196 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Media.php
+++ b/engine/Shopware/Components/Api/Resource/Media.php
@@ -28,6 +28,7 @@ use Doctrine\ORM\ORMException;
 use Shopware\Components\Api\Exception as ApiException;
 use Shopware\Components\Random;
 use Shopware\Components\Thumbnail\Manager;
+use Shopware\Models\Attribute\Media as MediaAttribute;
 use Shopware\Models\Media\Album;
 use Shopware\Models\Media\Media as MediaModel;
 use Symfony\Component\HttpFoundation\File\File;
@@ -127,6 +128,12 @@ class Media extends Resource
 
         $media = new MediaModel();
         $media->fromArray($params);
+        $attribute = new MediaAttribute();
+
+        if (isset($params['attribute']) && is_array($params['attribute'])) {
+            $attribute->fromArray($params['attribute']);
+        }
+        $media->setAttribute($attribute);
 
         $path = $this->prepareFilePath($media->getPath(), $media->getFileName());
         $media->setPath($path);
@@ -136,6 +143,7 @@ class Media extends Resource
             throw new ApiException\ValidationException($violations);
         }
 
+        $this->getManager()->persist($attribute);
         $this->getManager()->persist($media);
         $this->flush();
 
@@ -184,6 +192,15 @@ class Media extends Resource
                 @unlink($path);
                 throw new ApiException\CustomValidationException($exception->getMessage());
             }
+        }
+
+        if (!empty($params['attribute'])) {
+            $attribute = $media->getAttribute();
+            $attribute->fromArray($params['attribute']);
+
+            $media->setAttribute($attribute);
+            $this->getManager()->persist($attribute);
+            $this->getManager()->flush();
         }
 
         return $media;
--- a/tests/Functional/Api/OrderTest.php
+++ b/tests/Functional/Api/OrderTest.php
@@ -25,57 +25,45 @@
 namespace Shopware\Tests\Functional\Api;
 
 /**
- * @covers \Shopware_Controllers_Api_GenerateArticleImages
+ * @covers \Shopware_Controllers_Api_Orders
  */
-class GenerateArticleImagesTest extends AbstractApiTestCase
+class OrderTest extends AbstractApiTestCase
 {
-    public function testRequestWithoutAuthenticationShouldReturnError()
+    public function testRequestWithoutAuthenticationShouldReturnError(): void
     {
-        $this->client->request('GET', '/api/generateArticleImages');
-        $response = $this->client->getResponse();
-
-        static::assertEquals('application/json', $response->getHeader('Content-Type'));
-        static::assertEquals(401, $response->getStatusCode());
-
-        $result = $response->getBody();
-        $result = json_decode($result, true);
-
-        static::assertArrayHasKey('success', $result);
-        static::assertFalse($result['success']);
-        static::assertArrayHasKey('message', $result);
-    }
-
-    public function testBatchDeleteShouldFail(): void
-    {
-        $this->authenticatedApiRequest('DELETE', '/api/generateArticleImages');
+        $this->client->request('GET', '/api/orders/');
         $response = $this->client->getResponse();
 
         static::assertEquals('application/json', $response->headers->get('Content-Type'));
         static::assertEquals(null, $response->headers->get('Set-Cookie'));
-        static::assertEquals(405, $response->getStatusCode());
+        static::assertEquals(401, $response->getStatusCode());
 
         $result = $response->getContent();
+
         $result = json_decode($result, true);
 
         static::assertArrayHasKey('success', $result);
         static::assertFalse($result['success']);
-        static::assertEquals('This resource has no support for batch operations.', $result['message']);
+
+        static::assertArrayHasKey('message', $result);
     }
 
-    public function testBatchPutShouldFail()
+    public function testGetOrderWithInvalidIdShouldReturnMessage(): void
     {
-        $this->authenticatedApiRequest('PUT', '/api/generateArticleImages');
+        $id = 9999999999;
+        $this->authenticatedApiRequest('GET', '/api/orders/' . $id);
         $response = $this->client->getResponse();
 
-        static::assertEquals('application/json', $response->headers->get('content-type'));
-        static::assertEquals(null, $response->headers->get('set-cookie'));
-        static::assertEquals(405, $response->getStatusCode());
+        static::assertEquals('application/json', $response->getHeader('Content-Type'));
+        static::assertEquals(404, $response->getStatusCode());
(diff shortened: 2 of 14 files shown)


engine/Shopware/Components/Api/Resource/Media.php
```diff
@@ -127,6 +128,12 @@ public function create(array $params)
 
         $media = new MediaModel();
         $media->fromArray($params);
+        $attribute = new MediaAttribute();
```
shyim: This will overwrite all attributes

engine/Shopware/Components/Api/Resource/Media.php
```diff
@@ -186,6 +194,15 @@ public function update($id, array $params)
             }
         }
 
+        if (!empty($params['attribute'])) {
+            $attribute = $media->getAttribute();
```
shyim, line 198: getAttribute could be null

engine/Shopware/Components/Api/Resource/Media.php
```diff
@@ -186,6 +194,15 @@ public function update($id, array $params)
...
+            $attribute->fromArray($params['attribute']);
+
+            $media->setAttribute($attribute);
+            $this->getManager()->persist($attribute);
+            $this->getManager()->flush();
```
shyim, line 203: $this->flush() to consider flush mode :wink:

tests/phpunit_api.xml.dist
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
shyim: why own phpunit file, just use functional for that

tests/Api/OrderTest.php
```diff
@@ -0,0 +1,76 @@
...
+        $response = $this->getHttpClient()
+            ->setUri($this->apiBaseUrl . self::API_PATH . $id)
+            ->request('GET');
+
+        static::assertEquals('application/json', $response->getHeader('Content-Type'));
+        static::assertEquals(404, $response->getStatus());
+
+        $result = $response->getBody();
+
+        $result = Zend_Json::decode($result);
```
shyim: json_decode :blush:

tests/Api/bootstrap.php
```diff
@@ -0,0 +1,58 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+require_once __DIR__ . '/../../autoload.php';
+require_once __DIR__ . '/Traits/ApiSetupTrait.php';
```
shyim: Thats uggly, please use namespace with autoload...

tests/Api/Traits/ApiSetupTrait.php
```diff
@@ -0,0 +1,97 @@
...
+     * @throws \Zend_Http_Client_Exception
+     *
+     * @return Zend_Http_Client
+     */
+    public function getHttpClient(bool $auth = true)
+    {
+        $username = 'demo';
+        $password = sha1('demo');
+
+        $adapter = new Zend_Http_Client_Adapter_Curl();
```
shyim: Use Shopware HttpClient for that

tests/Api/Traits/ApiSetupTrait.php
```diff
@@ -0,0 +1,97 @@
...
+        $helper = Shopware();
+
+        $hostname = $helper->Shop()->getHost();
+        if (empty($hostname)) {
+            static::markTestSkipped(
+                'Hostname is not available.'
+            );
+        }
+
+        $this->apiBaseUrl = 'http://' . $hostname . $helper->Shop()->getBasePath() . '/api';
```
shyim: Consider secure flag on scheme

tests/Api/Traits/ApiSetupTrait.php
```diff
@@ -0,0 +1,97 @@
...
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Tests\Api\Traits;
+
+use Zend_Http_Client;
+use Zend_Http_Client_Adapter_Curl;
+use Zend_Http_Client_Adapter_Exception;
+
+trait ApiSetupTrait
```
shyim: Maybe AbstractApiTestCase and base all other cases on it

tests/Api/Traits/ApiSetupTrait.php
pascalheidmann-bedarf: I don't want to rewrite everything in this PR. Might change that in a later one but this would explode it too much in my opinion.

tests/Api/bootstrap.php
pascalheidmann-bedarf: Had the problem that it wouldn't autoload whatever I tried. But might be not a problem if I replace it with an abstract class.

tests/Api/Traits/ApiSetupTrait.php
pascalheidmann-bedarf: @shyim is there any documentation to that _HttpClient_? My IDE hadn't even found anything when I searched for it 😅

tests/Api/Traits/ApiSetupTrait.php
pascalheidmann-bedarf: Hmmm looks a bit like Travis CI doesn't like that 🙈

dev-ops/common/actions/unit-fast.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 #DESCRIPTION: execute unit tests without coverage
 
-vendor/bin/phpunit tests/Functional/ --config tests/phpunit.xml.dist --log-junit build/artifacts/test-log.xml --exclude-group=elasticSearch
 vendor/bin/phpunit tests/Unit/ --config tests/phpunit_unit.xml.dist --log-junit build/artifacts/test-log.xml
+vendor/bin/phpunit tests/Functional/ --config tests/phpunit.xml.dist --log-junit build/artifacts/test-log.xml --exclude-group=elasticSearch
+vendor/bin/phpunit tests/Api/ --config tests/phpunit.xml.dist --log-junit build/artifacts/test-log.xml --exclude-group=elasticSearch
```
shyim: Add it please to the functional tests

tests/Api/Traits/ApiSetupTrait.php
shyim: \Shopware\Components\HttpClient\HttpClientInterface

tests/Functional/Api/VersionTest.php
```diff
@@ -22,70 +22,27 @@
...
-        Shopware()->Db()->query('UPDATE s_core_auth SET apiKey = ? WHERE username LIKE "demo"', [sha1('demo')]);
-    }
-
-    /**
-     * @throws Zend_Http_Client_Adapter_Exception
-     *
-     * @return Zend_Http_Client
-     */
-    public function getHttpClient()
+    public function getApiResource(): string
```
shyim: protected class constant, can you access then with static::NAME

pascalheidmann-bedarf: I haven't tested it lately but AFAIK redeclaring a const will cause an php error :/

pascalheidmann-bedarf:
Right now the tests fail on Travis CI, on my local machine they work - it seams like there is some problem with name resolution for localhost within Travis in combination with `curl`.
As a possible solution @shyim suggested to make use of Symfony's internal routing like the controller tests do. But it seems like the lifecycle there won't call the `postDispatch` functions if an exception occurs - but as the API makes heavily use of the predefined exception handling they won't be catched correctly and every assertion fails.

Therefore any solution from @shopwareBot or anyone else is warmly welcome!

shyim:
Hey,
I have refactored all tests. You can find it here as patch https://dpaste.de/Drb6
Hopefully will be merged tomorrow.

You have just to create a new Test class and extend from AbstractApiTestCase then you can do all api operations with authentificatedApiRequest method like

Example:
```
$this->authenticatedApiRequest('GET', '/api/paymentmethods/' . $id);
$response = $this->client->getResponse();

$result = $response->getContent();
$result = json_decode($result, true);
```

pascalheidmann-bedarf:
Updated test cases to make use of @shyim solution - now everything seems to work as intended 👍

tests/Functional/Api/OrderTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $result = json_decode($result, true);
+
+        static::assertArrayHasKey('success', $result);
+        static::assertFalse($result['success']);
+
+        static::assertArrayHasKey('message', $result);
+    }
+
+    public function testGetCustomerGroupWithInvalidIdShouldReturnMessage(): void
```
shyim: customergroup?

pascalheidmann-bedarf: I don't know what you have seen there... 🙈

pascalheidmann-bedarf:
@shopwareBot is this in an acceptable state for now or still anything `Incomplete`? ;)

pascalheidmann-bedarf:

shyim:

PR #2238, opened 2019-10-18
Fix mail context filter

sbrueggenolte:

Because the mail context filter ('TemplateMail_CreateMail_MailContext') has no effect on the context, that is actually being provided to the mail template.

It fixes the issue mentioned above. The filter event now filters the variable $context, that is later being passed to the mail template.

- Subscribe the filter event 'TemplateMail_CreateMail_MailContext'
- Change the value of context variables or add you own variables
- Add an ouput of these variables to a random mail template
- Trigger this mail (e.g. playce an order to trigger the sORDER mail. Do not send a test mail from within the mail template manager in the backend!)

n/a

n/a

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/TemplateMail.php
+++ b/engine/Shopware/Components/TemplateMail.php
@@ -232,16 +232,16 @@ class Shopware_Components_TemplateMail
             ];
         }
 
-        // Save current context to mail model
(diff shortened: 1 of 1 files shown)


mitelg:
thanks for your contribution @sbrueggenolte :+1: :tada: :blue_heart:

PR #2240, opened 2019-10-21
Add getter and setter for $formatDateTime in json renderer

jkrzefski:
The variable `$formatDateTime` already exists, but cannot be set externally yet. So there is no way to not use date formatting.

Add accessor methods for the variable. Also some necessary refactoring stuff to get the cs-fixer off my back.

1. 🤔 Try to use the json renderer without date formatting.
2. 😁 Be happy to see `$formatDateTime` and its usage in a condition.
3. 😭 Have all your hopes and dreams crushed by realizing there is no way to set it.

None.

None.

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Controller/Plugins/Json/Bootstrap.php
+++ b/engine/Library/Enlight/Controller/Plugins/Json/Bootstrap.php
@@ -79,13 +79,11 @@ class Enlight_Controller_Plugins_Json_Bootstrap extends Enlight_Plugin_Bootstrap
     /**
      * Called from the Event Manager after the dispatch process
      *
-     * @param Enlight_Event_EventArgs $args
-     *
      * @return bool
      */
     public function onPostDispatch(Enlight_Event_EventArgs $args)
     {
-        /** @var $controller Enlight_Controller_Action $controller */
+        /** @var Enlight_Controller_Action $controller $controller */
         $subject = $args->get('subject');
         $response = $subject->Response();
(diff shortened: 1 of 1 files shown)


jkrzefski:
No idea why travis fails. Can someone take a look at it, please?

mitelg:
@jkrzefski try a rebase

engine/Library/Enlight/Controller/Plugins/Json/Bootstrap.php
```diff
@@ -1,20 +1,25 @@
 <?php
 /**
- * Enlight
+ * Shopware 5
```
shyim: No!

jkrzefski: I just applied `composer run cs-fix` to this file and it changed the header comment. I thought, that was expected.

jkrzefski:
@mitelg I am up to date. There is nothing to rebase onto.

engine/Library/Enlight/Controller/Plugins/Json/Bootstrap.php
shyim: It has to stay at enlight

mitelg:
@jkrzefski the failing is because the code style is broken in tests/Functional/Components/Api/OrderTest.php
I will create a PR for that, as it is not related to your changes

shyim:

PR #2241, opened 2019-10-28
Make ListProductService interchangeable with ProductService

jkrzefski:
When I have a service that normally uses a `ListProductService` and now I need more detailed information in the resulting structs, I now have the option to change the service definition to use a `ProductService` for the exact same code. E.g. the `ProductSearch` could then have `Product` instances instead of `ListProduct` instances in its results. That will not break anything, because `Product` extends `ListProduct`, so it can be used in the same way. The difference is that now I can decorate the `ProductSearch` and work with the information of `Product` structs.

`interface ProductServiceInterface extends ListProductServiceInterface`

See 1.

None.

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Service/ProductServiceInterface.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Service/ProductServiceInterface.php
@@ -27,7 +27,7 @@ namespace Shopware\Bundle\StoreFrontBundle\Service;
 use Shopware\Bundle\StoreFrontBundle\Struct\Product;
 use Shopware\Bundle\StoreFrontBundle\Struct\ProductContextInterface;
 
-interface ProductServiceInterface
+interface ProductServiceInterface extends ListProductServiceInterface
 {
     /**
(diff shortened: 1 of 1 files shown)


JoshuaBehrens:
Isn't it a bit irritating that the interface is extending an other one and serves a different level of detail? Extending an interface to add new things is cool but change signatures from existing ones is a bit weird, isn't it? If we had generics or shopware6 with typed collections this one would not even be possible. Just a :duck::keyboard: thing that would not work in the future as static typing fills php more and more.

jkrzefski:
I don't see it as that much of a problem. The resulting `Product` struct of the `ProductServiceInterface` even extends the `ListProduct` struct of the `ListProductInterface`. So the `ProductService` itself and also the results from each of its public methods are usable in exactly the same way and should therefore not break a thing.

Also I believe this is somewhat similar to how the (now deprecated) `ProductContext` was replaced with the newer `ShopContext`. Both of them still exist, but some services still require the `ProductContext`, but get the `ShopContext` instead. An IDE suggests that these calls are incompatible, but as long as it is just hinted in phpdocs, it is not an issue.

shyim:

PR #2243, opened 2019-10-29
Check for invalid cache result

fivetide:
In certain circumstances - i.e. when using Redis-Backend with LRU-Eviction active - loading a cached menu can return false, despite a successful test in the condition prior. 

Check if we actually received a proper menu-array from cache

- Set up Shopware to use redis backend.
- Load any page once.
- Remove the cache item from redis using cli, but keep the set containing tags.
- Reload the page. $menu will be "false" 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Bootstrap.php
@@ -166,7 +166,9 @@ class Shopware_Plugins_Frontend_AdvancedMenu_Bootstrap extends Shopware_Componen
 
         if ($this->Config()->get('caching') && $cache->test($cacheKey)) {
             $menu = $cache->load($cacheKey, true);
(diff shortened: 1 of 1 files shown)


fivetide:
Hey @shopwareBot ,
> Yet, we're not quite sure if this fix is the right one. This feels like fixing a symptom, not the cause itself.

I agree, this is fixing a symptom, not the cause itself.

>  The prior condition shouldn't be true in the first place here, should it?

Agreed, it should not be.
Since the issue ist with Zend Framwork - and the version used by Shopware having reached EOL three years ago, i fail to see a different approach except doing an actual upgrade Zend Framwork.

fivetide:
@shyim @shopwareBot , do you see any different route to solve this?

shyim:
Hey @fivetide,
We are currently internally working on an Update to Symfony 4.4 and migrating some Zend components to Symfony. One of it is the Caching component. With this change will this issue fixed globally.

PR #2244, opened 2019-10-30
SW-24821 / Disable submit button to prevent multiple form submits

jmatthiesen81:
This change is necessary to prevent multiple submit from the registration form. This change fixes the issue https://issues.shopware.com/issues/SW-24821.

This change disables the submit button after the first form submit was triggered.

Click the registration submit button more than one time, the user is created on the first request but the request will be canceled after the next click. This results in an error, because the user email address now already exists in the database, created by the first request.

https://issues.shopware.com/issues/SW-24821

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/register/index.tpl
+++ b/themes/Frontend/Bare/frontend/register/index.tpl
@@ -215,7 +215,7 @@
                     {block name='frontend_register_index_form_submit'}
                         {* Submit button *}
                         <div class="register--action">
-                            <button type="submit" class="register--submit btn is--primary is--large is--icon-right" name="Submit">{s name="RegisterIndexNewActionSubmit"}{/s} <i class="icon--arrow-right"></i></button>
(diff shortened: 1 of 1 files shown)


pascalheidmann-bedarf:
Is that form ajaxified? You then might want to reenable if the request fails because of network problems etc.

shyim:
Maybe use data-preloader-button instead. Like in confirm page

jmatthiesen81:
> Is that form ajaxified? You then might want to reenable if the request fails because of network problems etc.

No it isn't.

jmatthiesen81:
> Maybe use data-preloader-button instead. Like in confirm page

That solution also works. Both disabling the submit button.

shyim:
I would prefer then the data-preloader-button :blush:

jmatthiesen81:
> I would prefer then the data-preloader-button

Done

shyim:

PR #2245, opened 2019-10-30
[SW-24808] Fix for duplicate names in plugin defined menu

lx-wnk:
If you want to install two plugins with the same menu entry

Removes an index on `name` inside
 `s_core_menu`

Use the provided plugins and try to install (activate) both
[examplePlugins.zip](https://github.com/shopware/shopware/files/3788431/examplePlugins.zip)

https://issues.shopware.com/issues/SW-24808

///

### 6. Checklist
  - Tested the backend accessibility

### DIFF
--- a/engine/Shopware/Components/Plugin/MenuSynchronizer.php
+++ b/engine/Shopware/Components/Plugin/MenuSynchronizer.php
@@ -81,6 +81,11 @@ class MenuSynchronizer
             $items[] = $this->createMenuItem($plugin, $parent, $menuItem);
         }
 
+        foreach ($menuNames as &$menuName) {
+            $menuName = $plugin->getName() . '_' . $menuName;
+        }
+        unset($menuName);
+
         $this->em->flush($items);
         $this->removeNotExistingEntries($plugin->getId(), $menuNames);
     }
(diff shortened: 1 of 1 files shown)


lx-wnk:
@shopwareBot you are absolutely right.
I don't know how I managed to not getting it work like this before.

PaddyS:
Hey @jinnoflife,

I'm afraid you misunderstood our intention. :-(
It's the plugin manufacturers responsibility to add their custom plugin prefix when adding a menu-entry.

We did not want this to be changed in the code in any way.
We're just wondering, you wanted to have duplicate menu entries, when you could have just added a prefix to the menu entry in your plugin.

So the question still remains:
Was there any specific reason, why you wanted to create a duplicated menu entry in your plugin, when you could have just appended your manufacturer prefix to the menu entry's name?

Sorry for any inconveniences!

lx-wnk:
@PaddyS I think I really miss understood.
I'll close this PR :sweat_smile:

I didn't wanted to create an entry. I just saw the ticket and wanted to resolve it.
Also wouldn't it be better to provide the prefix for every plugin instead of give it to the developers hand? This would also resolve the issue and could provide some usability for developers

PR #2246, opened 2019-10-30
[SW-24816] Fix for remaining additional name if product is no longer a variant-p…

lx-wnk:
Because of Point 3, [this](https://forum.shopware.com/discussion/54172/) and [that](https://issues.shopware.com/issues/SW-24816)

Removes the option from the article detail
- Checks if product was a configurator product before

Create a product, create variants for this product, delete all variants except the main variant, uncheck the variant checkbox, put the product in your basket and look at the article name.

https://issues.shopware.com/issues/SW-24816

///

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Article.php
+++ b/engine/Shopware/Controllers/Backend/Article.php
@@ -3099,12 +3099,25 @@ class Shopware_Controllers_Backend_Article extends Shopware_Controllers_Backend_
     {
         $data = $this->prepareAssociatedData($data, $article);
 
+        $removeMainDetailOptions = false;
+        if ($article->getConfiguratorSet() !== null && $data['configuratorSetId'] === null) {
+            $removeMainDetailOptions = true;
+        }
+
         $article->fromArray($data);
 
         Shopware()->Models()->persist($article);
         Shopware()->Models()->flush();
         if (empty($data['id']) && !empty($data['autoNumber'])) {
             $this->increaseAutoNumber($data['autoNumber'], $article->getMainDetail()->getNumber());
+        }
+
+        if ($removeMainDetailOptions) {
+            $this->getManager()->getDBALQueryBuilder()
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Article.php
```diff
@@ -3107,6 +3111,11 @@ protected function saveArticle($data, $article)
             $this->increaseAutoNumber($data['autoNumber'], $article->getMainDetail()->getNumber());
         }
 
+        if ($hadConfiguratorSet) {
+            $sql = 'DELETE FROM s_article_configurator_option_relations WHERE article_id = ' . $article->getMainDetail()->getId();
```
shyim: prepared statement

engine/Shopware/Controllers/Backend/Article.php
```diff
@@ -3107,6 +3111,11 @@ protected function saveArticle($data, $article)
...
+            Shopware()->Db()->query($sql);
```
shyim: Don't use Shopware() global

shyim: Don't use Zend_Db

lx-wnk: You are absolutely right.
I just don't wanted to refactor the whole classe because of this.
The usage of Shopware() and Zend is strong in this one.

If you want I can simply change it to `$this->getManager` but in this it will still use the global.

shyim: its enought when you don't add more of that :)

lx-wnk: Okay, I'll resolve this 😄

PaddyS:
Hey @jinnoflife,

did you try this with the API as well?
Szenario:
- Create product with configurator set
- Update the product to remove the configurator set.

Also, can you please add tests?

PR #2247, opened 2019-10-31
added ignoredExceptionClasses also to testkernel config due to fatal …

buenosdiaz:
…through non-existent parameter

Executing PHPUnit will cause a fatal error otherwise due to missing config parameters, which were recently added to the default config.

`Fatal error: Uncaught Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException: The service "errorsubscriber" has a dependency on a non-existent parameter "shopware.errorHandler.ignoredExceptionClasses".`

Add the missing config parameter also in the test kernel config.

Just execute PHPUnit in the doc root. Make sure that the cache is cleared.

-

-

### 6. Checklist

- [ - ] I have written tests and verified that they fail without my change (not neccessary)
- [ - ] This change has comments for package types, values, functions, and non-obvious lines of code (not neccessary)

### DIFF
--- a/tests/Functional/config.php
+++ b/tests/Functional/config.php
@@ -30,6 +30,17 @@ return array_replace_recursive($this->loadConfig($this->AppPath() . 'Configs/Def
     ],
     'errorhandler' => [
         'throwOnRecoverableError' => true,
+        'ignoredExceptionClasses' => [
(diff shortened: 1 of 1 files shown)


shyim:
We don't have any problems. Also travis is green. Also the configuration extend the normal one.

PR #2249, opened 2019-11-01
Fix migration for shop pages in update to 5.6

jkrzefski:
The new template files use the new template keys. When the keys are only changed during a fresh install, the shop pages will not be displayed after an update. This has become a common problem that many face while updating from earlier versions to 5.6.

This change moves the sql to rename the german shop pages out of the condition for the migration mode to be `update`, because with that condition the shop pages are not displayed after the update.

1. Have a shopware shop in version `< 5.6`
2. Update to `5.6`
3. See no shop pages in the footer

None.

None.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1404-remove-language-distinction-from-shop-pages.php
+++ b/_sql/migrations/1404-remove-language-distinction-from-shop-pages.php
@@ -43,7 +43,8 @@ SQL;
 
         if ($modus === self::MODUS_INSTALL) {
             $this->addSql($dropEnglish);
-            $this->addSql($renameGerman);
         }
+
+        $this->addSql($renameGerman);
     }
 }


jkrzefski:
Yes, I know: This is technically a destructive change that should not be executed automatically as per blue/green standard. But it is a breaking change in any way. If you do execute it, it might disrupt some customized settings that are rather rare. If you do not execute it, the shop pages will for sure not be displayed. So I believe there is no "perfect" way to handle this situation and I propose to take the lesser evil. Of course: If you happen to know a better way to handle this, please feel free to correct me.

JoshuaBehrens:
Is it not possible to merge the languages into the translation tables?

shyim:
The main reason why we don't do that is that migrations should **not modify customers data**

You can't know what the customer has configured/built there. Shopware is highly customizeable

JoshuaBehrens:
If I read the [simple change log](https://www.shopware.com/de/changelog-sw5/#5-5-0) (not the technical one) there is no hint about the footer menu. This is covered as a feature "Übersetzungsfunktion für Shopseiten realisiert" without a hint (in the ticket as well https://issues.shopware.com/issues/SW-17501) that it is very likely to break menus or page customizations.

I can follow the guide not to modify customers data but the communication of the drawbacks of this feature was bad IIRC.

PR #2250, opened 2019-11-04
Fix a typo.

supus:
Because it fixes a bug in the register-form.

Fix a bug/ typo.

1. In the basic settings, go to the "Login / registration"-page.
2. Here, set "Show additional address line 1" and "Show additional address line 1" to "Yes".
3. Next, set only "Treat additional address line 2 as required" to "Yes".
4. Now, go to the register-page of your shop (f.ex. www.shop.de/account). Fill out all required inputs, while keeping the "The shipping address does not match the billing address"-Checkbox unset.
5. After that, click on the submit button. => Nothing happens, the responsible JS-Plugin could not set the hidden shipping-additionalAddressLine2-field to not-required.

None.

None.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/register/shipping_fieldset.tpl
+++ b/themes/Frontend/Bare/frontend/register/shipping_fieldset.tpl
@@ -97,7 +97,7 @@
                         <div class="register--additional-line1">
                             <input autocomplete="section-shipping shipping address-line2"
                                    name="register[shipping][additionalAddressLine1]"
-                                   type="text"{if {config name=requireAdditionAddressLine2}} required="required" aria-required="true"{/if}
+                                   type="text"{if {config name=requireAdditionAddressLine1}} required="required" aria-required="true"{/if}
                                    placeholder="{s name='RegisterLabelAdditionalAddressLine1'}{/s}{if {config name=requireAdditionAddressLine1}}{s name="RequiredField" namespace="frontend/register/index"}{/s}{/if}"
                                    id="additionalAddressLine21"
                                    value="{$form_data.additionalAddressLine1|escape}"
--- a/UPGRADE-5.6.md
+++ b/UPGRADE-5.6.md
@@ -40,6 +40,7 @@ This changelog references changes done in Shopware 5.6 patch versions.
 * Changed `Shopware\Models\Customer\Customer`, set correct return type for `getregisterOptInId`
 * Changed the `TemplateMail_CreateMail_MailContext` filter to work correctly.
 * Changed `ProductServiceInterface` to extend from `ListProductServiceInterface`
+* Changed additionAddressLine1 in `themes/Frontend/Bare/frontend/register/shipping_fieldset.tpl` to fix a typo.
 
 ### Removals
 


themes/Frontend/Bare/frontend/register/shipping_fieldset.tpl
```diff
@@ -97,7 +97,7 @@
                         <div class="register--additional-line1">
                             <input autocomplete="section-shipping shipping address-line2"
```
mitelg, line 98: this seems also wrong

themes/Frontend/Bare/frontend/register/shipping_fieldset.tpl
```diff
@@ -97,7 +97,7 @@
...
                                    name="register[shipping][additionalAddressLine1]"
-                                   type="text"{if {config name=requireAdditionAddressLine2}} required="required" aria-required="true"{/if}
+                                   type="text"{if {config name=requireAdditionAddressLine1}} required="required" aria-required="true"{/if}
                                    placeholder="{s name='RegisterLabelAdditionalAddressLine1'}{/s}{if {config name=requireAdditionAddressLine1}}{s name="RequiredField" namespace="frontend/register/index"}{/s}{/if}"
                                    id="additionalAddressLine21"
```
mitelg, line 102: and what about the ID? 🙈

shyim: That would be an bc

pascalheidmann-bedarf: `additionalAddressLine21`?

supus: @mitelg @pascalheidmann-bedarf 

I think the ID must be "additionalAddressLine21", because every field in the shipping_fieldset.tpl is prefixed with a "2".

F.ex.:
- lastname2
- street2
- additionalAddressLine22
- etc.

@shyim I'm sorry, I don't know what an "bc" is. Can you explain?

themes/Frontend/Bare/frontend/register/shipping_fieldset.tpl
supus: @mitelg Do you mean the "address-line2" part?

I think it is supposed to be that way:
- The "register[shipping][additionalAddressLine2]"-input has an "address-line3"-class.
- in "frontend/address/form.tpl" shopware uses the same classes, while not having the bug that I fixed in the PR

themes/Frontend/Bare/frontend/register/shipping_fieldset.tpl
shyim: breaking change

supus: @shyim Do you mean, that my adjustments are a breaking-change or that changing the ID would be a bc?

If it's my adjustments, how exactly are they a bc and how do I prevent that?

PR #2251, opened 2019-11-10
Sort payment in backend in a natural order

JoshuaBehrens:
It is difficult to sort multiple payments as they are sorted by id in the backend.

The sorting of the payment rules in the left sidebar is now by position like in the frontend and by active state. The id is not visible in the listing as it is confusing by now. The id is still visible in the detail view.

image

1. Open payments to change order
2. Select different payments to find out their order
3. Rearrange sorting parameters
4. Click a lot
5. Be confused :confused: 
6. Save and open the frontend to check
7. Possibly re-enter task at step 2

### 4. Checklist


JoshuaBehrens:
There is still some help needed. Somehow the id is not visible in the detail state yet. My vague ExtJs knowledge has been overcome. But I think the idea to change this should not be stopped by a simple ExtJs mistake. So lets first discuss the change and then find someone to show the missing primary key of the payments.

themes/Backend/ExtJs/backend/payment/view/payment/form_panel.js
```diff
@@ -149,6 +149,11 @@ Ext.define('Shopware.apps.Payment.view.payment.FormPanel', {
             fieldLabel: '{s name=formpanel_pluginID_label}PluginID{/s}',
             name: 'pluginId',
             disabled: true
+        },{
+            xtype: 'textfield',
```
uehler: id is an int, so this should be numberfield

uehler: and there is already a hidden id field in line 87, maybe that's the problem

mitelg: you could remove the hidden flag, but make sure, that this input field is not editable

JoshuaBehrens: Oh. I didn't see that. Neither in the frontend backend nor in the backend backend. I change that in a minute

JoshuaBehrens: (read in French accent) 15 hours later

JoshuaBehrens:
Sorry that I didn't leave a comment about my last (force) push. Yes, it was right that id was already in the field list and couldn't be configured twice:
image

PaddyS:
Hey @JoshuaBehrens,

we've got two ideas about this:
1st: If you really want this to be a field, please move it to the top, it should be the first field in the form.
We'd rather have the second option though, which is:
2nd: Instead of adding a field, wouldn't it be neat to have a titled fieldset around that form, just like a category detail entry?

category

JoshuaBehrens:
I don't really want it to be a field but now you can just read the number so it has to be equally simple to read it. A field is sufficient for that.
The fieldset is much more glamorous. I try my ExtJs skills on that one. I prefer it as well.

JoshuaBehrens:
I somehow did it. Check it out
grafik

PaddyS:
Awesome, thanks! 💙

PaddyS:
Hey @JoshuaBehrens,

We've just merged your changes into an internal branch, which we might sync later today.
Thus, your changes will be available with either Shopware 5.6.4 or 5.7. :-)

JoshuaBehrens:
5.6.3 is ready? :O

shyim:
While we are in freeze, we are already working on the next patch version

PR #2252, opened 2019-11-11
Remove facebook plugin (deprecated 5.0) completely

JoshuaBehrens:
As in migration 630 a deprecated plugin can be safely removed.

Copies migration 630 and replace Google with Facebook. It **deletes** an s_core_plugin entry that has been [deprecated since 5.0](https://github.com/shopware/shopware/blame/406925fbb6a8e98beb94e71ceca1fd473a6edc31/UPGRADE-5.0.md#L500).
That could've been done almost 4 years ago with migration 630 (deprecated since April 2015).

1. Look for not installed plugins in backend plugin manager
2. Look for not installed plugins via SQL as it is easier to copy into a script
3. Be confused why you get one more result in SQL than in plugin manager
4. :male_detective: 

Finally I can link an [issue](https://issues.shopware.com/issues/SW-13025).

### 5. Checklist

### DIFF
--- a/_sql/migrations/1705-remove-facebook-plugin.php
+++ b/_sql/migrations/1705-remove-facebook-plugin.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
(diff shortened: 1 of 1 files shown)


_sql/migrations/1705-remove-facebook-plugin.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+class Migrations_Migration1705 extends Shopware\Components\Migrations\AbstractMigration
+{
+    public function up($modus)
+    {
+        $sql = <<<'EOD'
+SET @pluginId = (SELECT id FROM s_core_plugins WHERE name = 'Facebook' and namespace = 'Frontend' and `source` = 'Default' LIMIT 1);
+SET @formId = (SELECT id FROM s_core_config_forms WHERE plugin_id = @pluginId);
+
+DELETE FROM s_core_config_elements WHERE form_id = @formId;
```
shyim: This could be risky. When the Facebook plugin is missing, @formId will be null and will remove all elements where form_id is null

JoshuaBehrens: Indeed. Good catch. I amend that

shyim:
Please test next time your Migration before submitting, there was an typo :see_no_evil:

JoshuaBehrens:
Woops. Didn't see it, just copied things over from an other migration. On the other hand: why didn't the automation catch this? :D

PR #2253, opened 2019-11-13
Use file hash as cache buster

JoshuaBehrens:
Using the file hash is much more reasonable as cache buster for any file that is served over the web.

Adds a custom less function swhash in js and php which looks up the file in parameter and returns its hash. Current hash algorithm is md5.

1. Replace loading indicator animation file
2. No change happen
3. Image was not reloaded as cache was not busted

In the future anyone should use the hash of a file for cache busting

### 5. Checklist


JoshuaBehrens:
I added an app secret that probably is discussable. My thoughts on that: if you have a local copy of a shop running on the same url for reasons you would get the same cached theme files although it is an other instance of the shop. This way the cache busters are valid per installation and do not depend on anything.

I still need to integrate that altered value into the swhash functions.

_sql/migrations/1650-add-application-secret-configuration.php
```diff
@@ -0,0 +1,60 @@
+<?php declare(strict_types=1);
+
+use Shopware\Components\Migrations\AbstractMigration;
+
+class Migrations_Migration1650 extends AbstractMigration
+{
+    public function up($modus)
+    {
+        $secret = bin2hex(random_bytes(32));
```
shyim: This is useless, all new installations will have the same one

_sql/migrations/1650-add-application-secret-configuration.php
```diff
@@ -0,0 +1,60 @@
+<?php declare(strict_types=1);
+
+use Shopware\Components\Migrations\AbstractMigration;
+
+class Migrations_Migration1650 extends AbstractMigration
+{
+    public function up($modus)
```
shyim: void me

_sql/migrations/1650-add-application-secret-configuration.php
shyim: I guess you are searching something like this https://github.com/shopware/shopware/blob/60fcfbab6c0397d0317475313672968ece29ef67/engine/Shopware/Bundle/PluginInstallerBundle/Service/UniqueIdGenerator/UniqueIdGenerator.php

_sql/migrations/1650-add-application-secret-configuration.php
JoshuaBehrens: I just copied the 1649 :sweat_smile: but you are right

_sql/migrations/1650-add-application-secret-configuration.php
JoshuaBehrens: Can you explain why every installation gets the same random bytes?

shyim: The build process executes all migrations and dumps than a install.sql. This install.sql will be used for all installations.

JoshuaBehrens: How will the unique id generator help me out there? It looks like I have to generate the id NOT in any migration

shyim: Maybe use the uniqueIDGenerator service for your secret. Its generates a unique id for the installation and saves that as config element for later usage.

JoshuaBehrens: But when do I generate it? A migration would be a no brainer running once and everything is fine. With that I have to check it every time

JoshuaBehrens: Now I get it. It is not about how to generate the key in this migration but use the service as it already provides such an installation specific id. Sorry my mistake. NICE!

shyim:
Hey @JoshuaBehrens,
unfortunately the Grunt implementation does not work.

```
❯ node node_modules/.bin/grunt
Running "fileExists:js" (fileExists) task
Checking 64 files for existence...
OK

Running "less:development" (less) task
>> ../themes/Frontend/Responsive/frontend/_public/src/less/_components/loading-indicator.less: [L92:C37] error evaluating function `swhash`: Cannot read property '@shopware-revision' of null
Warning: Error compiling ../web/cache/all.less Use --force to continue.

Aborted due to warnings.
```

Can you check and fix it?

JoshuaBehrens:
Good finding. Somehow `grunt watch` was not affected. Please check my changes.

soebbing:
Hi @JoshuaBehrens, thank you again for your contribution! 💙

mitelg:
happy 40 accepted PRs anniversary :smiley: :tada: :birthday: :blue_heart:

JoshuaBehrens:
Already 😲❓ insane! Thank you 🙇 You are looking closer to the statistics than I am. I think I'd missed it without you 💙

PR #2254, opened 2019-11-13
Remove the need to begin javascript files with a semicolon

jkrzefski:
To account for errors of others it has always been a good idea to begin a javascript file with a semicolon. That way a third party could have forgotten the last semicolon in a file and the merged code would still work. So I figured, it would be clever to have this extra semicolon be added automatically when merging the files.

Remove a bunch of semicolons at the beginning of javascript files.

Not really steps but rather an explanation: see 1.

None.

None. (I think)

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Theme/Compiler.php
+++ b/engine/Shopware/Components/Theme/Compiler.php
@@ -290,7 +290,7 @@ class Compiler
         $javascriptFiles = $this->javascriptCollector->collectJavascriptFiles($template, $shop);
         $content = '';
         foreach ($javascriptFiles as $jsFile) {
-            $content .= file_get_contents($jsFile) . "\n";
+            $content .= file_get_contents($jsFile) . ";\n";
         }
 
         if ($settings->getCompressJs()) {
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.invalid-tos-jump.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.invalid-tos-jump.js
@@ -1,4 +1,4 @@
-;(function($, window) {
+(function($, window) {
     'use strict';
 
     /**
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.add-article.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.add-article.js
@@ -1,4 +1,4 @@
-;(function ($, window) {
+(function ($, window) {
     'use strict';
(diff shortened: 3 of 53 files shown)


themes/Frontend/Responsive/frontend/_public/src/js/jquery.invalid-tos-jump.js
```diff
@@ -1,53 +1,53 @@
...
-
-        jumpToInvalid: function () {
-            var me = this;
-
-            window.scroll(0, me.$el.offset().top - (window.innerHeight/2));
-            me.$jumpLabel.addClass(me.opts.errorClass);
-        }
-    });
-})(jQuery, window);
+(function($, window) {
```
soebbing, line 1: What's up with this file? Why have all lines been touched?

engine/Shopware/Components/Theme/Compiler.php
```diff
@@ -290,7 +290,7 @@ public function compileJavascript($timestamp, Shop\Template $template, Shop\Shop
         $javascriptFiles = $this->javascriptCollector->collectJavascriptFiles($template, $shop);
         $content = '';
         foreach ($javascriptFiles as $jsFile) {
-            $content .= file_get_contents($jsFile) . "\n";
+            $content .= file_get_contents($jsFile) . ";\n";
```
shyim, line 293: Is this also in grunt so?

themes/Frontend/Responsive/frontend/_public/src/js/jquery.invalid-tos-jump.js
jkrzefski: I believe this file used to have `CRLF` instead of `LF` for line breaks. If you wish, I can try to only change the first line in it.

soebbing: No, that won't be necessary I think. Were just curious :)

engine/Shopware/Components/Theme/Compiler.php
jkrzefski: Good point. No, I believe not. But I also do not know, how to implement this behavior in grunt tasks. Can you give me a hint?

shyim:

jkrzefski:
@shyim Thanks for accepting it. But this still has no solution for grunt. Does this mean, grunt could potentially produce broken results now?

PR #2255, opened 2019-11-14
Fix update method so the existing component can be found

Isengo1989:
When you want to update a EKW element, you should use the createOrUpdate Method showed here (https://developers.shopware.com/developers-guide/custom-shopping-world-elements/#registering-a-new-element) - this will sadly not work since the component model ist not saving the component Name given in the method, but the name in the $data Array (see DB Table -> s_library_component).

The name therefore can never be matched and it will never update. 

It will change the findOneBy search for name from $componentName to $data['name']. The update works fine afterwards. 

$componentName seems to be unused after that. So it could be removed afterwards (I can change the PR) Please give a short feedback.

Install a Plugin with the usage of the ComponentInstaller -> update the fields / or add one -> a new element will be created.

-

https://developers.shopware.com/developers-guide/custom-shopping-world-elements/#registering-a-new-element -> maybe when the $componentName will be removed.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Emotion/ComponentInstaller.php
+++ b/engine/Shopware/Components/Emotion/ComponentInstaller.php
@@ -68,7 +68,7 @@ class ComponentInstaller
         $repo = $this->em->getRepository(Component::class);
         /** @var Component|null $component */
         $component = $repo->findOneBy([
-            'name' => $componentName,
+            'name' => $data['name'],
             'pluginId' => $plugin->getId(),
         ]);
         if (!$component) {


engine/Shopware/Components/Emotion/ComponentInstaller.php
```diff
@@ -68,7 +68,7 @@ public function createOrUpdate($pluginName, $componentName, array $data)
         $repo = $this->em->getRepository(Component::class);
         /** @var Component|null $component */
         $component = $repo->findOneBy([
-            'name' => $componentName,
+            'name' => $data['name'],
```
shopwareBot, line 71: We'd like to stick to the `$componentName` variable for this.

Instead of replacing it with the name from `$data`, rather unset the `name` key from the array to ensure it's not being overridden.

Isengo1989: @shopwareBot works for me - the docs need to be edited for that too then. I will update the PR later

PaddyS:
Hey @Isengo1989,

still on it?

PR #2257, opened 2019-11-14
Pre-check selector condition in swPanelAutoResizer to prevent endless resize recursion

JoshuaBehrens:
Prevent an Uncaught RangeError: Maximum call stack size exceeded.

In a case that a `panelHeaderSelector`, `panelBodySelector` or `panelFooterSelector` are an invalid parameter don't invoke a recursive call.

Register the `swPanelAutoResizer` plugin with a partially filled configuration like this:
```js
$.subscribe('plugin/swEmotionLoader/onDeviceChange', function(event, plugin) {
    StateManager.addPlugin(plugin.$el.find('.foo-bar'), 'swPanelAutoResizer', {
        panelBodySelector: '.foo-bar--sub-element'
    });
});
```

This might be more a problem with an initialization problem but one can forcefully unset the values and still run into the error. So better check it in any case.

### 5. Checklist

- [x]  I have squashed any insignificant commits

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.panel-auto-resizer.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.panel-auto-resizer.js
@@ -205,9 +205,18 @@
 
             // shortcut to resize all
             if (typeof selector === 'undefined') {
-                me.resize(me.opts.panelHeaderSelector);
-                me.resize(me.opts.panelBodySelector);
-                me.resize(me.opts.panelFooterSelector);
+                if (typeof me.opts.panelHeaderSelector !== 'undefined') {
+                    me.resize(me.opts.panelHeaderSelector);
+                }
+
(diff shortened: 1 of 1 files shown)


JoshuaBehrens:
Sure I can. I still think that both issues have to be addressed but as I stated: one can also set the values to a bad value and eventually end up in a recursion error.

PaddyS:
@JoshuaBehrens

Why not both?

I can imagine there's a better way to solve that issue then. Thinking of the PHP way, you would use `in_array([...], 'undefined')` or something alike, maybe there's an equal way in JS?

JoshuaBehrens:
I assume that `Object.assign`/`{ ...defaults, ...opts }` would be my way to go but I am unsure about the browser compatibility. Is `Object.assign` a thing nowadays for a dated browser?

PaddyS:
Seems like `Object.assign` is not available in IE.

JoshuaBehrens:
I guess there will be a way to imitate that.

PR #2258, opened 2019-11-15
Autoload Resources/templates directory

JoshuaBehrens:
Once upon a time it was common to put js and less files under `Resources/views/frontend/_public/src` and register the files manually via subscriber. A wise man automatically loads these source files now from `Resources/frontend`. But why are the template files in `Resources/views` not automatically loaded?
To reduce the chance of a breaking change I chose the directory `Resources/templates` as these are template files that needs to be loaded. In case of specific reasons to not load them automatically one can still load them from elsewhere conditionally.

Subscribe `Enlight_Controller_Action_PreDispatch` and tell the controller template manager to use the plugins path.

1. Put `all.less` in `Resources/frontend/less`
2. Style is online
3. Put `foobar.js` into `Resources/frontend/js`
4. Dynamic DOM is online
5. Put `tpl` files into `Resources/views`
6. Nothing happens
7. Adds subscriber class as one does not want to pollute the bootstrap file
8. Add reference to subscriber service in `Resources/services.xml` (which gets loaded automatically!)

The information about this new directory has to be propagated somewhere.

### 5. Checklist

### DIFF
--- a/engine/Shopware/Components/Plugin.php
+++ b/engine/Shopware/Components/Plugin.php
@@ -47,6 +47,8 @@ abstract class Plugin extends Bundle implements SubscriberInterface
      */
     protected $pluginNamespace;
 
+    protected $autoloadViews = false;
+
     /**
      * @var bool
      */
@@ -84,6 +86,11 @@ abstract class Plugin extends Bundle implements SubscriberInterface
     final public function isActive()
     {
         return $this->isActive;
+    }
+
+    public function isAutoloadViews(): bool
+    {
+        return $this->autoloadViews;
     }
 
     /**
--- a/engine/Shopware/Components/Plugin/ResourceSubscriber.php
+++ b/engine/Shopware/Components/Plugin/ResourceSubscriber.php
@@ -26,6 +26,7 @@ namespace Shopware\Components\Plugin;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Enlight\Event\SubscriberInterface;
+use Enlight_Event_EventArgs;
 use Shopware\Components\Theme\LessDefinition;
 use Symfony\Component\Finder\Finder;
 
@@ -37,11 +38,17 @@ class ResourceSubscriber implements SubscriberInterface
     private $pluginPath;
 
     /**
+     * @var bool
+     */
+    private $loadViewsDirectory;
+
+    /**
      * @param string $pluginPath
      */
-    public function __construct($pluginPath)
+    public function __construct($pluginPath, ?bool $loadViewsDirectory = null)
     {
         $this->pluginPath = $pluginPath;
+        $this->loadViewsDirectory = $loadViewsDirectory ?? false;
     }
 
     /**
@@ -53,6 +60,7 @@ class ResourceSubscriber implements SubscriberInterface
             'Theme_Compiler_Collect_Plugin_Less' => 'onCollectLess',
             'Theme_Compiler_Collect_Plugin_Css' => 'onCollectCss',
             'Theme_Compiler_Collect_Plugin_Javascript' => 'onCollectJavascript',
+            'Theme_Inheritance_Template_Directories_Collected' => 'onRegisterTemplate',
         ];
     }
 
@@ -96,6 +104,24 @@ class ResourceSubscriber implements SubscriberInterface
             [],
             [$file]
         );
+    }
+
+    public function onRegisterTemplate(Enlight_Event_EventArgs $args): void
+    {
+        if (!$this->loadViewsDirectory) {
+            return;
+        }
+
+        $viewsDirectory = $this->pluginPath . '/Resources/views';
+
+        if (@is_dir($viewsDirectory)) {
+            $templates = (array) $args->getReturn();
+
+            if (!in_array($viewsDirectory, $templates, true)) {
+                $templates[] = $viewsDirectory;
+                $args->setReturn($templates);
+            }
+        }
     }
 
     /**
--- a/engine/Shopware/Kernel.php
+++ b/engine/Shopware/Kernel.php
@@ -254,7 +254,10 @@ class Kernel extends SymfonyKernel
             }
 
             $this->container->get('events')->addSubscriber($bundle);
-            $this->container->get('events')->addSubscriber(new Plugin\ResourceSubscriber($bundle->getPath()));
+            $this->container->get('events')->addSubscriber(new Plugin\ResourceSubscriber(
+                $bundle->getPath(),
+                $bundle instanceof Plugin && $bundle->isAutoloadViews()
+            ));
         }
 
         $this->booted = true;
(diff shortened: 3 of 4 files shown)


jkrzefski:
I dislike the idea of inventing a new standard location for template files. Especially since plugins are now glorified bundles, we should conform to symfony's standards as described in [Best Practices for Reusable Bundles (Symfony Bundles Docs)](https://symfony.com/doc/current/bundles/best_practices.html#directory-structure). I would prefer a simple switch that can easily be activated to autoload the template directory. Maybe a generic event subscriber that can be added to the plugin's `services.xml` in a similar fashion as the [`EmotionComponentViewSubscriber`](https://developers.shopware.com/developers-guide/plugin-quick-start/#add-backend-emotion-components).

JoshuaBehrens:
I also dislike inventing a new directory but everything else would be a BC I guess. If not! then I really like falling back to views, or View, or Views or maybe just view. But on the other hand why not bring back the templates folder?
Yikes. We just had the update not having to write id AND class to a service definition and now I have to add an id again just because I need a unique service. And I need a services.xml for that again. I'd like to skip editing the same files.

Need a views folder registered? 
image

engine/Shopware/Components/Plugin/ResourceSubscriber.php
```diff
@@ -53,6 +54,7 @@ public static function getSubscribedEvents()
             'Theme_Compiler_Collect_Plugin_Less' => 'onCollectLess',
             'Theme_Compiler_Collect_Plugin_Css' => 'onCollectCss',
             'Theme_Compiler_Collect_Plugin_Javascript' => 'onCollectJavascript',
+            'Enlight_Controller_Action_PreDispatch' => 'onRegisterTemplate',
```
shyim: And on cli?

JoshuaBehrens: That way I didn't have to get the template manager as dependency. What event do you suggest? I mean I can also implement it Plugin::boot :D but I think it suits very good right there

shyim: Di event for template

JoshuaBehrens: Yikes. I hate these events but it is the best matching here, right

JoshuaBehrens: `Shopware\Components\ShopRegistrationService::registerTheme` overrides template directories set in `Enlight_Bootstrap_AfterInitResource_template` :unamused:

JoshuaBehrens: The best event that is not overwritten by the theme that I found is `Enlight_Controller_Front_DispatchLoopStartup`. That does not cover cli though :confounded:

JoshuaBehrens:
I prefer it as method. Like `getPath` have a `getTemplateDirectoryPath`. Would you still return `getPath() . 'Resources/templates'`?

jkrzefski:
I prefer the idea of a simple boolean flag in the plugin class. That way, a default location of `Resources/views/` is greatly encouraged, but not enforced in a breaking manner. I dislike the idea of suggesting a new template directory like `Resources/templates/`.

JoshuaBehrens:
So default is false? Then I am fine with a property. I gonna change that. A hint on how to register the templates on CLI properly?

engine/Shopware/Components/Plugin/ResourceSubscriber.php
shyim: Try out Theme_Inheritance_Template_Directories_Collected

JoshuaBehrens: Still not triggered on a normal cli call but I think that is ok because if someone renders something they need the theme. So I guess that works out then.

JoshuaBehrens:
Did I even add tests for it? Damn I did it.

engine/Shopware/Components/Plugin.php
```diff
@@ -86,6 +88,11 @@ public static function getSubscribedEvents()
         return $this->isActive;
     }
 
+    public function isAutoloadViews(): bool
```
mitelg, line 91: maybe set this to `final` ?

engine/Shopware/Components/Plugin/ResourceSubscriber.php
```diff
@@ -36,12 +37,18 @@ class ResourceSubscriber implements SubscriberInterface
...
+    /**
+     * @var bool
+     */
+    private $loadViewsDirectory;
+
     /**
      * @param string $pluginPath
      */
-    public function __construct($pluginPath)
+    public function __construct($pluginPath, ?bool $loadViewsDirectory = null)
```
mitelg, line 48: grr! why `null` as default for a boolean value? how about false? :smile:

JoshuaBehrens: That way you tell others to just put `null` for getting the default value without having to know it. You can simply alter the default value later without hassling others that simply wanted to follow the default. If I force someone to set this parameter e.g. because an other following parameter needs to be set, I force them to use the current default value. A pattern I learned somewhere else (I searched for a reference but didn't find one, presumably in a language that does not support named parameters compared to js or c# or typescript that fakes that es6 behaviour earlier exactly like my code up there).

Do I need to change it? I mean I don't care how I solve it :D That is just my way of doing it.

engine/Shopware/Components/Plugin.php
JoshuaBehrens: Any reasons to add hurdles for returning a `bool`? As the constructor is already `final` this seems to be the only way to make the `autoloadViews` conditional (maybe `Bundle::boot` would be the last chance to change it but then I am dependant on execution order that the bundle has been booted before asked whether it likes to have its views loaded).

engine/Shopware/Components/Plugin/ResourceSubscriber.php
mitelg: mh interesting approach. I totally get the point 👍👍 should the SW5 team decide 😁

engine/Shopware/Components/Plugin.php
mitelg: was just an idea 😉

JoshuaBehrens: :blue_heart:

engine/Shopware/Components/Plugin/ResourceSubscriber.php
PaddyS: While I understand the idea of having a null default value, is there any specific reason to have a default value at all?

JoshuaBehrens: Just be not breaking :D I want to have it in 5.6 not 5.7

PaddyS: No need to care about breaking by adding parameters to a constructor. :-)

JoshuaBehrens: NICE ok then I just remove the default value

JoshuaBehrens:
Sorry I totally forgot about this one :( I had in mind that you at least ask for my activity before closing it.

PaddyS:
@JoshuaBehrens Not always, just quite often for active users.
Those were plenty of inactive PRs though. No bad feelings, right? :-)

JoshuaBehrens:
Being on a shopware event as ambassador. not-active user. oof 100

PR #2259, opened 2019-11-15
Filter orders by any payment

JoshuaBehrens:
Orders can be placed in the past using a nowadays inactive payment. But you can't filter these orders by them.

It allows to retrieve all payments from the server side, but the ExtJs store always send the same filter that has been applied by the server at first. Then this filter gets overwritten for this specific store to allow choosing inactive payments as well. The order is kept the same as inactive payments are just added below the active ones. It shouldn't be breaking as the store is working the same like before.

1. A payment fails to work properly
2. Payment gets disabled so no further orders using this payment get created
3. Now all failed orders have to be manually processed
4. Open orders
5. Set payment filter to (now inactive) faulty payment
6. You can't do step 5 without this pull request :anger: 

The store change has to be part of an UPGRADE file I assume as if anyone uses the exact endpoint manually it has a different outcome (for now).

### 5. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Base.php
+++ b/engine/Shopware/Controllers/Backend/Base.php
@@ -136,9 +136,18 @@ class Shopware_Controllers_Backend_Base extends Shopware_Controllers_Backend_Ext
         // Load shop repository
         /** @var \Shopware\Models\Payment\Repository $repository */
         $repository = Shopware()->Models()->getRepository(\Shopware\Models\Payment\Payment::class);
+        $filter = $this->Request()->getParam('filter', []);
+        $hasActiveFilter = in_array('active', array_column($filter, 'property'));
+
+        if (!$hasActiveFilter) {
+            $filter[] = [
+                'property' => 'active',
+                'value' => true
+            ];
+        }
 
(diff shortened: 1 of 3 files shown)


engine/Shopware/Controllers/Backend/Base.php
```diff
@@ -137,7 +137,7 @@ public function getPaymentsAction()
         /** @var \Shopware\Models\Payment\Repository $repository */
         $repository = Shopware()->Models()->getRepository(\Shopware\Models\Payment\Payment::class);
 
-        $query = $repository->getActivePaymentsQuery(
+        $query = $repository->getAllPaymentsQuery(
```
shyim, line 149: BC make maybe create new action in order controller

shyim (changes requested):

engine/Shopware/Controllers/Backend/Base.php
JoshuaBehrens: That is right out dirty to create an action that lists all although the action of concern has a documentation that reads like it does not care about any payment property.

I found a less breaking solution. Please comment on that as well

shyim:

PR #2260, opened 2019-11-15
Introduce Argon2 as new password encoder

pascalheidmann-bedarf:
Security

Add new encoders Argon2i and Argon2id.

While `argon2` is considered more secure than `bcrypt` I hadn't enabled it by default as this would require several more code changes and is BC breaking.

The availability of the new password encoders requires PHP 7.2/7.3 and argon2 enabled.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1655-add-argon2-hashing.php
+++ b/_sql/migrations/1655-add-argon2-hashing.php
@@ -1,0 +1,103 @@
+<?php declare(strict_types=1);
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+use Shopware\Components\Migrations\AbstractMigration;
+
+// see rfc for argon2 in PHP -> https://wiki.php.net/rfc/argon2_password_hash
+if (!defined('PASSWORD_ARGON2_DEFAULT_MEMORY_COST')) {
+    define('PASSWORD_ARGON2_DEFAULT_MEMORY_COST', 1 << 20); // 1MiB
+}
+
+if (!defined('PASSWORD_ARGON2_DEFAULT_TIME_COST')) {
+    define('PASSWORD_ARGON2_DEFAULT_TIME_COST', 2);
+}
+
+if (!defined('PASSWORD_ARGON2_DEFAULT_THREADS')) {
+    define('PASSWORD_ARGON2_DEFAULT_THREADS', 2);
+}
+
+class Migrations_Migration1655 extends AbstractMigration
+{
+    public function up($modus)
+    {
+        $this->addSql('SET @parentFormId = (SELECT id FROM s_core_config_forms WHERE NAME = "Passwörter" AND parent_id=(SELECT id FROM s_core_config_forms WHERE NAME="Core") LIMIT 1)');
+
+        $sql = 'INSERT IGNORE INTO `s_core_config_elements` (`form_id`, `name`, `value`, `label`, `description`, `type`, `required`, `position`, `scope`, `options`) VALUES ';
+        $sql .= sprintf(
+            "(@parentFormId, '%s', '%s', '%s', '%s', 'number', 1, 0, 0, '%s'),",
+            'argon2MemoryCost',
+            serialize(PASSWORD_ARGON2_DEFAULT_MEMORY_COST),
+            'Argon2-Speicher',
+            'Ein höherer Speicherverbrauch macht es einem möglichen Angreifer schwerer, ein passendes Klartext-Passwort zu erzeugen.',
+            serialize(['minValue' => (string) (1 << 20), 'maxValue' => (string) (1 << 62)])
+        );
+        $sql .= sprintf(
+            "(@parentFormId, '%s', '%s', '%s', '%s', 'number', 1, 0, 0, '%s'),",
+            'argon2TimeCost',
+            serialize(PASSWORD_ARGON2_DEFAULT_TIME_COST),
+            'Argon2-Zeit',
+            'Ein höherer Zeitaufwand macht es einem möglichen Angreifer schwerer, ein passendes Klartext-Passwort zu erzeugen.',
+            serialize(['minValue' => '1', 'maxValue' => '30'])
+        );
+        $sql .= sprintf(
+            "(@parentFormId, '%s', '%s', '%s', '%s', 'number', 1, 0, 0, '%s');",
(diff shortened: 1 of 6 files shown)


_sql/migrations/1655-add-argon2-hashing.php
```diff
@@ -0,0 +1,78 @@
...
+        $parentFormId = $this->getConnection()->query($parentSql)->fetchColumn();
+
+        $sql = 'INSERT IGNORE INTO `s_core_config_elements` (`form_id`, `name`, `value`, `label`, `description`, `type`, `required`, `position`, `scope`, `options`) VALUES';
+        $sql .= sprintf(
+            "(%d, '%s', '%s', '%s', '%s', 'number', 1, 0, 0, '%s'),",
+            $parentFormId,
+            'argon2MemoryCost',
+            serialize(PASSWORD_ARGON2_DEFAULT_MEMORY_COST),
+            'Argon2-Speicher',
+            'Ein höherer Speicherverbrauch macht es einem möglichen Angreifer schwerer, ein passendes Klartext-Passwort zu erzeugen.',
```
shyim, line 52: Translation?

shyim (changes requested):

engine/Shopware/Components/Password/Encoder/Argon2i.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return 'Argon2i';
+    }
+
+    /**
+     * @return bool
+     */
+    public function isCompatible()
+    {
+        return PHP_VERSION_ID >= 70200 && defined('PASSWORD_ARGON2I');
```
hlohaus: maybe only defined('PASSWORD_ARGON2I')

_sql/migrations/1655-add-argon2-hashing.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->addSql('SET @parentFormId = (SELECT id FROM s_core_config_forms WHERE NAME = "Passwörter" AND parent_id=(SELECT id FROM s_core_config_forms WHERE NAME="Core") LIMIT 1)');
+
+        $sql = 'INSERT IGNORE INTO `s_core_config_elements` (`form_id`, `name`, `value`, `label`, `description`, `type`, `required`, `position`, `scope`, `options`) VALUES ';
+        $sql .= sprintf(
+            "(%d, '%s', '%s', '%s', '%s', 'number', 1, 0, 0, '%s'),",
+            '@parentFormId',
+            'argon2MemoryCost',
+            serialize(PASSWORD_ARGON2_DEFAULT_MEMORY_COST),
+            'Argon2 memory',
+            'Higher memory usage increases the security against attackers.',
```
shopwareBot: Use german translations here.

_sql/migrations/1655-add-argon2-hashing.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->addSql('SET @elemargon2TimeCost = (SELECT id FROM s_core_config_elements WHERE name="argon2TimeCost" AND form_id = @parentFormId LIMIT 1');
+        $this->addSql('SET @elemargon2Threads = (SELECT id FROM s_core_config_elements WHERE name="argon2Threads" AND form_id = @parentFormId  LIMIT 1');
+
+        $sql = 'INSERT INTO `s_core_config_element_translations` (`element_id`, `locale_id`, `label`, `description`) VALUES ';
+        // memory
+        $sql .= sprintf(
+            '(%d, %d, "%s", "%s")',
+            '@elemArgon2MemoryCost',
+            1,
+            'Argon2 Speicher',
```
shopwareBot: Insert only english translations. Any specific reasons, why you inserted both?

engine/Shopware/Plugins/Default/Core/PasswordEncoder/Bootstrap.php
```diff
@@ -57,6 +74,20 @@ public function install()
         return true;
     }
 
+    /**
+     * @return array
+     */
+    public function getArgon2Options()
```
shopwareBot: Return types

engine/Shopware/Plugins/Default/Core/PasswordEncoder/Bootstrap.php
```diff
@@ -85,7 +116,9 @@ public function getSha256Options()
     }
 
     /**
-     * @return \Shopware\Components\Password\Manager
+     * @throws Enlight_Event_Exception
+     *
+     * @return Manager
```
shopwareBot: Return types

_sql/migrations/1655-add-argon2-hashing.php
pascalheidmann-bedarf: This sounds not very future proof/i18n ready as AFAIK those values will be used if no translation is present. Therefore users that eg. have a Spanish Shopware installation will be presented with the german translation - which they potentially have more troubles with than with english...

pascalheidmann-bedarf:
@shopwareBot why 5.7? This isn't breaking anything and 5.7 won't be out soon I guess?
I originally had this for `\Shopware\Components\Password\Manager::getDefaultEncoderName()`:
```php
foreach ($this->encoder as $encoder) {
                if (method_exists($encoder, 'isCompatible') && $encoder->isCompatible()) {
                    return $encoder->getName();
                }
            }
            $encoderName = 'sha256';
```
which would definitely break things but otherwise this is just an enhancement...

_sql/migrations/1655-add-argon2-hashing.php
shopwareBot: While you're right, this is not the subject of your PR.

engine/Shopware/Plugins/Default/Core/PasswordEncoder/Bootstrap.php
pascalheidmann-bedarf: This would break the signature of an already known function

tests/Unit/Components/Password/Encoder/Argon2iTest.php
```diff
@@ -0,0 +1,131 @@
...
+{
+    /**
+     * @var Argon2i
+     */
+    private $hasher;
+
+    /**
+     * Sets up the fixture, for example, opens a network connection.
+     * This method is called before a test is executed.
+     */
```
soebbing, line 40: I think you can drop this comment and the ones below.

engine/Shopware/Plugins/Default/Core/PasswordEncoder/Bootstrap.php
PaddyS: Since we're in the scope of a plugin here, this method couldn't be overridden anyway and thus doesn't break - does it?

mitelg: @PaddyS I think you are wrong here :blush: as this is a plugin based on the legacy plugin system, the bootstrap methods should be hookable. and because of that, there are extendable by plugins (theoretically :smile: )

but adding a return type should not break a hook. but I'm not sure about that

shyim: Plugins are not hookable

pascalheidmann-bedarf: I added typehints for the plugin's bootstrap file for all functions that aren't inherited. Hope that satisfy your kinks ;)

pascalheidmann-bedarf:

shyim:
Hey,
I tested your PR and my shop is broken when I enable Aragon2i. The registration page just reloads.
Also, all configuration options are missing.

Did you tested your changes?

pascalheidmann:
@shyim tested kind of 🙈 Not the migration itself again after rewrote which then had problems with `sprintf` and `%d` specifier. Should be now fixed.

pascalheidmann:
@shopwareBot I tested it on my local installation where the config options now have been applied to the correct form. Also the registration worked as expected. Therefore I consider it tested now, with both manual as well as automated tests ;)

shyim:

PR #2263, opened 2019-11-18
Convert state types to lowercase before accessing them

jkrzefski:
This adds compatibility for states that (for whatever reason) have uppercase letters in their type field.

Make `$type` and all keys of `$this->availableTypes` lowercase before `$this->availableTypes[$type]` is accessed.

1. Have `SwagImportExport` and `BestitAmazonPay` installed.
2. Export with the profile `default_order_main_data`.
3. See the error `Invalid type 'Payment' given.`.

None.

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/StateTranslatorService.php
+++ b/engine/Shopware/Components/StateTranslatorService.php
@@ -47,7 +47,7 @@ class StateTranslatorService implements StateTranslatorServiceInterface
         ]
     ) {
         $this->snippetManager = $snippetManager;
-        $this->availableTypes = $types;
+        $this->availableTypes = array_change_key_case($types, CASE_LOWER);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/StateTranslatorService.php
```diff
@@ -57,6 +57,9 @@ public function __construct(
      */
     public function translateState($type, array $state)
     {
+        $type = strtolower($type);
+        $this->availableTypes = array_change_key_case($this->availableTypes);
```
shopwareBot: Move this line into the constructor please.

jkrzefski: Done.

engine/Shopware/Components/StateTranslatorService.php
```diff
@@ -47,7 +47,7 @@ public function __construct(
         ]
     ) {
         $this->snippetManager = $snippetManager;
-        $this->availableTypes = $types;
+        $this->availableTypes = array_change_key_case($types);
```
pascalheidmann-bedarf: That function is so descriptive... 🙈 Can you add the second default parameter `CASE_LOWER` just to make the case clear?

jkrzefski: Done.

shyim:

PR #2266, opened 2019-11-20
SW-24901 - Assign variantID to articleDetailId in mainData attribute

fmaasde:
The articleDetailId from the main variant is also read out for the attributes in the mainData and therefore the duplicate entry occurs.

I assigned the variantID to the articleDetailId in the mainData attribute.

Create a new variant product in the backend and enter for example a text in attr1. Now use the apply standard data function and make sure you include the apply attribute configuration option.

https://issues.shopware.com/issues/SW-24901

Just a bugfix

### 6. Checklist


shyim:
Already fixed with https://github.com/shopware/shopware/commit/f158315ddb9ab6be6cd49e07fceb4461eb4de8f0

PR #2268, opened 2019-11-25
Optimize setting of files in Enlight request object to prevent symfony exception

ghost:
Since commit 77399107b08d6636c5eaa4b94bb1bc4b1fb57726 if a form is submitted without any files selected, symfony throws the following exception:
`InvalidArgumentException: An uploaded file must be an array or an instance of UploadedFile.`

So my suggestion is to remove any empty file array in the request object before setting it to the Enlight-Request object.

If you have any questions, please feel free to ask. I'd be happy if this could be incorporated into the code before the next release.

Regards,
Alexander Pankow

### DIFF
--- a/engine/Library/Enlight/Controller/Request/RequestHttp.php
+++ b/engine/Library/Enlight/Controller/Request/RequestHttp.php
@@ -902,4 +902,20 @@ class Enlight_Controller_Request_RequestHttp extends Request implements Enlight_
     {
         $this->content = $content;
     }
+
+    /**
+     * Sets files. Removes empty values from files array.
+     *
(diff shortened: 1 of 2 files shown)


bpanet:
@shyim Just for your information. Maybe you have an better idea?

I run into this problem while testing our "File upload for forms" plugin. Our plugin adds the possibility to add file upload fields to the cms forms.
So because the exception is thrown in the core I have a hardly way to prevent this exception in the plugin.

LiaraAlis:
Just for information: We transfered our account into a organization. So I will continue with my private account. ;)

shyim:

PR #2270, opened 2019-11-28
add option to specifiy if just SEO-URLs should be printed

tinect:
Grown historically and product related we have some products containing the same name. The SEO-URLs just contains the name, so not every product has its SEO-URLs. Currently, non-SEO-URLs are filtered out of the hreflang. Sure, these links aren't looking nice, but these links are printed into sitemap and so on. So this restriction should be adjustable.

Adds option to specify, if just SEO-URLs should be printed into href-lang.

### 6. Checklist


shyim:

PR #2271, opened 2019-11-28
Code cleanup + micro optimisations for Enlight library

pascalheidmann:
While Enlight isn't everyone darling anymore (for good reasons) it is still a big chunk of (legacy) Shopware. Improving both the quality and performance of several functions can therefore have a great impact.

Micro optimisations and cleaning up code (most of the time one will help the other)

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Config/Adapter/DbTable.php
+++ b/engine/Library/Enlight/Config/Adapter/DbTable.php
@@ -102,7 +102,7 @@ class Enlight_Config_Adapter_DbTable extends Enlight_Config_Adapter
     /**
      * Sets the options of an array.
      *
-     *
+     * @param array $options
      * @return Enlight_Config_Adapter
      */
     public function setOptions(array $options)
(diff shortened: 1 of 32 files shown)


engine/Library/Enlight/Config/Adapter/DbTable.php
```diff
@@ -384,4 +358,28 @@ protected function readSection($name, $section)
 
         return $data;
     }
+
+    /**
+     * @param array $section
```
shyim: phpdoc is not necessary when you can typehint anything

engine/Library/Enlight/Controller/Request/RequestHttp.php
```diff
@@ -677,7 +671,7 @@ public function getBaseUrl($raw = false)
             $this->baseUrl = $this->prepareBaseUrl();
         }
 
-        return ($raw == false) ? urldecode($this->baseUrl) : $this->baseUrl;
+        return $raw ? $this->baseUrl : urldecode($this->baseUrl);
```
shyim, line 674: Thats not the same

pascalheidmann: As it had been a non strict comparison (and I can't change it without breaking) it is still basically the same, just with inverted logic.

PR #2273, opened 2019-11-28
Use more strict comparisons if possible in Shopware Components

pascalheidmann:
Code quality in `/engine/Shopware/Components`

Use strict comparisons if possible, and replace magic numbers with more vocal consts.

https://github.com/shopware/shopware/pull/2271

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/PropertyGroup.php
+++ b/engine/Shopware/Components/Api/Resource/PropertyGroup.php
@@ -25,6 +25,7 @@
 namespace Shopware\Components\Api\Resource;
 
 use Exception;
+use Shopware\Bundle\StoreFrontBundle\Struct\Property\Set;
 use Shopware\Components\Api\Exception as ApiException;
 use Shopware\Components\Api\Exception\NotFoundException;
 use Shopware\Components\Api\Exception\ParameterMissingException;
@@ -48,10 +49,10 @@ class PropertyGroup extends Resource
     /**
      * @param int $id
      *
-     * @return array|Group
      *@throws NotFoundException
-     *
      * @throws ParameterMissingException
+     *
+     * @return array|Group
      */
     public function getOne($id)
     {
@@ -216,18 +217,16 @@ class PropertyGroup extends Resource
 
             if (!isset($params['sortmode']) || empty($params['sortmode'])) {
                 // Set sortmode
-                $params['sortmode'] = 0;
+                $params['sortmode'] = Set::SORT_ALPHANUMERIC;
             }
 
             // Sortmode equals the old article_count sorting?
-            if ((int) $params['sortmode'] === 2) {
+            if ($params['sortmode'] === Set::SORT_LEGACY) {
                 // Fallback to the default sorting
-                $params['sortmode'] = 0;
+                $params['sortmode'] = Set::SORT_ALPHANUMERIC;
             }
-        } else {
-            if (isset($params['name']) && empty($params['name'])) {
-                throw new ApiException\CustomValidationException('Name must not be empty');
-            }
+        } elseif (isset($params['name']) && empty($params['name'])) {
+            throw new ApiException\CustomValidationException('Name must not be empty');
         }
 
         return $params;
--- a/engine/Shopware/Components/Api/Resource/PropertyGroup.php
+++ b/engine/Shopware/Components/Api/Resource/PropertyGroup.php
@@ -24,7 +24,13 @@
 
 namespace Shopware\Components\Api\Resource;
 
+use Exception;
 use Shopware\Components\Api\Exception as ApiException;
+use Shopware\Components\Api\Exception\NotFoundException;
+use Shopware\Components\Api\Exception\ParameterMissingException;
+use Shopware\Components\Api\Exception\ValidationException;
+use Shopware\Models\Property\Group;
+use Shopware\Models\Property\Repository;
 
 /**
  * Property API Resource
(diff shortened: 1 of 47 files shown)


engine/Shopware/Components/Api/Resource/PropertyGroup.php
```diff
@@ -214,7 +214,7 @@ private function preparePropertyData(array $params, $propertyGroup = null)
             }
 
             // Sortmode equals the old article_count sorting?
-            if ($params['sortmode'] == 2) {
+            if ($params['sortmode'] === 2) {
```
shyim: cast maybe?

engine/Shopware/Components/DependencyInjection/LegacyPhpDumper.php
```diff
@@ -68,22 +68,39 @@ class LegacyPhpDumper extends Dumper
     const NON_FIRST_CHARS = 'abcdefghijklmnopqrstuvwxyz0123456789_';
 
     private $definitionVariables;
+
```
shyim: Don't touch this file. We need the ability to diff it with the class from symfony

pascalheidmann: I guess PHPstan was quite successful in reformating 😅

engine/Shopware/Components/Api/Resource/PropertyGroup.php
pascalheidmann: Good point.
But considering it, it looks like code coverage is missing here

pascalheidmann: After looking into it: sortmode had always to be `integer` according to https://developers.shopware.com/developers-guide/rest-api/api-resource-property-group/

mitelg:
are all changes saved by tests? we do not have strict type mode, so this would be valid code:
```
$nullableValue = false;
if ($nullableValue == null) { // <- true!
}
```
we have to ensure, that we don't change the current behaviour or create errors through restricting the types

pascalheidmann:
@mitelg some had crashed initially and had to be casted.
But most are comparisons between strings anyway where something other than a string already won't match.
The few occasions I found (mysql queries are horrible as they are always strings!) I casted for now.

engine/Shopware/Components/Plugin/ConfigWriter.php
```diff
@@ -88,7 +88,7 @@ public function saveConfigElement(Plugin $plugin, $name, $value, Shop $shop)
             throw new \Exception(sprintf('Config element "%s" not found.', $name));
         }
 
-        if ($element->getScope() == 0 && $shop->getId() !== 1) {
+        if ($element->getScope() === 0 && $shop->getId() !== 1) {
```
mitelg: don't we have a constant for the scopes?

engine/Shopware/Components/MultiEdit/Resource/Product/DqlHelper.php
```diff
@@ -780,7 +780,7 @@ public function formatValue($prefix, $field, $value)
         $info = $this->getColumnsForProductListing();
         $info = $info[ucfirst($prefix) . ucfirst($field['field'])];
 
-        if ($info['nullable'] && $value == '') {
+        if ($info['nullable'] && $value === '') {
```
mitelg, line 783: did you checked this? if $value was false this  matched

engine/Shopware/Components/Document.php
```diff
@@ -748,22 +756,22 @@ protected function saveDocument()
...
-            if ($typID != 4) {
+            if ($typID !== Document::TYPE_CANCEL) {
                 if (!empty($this->_document['numbers'])) {
-                    $numberrange = $this->_document['numbers'];
+                    $numberRange = $this->_document['numbers'];
                 } else {
                     // The typID is indexed with base 0, so we need increase the typID
-                    if (!in_array($typID, ['1', '2', '3'])) {
-                        $typID = $typID + 1;
+                    if (!in_array($typID, [1, 2, 3], true)) {
```
mitelg: why don't you use constants here?

engine/Shopware/Components/Document.php
```diff
@@ -677,7 +685,7 @@ protected function saveDocument()
             WHERE `type` = ? AND userID = ? AND orderID = ? LIMIT 1
             ';
             $amount = ($this->_order->order->taxfree ? true : $this->_config['netto']) ? round($this->_order->amountNetto, 2) : round($this->_order->amount, 2);
-            if ($typID == 4) {
+            if ($typID === 4) {
```
mitelg: constant

engine/Shopware/Components/Api/Resource/PropertyGroup.php
```diff
@@ -214,7 +220,7 @@ private function preparePropertyData(array $params, $propertyGroup = null)
             }
 
             // Sortmode equals the old article_count sorting?
-            if ($params['sortmode'] == 2) {
+            if ((int) $params['sortmode'] === 2) {
```
mitelg: are there constants for the sort mode?

engine/Shopware/Components/Plugin/ConfigWriter.php
pascalheidmann: ~No, sadly not. But I might introduce it... ;)~
I found them in _Element_ model and will use them

engine/Shopware/Components/Api/Resource/PropertyGroup.php
pascalheidmann: No, but actually I haven't found anything related to sortmode.
What are those actually values meant for?

engine/Shopware/Components/MultiEdit/Resource/Product/DqlHelper.php
pascalheidmann: The function's doc block explicitly says
_Will set value = null for empty *strings* and replace comma with period for decimals_.
Therefore following its specs that should be ok I think.

Also the only usage of that function is in `Shopware\Components\MultiEdit\Resource\Product::save()` which itself gets its parameters via Request params which are always strings.

engine/Shopware/Components/Api/Resource/PropertyGroup.php
mitelg: maybe @shyim has an idea

pascalheidmann: I found 3 consts in `Shopware\Bundle\StoreFrontBundle\Struct\Property\Set`, but those are values `0`, `1` and `3`.

PR #2274, opened 2019-11-28
Allow only unique attributes for each entity

pascalheidmann:
It is possible to create several attributes with the same name for the same entity. This results in strange problems like attributes not appearing etc. as the newest attribute entry will win.

Enforce the combination of entity and attribute to be unique via Doctrine Entity Annotation (+ migration)

1. Run
```
INSERT INTO `s_attribute_configuration` (`table_name`, `column_name`, `column_type`, `display_in_backend`)
VALUES ('s_articles_attributes', 'attr5', 'text', TRUE), ('s_articles_attributes', 'attr5', 'boolean', FALSE);
```
on a clean installation.
2. Regenerate models
3. Try to modify new attr5 on a product. You won't find its entry.

### 6. Checklist

### DIFF
--- a/_sql/migrations/1651-enforce-unique-attributes.php
+++ b/_sql/migrations/1651-enforce-unique-attributes.php
@@ -1,0 +1,35 @@
+<?php declare(strict_types=1);
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


shyim:
Will this case caught by the ui also?

pascalheidmann:
@shyim right now the attribute configuration UI will only show the newest entry if there are more than one.

shyim:
I mean what happens when you try to create two times the same attribute in backend. It should happen with your chance a sql exception. This should be catched also by the UI (ExtJs) with an message

pascalheidmann:
@shyim 
```
Duplicate column name
The name already exists in table s_core_paymentmeans_attributes. Please choose an unique column name.
```
So there is already a check in place but as some plugin devs doing nasty stuff (Viison in our case...) and not always using the crud service. This is just another layer of protection against stupidity 😄

PaddyS:
Hey @pascalheidmann,

Could you please add tests for this case?

pascalheidmann:
@PaddyS there isn't much really to test. Nonetheless I added two cases:
- regular usage via `CrudService` which will update as expected
- directly adding attribute column entries will result in an (expected) exception

shyim:

PR #2276, opened 2019-11-29
Use transactions for plugin uninstall, activate and deactivate

pascalheidmann:
No more half uninstalled plugins or broken (de-)activations

Wraps plugins handling inside a transactions like it already happens with plugin installations

Create a plugin that does something to your database in its uninstall routine, and then intentionally crashes.
Your database is now corrupted.

As @shyim mentioned this is potentially breaking some plugins that are using transactions. Therefore this has to be mentioned on the plugin page that now all actions are inside an transaction already.

### 6. Checklist


shyim:
It's an BC. Change target to 5.7.
When plugin already uses transaction in those methods it will get then an error, cause already a transaction is running

shyim:

PR #2278, opened 2019-12-05
Add cms attribute translations

sbrueggenolte:

Because attribute translations are not being read in the frontend for custom pages (CMS).

It reads the attribute translations and merges them with the base data translations before they are being prepared for the frontend view.

* Add a new translatable attribute field for custom pages (display in backend = true)
* Create/edit a custom page and set a translation value for the custom page attribute
* Open the page in the frontend using the proper shop for the entered translation

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sCms.php
+++ b/engine/Shopware/Core/sCms.php
@@ -111,7 +111,15 @@ class sCms implements \Enlight_Hook
                 $fallbackId = Shopware()->Shop()->getFallback()->getId();
             }
 
-            $translations = $this->translationComponent->readWithFallback($shopId, $fallbackId, 'page', $staticId);
+            $translations = \array_merge(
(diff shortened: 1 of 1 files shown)


shyim:
Hey @sbrueggenolte,
I tried to reproduce your Issue on the 5.6 Branch and its working already without your Patch.

The translations are saved in the `page` translation, which is currently already implemented.
image

I guess your issue is caused by a wrong writing to the translations.

sbrueggenolte:
Hey @shyim,
the translation is being written to the translation table, that's correct.
But when it comes to displaying the attribute field in the frontend, the translated value is not being loaded properly.
Please follow my reproduction description and add a smarty debug on the custom/index.tpl. 
We have just been able to reproduce this error in the latest Shopware release (5.6.4).

shyim:
Where is it missing? On the static page controller?

PR #2279, opened 2019-12-06
Fixed fetching of documents positions metadata

jonasklause:
So far the fetching of the documents position metadata was done by articleordernumber, but it has to be articleID. Because later on the given value will be used to retrieves the ordernumber by searching with a given articleID.

see above

The metadata for document positions has been wrong, if there was an article with an detailID which matches the given articleordernumber. Which is done in sArticles::sGetPromotionById() and sArticles::getPromotionNumberByMode().

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -495,7 +495,7 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
         $articleModule = Shopware()->Modules()->Articles();
         foreach ($positions as &$position) {
             if ($position['modus'] == 0) {
-                $position['meta'] = $articleModule->sGetPromotionById('fix', 0, $position['articleordernumber']);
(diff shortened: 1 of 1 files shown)


shyim:
That's not the same. With your change we fetch main variant instead specific one

aragon999:
Maybe the complete thing should be refactored using the `StoreFrontBundle`. Otherwise one is probably not able to fix this problem. Maybe replacing the function with the following, would be sufficient: https://github.com/shopware/shopware/blob/8d2e95a71c217581d179949a1e1e60fea91a9b66/engine/Shopware/Core/sArticles.php#L1252

Is the limitation, that one should not use pure numeric product numbers, still valid? Or was this removed since by now the regex for the product numbers is adjustable?

hlohaus:
Yes, can use this code from me in basket:
https://github.com/shopware/shopware/blob/5.6/engine/Shopware/Core/sBasket.php#L2502
It's also faster 💃

PR #2280, opened 2019-12-06
SW-24987 fix empty array key in request cookies

htuscher:
Shops break if you have a cookie that contains `.*s_user`.

It prevents the `InputFilter` from generating empty array keys in the request cookies.

Set a cookie on your domain e.g. `ajs_user=something`. The shop will break with 
```
PHP message: PHP Fatal error:  Uncaught InvalidArgumentException: The cookie name cannot be empty. in /var/www/clients/client1/web14/web/vendor/symfony/http-foundation/Cookie.php:110
```

OR 

```
PHP message: PHP Fatal error:  Uncaught Zend\Diactoros\Exception\InvalidArgumentException: Invalid header value type; must be a string or numeric; received NULL in /var/www/clients/client1/web14/web/vendor/zendframework/zend-diactoros/src/HeaderSecurity.php:139
```

https://issues.shopware.com/issues/SW-24987

Nothing

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/InputFilter/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Frontend/InputFilter/Bootstrap.php
@@ -150,11 +150,15 @@ class Shopware_Plugins_Frontend_InputFilter_Bootstrap extends Shopware_Component
             foreach ($val as $k => $v) {
                 unset($process[$key][$k]);
                 $stripTags = in_array($k, $whiteList) ? false : $stripTagsConf;
+                $filteredKey = self::filterValue($k, $regex, $stripTags);
+                if ($filteredKey === '' || $filteredKey === null) {
+                  continue;   
+                }
                 if (is_array($v)) {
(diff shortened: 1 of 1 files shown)


engine/Shopware/Plugins/Default/Frontend/InputFilter/Bootstrap.php
```diff
@@ -150,11 +150,15 @@ public function onRouteShutdown(Enlight_Controller_EventArgs $args)
             foreach ($val as $k => $v) {
                 unset($process[$key][$k]);
                 $stripTags = in_array($k, $whiteList) ? false : $stripTagsConf;
+                $filteredKey = self::filterValue($k, $regex, $stripTags);
+                if ($filteredKey === '' || $filteredKey === null) {
+                  continue;
```
hlohaus: maybe a unset would be better. so it would no be filtered anymore.

soebbing:
Thank you very much for your contribution, @hhoechtl! 💙

PR #2281, opened 2019-12-06
Support composer installer

mschakulat:
Allow installing shopware as dependency and to overwrite the install path and name

Sometimes you need to install shopware as dependency and integrate it in an existing project

Enables support for composer installers, so that you can change the install path/name

### 6. Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,7 +81,8 @@
         "superbalist/flysystem-google-storage": "6.0.0",
         "stecman/symfony-console-completion": "0.10.1",
         "beberlei/doctrineextensions": "1.2.1",
-        "phpstan/phpstan-shim": "0.11.19"
+        "phpstan/phpstan-shim": "0.11.19",
+        "oomphinc/composer-installers-extender": "~1.1.2"
     },
     "suggest": {
         "ext-apcu": "*",
@@ -149,5 +150,8 @@
         "post-update-cmd": "./build/composer-post-update-cmd.sh",
         "test": "phpunit -c tests/ --colors=always",
         "test-unit": "phpunit -c tests/phpunit_unit.xml.dist --colors=always"
+    },
+    "extra": {
+        "installer-name": "shopware"
     }
 }


shyim:
Why do we need that?
Is https://github.com/shopware/composer-project not something you want?

mschakulat:
> Why do we need that?
> Is https://github.com/shopware/composer-project not something you want?

No. The composer project forces a specific project structure. We want to be able to specify the structure by ourselves. For example:

- app
- bin
- shopware
- ...

The existing composer project forces a specific application structure.

Btw: It would be really helpfull for projects, which were build before shopware 5.4, since there was no composer project.

soebbing:
But the structure the project enforces is the structure Shopware needs to work properly. What is the benefit to specify another structure if Shopware won't work?

mschakulat:
What would be the downside of being able to do that? The advantage would be that you CAN use it. No one MUST use it.

teiling88:
> What would be the downside of being able to do that?

One more dependency to care about?

> We want to be able to specify the structure by ourselves.

The composer-project ist boilerplate code, you can change everything to your needs :-)

mschakulat:
> The composer-project ist boilerplate code, you can change everything to your needs :-)

Again: it is about EXISTING projects. I know we could use the composer project, if we start a new one. But how does that help me to integrate shopware as a dependency in existing projects?

soebbing:
The downside I see is setting a wrong expectation. If I am able to just install Shopware using your change, but Shopware isn't able to function as expected, that might lead developers into wondering what they are doing wrong.

What I do not understand currently: What existing projects might need a copy of Shopware code that isn't able to work due to where it is installed?

mschakulat:
I wonder why you'd think it wouldn't work. We have exactly such a setup and it runs perfectly. And basically it is nothing else than the composer project, only with the difference that this one can also be used for existing projects.

mitelg:
@mschakulat did you try to install plugins, themes etc?

soebbing:
Well, you pretty sure needed some changes to e.g. set the Version of Shopware. Are plugins just installed inside the of the `shopware` dependency? What about logs and caches? You might have to create some symlinks to make that work.

I'm not saying that it can't be fixed, just that it won't work as one might expect it to out of the box.

PR #2282, opened 2019-12-06
Remove PHPUnit 7/8 switch

pascalheidmann:
With `@before` and `@after` annotation we are able to avoid the problems with the changed function signature PHPUnit introduced with the switch to PHPUnit 8.

Replace build-in `setUp` and `tearDown` methods in Enlight's `TestCase` classes with equivalent functions with `@before` and `@after` annotation.
This worked out of the box with ~2579 of 2580 testcases but `Shopware\Tests\Functional\Plugins\Core\HttpCache\BootstrapTest::testCacheableRoute` which had some weird plugin reloading bug. By fixing this I also introduced the `DatabaseTransactionBehaviour` trait so there won't be any changes to the database which can cause side effects and require otherwise manual cleanup.

### 6. Checklist


shyim:
As already in Gitter mentioned we can't do that.
This would be a breaking change. That bridge is already removed in 5.7 branch.

pascalheidmann:
@shyim what exactly would be breaking?

shyim:
Ahh nvm. 
Did you test it with both PHPUnit versions?

pascalheidmann:
@shyim both phpunit 7.5 (?, prebundled) and 8.5 run through but there had been some warnings with version 8 of deprecated assertions. But that is another story I guess (or are they already fixed in 5.7?)

shyim:
Deprecations already gone at 5.7 branch.

shyim:

PR #2283, opened 2019-12-06
Add sw:admin:login for instant backend login

jkrzefski:
It is not necessary. This is merely for convenience.

A new command is introduced to the CLI: `sw:admin:login`
It generates a one-time-password for a backend user that is valid for 30 seconds. The username and the generated password are composed into a url to the backend login controller.

1. Run `sw:admin:login` in CLI.
2. Open the link.
3. You are now logged into the backend 🚀 

None.

I don't know.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Commands/AdminLoginCommand.php
+++ b/engine/Shopware/Commands/AdminLoginCommand.php
@@ -1,0 +1,150 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Commands;
+
+use Shopware\Components\Model\ModelManager;
+use Shopware\Components\Password\Manager;
+use Shopware\Components\Random;
+use Shopware\Components\Routing\Context;
+use Shopware\Components\Routing\RouterInterface;
+use Shopware\Models\Shop\Shop;
+use Shopware\Models\User\Role;
+use Shopware\Models\User\User;
+use Shopware_Components_Config;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Throwable;
+
+class AdminLoginCommand extends ShopwareCommand
+{
+    /**
+     * @var Manager
+     */
+    private $passwordRegistry;
+
+    /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    /**
+     * @var ModelManager
+     */
+    private $modelManager;
+
+    /**
+     * @var Shopware_Components_Config
+     */
+    private $config;
+
+    public function __construct(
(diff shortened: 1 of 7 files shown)


engine/Shopware/Commands/AdminLoginCommand.php
```diff
@@ -0,0 +1,124 @@
...
+
+        $user->setLockedUntil(null);
+        $user->setFailedLogins(0);
+        $user->setOtp($encoder->encodePassword($otp));
+        $user->setOtpActivation(date_create());
+
+        $this->modelManager->flush();
+
+        $shop = $this->modelManager->getRepository(Shop::class)->getActiveDefault();
+        $this->shopRegistrationService->registerResources($shop);
```
shyim: why do we need that?

jkrzefski: I generate a url to the backend. If I do not do this, the host of the url would be `localhost`.

hlohaus: you can do $router->getContext()->createFromShop
https://github.com/shopware/shopware/blob/5.6/engine/Shopware/Components/Routing/Context.php#L384

hlohaus (changes requested):

engine/Shopware/Commands/AdminLoginCommand.php
jkrzefski: Done.

engine/Shopware/Commands/AdminLoginCommand.php
```diff
@@ -0,0 +1,128 @@
...
+
+        if (!$user instanceof User) {
+            $output->writeln('No active user could be found.');
+
+            return 1;
+        }
+
+        try {
+            // generate a cryptographically secure one-time-password
+            $otp = bin2hex(random_bytes(32));
```
shyim: Maybe use https://github.com/shopware/shopware/blob/5.6/engine/Shopware/Components/Random.php#L115

jkrzefski: Done.

engine/Shopware/Commands/AdminLoginCommand.php
```diff
@@ -0,0 +1,123 @@
...
+
+        $this->passwordRegistry = $passwordRegistry;
+        $this->router = $router;
+        $this->modelManager = $modelManager;
+        $this->config = $config;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $user = $this->modelManager->getRepository(User::class)->findOneBy([
```
uehler: this will return any active backend user but won't there be 1-n users?

uehler: so anyone who calls that command will be logged in as the same user. Wouldn't it be better to require to provide a username?

jkrzefski: Yes, this is true. My intention was to login with the first active backend user that is found. If you would rather have the username as an argument, I can do that too. In that case, I'd keep the current line as fallback, in case no username is given.

jkrzefski: Or should I pick the first **admin** user, that I find?

jkrzefski: @uehler I added an optional argument for the username now. If no username is provided, the command will find the first admin user in the database.

engine/Shopware/Controllers/Backend/Login.php
```diff
@@ -105,6 +105,19 @@ public function loginAction()
...
+        if ($this->Request()->isGet()) {
+            // GET-method indicates, that this is not the default xhr from the login form
+            // instead this request is likely to be an otp-login, so we should redirect to the backend
+            $this->redirect([
+                'module' => 'backend',
+                'controller' => 'index',
+                'action' => 'index',
+                // this is necessary, because with 302 chrome will not save the session cookie
+                // @see https://bugs.chromium.org/p/chromium/issues/detail?id=696204
+                'code' => 307,
```
jkrzefski, line 118: I made something of a mistake here and it is driving me crazy. I accidentally put `code` into the url-parameters instead of the redirect-options (first argument instead of second) and the login works. This is a normal 302-redirect, so my explanation in the comment above is not correct. Having the code in the parameters leads to a url like this: `backend/index/index/code/307` which is obviously not intended. But if I remove this line, the login will not work anymore. It appears that the session-cookie is lost in redirection then. Can someone take a look at this and tell me if they can reproduce this issue? This issue is not even browser-specific anymore. It simply won't work without this line, but I don't know why.

PR #2284, opened 2019-12-06
Autoload Resources/templates directory

JoshuaBehrens:
image
[image](https://github.com/shopware/shopware/pull/2258)
### 0. Why do you post this again?
The previous [pull request 2258](https://github.com/shopware/shopware/pull/2258) just expired.

Once upon a time it was common to put js and less files under `Resources/views/frontend/_public/src` and register the files manually via subscriber. A wise man automatically loads these source files now from `Resources/frontend`. But why are the template files in `Resources/views` not automatically loaded?
To reduce the chance of a breaking change I chose the directory `Resources/templates` as these are template files that needs to be loaded. In case of specific reasons to not load them automatically one can still load them from elsewhere conditionally.

Subscribe `Enlight_Controller_Action_PreDispatch` and tell the controller template manager to use the plugins path.

1. Put `all.less` in `Resources/frontend/less`
2. Style is online
3. Put `foobar.js` into `Resources/frontend/js`
4. Dynamic DOM is online
5. Put `tpl` files into `Resources/views`
6. Nothing happens
7. Adds subscriber class as one does not want to pollute the bootstrap file
8. Add reference to subscriber service in `Resources/services.xml` (which gets loaded automatically!)

The information about this new directory has to be propagated somewhere.

### 5. Checklist
* [x]  I have written tests and verified that they fail without my change
* [x]  I have squashed any insignificant commits
* [ ]  This change has comments for package types, values, functions, and non-obvious lines of code
* [x]  I have read the contribution requirements and fulfil them.

### DIFF
--- a/engine/Shopware/Components/Plugin.php
+++ b/engine/Shopware/Components/Plugin.php
@@ -88,7 +88,7 @@ abstract class Plugin extends Bundle implements SubscriberInterface
         return $this->isActive;
     }
 
-    public function isAutoloadViews(): bool
+    public function hasAutoloadViews(): bool
     {
         return $this->autoloadViews;
     }
--- a/engine/Shopware/Components/Plugin.php
+++ b/engine/Shopware/Components/Plugin.php
@@ -47,6 +47,8 @@ abstract class Plugin extends Bundle implements SubscriberInterface
      */
     protected $pluginNamespace;
 
+    protected $autoloadViews = false;
+
     /**
      * @var bool
      */
@@ -84,6 +86,11 @@ abstract class Plugin extends Bundle implements SubscriberInterface
     final public function isActive()
     {
         return $this->isActive;
+    }
+
+    public function isAutoloadViews(): bool
+    {
+        return $this->autoloadViews;
     }
 
     /**
--- a/engine/Shopware/Components/Plugin/ResourceSubscriber.php
+++ b/engine/Shopware/Components/Plugin/ResourceSubscriber.php
@@ -111,13 +111,15 @@ class ResourceSubscriber implements SubscriberInterface
 
         $viewsDirectory = $this->pluginPath . '/Resources/views';
 
-        if (@is_dir($viewsDirectory)) {
-            $templates = (array) $args->getReturn();
+        if (!(@is_dir($viewsDirectory))) {
+            return;
+        }
 
-            if (!in_array($viewsDirectory, $templates, true)) {
-                $templates[] = $viewsDirectory;
-                $args->setReturn($templates);
-            }
+        $templates = (array)$args->getReturn();
+
+        if (!in_array($viewsDirectory, $templates, true)) {
+            $templates[] = $viewsDirectory;
+            $args->setReturn($templates);
         }
     }
 
--- a/engine/Shopware/Components/Plugin/ResourceSubscriber.php
+++ b/engine/Shopware/Components/Plugin/ResourceSubscriber.php
@@ -26,7 +26,9 @@ namespace Shopware\Components\Plugin;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Enlight\Event\SubscriberInterface;
+use Enlight_Controller_ActionEventArgs;
 use Enlight_Event_EventArgs;
+use Enlight_Exception;
 use Shopware\Components\Theme\LessDefinition;
 use Symfony\Component\Finder\Finder;
 
@@ -58,6 +60,7 @@ class ResourceSubscriber implements SubscriberInterface
             'Theme_Compiler_Collect_Plugin_Css' => 'onCollectCss',
             'Theme_Compiler_Collect_Plugin_Javascript' => 'onCollectJavascript',
             'Theme_Inheritance_Template_Directories_Collected' => 'onRegisterTemplate',
+            'Enlight_Controller_Action_PreDispatch_Backend' => 'onRegisterControllerTemplate',
         ];
     }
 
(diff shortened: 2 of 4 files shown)


engine/Shopware/Components/Plugin.php
```diff
@@ -86,6 +88,11 @@ public static function getSubscribedEvents()
         return $this->isActive;
     }
 
+    public function isAutoloadViews(): bool
```
PaddyS: Probably `isAutoloadingViews`?
I do get the reference to the property name, but I guess it should still be clear enough.

engine/Shopware/Components/Plugin/ResourceSubscriber.php
```diff
@@ -98,6 +103,24 @@ public function onCollectLess()
...
 
+    public function onRegisterTemplate(Enlight_Event_EventArgs $args): void
+    {
+        if (!$this->loadViewsDirectory) {
+            return;
+        }
+
+        $viewsDirectory = $this->pluginPath . '/Resources/views';
+
+        if (@is_dir($viewsDirectory)) {
```
PaddyS: Early return? :-)

JoshuaBehrens: Reduce indentions? I like your style!

engine/Shopware/Components/Plugin.php
JoshuaBehrens: I like the `isAutoloadingViews` as well. Should I rename the related property and variables as well? Consistency.

engine/Shopware/Components/Plugin/ResourceSubscriber.php
shyim: Maybe try to move this to CompilerPass. So we can save time here.

shyim: Maybe try to move this to CompilerPass. So we can save time here.<

JoshuaBehrens: You mean adding a subscriber via compiler pass? Sure I can move that

shyim: Just check here also if the template folder exists? 
https://github.com/shopware/shopware/pull/2284/files#diff-b0f8691023c0f7345c9827c5272f94fdR259

engine/Shopware/Components/Plugin.php
PaddyS: Not quite sure about renaming the property. Going for `$whatever->autoloadViews` seems better than `$whatever->autoloadingViews`. Meanwhile, `$whatever->isAutoloadingViews()` sounds better than `$whatever->isAutoloadViews()`.

Either way we got a bad result.
Maybe change the method to `hasAutoloadViews`? (idea sponsored by @soebbing)

engine/Shopware/Components/Plugin/ResourceSubscriber.php
JoshuaBehrens: I don't see the time saving. I have to check the directory anyway. Does that really matter putting it there.

JoshuaBehrens:
I reacted to the community reviews. Please review again.

engine/Shopware/Components/Plugin/ResourceSubscriber.php
```diff
@@ -53,6 +57,7 @@ public static function getSubscribedEvents()
             'Theme_Compiler_Collect_Plugin_Less' => 'onCollectLess',
             'Theme_Compiler_Collect_Plugin_Css' => 'onCollectCss',
             'Theme_Compiler_Collect_Plugin_Javascript' => 'onCollectJavascript',
+            'Theme_Inheritance_Template_Directories_Collected' => 'onRegisterTemplate',
```
shopwareBot, line 62: Did you tested it with backend applications?

engine/Shopware/Components/Plugin/ResourceSubscriber.php
```diff
@@ -98,6 +103,26 @@ public function onCollectLess()
...
+            return;
+        }
+
+        $viewsDirectory = $this->pluginPath . '/Resources/views';
+
+        if (!(@is_dir($viewsDirectory))) {
+            return;
+        }
+
+        $templates = (array)$args->getReturn();
```
shopwareBot: Did you let the php-cs-fixer run?

JoshuaBehrens: It should've been run. I use the githooks. But I will check.

engine/Shopware/Components/Plugin/ResourceSubscriber.php
JoshuaBehrens: No but is this needed? I learned that one had to cherry pick some templates for specific actions. Or are we talking lightweight?

JoshuaBehrens: I wanted to test but got stuck [here](https://github.com/shopware/shopware/pull/2297). Please keep open for a while so I can test later again.

JoshuaBehrens: I could now do the testing. I use the lightweight example plugin to test. It was obviously not working. I'd go either `Enlight_Controller_Action_Backend_preDispatch` or `Enlight_Controller_Dispatcher_ControllerPath_Backend_*`. Opinions?

JoshuaBehrens: I chose the backend pre dispatch event and tested against SwagLightweightModule. Works fine

engine/Shopware/Components/Plugin/ResourceSubscriber.php
JoshuaBehrens: Fixed it

shyim:

I have refactored the ResourceSubscriber also little bit to reduce disk calls

JoshuaBehrens:
Nice thoughts on it 👍 Happy coding less :)

PR #2285, opened 2019-12-09
Export Statistics - getRatingAction

pascalguhl:
Because of this ticket: 
https://issues.shopware.com/issues/SW-23749
Not possible to export more than 25 items at Conversion overview

set the limit to 25 if the list is exported to a csv

1. Click Marketing
2. Click Analysis
3. Click Analysis
4. Click Conversion overview
5. Click Export - Try to export all items of the list
6. Open the exported csv -> only 25 items

https://issues.shopware.com/issues/SW-23749

### 6. Checklist


engine/Shopware/Controllers/Backend/Analytics.php
```diff
@@ -276,11 +276,17 @@ public function getRatingAction()
             }
         }
 
+        // Sets the correct limit
+        $limit = 25;
```
jkrzefski, line 280:
```suggestion
        $limit = (int) $this->Request()->getParam('limit', 25);
```

You should use the request parameter as initial value. Otherwise the user cannot select a larger page size.

jkrzefski (changes requested):

engine/Shopware/Controllers/Backend/Analytics.php
shyim: $this->Request()->query->getInt('limit', 25);

jkrzefski:
For the sake of completeness, I would like to add: Please squash your commits before you tick the checkbox `I have squashed any insignificant commits`. 😉

PR #2286, opened 2019-12-09
Add database session config field in configuration

JoshuaBehrens:
In case you need to change e.g. `wait_timeout` for shopware but your hosting environment does not offer the choice of editing the my.cnf. Just add a `'db' => [ 'session' => [ 'wait_timeout' => 1 << 16 ] ]` to wait for a decent amount of seconds before your database or hosting partner is about to die.

Simply sets these values on the database as well everytime a connection is created.

1. Have inperformant SQL query
2. MySQL gone away
3. Increase timeout in my.cnf
4. my.cnf is readonly

Maybe a hint in the dev cheatsheet or hosting docs?

### 5. Checklist


shyim:

JoshuaBehrens:
You're always welcome :) 💙

PR #2287, opened 2019-12-10
Fix blog statistic with enabled http cache

hlohaus:

blog statistic doesn't work if http cache is enabled.

Move code from controller to statistic plugin.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/Statistics/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Frontend/Statistics/Bootstrap.php
@@ -23,6 +23,7 @@
  */
 
 use Symfony\Component\HttpFoundation\Cookie;
+use Symfony\Component\HttpFoundation\Request;
 
 /**
  * Shopware Statistics Plugin
@@ -131,6 +132,7 @@ ShopWiki;Bot;WebAlta;;abachobot;architext;ask jeeves;frooglebot;googlebot;lycos;
             $this->refreshLog($request);
             $this->refreshReferer($request);
             $this->refreshArticleImpression($request);
+            $this->refreshBlog($request);
             $this->refreshCurrentUsers($request);
             $this->refreshPartner($request, $response);
         }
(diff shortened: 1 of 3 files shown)


engine/Shopware/Plugins/Default/Frontend/Statistics/Bootstrap.php
```diff
@@ -384,4 +386,31 @@ public function refreshPartner($request, $response)
...
+        /** @var Enlight_Components_Session_Namespace $session */
+        $session = $this->get('session');
+        $blogArticleId = (int) $request->get('blogId');
+        if (empty($blogArticleId)) {
+            return;
+        }
+
+        // Count the views of this blog item
+        $visitedBlogItems = $session->get('visitedBlogItems');
+        if (!in_array($blogArticleId, $visitedBlogItems, true)) {
```
soebbing: You can turn this into an early return (`if (in_array(...)) `)

engine/Shopware/Plugins/Default/Frontend/Statistics/Bootstrap.php
```diff
@@ -384,4 +386,31 @@ public function refreshPartner($request, $response)
             }
         }
     }
+
+    private function refreshBlog(Request $request)
+    {
+        /** @var \Doctrine\DBAL\Connection $connection */
+        $connection = $this->get('dbal_connection');
+        /** @var Enlight_Components_Session_Namespace $session */
+        $session = $this->get('session');
```
soebbing, line 399: Move the retrieval of the `session` service down below the early return

soebbing (changes requested):

engine/Shopware/Plugins/Default/Frontend/Statistics/Bootstrap.php
```diff
@@ -384,4 +386,32 @@ public function refreshPartner($request, $response)
             }
         }
     }
+
+    private function refreshBlog(Request $request)
+    {
+        $blogArticleId = (int) $request->get('blogId');
```
shyim: $request->query->getInt

soebbing:
Hi @hlohaus, thank you for your contribution!

PR #2290, opened 2019-12-12
Make position reset for product slider optional

JoshuaBehrens:
In case of an update on a page the sliders can occasionally be updated. This will slide them to the first item. This also can reset the position although no change happened. Either situation is appreciatable so it is now optionally disablable.

Add an option to deactivate that setPosition call. The default is behaving like before so no real change in bahaviour unless configured differently.

1. Have a product slider at the top of the page
2. Slide to the right
3. Scroll down where an update is triggered to product sliders
4. Scroll up and the slider is back to the start
5. What product was I looking again? 😖 
6. Have @Yarisade in your team debugging it down to the problematic area

### 4. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.product-slider.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.product-slider.js
@@ -291,7 +291,15 @@
              * @property initOnEvent
              * @type {String}
              */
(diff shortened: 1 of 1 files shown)


shyim:

JoshuaBehrens:
You are very welcome today 💙

PR #2291, opened 2019-12-18
Reunite ConfigureContainerAwareCommands with the other compiler passes

JoshuaBehrens:
When using a compiler-aware-command (a command implementing the CompilerAwareInterface) from the service container it has no valid reference to the service container. So it is not properly initialized.

Readd the `ConfigureContainerAwareCommands` to the service container, remove the redundant `setContainer` call in `Shopware\Components\Console\Application::doRunCommand` and add a test to check whether things stay that way.

1. Be up to use auto completion feature
2. Type `sw:pl<tab>:act<tab> <tab>`
3. Get `Call to a member function get() on null in PluginActivateCommand.php:53`
4. On null? What is happening there?
5. Find `getContainer()->get` failing
6. But why is it not initialized? When is `setContainer` called?
7. Find `Shopware\Components\Console\Application::doRunCommand` and `ConfigureContainerAwareCommands`. Wait didn't I implement the `ConfigureContainerAwareCommands`?
8. Check auto completion pull request. Yes I did. Did I forget adding it? 😳
9. [Check diff from pull request](https://github.com/niemand-online/shopware/blame/e61f879bd28302fbe7732e20c8e122651e4a1c8a/engine/Shopware/Kernel.php#L621) 🙂 I did add it back then
10. Is it still present? NO?!
[grafik](https://knowyourmeme.com/memes/confused-nick-young)

### 4. Checklist

### DIFF
--- a/engine/Shopware/Components/Console/Application.php
+++ b/engine/Shopware/Components/Console/Application.php
@@ -184,6 +184,10 @@ class Application extends BaseApplication
             if ($command instanceof Command) {
                 $this->add($command);
             }
+
+            if ($command instanceof ContainerAwareInterface) {
+                $command->setContainer($this->getKernel()->getContainer());
+            }
         }
     }
 
--- a/engine/Shopware/Components/Console/Application.php
+++ b/engine/Shopware/Components/Console/Application.php
@@ -113,10 +113,6 @@ class Application extends BaseApplication
      */
     protected function doRunCommand(Command $command, InputInterface $input, OutputInterface $output)
     {
-        if ($command instanceof ContainerAwareInterface) {
-            $command->setContainer($this->kernel->getContainer());
-        }
-
         /** @var \Enlight_Event_EventManager $eventManager */
         $eventManager = $this->kernel->getContainer()->get('events');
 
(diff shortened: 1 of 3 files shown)


JoshuaBehrens:
I have a potential test but it fails on testing just this one test (that the shopware kernel is not initialized) but it succeeds on complete test run. I'll add it here later for review

shyim:
ContainerAware should be not used anymore. You should pass services to constructor for those actions. So I would prefer that way.

JoshuaBehrens:
Let's split it up. I readd the check in doRunCommand. So this can be easily merged without any breaks and the completion (and maybe other things) work again. You start to mark the interface as deprecated and we improve the commands one by one that they stop using the interface at all. I think we all don't want to make this a breaking PR and wait for two versions until it is merged.

shyim:
This will break commands registered by events, because they will never receive the container

JoshuaBehrens:
That is why I said
> I readd the check in doRunCommand

so everything else will work like before.

There are commands by events?

JoshuaBehrens:
No I did not. What is special about it? What do I have to look for?

JoshuaBehrens:
Tested against Shopware_Console_Add_Command and SwagImportExport

JoshuaBehrens:
Thanks @shyim for house keeping :)

PR #2292, opened 2019-12-18
Add new event before a customer is actually registered

t2oh4e:
Because there is currently no clean way to abort a registration process beside the form validation. You can decorate the RegisterService but since this does not return anything you can not pass back any information to the register controller to initiate a custom handling. I had the need to do this because I wanted to abort registrations after the form validation is run but before the customer is actually saved.

Add a new event to handle custom errors (or messages in general) during the registration process.

Try to abort a registration before a customer is saved to the database.
Example subscriber that could make use of the new event:
```php
class Test implements SubscriberInterface {
    {
        return [
            'Shopware_Modules_Admin_SaveRegister_BeforeRegister' => 'onBeforeRegister',
        ];
    }

    public function onBeforeRegister(\Enlight_Event_EventArgs $args) {
        $errors = $args->getReturn();
        $customer = $args->get('customer');
        $billing = $args->get('billing');
        $shipping = $args->get('shipping');

        // Do some custom validation stuff with the customer or the addresses
        // e.g. sync them with another system and react to API errors of that system.

        // Merge to make sure you keep errors that may have been added by other plugins previously.
        return array_merge($errors, [
           'occurred' => true,
           'personal' => [
               //reuse error structure from register controller which will be displayed in the frontend form
               'firstname' => 'Something is wrong here',
           ],
           'my_custom_errors' => [
               // Add your own errors. They will be passed to the view and you can handle them there.
           ]
        ]);
    }
}
```

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Register.php
+++ b/engine/Shopware/Controllers/Frontend/Register.php
@@ -147,15 +147,7 @@ class Shopware_Controllers_Frontend_Register extends Enlight_Controller_Action
         );
 
         if ($errors['occurred']) {
-            unset(
-                $data['register']['personal']['password'],
-                $data['register']['personal']['passwordConfirmation'],
-                $data['register']['personal']['emailConfirmation']
-            );
-
-            $this->View()->assign('errors', $errors);
-            $this->View()->assign($data);
-            $this->forward('index');
+            $this->handleRegisterError($data, $errors);
 
             return;
         }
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Frontend/Register.php
```diff
@@ -185,6 +177,26 @@ public function saveRegisterAction()
...
 
+        /** @var Enlight_Event_EventManager $eventManager */
+        $eventManager = $this->get('events');
+
+        $errors = ['occurred' => false];
+        $errors = $eventManager->filter(
+            'Shopware_Modules_Admin_SaveRegister_BeforeRegister',
+            $errors,
+            [
+                'customer' => $customer,
```
shyim, line 189: subject missing

t2oh4e: Added.

t2oh4e:
@shopwareBot Sorry I was on vacations :) I added the missing subject requested by shyim.

mitelg:
shouldn't we avoid to add `subject` to events? so in case of refactoring, it would be easier to move the code :wink:

t2oh4e:
@mitelg sounds legit. I found both ways (with and without subject) in the existing code. I can remove it again. Shall I?

shyim:

PR #2293, opened 2019-12-18
Message only has text/html MIME parts

hlohaus:

See: https://forum.shopware.com/discussion/57909/shopware-e-mails-erhalten-negative-spam-einschaetzung-durch-spamassassin

### 6. Checklist


soebbing:
Hi @hlohaus, what do you think about a litte parsing of the HTM mail? Something like `br2nl` plus `striptags` to still contain the content of the mail without cluttering the textmail with html.

hlohaus:
Not really nice. So the shop admin of no controlling of plain text.

shyim:
The normal user will not understand why has to write the same text two times

PR #2296, opened 2020-01-03
Encode arrays passed to routing assembly which are otherwise lost

pascalheidmann:
Try passing an array to the Router assemble method. Using the default generator & matcher the array values will be lost.

Encode array values

Run `Shopware()->Container()->get('router')->assemble(['foo' => ['test' => 'value']]);`
The resulting url will be looking something like this: `https://example.tld/index/index/foo/`
The value of foo is lost.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Routing/Generators/DefaultGenerator.php
+++ b/engine/Shopware/Components/Routing/Generators/DefaultGenerator.php
@@ -63,17 +63,9 @@ class DefaultGenerator implements GeneratorInterface
             unset($params['_seo']);
         }
 
-        $module = isset($params[$context->getModuleKey()])
-            ? $params[$context->getModuleKey()]
-            : $this->dispatcher->getDefaultModule();
-
-        $controller = isset($params[$context->getControllerKey()])
-            ? $params[$context->getControllerKey()]
-            : $this->dispatcher->getDefaultControllerName();
-
-        $action = isset($params[$context->getActionKey()])
-            ? $params[$context->getActionKey()]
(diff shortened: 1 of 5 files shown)


engine/Shopware/Components/Routing/Generators/DefaultGenerator.php
```diff
@@ -94,7 +86,7 @@ public function generate(array $params, Context $context)
 
         foreach ($params as $key => $value) {
             $route[] = $key;
-            $route[] = $value;
+            $route[] = is_array($value) || is_object($value) ? http_build_query($value) : $value;
```
shyim: Is there any specific reason why allowing objects? It feels wrong building that as query parameters

pascalheidmann: To allow using `\StdClass`. Otherwise, yes, no real reason besides that `http_build_query` allows objects to be passed ;)
I can remove it but then: what should happen with an object as this will silently fail again?

shyim: What about a trigger_error with an E_USER_DEPRECATED. And with 5.7 a exception.

pascalheidmann: Good point. I have added the triggering and modified the test accordingly. As this required replacing the _error_handler_ I had to move it to its own testfile to be able to clean up afterwards.

shyim:

PR #2299, opened 2020-01-15
add invoiceShippingTaxRate to order Whitelist

in-voker:
If the invoiceShippingTaxRate is not in whitelist, it's not possible to create an order with the API and set the tax rate for shipping. (default is always 0)

add the invoiceShippingTaxRate to whitelist of the order api

none

### 6. Checklist


soebbing:
Hi @in-voker, thank you for your PR! Could you please also modify the tests in `\Shopware\Tests\Functional\Components\Api\OrderTest` to make sure the field is being imported and returned on requests?

PR #2300, opened 2020-01-15
SSL fix

christopherholst:

**Loads page resources using protocol relative URIs**

change from http to https

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/index/index.tpl
+++ b/themes/Frontend/Bare/frontend/index/index.tpl
@@ -4,7 +4,7 @@
 {/block}
 
 {block name='frontend_index_html'}
-<html class="no-js" lang="{s name='IndexXmlLang'}{/s}" itemscope="itemscope" itemtype="http://schema.org/WebPage">
+<html class="no-js" lang="{s name='IndexXmlLang'}{/s}" itemscope="itemscope" itemtype="https://schema.org/WebPage">
 {/block}
 
 {block name='frontend_index_header'}


larsbo:
I think this change doesn’t solve anything at all. The schema.org url in the source code is just an information for search bots. There will be no request that can modified by an attacker. 

Schema.org itself use the http version inside their examples (microdata example):
https://schema.org/WebPage#WebPage-gen-215

soebbing:
See the FAQ for schema.org: https://schema.org/docs/faq.html#19 - it doesn't matter really

tinect:
I agree with @larsbo.

Why there has been "fixed" just one occurrence? :-)
I don't see any need to adjust just one.

soebbing:
Hi @snHolst, thanks for your contribution! 💙

larsbo:
@soebbing what about the other occurences? 😮 
https://github.com/shopware/shopware/search?q=%22http%3A%2F%2Fschema.org%2F%22&unscoped_q=%22http%3A%2F%2Fschema.org%2F%22

soebbing:
Good point. Want to make a PR? :)

larsbo:
No, thanks. Btw @tinect mentioned this before https://github.com/shopware/shopware/pull/2300#issuecomment-575510008 😉 

I'm more interested in how this change align with your contribution guidelines and why other (more beneficial) PRs are closed with the reason `We decided to close this PR because of time constraints and limited overall benefit to the product.`
I dont't want to say this looks unfair but well... some somewhat arbitrary.

soebbing:
You're right. But since this PR didn't hurt and the first part of the reason you've mentioned ("time constraints") currently allowed us to merge this, we did so.
I agree that this PR is neither the most valuable nor the most complete one, but we try to be more inclusive and lower the bar for contributions as long as they improve the product somewhat. Maybe @snHolst feels compelled to complete what he started? 😄

larsbo:
> we try to be more inclusive and lower the bar for contributions as long as they improve the product somewhat.

I'm very happy about this! 😃 
Than you for the clarification.

christopherholst:
@soebbing before I deal with it, I first wanted to see if this one pr is accepted by you at all. Now that you have accepted this, I can also update the rest. I have already integrated it in our store and I actually think it makes sense in times when SSL has simply become essential.

PR #2302, opened 2020-01-28
SSL fix for w3.org, schema.org and facebook.com

christopherholst:

Loads page resources using protocol relative URIs

change from http to https urls where https urls are available

### 6. Checklist


christopherholst:
@soebbing please, here are all fixes with a check whether this url exists as https.

tinect:
Is it on purpose and verified you edited namespaces (w3.org), too? This won't be requested, as it's just a "text" (namespace). I cannot get any information these can be valid.

Edit: Changing namespace in svg will break them - at least in preview in chrome and github.

tinect (changes requested):
Please revert all w3.org-changes, these are invalid...
https://stackoverflow.com/questions/30707609/xml-namespace-uri-with-https/30761004#30761004
https://github.com/validator/validator/issues/692
https://stackoverflow.com/questions/22976359/xml-namespace-a-security-issue

tinect (commented):
This makes it more broken, as you added non-namespace `https://www.w3.org/TR/SVG2/`

PR #2303, opened 2020-01-29
SSL fix from http to https for schema.org and facebook.com

christopherholst:

Loads page resources using protocol relative URIs

change from http to https urls where https urls are available

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/_includes/rating.tpl
+++ b/themes/Frontend/Bare/frontend/_includes/rating.tpl
@@ -81,9 +81,9 @@
 {* Microdata depending on type *}
 {block name='frontend_rating_microdata_type'}
     {if $isType === 'single'}
-        {$data='itemprop="reviewRating" itemscope itemtype="http://schema.org/Rating"'}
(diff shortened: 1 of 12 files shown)


tinect:
Just to be sure: There isn't any request to schema.org, so this change is just for style :-)

mitelg:
@tinect and totally useless for schema.org? :thinking:

larsbo:
@mitelg your buddy Hendrik checked it in a previous PR: https://github.com/shopware/shopware/pull/2300#issuecomment-575067424

tinect:
> @tinect and totally useless for schema.org? 🤔

@mitelg no, it's just typo ;-) Can be merged. I just wanted to make clear changing schema.org-declarations hasn't anything to do with security.

soebbing:
There are two possible benefits I see in this PR, both rather negligible: For one, other developers seeing the reference in the source use https when visiting the schema.org-reference. And also there won't be another PR in the foreseeable future regarding this change/. :)

soebbing:
Hi @snHolst, thanks for pulling through and change the protocol everywhere! 💙

PR #2304, opened 2020-01-30
Check the action name of RefreshStatistics against the lower case to prevent an error 500

ghost:
Check the action name of RefreshStatistics against the lower case to prevent an error 500

This change is necessary because at irregular intervals some browsers call the URL `widgets/index/refreshstatistic` instead of `widgets/index/refreshStatistic`. Without this change, this leads to an Error 500.

All checks for the action name have been provided with a strtolower and it is always checked against the lower case.

call `widgets/index/refreshstatistic` instead of `widgets/index/refreshStatistic`

nothing

### 6. Checklist


engine/Shopware/Components/HttpCache/AppCache.php
```diff
@@ -99,7 +99,9 @@ public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQ
             return $this->pass($request, $catch);
         }
 
-        if (strpos($request->getPathInfo(), '/widgets/index/refreshStatistic') === 0) {
```
shyim, line 105: stripos?

viosys: thanks

pascalheidmann: As `stripos` is case insensitive you can still use `'/widgets/index/refreshStatistic'` as comparison and there should be no need for that `@noinspection` annotation.

(same goes for other instances)

pascalheidmann (approved):

shyim:

PR #2305, opened 2020-01-30
Bugfix / missing feature: shipping time field is now language shop / subshop-aware

ghost:
Before this change it was not possible to have different delivery times in the sub or language shop.

It uses the correct shipping time from the current shop, which the shop customer is currently on, instead of using the origin value from ``s_articles_details``. This is now pretty much the same as it is already on product detail page.

1. sign in to backend
2. Create a language shop, if not already.
3. Make a shipping time translation with another value as the origin one in the language shop you created or used with step 2.
4. Clear all caches.
5. Go to shop frontend in that language shop and add the item to basket.
6. You should now see in basket views, that there is the expected value. Without the change it would be the value from the main shop.

For plugin developers a hint in the changelogs would be really helpful, since the prepared statement question markers changed in their corresponding order.

### 6. Checklist

As described above, this change can be easily tested using a language shop. So, no other tests necessary here...

Also, since this change only belongs to an filter event this could be easily placed in a plugin for older versions of Shopware 5. My changes are downwards-compatible, since it looks also in the old translation tables.

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2843,13 +2843,13 @@ SQL;
             ad.purchaseunit,
             COALESCE (ad.unitID, mad.unitID) AS unitID,
             ad.laststock,
-            ad.shippingtime,
             COALESCE(
-                coreTransArticle.objectdata,
-                coreTransDetail.objectdata,
-                transDetail.shippingtime,
                 transArticle.shippingtime,
                 ad.shippingtime
+            ) AS shippingtime,
+            COALESCE(
+                coreTransDetail.objectdata,
+                coreTransArticle.objectdata
             ) AS transShippingTime,
             ad.releasedate,
             ad.releasedate AS sReleaseDate,
@@ -2915,10 +2915,11 @@ SQL;
 
                 if (!empty($shippingTime)) {
                     $basketItems[$basketPosition]['shippingtime'] = $shippingTime['txtshippingtime'];
-                    // unset the origin one, since we don't need it anymore, to free memory:
-                    unset($basketItems[$basketPosition]['transShippingTime']);
                 }
             }
+
+            // unset the origin one, since we don't need it anymore, to free memory:
+            unset($basketItems[$basketPosition]['transShippingTime']);
         }
 
         return $basketItems;
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2651,6 +2651,7 @@ SQL;
             // Get additional basket meta data for each product
             if ($getProducts[$key]['modus'] == 0) {
                 $getProducts[$key]['additional_details'] = $additionalDetails[$getProducts[$key]['ordernumber']];
+                $getProducts[$key]['shippingtime'] = $additionalDetails[$getProducts[$key]['ordernumber']]['shippingtime'];
             }
 
             $getUnitData = [];
@@ -2843,14 +2844,7 @@ SQL;
             ad.purchaseunit,
             COALESCE (ad.unitID, mad.unitID) AS unitID,
             ad.laststock,
-            COALESCE(
-                transArticle.shippingtime,
-                ad.shippingtime
-            ) AS shippingtime,
-            COALESCE(
-                coreTransDetail.objectdata,
-                coreTransArticle.objectdata
-            ) AS transShippingTime,
+            ad.shippingtime,
             ad.releasedate,
             ad.releasedate AS sReleaseDate,
             COALESCE (ad.ean, mad.ean) AS ean,
@@ -2866,63 +2860,17 @@ SQL;
         LEFT JOIN s_articles_details AS ad ON ad.ordernumber = s_order_basket.ordernumber
         LEFT JOIN s_articles a ON (a.id = ad.articleID)
         LEFT JOIN s_articles_details AS mad ON mad.id = a.main_detail_id
-        LEFT JOIN s_order_basket_attributes ON s_order_basket.id = s_order_basket_attributes.basketID'
-            // article shipping time translation stuff below:
-        .'
-        LEFT JOIN s_articles_translations AS transArticle ON transArticle.articleID = ad.articleID
-            AND a.main_detail_id = ad.id
-            AND transArticle.languageID = ?
-        LEFT JOIN s_articles_translations AS transDetail ON transDetail.articleID = ad.id
-            AND a.main_detail_id <> ad.id
-            AND transDetail.languageID = ?
-        LEFT JOIN s_core_translations AS coreTransArticle ON coreTransArticle.objectkey = a.id
-            AND a.main_detail_id = ad.id
-            AND coreTransArticle.objecttype = \'article\'
-            AND coreTransArticle.objectlanguage = ?
-        LEFT JOIN s_core_translations AS coreTransDetail ON coreTransDetail.objectkey = ad.id
-            AND a.main_detail_id <> ad.id
-            AND coreTransDetail.objecttype = \'variant\'
-            AND coreTransDetail.objectlanguage = ?
+        LEFT JOIN s_order_basket_attributes ON s_order_basket.id = s_order_basket_attributes.basketID
         WHERE sessionID=?
         ORDER BY id ASC, datum DESC
         ';
-        $sessionId = $this->session->get('sessionId');
-        $shopId = Shopware()->Shop()->getId();
         $sql = $this->eventManager->filter(
             'Shopware_Modules_Basket_GetBasket_FilterSQL',
             $sql,
-            [
-                'subject' => $this,
-                'sessionId' => &$sessionId,
-                'shopID' => &$shopId
-            ]
+            ['subject' => $this]
         );
 
-        $basketItems = $this->db->fetchAll(
-            $sql,
-            [
-                $shopId,
-                $shopId,
-                $shopId,
-                $shopId,
-                $sessionId
-            ]
-        );
-
-        foreach ($basketItems as $basketPosition => $basketItem) {
-            if (!empty($basketItem['transShippingTime'])) {
-                $shippingTime = @unserialize($basketItem['transShippingTime']);
-
-                if (!empty($shippingTime)) {
-                    $basketItems[$basketPosition]['shippingtime'] = $shippingTime['txtshippingtime'];
-                }
-            }
-
-            // unset the origin one, since we don't need it anymore, to free memory:
-            unset($basketItems[$basketPosition]['transShippingTime']);
-        }
-
-        return $basketItems;
+        return $this->db->fetchAll($sql, [$this->session->get('sessionId')]);
     }
 
     /**
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -2843,7 +2843,12 @@ SQL;
             ad.purchaseunit,
             COALESCE (ad.unitID, mad.unitID) AS unitID,
             ad.laststock,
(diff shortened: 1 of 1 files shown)


engine/Shopware/Core/sBasket.php
```diff
@@ -2843,7 +2843,13 @@ private function loadBasketProducts()
             ad.purchaseunit,
             COALESCE (ad.unitID, mad.unitID) AS unitID,
             ad.laststock,
-            ad.shippingtime,
+            COALESCE(
+                coreTransArticle.objectdata,
```
hlohaus: what if the product translation haven't a translation for shippingtime?

engine/Shopware/Core/sBasket.php
```diff
@@ -2859,17 +2865,67 @@ private function loadBasketProducts()
...
+                $shopId,
+                $shopId,
+                $shopId,
+                $sessionId
+            ]
+        );
+
+        foreach ($basketItems as $basketPosition => $basketItem) {
+            if (
+                $basketItem['shippingtime'] !== null
```
hlohaus: !empty()

engine/Shopware/Core/sBasket.php
```diff
@@ -2859,17 +2865,67 @@ private function loadBasketProducts()
...
+                && $basketItem['shippingtime']
+            ) {
+                $shippingTime = @unserialize($basketItem['shippingtime']);
```
hlohaus: please use a second parameter in the sql result for translations.

engine/Shopware/Core/sBasket.php
```diff
@@ -2859,17 +2865,67 @@ private function loadBasketProducts()
...
+
+                if (
+                    $shippingTime !== false
+                    && is_array($shippingTime)
+                    && isset($shippingTime['txtshippingtime'])
```
hlohaus: isset is enough or !empty

hlohaus (changes requested):
see comments

engine/Shopware/Core/sBasket.php
alpham8: Then an empty result is used as before

engine/Shopware/Core/sBasket.php
alpham8: Nope, because ``empty(false) // => true``.

engine/Shopware/Core/sBasket.php
alpham8: You mean simply another column in select?

engine/Shopware/Core/sBasket.php
```diff
@@ -2859,17 +2865,67 @@ private function loadBasketProducts()
...
+        $sessionId = $this->session->get('sessionId');
+        $shopId = Shopware()->Shop()->getId();
         $sql = $this->eventManager->filter(
             'Shopware_Modules_Basket_GetBasket_FilterSQL',
             $sql,
-            ['subject' => $this]
+            [
+                'subject' => $this,
+                'sessionId' => &$sessionId,
+                'shopID' => &$shopId
```
hlohaus: remove &

engine/Shopware/Core/sBasket.php
alpham8: I'd guess not really, since an error happened while ``unserialize``, it returns indeed ``false``. And it needs to be an array to access it correct in the following lines.

engine/Shopware/Core/sBasket.php
hlohaus: basketItem['shippingtime'] !== null && $basketItem['shippingtime'] are the same as !empty($basketItem['shippingtime'])

engine/Shopware/Core/sBasket.php
alpham8: I made this, because I saw no other way to get that variables filtered, too. I mean get it as return parameter back from the filter event. I'd guess this is the only way by reference to do it so.

engine/Shopware/Core/sBasket.php
alpham8: Then an empty result is used as before

engine/Shopware/Core/sBasket.php
hlohaus: yes, please. then the first comment will be fixed too

engine/Shopware/Core/sBasket.php
hlohaus: please use !empty it's the same

engine/Shopware/Core/sBasket.php
hlohaus: and it works? :o

alpham8: It should. PHP arrays are copy-by-value (or "intelligent" compiler guess). But if you copy a reference, it still points to the same variable in the stack... So in the event you got another reference than the origin one, but it points to the same assignment in the stack

engine/Shopware/Core/sBasket.php
```diff
@@ -2843,7 +2843,13 @@ private function loadBasketProducts()
             ad.purchaseunit,
             COALESCE (ad.unitID, mad.unitID) AS unitID,
             ad.laststock,
-            ad.shippingtime,
+            COALESCE(
+                coreTransArticle.objectdata,
+                coreTransDetail.objectdata,
+                transDetail.shippingtime,
```
hlohaus: that won't work. the are only a table for product translations.

alpham8: ... and it's about products, isn't it?

ghost:
@hlohaus I made the changes from your review.
PS: Sorry about the account confusion. I accidentally commented your comments with my private account, not with the business one, with which I made the PR this time.

engine/Shopware/Core/sBasket.php
hlohaus: it's join the variant

engine/Shopware/Core/sBasket.php
```diff
@@ -2844,6 +2844,13 @@ private function loadBasketProducts()
             COALESCE (ad.unitID, mad.unitID) AS unitID,
             ad.laststock,
             ad.shippingtime,
+            COALESCE(
+                coreTransArticle.objectdata,
+                coreTransDetail.objectdata,
+                transDetail.shippingtime,
+                transArticle.shippingtime,
+                ad.shippingtime
+            ) AS transShippingTime,
```
hlohaus:
```
            COALESCE(
                transArticle.shippingtime,
                ad.shippingtime
            ) AS shippingtime,
            COALESCE(
                coreTransDetail.objectdata,
                coreTransArticle.objectdata
            ) AS transShippingTime,
```

bmalpham8: But your solution would remove the new feature to get the translated shipping time string???

``COALESCE`` takes the first not ``null`` argument. So, it would first take the main article translation, then the variant translation (an article could only be main article or a variant, so the position doesn't matter here), then variant untranslated shipping time and then the main article untranslated shipping time. If everything is ``null``, then ``null`` is returned. So, the same as before this change, which should'nt hurt the Shopware core...

engine/Shopware/Core/sBasket.php
mitelg: hey @hlohaus 
we really appreciate your engagement in doing code reviews, but please try to be polite to the contributors :blush: you could always suggests things, which **might** be more correct in your opinion, but what someone has to do or not, should be up to us :wink:

hlohaus: 👍 oki

hlohaus (approved):

engine/Shopware/Core/sBasket.php
ghost: You're both welcome 😄 I'm used to know these kinda strict code reviews, so nothin new for me 😉

hlohaus:
@bmalpham8 good work :)

engine/Shopware/Core/sBasket.php
shopwareBot: Hey @bmalpham8,

still we'd like you to remove it. The third parameter is **not** supposed to be filterable, it's just some kind of "extra context". Only the second parameter is supposed to change with a filter event.
That's the whole point of our filter events.

Of course we are aware, that you can still change the third parameter if you pass in objects.

engine/Shopware/Core/sBasket.php
```diff
@@ -2859,17 +2866,63 @@ private function loadBasketProducts()
         LEFT JOIN s_articles_details AS ad ON ad.ordernumber = s_order_basket.ordernumber
         LEFT JOIN s_articles a ON (a.id = ad.articleID)
         LEFT JOIN s_articles_details AS mad ON mad.id = a.main_detail_id
-        LEFT JOIN s_order_basket_attributes ON s_order_basket.id = s_order_basket_attributes.basketID
+        LEFT JOIN s_order_basket_attributes ON s_order_basket.id = s_order_basket_attributes.basketID'
```
shopwareBot: This actually adds some performance issues. Adding more and more joins to an already (quite) huge and critical SQL statement might lead to an exponential loss of performance in bigger systems.

Please try to find another spot to add those translations.

Have a look [here](https://github.com/shopware/shopware/blob/5.6/engine/Shopware/Bundle/StoreFrontBundle/Service/Core/ListProductService.php#L124-L193) to see how we deal with that.
Instead of adding dozens of joins, we read the IDs and their data of every entity first and then re-assign them afterwards in PHP.

ghost: This might be true for huge lists. But we're talking about basket. How many items you typically have in your basket? I'd guess less than 100. So, in conclusion, I don't see the performance issue here...

hlohaus: performance... 👍

engine/Shopware/Core/sBasket.php
```diff
@@ -2859,17 +2866,63 @@ private function loadBasketProducts()
...
+            // article shipping time translation stuff below:
+        .'
+        LEFT JOIN s_articles_translations AS transArticle ON transArticle.articleID = ad.articleID
+            AND a.main_detail_id = ad.id
+            AND transArticle.languageID = ?
+        LEFT JOIN s_articles_translations AS transDetail ON transDetail.articleID = ad.id
```
hlohaus: this join can be removed. it isn't in the select and it won't be work.

hlohaus: this join can be removed. it isn't in the select

engine/Shopware/Core/sBasket.php
ghost: Ok so, I would place this in a service. But the question is, would you recommend me to extend an existing one or create a new one?

engine/Shopware/Core/sBasket.php
shopwareBot: @alpham8 Still on it?

ghost: Please review my other comment about service usage - I'd like to do it all at once.

shyim:
Hey @bmalpham8,

I checked the problem and could fix it with one additional assignment in the basket loop. The shippingtime is translated already present but not used.

Can you check is this fixes your issue?

Just add after this line https://github.com/shopware/shopware/blob/9bf3ffaf22eb87b8506cf5a10b1e87e731e1336d/engine/Shopware/Core/sBasket.php#L2610

```
$getProducts[$key]['shippingtime'] = $additionalDetails[$getProducts[$key]['ordernumber']]['shippingtime'];
```

ghost:
@shyim I tested your line and for me it worked for main articles and variants. Thank you very much, that solves the problem in a simple way :-)

shyim:

PR #2306, opened 2020-02-06
Remove unused dependencies

JoshuaBehrens:
Loading dependencies in a plugin that require more recent versions of it are blocked by the existing ones required in a shopware composer project. Having these in a non-composer installation the plugin would just break.

Remove dependencies that seem not to be used anymore. (regarding the usage of the namespace)

1. Require `php-http/client-common ^2.0` in plugin
2. Composer breaks

A removal hint of dependencies needs to be added so if a plugin was using a not-used dependency it knows it should just use Guzzle.

### 5. Checklist


shyim:
Hey @JoshuaBehrens,
we cannot drop packaged libraries in a patch release. These dependencies are already removed in 5.7 branch.

JoshuaBehrens:
@shyim always faster than the tests 😁 I get it. Not fixing my issue though 😖

shyim:
I am closing this then because we have already removed that library in Shopware 5.7

Maybe a solution could be to prefix your dependencies with php scooper
https://github.com/humbug/php-scoper

JoshuaBehrens:
Thanks for knowing this project by your heart. I am just unhappy that this didn't happen earlier 😊

PR #2307, opened 2020-02-08
Action overriding profit from Enlight_Controller_ActionEventArgs

JoshuaBehrens:
Every dispatched event from `Enlight_Controller_Action` passes an instance of `Enlight_Controller_ActionEventArgs` as argument except the overriding notifyUntil. This should profit from it as well as it provides better typehinting.

Replaces `['subject' => $this]` with `$args` which has the value of `['subject' => $this, 'request' => $this->Request(), 'response' => $this->Response()]`

1. Subscribe different module or controller dispatch event
2. Override one action and use the same argument typehint
3. `TypeError: Argument 1 must be an instance of Enlight_Controller_ActionEventArgs, instance of Enlight_Event_EventArgs given`

A change in haehnchens idea plugin could be made

### 5. Checklist


shyim:

JoshuaBehrens:
You're welcome 💙

PR #2308, opened 2020-02-11
unlimit searchable orders

fivetide:
In the backend, you can not search for all orders of a certain customer, if that customer has more than 400 orders.

Remove the result limit in \Shopware\Models\Order\Repository::searchCustomers

- Have a customer with more than 400 orders.
- search for that customer's email in the search-mask in Backend -> Orders

### 6. Checklist


shyim:
Did you try large amount of entries. How does this reflect in searching performance?

fivetide:
@shyim we have not seen measurable impact on search performance. When searching for our affected case (one customer with ~1.600 Orders), we had request times around 1100ms either way.

shyim:

PR #2309, opened 2020-02-11
Missing controller attribute check used the same message as the missi…

rf-elio:
Wrong error message for a missing controller attribute in the "shopware.controllers" tag was displayed.

Displays the correct error message.

Tag an controller without the controller attribute and you see the same error message that you get by not providing the module attribute.

### 6. Checklist


shyim:

PR #2311, opened 2020-02-12
Remove deprecation comments for version 6.0 in favor of stability

jkrzefski:
This project will never reach version 6.0 in favor of [shopware/platform](https://github.com/shopware/platform). That means, these deprecations will never be removed unless the target version for their removal is reduced. Without this change these deprecations are meaningless.

Change the target version of deprecations from 6.0 to 5.8.

None.

None.

None.

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Application.php
+++ b/engine/Library/Enlight/Application.php
@@ -24,7 +24,6 @@
  * the configuration automatically. It loads the different resources, for example classes, loader or the
  * managers for the different packages (Hook, Plugin, Event).
  *
- * @deprecated since 5.2, to be removed in 6.0
  * @category   Enlight
  * @package    Enlight_Application
  * @copyright  Copyright (c) 2011, shopware AG (http://www.shopware.de)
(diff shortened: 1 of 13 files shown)


JoshuaBehrens:
So deprecation comments regarding 6.0 are not worth anything?

We see deprecation marks since 5.2 (3+ years). Believe in it. And now we just get ditched with a shopware 6?

So there is no intention to improve the overall quality of the product at these points anymore?

Is shopware 5 itself now deprecated?!

I know I exaggerate

mitelg:
I think it would be ok, if there is a deprecation change to SW5.8 and in the same step changing all code to the alternative.
Another PR for the 5.8 branch, which removes the old code.

Maybe not all at once, which was done here, but maybe step by step

what do you think @shopwareBot ?

shyim:
Hey,

on the last scd we said we want to focus on compatibility and stability. With the deprecations inside this PR we don't win any benefit. It looks better, but makes work for plugin developers. So in our limited time we want to focus on that and break things **only** if it's really necessary.

These deprecations are still in use (searched in all plugins in store)

In the next minor version the breaks are only from dependencies we had to update to ensure the compatibility in the support time.

So I hope I could explain our point of view. This all does **not** mean that Shopware 5 is deprecated. It's still in development, but we are focusing on compatibility and stability.
That's the whole reason why we decided to remove the deprecation.

jkrzefski:
@shopwareBot @shyim I now have removed the deprecation comments.

shyim:

PR #2313, opened 2020-02-19
Replace "beliebig" in Price with const

pascalheidmann:
`beliebig` is a really bad value as it isn't descriptive, messing with type (`string` instead of `int`), repetitive (-> typo prone) and should be removed for better i18n.

This PR isn't replacing the value `beliebig` yet but at least adds an abstraction by replacing all instances (in PHP) with a reference to a newly introduced const `Shopware\Models\Article\Price:: NO_PRICE_LIMIT`. This way the change isn't breaking.
In the future another PR should replace the default value of `Shopware\Models\Article\Price::$to` with `null` as this is a better fit.

Using the const instead of default value is recommended.

### 6. Checklist


pascalheidmann:
Right now this is failing cause there are some very wild `$query->getSQL()` actions are used to do subqueries without copying their bound parameters...

PR #2314, opened 2020-02-19
Allow filtering log files by name in backend

pascalheidmann:
Filtering for log files in backend isn't working.

Adds filtering file names by already provided `query` parameter.

1. Go to `Settings` -> `Logfile` and then to `System log` tab
2. Try searching for a file -> won't work
3. Apply PR and try again -> will filter correctly

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Log.php
+++ b/engine/Shopware/Controllers/Backend/Log.php
@@ -191,7 +191,7 @@ class Shopware_Controllers_Backend_Log extends Shopware_Controllers_Backend_ExtJ
         // filter against input
         $query = trim($this->Request()->getParam('query', ''));
         foreach ($files as $k => $file) {
-            if (!empty($query) && mb_stripos($file[0], $query) === false) {
+            if ($query !== '' && mb_stripos($file[0], $query) === false) {
                 continue;
             }
             $files[$k] = [
--- a/engine/Shopware/Controllers/Backend/Log.php
+++ b/engine/Shopware/Controllers/Backend/Log.php
@@ -189,23 +189,19 @@ class Shopware_Controllers_Backend_Log extends Shopware_Controllers_Backend_ExtJ
         $defaultFile = $this->getDefaultLogFile($files);
 
         // filter against input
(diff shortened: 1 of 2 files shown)


engine/Shopware/Controllers/Backend/Log.php
```diff
@@ -188,13 +188,22 @@ public function getLogFileListAction()
         $files = $this->getLogFiles($logDir);
         $defaultFile = $this->getDefaultLogFile($files);
 
-        $files = array_map(function ($file) use ($defaultFile) {
+        // filter against input
+        $query = $this->Request()->getParam('query');
+        if (!empty($query)) {
+            $query = trim($query);
+            $files = array_filter($files, static function ($file) use ($query) {
```
mitelg: isn't it possible to combine these two loops? :thinking:

pascalheidmann: You mean `array_filter` and `array_map`? Sadly no, at least not without resolving to `foreach`...

pascalheidmann:
I combined the `array_filter` and `array_map` loops into a single foreach and added testcases for the controller action.

engine/Shopware/Controllers/Backend/Log.php
```diff
@@ -188,15 +188,20 @@ public function getLogFileListAction()
         $files = $this->getLogFiles($logDir);
         $defaultFile = $this->getDefaultLogFile($files);
 
-        $files = array_map(function ($file) use ($defaultFile) {
-            return [
+        // filter against input
+        $query = trim($this->Request()->getParam('query', ''));
+        foreach ($files as $k => $file) {
+            if (!empty($query) && mb_stripos($file[0], $query) === false) {
```
mitelg:
```suggestion
            if ($query !== '' && mb_stripos($file[0], $query) === false) {
```
I would check against an empty string, so it is more explicit. `empty()` checks to much. e.g. the string `'0'` will evaluate to empty and this is not intended, is it?

shyim:

PR #2315, opened 2020-02-19
Fix id mapping when copying emotions and their translations

jeboehm:
The fields start_row, start_col, end_row, end_col are always set to '1' in s_emotion_element. This leads to a non unique key that is used to map and copy the translation objects. All translation objects will be saved with the key 1111 in copyElementTranslations::$oldObjectKeys.
When an emotion is copied then all element translations except of one are getting lost.

My change reads all the viewport positioning elements and concatinates the values into an identifier string. Each element gets its own unique key again.
The change is also backward compatible with emotions without an exact viewport positioning.

- Create a new emotion with two different text elements.
- Translate each element with different texts
- Copy the emotion
- Review the copied emotion. The translations are either completely lost or mapped incorrectly.

-

-

### 6. Checklist

- [✅] I have squashed any insignificant commits
- [✅] I have read the contribution requirements and fulfil them.

### DIFF
--- a/engine/Shopware/Controllers/Backend/Emotion.php
+++ b/engine/Shopware/Controllers/Backend/Emotion.php
@@ -1514,7 +1514,20 @@ EOD;
      */
     private function getElementIdentifier(Element $el)
     {
-        return $el->getStartCol() . $el->getStartRow() . $el->getEndCol() . $el->getEndRow();
+        $identifier = '';
+
+        foreach ($el->getViewports() as $viewport) {
+            /** @var \Shopware\Models\Emotion\ElementViewport $viewport */
+            $identifier .= $viewport->getAlias()
+                .$viewport->getStartRow()
+                .$viewport->getStartCol()
+                .$viewport->getEndRow()
+                .$viewport->getEndCol();
+        }
+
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Emotion.php
```diff
@@ -1510,11 +1510,26 @@ private function copyElementTranslations(Emotion $emotion, Emotion $clonedEmotio
     /**
      * creates a unique identifier string based on the grid position
      *
+     * @param Element $el
```
shyim: We don't add ``@param`` for typed parameters

jeboehm: Fixed!

PaddyS:
Hey @jeboehm,

can you please test if your changes still work with ZIPs with the old code?

jeboehm:
Hi @PaddyS 
I've successfully exported an emotion with the old code base and imported it on the new base.

shyim:
Hey @jeboehm,
can you please sign the CLA?

jeboehm:
Done!

signed

shyim:
Anddd merged.

PR #2316, opened 2020-02-20
Add command for zip files

JoshuaBehrens:
There is still no really good way to version community store plugins.
1. sw:store:download has a bad versioning support and requires credentials
2. packages.friendsofshopware.com is not officially supported

Now you can at least version a downloaded plugin file and simply do:
`bin/console sw:plugin:zip-import SwagBundle.zip && bin/console sw:plugin:install --activate --clear-cache SwagBundle`

Add a new command to import plugin zip files.

1. Have a previously downloaded plugin zip file
2. Open /backend
3. Open plugin manager
4. Upload zip file
5. Install plugin
6. Activate plugin
7. Be annoyed that it is not possible via CLI in an easy way to do the same

### 4. Checklist

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Commands/PluginZipImportCommand.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Commands/PluginZipImportCommand.php
@@ -1,0 +1,101 @@
+<?php declare(strict_types=1);
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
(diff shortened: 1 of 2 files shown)


engine/Shopware/Bundle/PluginInstallerBundle/Commands/PluginZipImportCommand.php
```diff
@@ -0,0 +1,101 @@
...
+        parent::__construct();
+        $this->downloadService = $downloadService;
+        $this->installerService = $installerService;
+    }
+
+    protected function configure()
+    {
+        parent::configure();
+        $this
+            ->setName('sw:plugin:zip-import')
```
shyim, line 58: Not needed

engine/Shopware/Bundle/PluginInstallerBundle/Commands/PluginZipImportCommand.php
```diff
@@ -0,0 +1,101 @@
...
+            $output->writeln('File does not exist');
+
+            return 1;
+        }
+
+        $information = pathinfo($zipFile);
+        $pluginFileName = $information['basename'];
+
+        $this->downloadService->extractPluginZip($zipFile, $pluginFileName);
+        $output->writeln('Plugin zip file ' . $pluginFileName . ' has been successfully imported');
```
shyim, line 92: sprintf

shyim (changes requested):

engine/Shopware/Bundle/PluginInstallerBundle/Commands/PluginZipImportCommand.php
JoshuaBehrens: Why do all the other commands still do it? And why is it not needed?

shyim: You have defined that in the service definition

JoshuaBehrens: That was my only way to get it registered. Without the command attribute at the tag it was not loaded. What does it do? But sure I can remove it.

PR #2318, opened 2020-02-21
Add metadata to gcp adapter

wbob:

it allows to modify the objects metadata. I needed it to serve files with Cache-Control headers

add a config array

it does for GCP what SW-24932 / shopware/shopware@bf691c24 does for AWS S3

none, the Default Config is self explanatory

### 6. Checklist

Opinionated: the Cache-Control header ist set to a week, similar to the nginx "expires 1w" directive for media assets.

### DIFF
--- a/engine/Shopware/Configs/Default.php
+++ b/engine/Shopware/Configs/Default.php
@@ -114,6 +114,9 @@ return array_replace_recursive([
                 'keyFilePath' => '',
                 'bucket' => '',
                 'root' => '',
+                'metadata' => [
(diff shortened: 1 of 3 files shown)


engine/Shopware/Configs/Default.php
```diff
@@ -114,6 +114,9 @@
                 'keyFilePath' => '',
                 'bucket' => '',
                 'root' => '',
+                'metadata' => [
+                    'Cache-Control' => 'public, max-age=604800'
```
shyim, line 118: Please remove that. We cannot add something like this in a patch release.

PR #2319, opened 2020-02-21
Move highfreq refresh statistics delete to lowfreq Clearing cronjob

wbob:

when enabling the refreshStatistics function for a higher traffic site the DELETE turned out to be a bottleneck

it moves the 2nd DELETE in the function to the Clearing cronjob. For the pageimpression / uniquevisitor count a daily cleanup is sufficient as it filters with its own WHERE-condition in the pool table before counting to their target. For the currentuser table deleting on rand()%10 of requests does make sense, not for the daily pool though. If the cronjobs runs at the end of the day, it still keeps the pool table small for inserts.
Git will not tell the story of that code, I can find it in sw 3.5.6 still. I guess the first DELETE was present before the second was added.

haven't found any, in general performance topics

### 6. Checklist

Re checklist: I tested the clearing cronjob to successfully delete old entries, but didn't write tests.

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/CronRefresh/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Frontend/CronRefresh/Bootstrap.php
@@ -81,6 +81,11 @@ class Shopware_Plugins_Frontend_CronRefresh_Bootstrap extends Shopware_Component
         $data['referrer']['rows'] = $this->deleteOldReferrerData($this->get('config')->get('maximumReferrerAge'));
         $data['article_impression']['rows'] = $this->deleteOldArticleImpressionData($this->get('config')->get('maximumImpressionAge'));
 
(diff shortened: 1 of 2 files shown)


wbob:
We didn't see cleanupStatistic() impacting another higher traffic 10.3 mariadb much, only 10.2 with enabled binlogs. There were [improvements](https://mariadb.com/kb/en/changes-improvements-in-mariadb-103/#optimizerperformance) to 10.3 so an explanation could be a DELETE having less overhead (see the undo log changes to "persistent data structures").

Some numbers during regular traffic for casual comparison: refreshStatistic ~250ms, ~150ms DB time. +40ms with the DELETE on pool on 10.2, +15ms on 10.3. The DELETE on currentusers is one magnitude quicker as the table is smaller. The stats controller is among the most requested (top 3), so I think the change is worthwhile also for >10.2 mariadb users.

shyim:

PR #2320, opened 2020-02-25
Allow filtering customer groups and shops in order list

pascalheidmann:
You can't filter

Allow remote filtering

* Open order list in backend
* start searching for a customer group or shop in their filters
* will still show all entries

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Base.php
+++ b/engine/Shopware/Controllers/Backend/Base.php
@@ -197,8 +197,9 @@ class Shopware_Controllers_Backend_Base extends Shopware_Controllers_Backend_Ext
 
         $request = $this->Request();
         $filter = $request->getParam('filter', []);
-        if ($request->getParam('query')) {
-            $query = '%' . trim($request->getParam('query')) . '%';
+        $query = trim($request->getParam('query', ''));
+        if ($query !== '') {
+            $query = '%' . $query . '%';
             $filter[] = [
                 'property' => 'groups.key',
                 'value' => $query,
@@ -701,8 +702,9 @@ class Shopware_Controllers_Backend_Base extends Shopware_Controllers_Backend_Ext
 
         $request = $this->Request();
         $filter = $request->getParam('filter', []);
-        if ($request->getParam('query')) {
-            $query = '%' . trim($request->getParam('query')) . '%';
+        $query = trim($request->getParam('query', ''));
+        if ($query !== '') {
+            $query = '%' . $query . '%';
             $filter[] = [
                 'property' => 'shop.name',
                 'expression' => 'LIKE',
--- a/engine/Shopware/Controllers/Backend/Base.php
+++ b/engine/Shopware/Controllers/Backend/Base.php
@@ -194,11 +194,31 @@ class Shopware_Controllers_Backend_Base extends Shopware_Controllers_Backend_Ext
             'groups.taxInput as taxInput',
             'groups.mode as mode',
         ]);
-        $builder->addFilter($this->Request()->getParam('filter', []));
-        $builder->addOrderBy($this->Request()->getParam('sort', []));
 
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Base.php
```diff
@@ -194,11 +194,31 @@ public function getCustomerGroupsAction()
...
             'groups.mode as mode',
         ]);
-        $builder->addFilter($this->Request()->getParam('filter', []));
-        $builder->addOrderBy($this->Request()->getParam('sort', []));
 
-        $builder->setFirstResult($this->Request()->getParam('start'))
-            ->setMaxResults($this->Request()->getParam('limit'));
+        $request = $this->Request();
+        $filter = $request->getParam('filter', []);
+        if (empty($filter) && $request->getParam('query')) {
```
PaddyS: What's the reason for `empty($filter)`?

pascalheidmann: I added this check to prevent clashing with filters that change the same but after considering this shouldn't be problem so I'll remove it

engine/Shopware/Controllers/Backend/Base.php
```diff
@@ -194,11 +194,31 @@ public function getCustomerGroupsAction()
...
             'groups.mode as mode',
         ]);
-        $builder->addFilter($this->Request()->getParam('filter', []));
-        $builder->addOrderBy($this->Request()->getParam('sort', []));
 
-        $builder->setFirstResult($this->Request()->getParam('start'))
-            ->setMaxResults($this->Request()->getParam('limit'));
+        $request = $this->Request();
+        $filter = $request->getParam('filter', []);
+        if ($request->getParam('query')) {
```
mitelg: store the trimmed `query` parameter in a variable, so you don't need to get it twice. same below

engine/Shopware/Controllers/Backend/Base.php
```diff
@@ -194,11 +194,32 @@ public function getCustomerGroupsAction()
...
         ]);
-        $builder->addFilter($this->Request()->getParam('filter', []));
-        $builder->addOrderBy($this->Request()->getParam('sort', []));
 
-        $builder->setFirstResult($this->Request()->getParam('start'))
-            ->setMaxResults($this->Request()->getParam('limit'));
+        $request = $this->Request();
+        $filter = $request->getParam('filter', []);
+
+        if ($query = trim($request->getParam('query', ''))) {
```
mitelg:
```suggestion
        $query = trim($request->getParam('query', ''))
        if ($query !== '') {
```
assignments in if :-1: 

you also would have implicit castings on the result, which will be used for the if check. better use explicit comparisons

engine/Shopware/Controllers/Backend/Base.php
```diff
@@ -679,14 +700,24 @@ public function getShopsAction()
         /** @var \Shopware\Models\Shop\Repository $repository */
         $repository = Shopware()->Models()->getRepository(\Shopware\Models\Shop\Shop::class);
 
+        $request = $this->Request();
+        $filter = $request->getParam('filter', []);
+        if ($request->getParam('query')) {
+            $query = '%' . trim($request->getParam('query')) . '%';
```
mitelg:
```suggestion
        $query = trim($request->getParam('query', ''))
        if ($query !== '') {
            $query = '%' . $query . '%';
```

shyim:
The CI is failing

pascalheidmann:
@shyim yeah cause the database wasn't working, I guess because of github's recents hickups...
Can you restart the CI tests?

pascalheidmann:
Ok, my fault 😅 
There had been two semicolons missing in the last original commits which then broke the class...

mitelg:
thanks for your contribution @pascalheidmann :+1: :tada: :blue_heart:

PR #2322, opened 2020-02-26
Check against valid category ID in Shopware_Controllers_Backend_Category::saveDetail

pascalheidmann:
Trying to update a category that has been removed by another user will result in an uncatched error.

Check if a category could be found and otherwise return a more descriptive error to the end user.

- Use 2 tabs/windows with the backend and open the same category
- modify something in first tab
- delete category in second tab
- try saving in first tab
-> error

### 6. Checklist


engine/Shopware/Controllers/Backend/Category.php
```diff
@@ -535,6 +535,10 @@ public function saveDetail()
         } else {
             $categoryModel = $this->getRepository()->find($categoryId);
         }
+        
+        if (!$categoryModel) {
+            $this->View()->assign(['success' => false, 'message' => 'Invalid categoryId']);
```
shyim: You didn't tested that. When its missing it's still running to the underlying code which will throw an error cause of null.

Please also translate the error and maybe add a single test case. Should be easy :)

pascalheidmann: Editing in github editor is a bad idea... :D
Yeah, will add at least a return statement

pascalheidmann: Now the PR got a bit bigger than anticipated...
I removed all `Shopware()->Models()` calls and replaced those with DI provided instances ;)

tests/Functional/Controllers/Backend/CategoryTest.php
```diff
@@ -227,9 +240,10 @@ private function getDummyData()
     /**
      * Helper method to create the dummy object
      *
-     * @return Category
+     * @throws \Doctrine\ORM\ORMException
```
PaddyS: Remove those please

tests/Functional/Controllers/Backend/CategoryTest.php
```diff
@@ -206,12 +206,25 @@ public function testDelete($id)
...
+        unset($params['parentId']);
+        $params['id'] = -10000;
+        $params['articles'] = [];
+        $params['customerGroups'] = [];
+
+        // Test new category
+        $this->Request()->setParams($params);
+        $this->dispatch('backend/Category/createDetail');
+        static::assertFalse($this->View()->success);
+        static::assertNotEmpty($this->View()->message);
```
PaddyS, line 221: Please check for exact message

pascalheidmann: I would like to but this would require the current language settings I guess. Or can I just call 
`Shopware()->Container()->get('snippets')->getNamespace('backend/category/main')->get('saveDetailInvalidCategoryId', 'Invalid categoryId')` and compare with the returned message?

engine/Shopware/Controllers/Backend/Category.php
```diff
@@ -42,7 +45,25 @@ class Shopware_Controllers_Backend_Category extends Shopware_Controllers_Backend
     /**
      * @var \Shopware\Models\Customer\Repository
      */
-    protected $customerRepository = null;
+    protected $customerRepository;
```
shopwareBot, line 48: @pascalheidmann 
Can you please revert the refactoring back since this could lead to unwanted side effects.

pascalheidmann: @shopwareBot are you referring to general refactoring or explicitly this change? This change isn't changing anything logically as class variables without any value are implicitly always `null`.

PaddyS: Hey @pascalheidmann,

we'd like you to remove all refactoring and just keep it as simple as possible. Just leave your `if` condition and then it's fine. :-)

PR #2326, opened 2020-03-10
Add translations to attributes with type=combo to their values

pascalheidmann:
While you can translate labels of an attribute field you can't translate the values itself on a combo field.

Iterate over `arrayStore` in `options` and try to translate those based on their key 

Create a new attribute like 
```
$crudService = $this->container->get('shopware_attribute.crud_service');
        $crudService->update(
            's_articles_attributes',
            'my_cool_plugin',
            'combobox',
            [
                'label' => 'Dropdown',
                'displayInBackend' => true,
                'arrayStore' => [
                    ['key' => 1, 'value' => 'Value 1'],
                    ['key' => 2, 'value' => 'Value 2'],
                    ['key' => 3, 'value' => 'Value 3'],
                ],
            ]
        );
```
Afterwards you can translate or import translations for the label but not its values. With this PR those will be translated

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/AttributeBundle/Controllers/Backend/AttributeData.php
+++ b/engine/Shopware/Bundle/AttributeBundle/Controllers/Backend/AttributeData.php
@@ -115,7 +115,7 @@ class Shopware_Controllers_Backend_AttributeData extends Shopware_Controllers_Ba
     /**
      * @param ConfigurationStruct[] $columns
      */
-    private function translateColumns($columns)
+    private function translateColumns(array $columns): void
     {
         $snippets = $this->container->get('snippets')->getNamespace('backend/attribute_columns');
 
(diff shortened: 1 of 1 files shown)


PaddyS:
Hey @pascalheidmann,

thank you for your pull request.

Unfortunately I am failing to understand, why you would want to translate actual values.
Could you please take some time to enlighten me on that?

shyim:
Your example and the changed code do not match.
Can you stop creating pull requests without testing it 🙈

pascalheidmann:
@PaddyS the naming might be confusing as the `key` in the `arrayStore` is the actual value and `value` is the label which should be translated.

@shyim you are right, I played with that service around but looks like I then committed the wrong changes 🙈I changed that so that this should work as expected.

shyim:

PR #2327, opened 2020-03-10
fix routing problem on nginx as reverse proxy

linushstge:
This change prevents caching issues if shopware runs behind a reverse proxy like nginx. 

This change force the openPdfAction to be interpreted as controller action instead of a pdf file.
This fix adds a trailing slash for the openPdfAction before request params will be added.

* Setup nginx as reverse proxy for serving static files based on the request URL
* Perform an order
* Create a document
* Open the document in the backend 
* URL will be generated: shop.com/backend/**openPdf**?id=<file-hash>
* Response will be a 404 

* https://forum.shopware.com/discussion/61983/kann-pdf-belege-aus-dem-backend-heraus-nicht-aufrufen
* https://talk.plesk.com/threads/issue-with-pdf-files.353074/

No documentation has to be changed

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/order/view/list/document.js
+++ b/themes/Backend/ExtJs/backend/order/view/list/document.js
@@ -242,7 +242,7 @@ Ext.define('Shopware.apps.Order.view.list.Document', {
         var spec = {
             tag: 'a',
             html: display,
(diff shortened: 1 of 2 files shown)


shyim:

PR #2328, opened 2020-03-10
Move usage of useShortDescriptionInListing from sArticles to LegacyStructConverter

jkrzefski:
The config value `useShortDescriptionInListing` is only honored when requesting the listing via the `frontend/listing` controller. When the user starts scrolling or applies a filter to the listing, the new result is loaded from the `widgets/listing` controller. Because the frontend controller uses `sAdmin` and the widget controller does not, the description will be different.

I moved the condition on whether to use the description or the short description from `sArticles` into the `LegacyStructConverter`. This way the condition is applied in both situations.

1. Set `useShortDescriptionInListing` to true.
2. Have a product with a description and a short description (that has more than 5 characters).
3. Open a listing with said product. Notice that you will see the short description.
4. Apply a filter that will match with said product and let ajax load the filter results. Notice that you will see the long description.

None.

None.

### 6. Checklist


shyim:

PR #2329, opened 2020-03-12
fix related slider block

uehler:
the frontend_detail_index_similar_slider_content block belongs to the similar.tpl

changes the frontend_detail_index_similar_slider_content in the related.tpl to frontend_detail_index_related_slider_content

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/detail/tabs/related.tpl
+++ b/themes/Frontend/Bare/frontend/detail/tabs/related.tpl
@@ -2,9 +2,12 @@
 
 {if $sArticle.sRelatedArticles && !$sArticle.crossbundlelook}
     {* Related products - Content *}
-    {block name="frontend_detail_index_similar_slider_content"}
-        <div class="related--content">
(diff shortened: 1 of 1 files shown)


themes/Frontend/Bare/frontend/detail/tabs/related.tpl
```diff
@@ -2,9 +2,9 @@
 
 {if $sArticle.sRelatedArticles && !$sArticle.crossbundlelook}
     {* Related products - Content *}
-    {block name="frontend_detail_index_similar_slider_content"}
+    {block name="frontend_detail_index_related_slider_content"}
```
mitelg, line 5: isn't this a break?

uehler: maybe 😅
so target shopware 5.7?

mitelg: no, but you have to preserve the old block and the new block surrounds the old one 

then deprecate the old block

shyim:

PR #2331, opened 2020-03-16
Check against valid category ID in Shopware_Controllers_Backend_Category::saveDetail

pascalheidmann:
Trying to update a category that has been removed by another user will result in an uncatched error.

Check if a category could be found and otherwise return a more descriptive error to the end user.

- Use 2 tabs/windows with the backend and open the same category
- modify something in first tab
- delete category in second tab
- try saving in first tab
-> error

This is #2322 but without refactoring & translation. I will add a follow up PR with that changes.

### 6. Checklist


shyim:

PR #2332, opened 2020-03-16
Follow up to #2331: Refactoring backend controller "Category" & testcase

pascalheidmann:
Code quality & translation 

Building upon #2331 , adding:
- Introduce translation to `Shopware_Controllers_Backend_Category` and its test case
- Refactor both the controller and its test case

This is a follow up to #2331 

### 6. Checklist


engine/Shopware/Controllers/Backend/Category.php
```diff
@@ -42,7 +50,25 @@ class Shopware_Controllers_Backend_Category extends Shopware_Controllers_Backend
...
+     * @var ModelManager
+     */
+    private $em;
+
+    /**
+     * @var Shopware_Components_Snippet_Manager
+     */
+    private $translation;
+
+    public function __construct(ModelManager $em, Shopware_Components_Snippet_Manager $translation)
```
shyim: Can you alias the SnippetManager?

pascalheidmann: I would prefer introducing an interface and using that instead but that would blow up this ticket's intention ;)

shyim: Maybe we can alias it in 5.7 branch. So both would be possible :)

snippets/backend/category/main.ini
```diff
@@ -136,6 +137,7 @@ category/action/add/success = "[0]x Artikel Zuordnungen gespeichert"
 category/action/remove/failure = "Während des Löschens der Artikel ist folgender Fehler aufgetreten:"
 category/action/remove/success = "[0]x Artikel Zuordnungen gelöscht"
 main/window/button/save = "Speichern"
+saveDetailInvalidCategoryId = "Ungültige CategoryId"
```
mitelg:
```suggestion
saveDetailInvalidCategoryId = "Ungültige Kategorie-ID"
```

shyim:

PR #2333, opened 2020-03-16
New order filter "supplier"

pascalheidmann:
Sometimes filtering orders by article number isn't enough.

This PR is based on an internal ticket from us done by @NilsEvers and adds a new `suppliers` filter to order list.

### 6. Checklist


shyim:
Please test your changes also with enabled Elasticsearch backend. These filters have to work in both.

pascalheidmann:
@shyim after digging for at least 1 hour around I cannot find a useful way to make this feature available with ES. But this is a result of the flat structure which already makes filtering by details (and therefore by article) not possible with ES which is already part of shopware for years.

I can hide those fields on ES installations if this would make support easier ;)

shyim:
Can you add a test case for your new filter option? I will build the Elasticsearch part then after importing

shyim:

PR #2334, opened 2020-03-18
Seo remove comments

hlohaus:

https://forum.shopware.com/discussion/comment/267138/#Comment_267138

1. The new "http compression" is slow. Use the default compression for esi elements too.
2. Add the config to the performance module.

see: 1.

open:

view-source:https://www.arktis.de/widgets/emotion/index/emotionId/338/controllerName/index

see that the source code is not optimized.

### 6. Checklist


engine/Shopware/Controllers/Backend/Performance.php
```diff
@@ -571,6 +571,7 @@ protected function prepareConfigData()
                 'disableArticleNavigation',
                 'http2Push',
                 'minifyHtml',
+                'seoremovecomments',
```
mitelg, line 574: where is this new config used?

hlohaus: It's not new.  It used as sSEOREMOVECOMMENTS.

shyim:
Hey @hlohaus ,
can you please fill the pull request template?

Sadrak:
Hi, i wrote a simple test and the performance is way better with this change, but also the sSEOREMOVECOMMENTS removed whitespaces in script|pre|textarea, so this is destructive! Also #2337 ist 4x faster than #2334.

Also the name is missleading. sSEOREMOVECOMMENTS should only remove comments i guess.

Perhaps we should combine both PR #2334 and #2337 to one proper solution.

Also i don't understand why the checkout should not be optimized?!

Edit: perhaps i am stupid, but this PR did not change the underlying html compress algo at all? I am simple confused now.

What this PR make exactly?

hlohaus:
yes, but:

1. sSEOREMOVECOMMENTS would be executed in default. so i can't be more destructive.
2. The name is missleading, yes, but that was it before too.
3. the checkout was not optimized before too.

Maybe we can add an interface for controllers to exclude them.

shyim:
Hey,
Can you remove the configuration from the Basic settings? The configuration should be only located in one module.

jonashrem:
I still see, the html compressor beeing slow

grafik

can we have another look into this PR (or other possibilities)

mitelg:
hey @jonashrem 

feel free to create another PR or create a ticket for that :slightly_smiling_face:

Sadrak:
@mitelg 

i provided everything to the shopware devs to fix this bug. This is kind of mean to ask for more help ... please simple fix it or stay at wontfix (https://github.com/shopware/shopware/pull/2337#issuecomment-827589053).

mitelg:
hey @Sadrak 

I didn't want to be mean :slightly_smiling_face: I was not aware of the other PR and the resulting ticket on our side, sorry for that!

If you still need the HTML compression, you may want to add information to the ticket, which is still TODO, what you need as extensions points here, so it would be possible to implement an own solution with a plugin.

Sadrak:
Hello @mitelg 

sorry for the harsh word, but i spent a lot of time to understand this bug, find a solution, document everything ... and then get canceled from shopware. And then see a request to another user who should place time in a new PR/ticket.

I understand that simple replacing the buggy code with my solution is problematic, but in the first place the bug occur due to a direct copy&paste from stackoverflow ... so there was time to implement a feature without tests and when a bug occur and a solution arise this solution was declined due to missing tests (perhaps shopware should have "contributed" the tests in such a case) ...

Sorry to sound frustrated, but this was simple frustrating as a developer/user ...

jonashrem:
As there are already two PRs and an internal ticket ( https://issues.shopware.com/issues/SW-26043 ). I'm not sure, what I can provide here. So I would agree to @Sadrak in this regard.

For my understanding, it was already decided how to go in SW-26043, but it has not been implemented yet.

If there is anything I can help with, let me know, but I don't see more tickets/PRs is much useful here.

mitelg:
hey @Sadrak @jonashrem 

I am really sorry! As mentioned above I was not aware of the other PR and the ticket in the time I wrote my first answer here. 

This should not be an excuse, but we changed the responsibility on Shopware 5 projects last year, so this was simply not in my mind.

Nevertheless the decision was made to remove this feature again. I will add a hint so that we will implement extensions points, so that the compression could be provided by plugins.

jonashrem:
@mitelg Thanks!

No worries, I was not offended. I just wanted to check the status, as I saw an issue with the html compression and noticed there were already PRs and tickets.

PR #2336, opened 2020-03-23
'Cron Product Export' plugin, export only the unvalidated Cache files

ahmad-saad:
without this change the Export cronjob will create a new Export feed every time I call it.
for example: 
Feed_1 interval : 1 day
Feed_2 interval: 30 minutes
Export cronjob interval: 15 minutes
with the current logic every time I call the cronjob it will ignore the feed interval and create a new feed file.  

after this change the cronjob will recreate only the unvalidated cache files. 

simply create the feeds with the interval in my example and run the cronjob

there are no issues

no need

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Core/CronProductExport/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Core/CronProductExport/Bootstrap.php
@@ -77,8 +77,19 @@ class Shopware_Plugins_Core_CronProductExport_Bootstrap extends Shopware_Compone
         $activeFeeds = $productFeedRepository->getActiveListQuery()->getResult();
         foreach ($activeFeeds as $feedModel) {
             /** @var Shopware\Models\ProductFeed\ProductFeed $feedModel */
+            $fileName = $feedModel->getHash() . '_' . $feedModel->getFileName();
+            $filePath = $cacheDir . $fileName;
             if ($feedModel->getInterval() === 0) {
                 continue;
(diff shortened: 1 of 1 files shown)


shyim:
Please fix the code style issues

engine/Shopware/Plugins/Default/Core/CronProductExport/Bootstrap.php
```diff
@@ -77,8 +77,20 @@ public function exportProductFiles()
         $activeFeeds = $productFeedRepository->getActiveListQuery()->getResult();
         foreach ($activeFeeds as $feedModel) {
             /** @var Shopware\Models\ProductFeed\ProductFeed $feedModel */
+            $fileName = $feedModel->getHash() . '_' . $feedModel->getFileName();
+            $filePath = $cacheDir . $fileName;
+
             if ($feedModel->getInterval() === 0) {
                 continue;
+            } elseif ($feedModel->getInterval() > 0) {
```
shyim, line 85: Why elseif?

ahmad-saad: hallo @shyim  ,

the elseif condition hier is required to handle the 'only cron' option ( interval = -1 ), and by this option the code must always create the feed file by every request.

Regard's
Ahmad

shyim:

mitelg:
merged with b446263ce57ebd869010ff7efdfb8d24b8940f05

PR #2337, opened 2020-03-26
"html compression" use a very slow regex-only approach

Sadrak:
The current implementation is VERY slow for bigger pages. I.e. on our shop only the minify take ~3 seconds on some pages.

It replace the regex-only approach to erase whitespaces with an algorithm that in first place replace every untouchable HTML-tags with placeholders, then replace all whitespaces and then refill the placeholders with the original content.

The change itself did not change anything else. The results from this new algorithm is identical to the previous implementation.

Enable html compression on a bigger shop. Worst results are when the filters are enabled and your products have many attributes, so the page has ~700.000 removable whitespaces in it.

After some benchmarking this new algorithm drop the CPU time for our usecase from ~3seconds to 1.3ms.
Original contentsize was 940.000 bytes with ~710.000 unnecessary whitespaces.

https://forum.shopware.com/discussion/66883/

none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Template/HtmlMinCompressor.php
+++ b/engine/Shopware/Components/Template/HtmlMinCompressor.php
@@ -28,6 +28,43 @@ class HtmlMinCompressor implements HtmlCompressorInterface
 {
     public function minify(string $content): string
     {
-        return preg_replace('#(?ix)(?>[^\S ]\s*|\s{2,})(?=(?:(?:[^<]++|<(?!/?(?:textarea|pre|script)\b))*+)(?:<(?>textarea|pre|script)\b|\z))#', ' ', $content);
+        // List of untouchable HTML-tags.
+        $unchanged = 'script|pre|textarea';
+        // It is assumed that this placeholder could not appear organically in your
+        // output. If it can, you may have an XSS problem.
+        $placeholder = "@@<'-pLaChLdR-'>@@";
+
+        // Some helper variables.
+        $unchangedBlocks  = [];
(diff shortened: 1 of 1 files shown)


Sadrak:
Hi,

i changed the behavior of the method to only remove whitespaces at the beginning and complete empty lines.

The old solution has much more potential to blow something up (like attributes with whitespaces ; ignore the . in the example, github remove also whitespaces ...).
```<option value="this . . . . will . . . . . break">option1</option>```

In my test scenario the difference is minimal (700 more whitespaces left from 710.000 total).

I would prefer this solution.
Bonus: as a human i can still read the content (because many newlines are preserved)

shyim:
Hey @Sadrak,
can you provide us some tests to your regex? Unfortunately we don't have currently time to test it manually.
So we are sure that everything works as expected.

Sadrak:
> Hey @Sadrak,
> can you provide us some tests to your regex? Unfortunately we don't have currently time to test it manually.
> So we are sure that everything works as expected.

Did you have some tests for the old behavior? Did you have a howto for writing tests?

Sadrak:
@shyim this issue still hit customers very bad. I get emails thanking me for my work ...

You should really invest some time here to:
* write own test for current behavior (so i can prove my change work as expected)
* or make a codereview and go on without a test (you did not have a test for the initial slow regex too ...)

shyim:
Hey @Sadrak,

We have long discussed this feature internally and decided to remove the HTML Minifier with Ticket https://issues.shopware.com/issues/SW-26043. Sorry for the late response 😢.

PR #2338, opened 2020-03-27
Change EventSuffix to be unique for each plugin

screeny05:
Make registered jQuery-plugin events unique

Add a GUID to the eventSuffix of jQuery-plugins

1. Have multiple instances of the same plugin on the page
2. Which add event-listeners to the same element (e.g. body)
3. Destroy one instance
4. Both plugins will have their event-listeners removed

Code (simplified):

```html
<html>
<body>
<div data-demo-plugin="true" data-id="1"></div>
<div data-demo-plugin="true" data-id="2"></div>
</body>
</html>
```

```js
$.plugin('demoPlugin', {
  init: function(){
    this._on($('body'), 'click', function(){ console.log('foo'); });
  },
  [...]
});
```

Clicking the body will print `foo` twice.

Somewhere in your code:

```js
StateManager.destroyPlugin('[data-id="1"]', 'demoPlugin');
```

Now clicking the body won't print `foo` anymore, because destroying one plugin caused a call like this:

```js
$('body').off('click.demoPlugin');
```

Which removes both listeners.

-

-

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.plugin-base.js
@@ -77,7 +77,7 @@
         /**
          * @property {string} eventSuffix - Suffix which will be appended to the eventType to get namespaced events
          */
-        me.eventSuffix = '.' + name;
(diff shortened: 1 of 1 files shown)


shyim:

PR #2289, opened 2019-12-12
Fixed fetching of documents positions metadata

hlohaus:
### 1. Why is this change necessary?

See #2279 

### 6. Checklist

- [ ] I have written tests and verified that they fail without my change
- [ ] I have squashed any insignificant commits
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code
- [ ] I have read the contribution requirements and fulfil them.

PR #2339, opened 2020-03-30
Fixing document showing wrong article data

the-habu:
https://forum.shopware.com/discussion/67158/template-problem?new=1
https://forum.shopware.com/discussion/comment/267721/#Comment_267721

As encountered by a few users, using certain article data in templates might lead to a strange bug, where certain articles show wrong information.

Here is what I found out why this happens:
In the end, Document.php would call sGetPromotionById() with an articleordernumber, which in turn will call getOrderNumberByProductId() with this articleordernumber. The existence of the problem can already be seen with just this one sentence, as it tries to get an order number by using a product id, which in reality is already an order number.

Providing the articleID instead of the articleordernumber should work around the problem. In general it does not seem smart, relying on is_numeric to determine which field should be used to query an article. I'd even go as far as calling it bad design, but I might be wrong not having much PHP and/or a lot backend coding experience.

Here is an excerpt of the data that is provided to the template in our case. You can clearly see, that the articleordernumber ends up to be the articleID in meta.

```
Array (
    [id] => 4080
    [orderID] => 1771
    [ordernumber] => 100006937
    [articleID] => 8433
    [articleordernumber] => 0006907
    [price] => 6.28
    [quantity] => 1
    [name] => Bepanthen - Wund- und Heilsalbe
    [status] => 0
    [shipped] => 0
    [shippedgroup] => 0
    [releasedate] =>
    [modus] => 0
    [esdarticle] => 0
    [taxID] => 1
    [tax_rate] => 10
    [config] => [ean] => [unit] => g
    [pack_unit] =>
    [articleDetailID] => 8434
    [articleTaxID] => 1
    [kind] => 1
    [attr1] =>
    ...
    [attr20] =>
    [attributes] => Array (
        [id] => 4080
        [detailID] => 4080
        [attribute1] =>
        ...
        [attribute6] =>
        [swag_promotion_item_discount] => 1.02
        [swag_promotion_direct_item_discount] => 1.02
        [swag_promotion_direct_promotions] => [5]
        [bundle_id] =>
        [bundle_article_ordernumber] =>
        [bundle_package_id] =>
        [pickware_canceled_quantity] => 0 
    )
    [tax] => 10
    [netto] => 5.7090909090909
    [amount_netto] => 5.71
    [amount] => 6.28
    [meta] => Array (
        [articleID] => 6907
        [articleDetailsID] => 6907
        [ordernumber] => 4575909
        [highlight] =>
        [description] => Stillt den Hustenreiz – entspannt den Rachen. Lindert sofort Hustenreiz und Reizrachen 3-fach
reizlindernde Island Formel Wirkt sofort reizlindernd Ideal für Tag und Nacht Ohne Alkohol und Zucker Rein pflanzlich
        ...
    )
```

We created an issue @ shopware but it ended up getting no attention as we had an insufficient support plan. We posted in the forum and found another entry with the same problem. I think there will be more people suffering from this (the error might only be visible to the someone having a really close look at the results) with a respectable dark figure. In our case it is really bad as we are an online pharmacy.

https://forum.shopware.com/discussion/67158/template-problem?new=1
https://forum.shopware.com/discussion/comment/267721/#Comment_267721

None

### 6. Checklist

- [  ] I have written tests and verified that they fail without my change
I have not written any tests. I am developing on windows and have very little PHP development experience. I ran test-unit though and could not find that my changes have a negative impact.
- [  ] This change has comments for package types, values, functions, and non-obvious lines of code

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -495,7 +495,7 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
         $articleModule = Shopware()->Modules()->Articles();
         foreach ($positions as &$position) {
             if ($position['modus'] == 0) {
-                $position['meta'] = $articleModule->sGetPromotionById('fix', 0, $position['articleordernumber']);
+                $position['meta'] = $articleModule->sGetPromotionById('fix', 0, $position['articleID']);
             }
         }
 


engine/Shopware/Components/Document.php
```diff
@@ -495,7 +495,7 @@ protected function assignValues4x()
         $articleModule = Shopware()->Modules()->Articles();
         foreach ($positions as &$position) {
             if ($position['modus'] == 0) {
-                $position['meta'] = $articleModule->sGetPromotionById('fix', 0, $position['articleordernumber']);
+                $position['meta'] = $articleModule->sGetPromotionById('fix', 0, $position['articleID']);
```
mitelg, line 498: did you test this with variants? 
because the orderNumber of a products is related to the variants of the products, but the `articleID` is only related to the main product. so if you change this, you might get data from the main product instead of the specific variant

the-habu: > did you test this with variants?

No I did not. I will ask my colleagues to do so.

> because the orderNumber of a products is related to the variants of the products, but the `articleID` is only related to the main product. so if you change this, you might get data from the main product instead of the specific variant

Good to know!

I could not come up with a better quick work around, everything else is deeper in the chain thus will have more impact I can not forsee.

Do you have ideas for a solution?
Does the PR make the existence of the problem clear?
Why (at least seems like) do all other calls provide an id whereas the Document provides the articleordernumber?

Only thing I can imagine doing better right now, is providing an optional parameter which specifies which kind of ID or number is provided.

Thank you for your time and brains!

shyim:
This has been already fixed in the 5.7 with the Pull Request https://github.com/shopware/shopware/pull/2289.
Can you please verify this?

the-habu:
> This has been already fixed in the 5.7 with the Pull Request #2289.
> Can you please verify this?

I will, when will it become officially available via the update service?

Thing is, we are on overload already, having more customer requests than ever. We currently have no staging environment and no resources to do a verification. If it'll take long for 5.7 to become official, I might get the resources to test it at some point...

But you're right, it seems to address the same problem more exhaustingly, thus better for the product.

engine/Shopware/Components/Document.php
the-habu: OK, so the webshop does not use variants at all and I can't check the impact. I hope 5.7 fixes it and would leave the PR now as is as a reminder, until I can verify.

hlohaus:
It should be fixed. The pull request #2289  has the intention to fix this issue.
Maybe we can copy the order bundle to 5.6? @shopwareBot

keinereiner:
we have still the same problem with (position.meta.attributes.core->get('xxxx')). It reference another product as in position written. we use shopware 5.6.x

hlohaus:
@keinereiner The fix is only the 5.7 branch.

shyim:
As we already said it's already fixed in 5.7 and we are currently not intending to backport those changes.

the-habu:
> As we already said it's already fixed in 5.7 and we are currently not intending to backport those changes.

Sure you did, but you never answered my question when 5.7 will be official... We've been hunting this for quite some time with no help from support.

shyim:
It's currently planned to release 5.7 in autumn this year

PR #2340, opened 2020-04-01
fix CronJob EndDate Parsing 

DerKaito:

The end-date in the Shopware-Crons is not parsed correctly. I my Case 1st April will parsed as the 4th January. That happen in a cron that used the getEnd function and in the command sw:cron:list

It checks the param is a string and give the Zend_Date the parsing format ISO_8601.
The same code is used in the setNext function some lines above.  

This happen in sw:cron:list or when you use the getEnd function in your CronJob.

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Components/Cron/Job.php
+++ b/engine/Library/Enlight/Components/Cron/Job.php
@@ -324,7 +324,7 @@ class Enlight_Components_Cron_Job
     public function setStart($start = null)
     {
         if (!$start instanceof Zend_Date) {
-            $start = new Zend_Date($start);
+            $start = new Zend_Date($start, is_string($start) ? Zend_Date::ISO_8601 : null);
         }
         $this->start = $start;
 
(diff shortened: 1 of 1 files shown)


engine/Library/Enlight/Components/Cron/Job.php
```diff
@@ -344,14 +344,14 @@ public function getEnd()
...
-     * @param null|Zend_Date $end
+     * @param string|null|Zend_Date $end
      *
      * @return Enlight_Components_Cron_Job
      */
     public function setEnd($end = null)
     {
         if ($end !== null && !$end instanceof Zend_Date) {
-            $end = new Zend_Date($end);
+            $end = new Zend_Date($end, is_string($end) ? Zend_Date::ISO_8601 : null);
```
shyim, line 354: Can you fix this also in `setStart`?

DerKaito: sure

shyim:

PR #2341, opened 2020-04-01
Update rss.tpl

ededagic:
Without escape it seems that some characters can cause rss feed break

Without escape some characters ma break the rss feed.

it escapes the sCategoryContentdescription value in the view

use some weird polish characters in the description

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/blog/rss.tpl
+++ b/themes/Frontend/Bare/frontend/blog/rss.tpl
@@ -5,7 +5,7 @@
         <atom:link href="{$sCategoryContent.rssFeed}" rel="self" type="application/rss+xml"/>
         <title>{block name='frontend_atom_title'}{s name="BlogRssFeedHeader"}{$sCategoryContent.description|escape:'hexentity'}{/s}{/block}</title>
         <link>{url controller='index'}</link>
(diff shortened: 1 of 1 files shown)


themes/Frontend/Bare/frontend/blog/rss.tpl
```diff
@@ -5,7 +5,7 @@
         <atom:link href="{$sCategoryContent.rssFeed}" rel="self" type="application/rss+xml"/>
         <title>{block name='frontend_atom_title'}{s name="BlogRssFeedHeader"}{$sCategoryContent.description|escape:'hexentity'}{/s}{/block}</title>
         <link>{url controller='index'}</link>
-        <description>{$sShopname|escapeHtml} - {$sCategoryContent.description}</description>
+        <description>{$sShopname|escapeHtml} - {$sCategoryContent.description|escape}</description>
```
shyim, line 8: Why `escape` here instead of `escapeHtml`

ededagic: No specific reason, escape just worked, escapeHtml not tested.

PaddyS: Could you please change this then and test it? :-)

PR #2342, opened 2020-04-01
API improvement

Algorithman:
When adding products with image(s), also apply a albumId to each (albumId -1 if not provided)

When filling/updating the shop via Article API all images were put into the main album (-1).
Sorting them by hand with 1000's of pictures was taking too long.
And the function internalCreateMediaByFileLink already was accepting an albumId as parameter.

If a albumId is provided in the Article POST (inside the Image object besides the link parameter) the image will be put into the correct album

Add a Article via POST and provide a image with it. For now you can't provide the albumId.

REST-API https://developers.shopware.com/developers-guide/rest-api/examples/article/
Adding a complete example for a new product with an image would be nice.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Article.php
+++ b/engine/Shopware/Components/Api/Resource/Article.php
@@ -2350,9 +2350,17 @@ class Article extends Resource implements BatchInterface
             );
 
             if (isset($imageData['link'])) {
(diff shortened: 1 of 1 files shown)


shyim:

PR #2344, opened 2020-04-12
Zend: Update HTTP version check regex for HTTP/2

myfluxi:
I'm seeing more sites use HTTP/2 now, his causes our Zend
implementation to fail. Use a different regex that makes
the dot and the word (digit) after optional.

Tests: Zuora API calls do not raise an exception. All other sites still work.

Sites that use HTTP/2 protocol raise an exception because a regex check fails.

The regex is updated insofar the dot and following word (digit) is optional.
That way we can use all the versions that exist in the wild (1.1, 1.2, 2) but also 3.

Create a zend client and call a site that uses protocol version 2
(we observed this with Zuora).

-

-

### 6. Checklist

- [ x ] I have written tests and verified that they fail without my change
- [ - ] I have squashed any insignificant commits
- [ - ] This change has comments for package types, values, functions, and non-obvious lines of code

### DIFF
--- a/engine/Library/Zend/Http/Response.php
+++ b/engine/Library/Zend/Http/Response.php
@@ -175,7 +175,7 @@ class Zend_Http_Response
         $this->body = $body;
 
         // Set the HTTP version
-        if (! preg_match('|^\d\.\d$|', $version)) {
+        if (!preg_match('|^\d(\.\d)?$|', $version)) {
             throw new Zend_Http_Exception("Invalid HTTP response version: $version");
         }
 


shyim:
Why are you still using this Http Client? We want maybe deprecate this component in 5.7

myfluxi:
Oh, you are right. We are finally migrating from 5.1.x and have quite a bit of leftover code.

As a proper solution, I quickly moved to a Guzzle implementation.

Then again, it could be fixed here if it is still an issue in 5.6.7 if there is one or 5.7.

shyim:
Thanks! 💙

PR #2346, opened 2020-04-16
Global cronjob events

dejanvujanovic:
It will allow each cronjob to be tracked instead of just an individual one.

The current implementation allows you to subscribe to only one cronjob, by adding a cronjob action at the end of the event. If you want to track more than one or all the cronjobs, you need to subscribe to each cronjob individually through the plugin. With these changes you can subscribe to only one event and thus you can still follow all or just one cronjob.

Subscribe to all cronjobs which are present inside the shop.

No documentation change is needed.

### 6. Checklist

### DIFF
--- a/engine/Library/Enlight/Components/Cron/Manager.php
+++ b/engine/Library/Enlight/Components/Cron/Manager.php
@@ -282,7 +282,7 @@ class Enlight_Components_Cron_Manager
             }
 
             $this->endJob($job);
-            $this->eventManager->notify('Shopware_CronJob_Finished_' . $job->getAction(), [
+            $this->eventManager->notify('Shopware_CronJob_Finished', [
                 'subject' => $this,
                 'job' => $job,
             ]);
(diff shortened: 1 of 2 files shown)


engine/Library/Enlight/Components/Cron/Manager.php
```diff
@@ -282,7 +282,7 @@ public function runJob(Enlight_Components_Cron_Job $job)
             }
 
             $this->endJob($job);
-            $this->eventManager->notify('Shopware_CronJob_Finished_' . $job->getAction(), [
+            $this->eventManager->notify('Shopware_CronJob_Finished', [
```
shyim, line 290: This is an BC Break. Please fire two events

dejanvujanovic: I added them as two separate events.

shyim:

PR #2347, opened 2020-05-04
Merge translations on update

hlohaus:

On product update the translations get lost.

Activate the merge modus on update.

### 6. Checklist


shyim:
Is that not a breaking change?

hlohaus:
> Is that not a breaking change?

It's maybe a bug. So it's maybe okay?

shyim:
Hey,
@hlohaus This change could leak to unwanted behaviour when we apply this in a patch version when someone depends on the old change.

So I would suggest having a parameter like already existing `__options_image` with replace => false. Were we have the same behaviour after this change, but when the user wants to merge it, he could enable it :)

What do you think about that?

PR #2348, opened 2020-05-04
Improve performance a little bit

hlohaus:

Don't use sCheckUser method on detail page,
because it's slow and not really used.

Only check the session state. It is enough.

### 6. Checklist


uehler:
Because the detail page is cached, you could drop it completely, because the logged in state is useless on a cached page

hlohaus:
> Because the detail page is cached, you could drop it completely, because the logged in state is useless on a cached page

It's used in the rating action. Additionally it would be breaking change.

engine/Shopware/Controllers/Frontend/Detail.php
```diff
@@ -75,7 +75,7 @@ public function indexAction()
         $this->View()->assign('sAction', isset($this->View()->sAction) ? $this->View()->sAction : 'index', true);
         $this->View()->assign('sErrorFlag', isset($this->View()->sErrorFlag) ? $this->View()->sErrorFlag : [], true);
         $this->View()->assign('sFormData', isset($this->View()->sFormData) ? $this->View()->sFormData : [], true);
-        $this->View()->assign('userLoggedIn', Shopware()->Modules()->Admin()->sCheckUser());
+        $this->View()->assign('userLoggedIn', !!Shopware()->Session()->offsetGet('sUserId'));
```
shopwareBot: Is there any reason you're not using `(bool)` here?

shyim:

PR #2349, opened 2020-05-05
Notepad fix, when in cookie preferences comfort function is disabled.

Algorithman:
With disabled comfort cookies (sUniqueID) the notepad doesn't work as intended. It creates a new unique Id every time. A fallback to sessionId is more appropriate here.
At the moment, the table s_order_notes is filling with different unique Id's for customers who didn't activate the comfort cookies.

In sAddNote it checks if the comfort cookie sUniqueID is set to true -> behaviour as is, false -> use sessionId as $uniqueId. There might be a better way to access the cookiePreferences, edit as needed.

Disable comfort cookie sUniqueID in cookie preferences and click on the notepad link in the detail page of a article. Do it multiple times, the count on the icon goes up, since every entry has a unique Id. Also to remove all of them, the customer has to delete the same article over and over again for each unique Id. 
Note: in the notepad only one entry is shown, despite there can be multiple of them.

Probably (or partly) https://issues.shopware.com/issues/SW-24776

Probably none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -1467,7 +1467,9 @@ SQL;
         $cookieData = $this->front->Request()->getCookie();
         $uniqueId = $this->front->Request()->getCookie('sUniqueID');
 
-        if (!empty($cookieData) && empty($uniqueId)) {
+        $cookiePreferences = json_decode($cookieData['cookiePreferences']);
+
+        if (!empty($cookieData) && ($cookiePreferences->groups->comfort->cookies->sUniqueID->active===true) &&  empty($uniqueId)) {
             $basePath = Shopware()->Shop()->getBasePath();
             if ($basePath === null || $basePath === '') {
                 $basePath = '/';
@@ -1478,11 +1480,13 @@ SQL;
                     'sUniqueID',
                     $uniqueId,
                     time() + (86400 * 360),
-                    $basePath,
-                    null,
-                    $this->front->Request()->isSecure()
+                    $basePath
                 )
             );
+        }
+        else
+        {
+            $uniqueId = $this->session->get('sessionId');
         }
 
         // Check if this product is already noted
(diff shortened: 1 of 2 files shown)


Algorithman:
Forgot: In sDeleteNote it was the same Problem, only sUniqueID was used to delete the entry. Also added a fallback to sessionId there.

Algorithman:
@klarstil @janbuecker @wesolowski @arnoldstoba @PascalThesing 
Can someone of you please adjust the testsAddNotes in tests/Functional/Core/sBasketTests.php?

Shouldn't take too long, but i don't have the knowledge to add it myself (i don't know the internal functions to set cookies/cookie preferences for example)

PaddyS:
Hey @Algorithman,

thank you for your Pull Request.

I thought about your approach and don't think this is the appropiate solution.
Currently you're working around a somewhat "intended" feature. If the user didn't accept the cookie, the notes **should not work**.

Of course it's bad, that this also results in our database being filled with useless IDs and the customer has no idea why the notes aren't working.

A better solution would be to show some kind of "modal window" including an explanation why this feature isn't working with the given cookie configuration.
Cool would be an additional button to accept the necessary cookie in the modal.

What do you think?

Algorithman:
Probably best not showing the notes link/icons at all if the consent cookie is not set.
At least that's what i'll do, a broken notepad is worse than none (especially since on this project we only have known customers B2B)
Further testing also showed me errors (invalid form token) if one had something in the notepad (with the session id) and then accepted the sUniqueID cookie and then trying to delete it.

I did write it this way, since there already was a fallback to the sessionId, but i see, thats not working.

PaddyS:
Hey @Algorithman,

that's also fine for me, nice idea! :-)
Could you change your PR to simply hiding all icons related to notes if the cookie isn't accepted yet?

Also, then you wouldn't have to deal with the failing PHPUnit tests as well, since there are no PHP changes necessary. Most likely you will have to change this via Javascript.

Could you do that instead? :-)

PR #2353, opened 2020-05-16
HttpClient: Add Guzzle 6.x compatibility

myfluxi:
Also cleanup unused arguments to AccountManagerService.

Practically all packages require Guzzle v6.x as minimum version. Shopware 5 is stuck at v5.3.x

This change add Guzzle v6.x compatibility while preserving compatibility to v5.3.x, thus not putting existing third party plugins' dependencies into danger.

Install any composer package with Guzzle v6.x requirement. Start plugin manager from backend, warm-up cache from backend.

I guess this could be a silent change as it does not change default behaviour.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/DependencyInjection/services.xml
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/DependencyInjection/services.xml
@@ -95,8 +95,6 @@
             <argument type="service" id="shopware_plugininstaller.plugin_installer_struct_hydrator" />
             <argument type="service" id="snippets" />
             <argument type="service" id="models" />
-            <argument type="service" id="guzzle_http_client_factory" />
-            <argument>%shopware.store.apiEndpoint%</argument>
         </service>
 
         <service id="shopware_plugininstaller.plugin_licence_service" class="Shopware\Bundle\PluginInstallerBundle\Service\PluginLicenceService">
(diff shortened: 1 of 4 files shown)


shyim:
That looks for me for a very terrible idea. 
We require explict the Guzzle 5 version. So the core should also use only features of that and should not try to feature detect the right version.

When plugins bring Guzzle 6 or newer, it's wrong.  You bring the Core in a unknown situation, you cannot know which version is active. Does any other library collide with that? When you need that for a Library you should search for an another library which makes use of PSR18 where such problems do not exists.
Also, a solution could be (when it's not possible to use the right way (**PSR-18**)  you should maybe try to handle that with prefixes [https://github.com/humbug/php-scoper].

BTW: Guzzle will be updated in 5.7 (see branch)

PS: You didn't accepted the CLA

myfluxi:
> BTW: Guzzle will be updated in 5.7 (see branch)

"guzzlehttp/guzzle": "6.4.1",

I didn't notice this, excellent!

myfluxi:
Just asking: setDefaultOption() is not available in Guzzle v6:

https://github.com/shopware/shopware/blob/5.7/engine/Shopware/Components/HttpClient/GuzzleFactory.php#L41

shyim:
Good catch. That file exists only in packages 🙈

PR #2354, opened 2020-05-18
Fix SQL Statement

ghost:

Cause it is not possible to use multiple theme/shop/documents/index.tpl with the same language

It change a SQL statement. Instead of comparing a shop language it compare the id

Create third shop (German, English, Swiss)
Create second theme (German, Swiss)
Create theme/swiss/shop/documents/index.tpl

https://issues.shopware.com/issues/SW-25438

none

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Document.php
+++ b/engine/Shopware/Components/Document.php
@@ -201,7 +201,7 @@ class Shopware_Components_Document extends Enlight_Class implements Enlight_Hook
                  LEFT JOIN s_core_shops m
                      ON m.id=s.main_id
                      OR (s.main_id IS NULL AND m.id=s.id)
-                 WHERE s_order.language = s.id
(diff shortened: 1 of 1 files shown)


shyim:
Can you please accept the cla?

shyim:
Hey @dennisottenbacher,

I did tested your change and the result looks wrong now to me.

I built a Shop structure

* Normal Shop (German)
* Second Shop (German)
 * English Language Shop
And did a order in the English language shop. The Query selects me the templates for the Subshop, thats okay, cause language shops can't have a own template. 

Your change looks very similar, **but** locale and isocode are now from subshop instead from language shop.

shyim:
We have discussed it in a Video Call. The issue was an importer. So we don't need to change here something :blush:

PR #2357, opened 2020-05-28
Remove second non integer exception parameter

revoltek-daniel:
The second parameter of an exception should be an integer, if this exception is thrown the $cmd variable contains an non numeric string and leads to an php error:
```Wrong parameters for Zend_Mail_Protocol_Exception([string $message [, long $code [, Throwable $previous = NULL]]])```

removed the second code parameter and pass the $cmd variable in the exception message

return code of the mail server did not match the expected codes

### 6. Checklist


shyim:
Can you please accept the cla? 😊

shyim:

PR #2362, opened 2020-07-06
Added setter for languageId

eeromay:
Setter for languageId

Setter for languageId is missing in Api.
In the documentation is languageId set via post, but has no effect.

sets languageId via api

post customer languageId to shop with different id than 1

No documentation changes needed, the documentation is correct but the endpoint is incorrect.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Customer/Customer.php
+++ b/engine/Shopware/Models/Customer/Customer.php
@@ -1182,6 +1182,14 @@ class Customer extends LazyFetchModelEntity
     }
 
     /**
+     * @param int $languageId
+     */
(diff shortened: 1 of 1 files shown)


engine/Shopware/Models/Customer/Customer.php
```diff
@@ -1181,6 +1181,14 @@ public function getLanguageSubShop()
         return $this->languageSubShop;
     }
 
+    /**
+     * @param int $languageId
```
shyim: add typehint and remove phpdoc

shyim:

PR #2363, opened 2020-07-06
Update jquery.image-slider.js

sergejreznik:
Prevent mobile browsers from throwing errors like `Unable to preventDefault inside passive event listener invocation.` and calling the event twice.

Prevent mobile browsers from throwing errors like `Unable to preventDefault inside passive event listener invocation.` and calling the event twice.

It changes the eventlistener from touchstart to touchend.

On some devices and projects it's calling the image slider twice when clicking the arrow on mobile devices. To prevent this behaviour it's necessary to call `touchend` and not `touchstart` 

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.image-slider.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.image-slider.js
@@ -635,8 +635,8 @@
             }
 
             if (opts.arrowControls) {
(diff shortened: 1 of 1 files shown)


shyim:
Hey,
I am unable to reproduce it.
It setup a demo shop here: http://shyim.shopwaredemo.de/cat/index/sCategory/67
When I use the slider on iOS or Android. I don't get any errors in dev console

PR #2365, opened 2020-07-07
Add parentheses to purchasesteps-condition

jkrzefski:
Currently the integer cast only targets the `$quantity` variable instead of the result of the calculation `$quantity / $additionalInfo['purchasesteps']`. This is clearly not the intention of this code as quantities are always integer.

The integer cast not targets the calculations result. This makes the thing work.

Look at the code. It wouldn't make sense this way.

None.

None.

### 6. Checklist


shyim:

PR #2367, opened 2020-08-09
Set phone to model when creating fromAddress()

MrTschi:
Using the plugin "Shopware Backend Order", the billing address phone number will not get copied to shipping address.

Set phone number from a given address object into the shipping model

- Install Backend Order Plugin
- Place backend order and tick "use billing address" for the shipping address. Ensure, that the customer has a phone number within his billing address.

-

none

### 6. Checklist

- [model has no tests] I have written tests and verified that they fail without my change
- [yes] I have squashed any insignificant commits
- [?] This change has comments for package types, values, functions, and non-obvious lines of code
- [yes] I have read the contribution requirements and fulfil them.


shyim:

PR #2371, opened 2020-09-01
Update product.tpl - add block inside of form-tag

choeft:
As a plugin manufacturer, I want to be able to add additional inputs within the form. Right now, this can only be done by replacing the block "frontend_checkout_cart_item_delete_article" completely, which may lead to incompatibilities with other plugins. A new block would be of much help.

As a plugin manufacturer, I want to be able to add additional inputs within the form. Right now, this can only be done by replacing the block "frontend_checkout_cart_item_delete_article" completely, which may lead to incompatibilities with other plugins. A new block would be of much help.

Adds a new block inside the form tag.

n/a

n/a

n/a

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/checkout/items/product.tpl
+++ b/themes/Frontend/Bare/frontend/checkout/items/product.tpl
@@ -176,6 +176,8 @@
                         title="{$snippetCartItemLinkDelete|escape}">
                     <i class="icon--cross"></i>
                 </button>
+                {block name='frontend_checkout_cart_item_delete_article_form_inner'}        
+                {/block}
             </form>
         </div>
     {/block}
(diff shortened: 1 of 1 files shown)


themes/Frontend/Bare/frontend/checkout/items/product.tpl
```diff
@@ -176,6 +176,8 @@
                         title="{$snippetCartItemLinkDelete|escape}">
                     <i class="icon--cross"></i>
                 </button>
+                {block name='frontend_checkout_cart_item_delete_article_form_inner'}
```
aragon999: Why don't you create a block around the button, or the complete form content but rather create an empty block?

choeft: Hi Max,

thanks for your feedback. You are right, creating it around the button makes more sense. I changed it.

shyim:

PR #2375, opened 2020-09-14
Add attributes to user api list action

larsbo:
# 1. Why is this change necessary?
Like other API resources the user API list action should display the attributes of the users.

It adds a left join to the users attributes table.

1. Open `/api/users`.
2. Look for an `attribute` field at the entries.
3. Cry because its missing 😢 

-

-

### 6. Checklist


shyim:

PR #2379, opened 2020-09-24
fix cookie removal for www domains

mkreusch:
## 1. Why is this change necessary?
Cookies that are set on .domain.tld currently do not get removed by the cookie removal service, if the shop runs on www.domain.tld
Since major cookies by tracking services (e.g. Google Analytics, Hotjar, ...) use cookies on .domain.tld the user's cookie preferences are bypassed. **This is currently causing serious privacy issues!**

It adds a www-stripped domain to the overwriting cookie, thereby making sure that both www and non-www domains are addressed. (https://stackoverflow.com/questions/1062963/how-do-browser-cookie-domains-work)

1. The shop has to run on a www domain and the Shopware cookie settings shall be set to either not allow any cookies before giving consent or only allow technically required cookies
2. Manually set a cookie on .yourdomain.tld
3. Open any shop page without accepting cookies
4. Notice that the cookie does not get removed

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/CookieBundle/Services/CookieRemoveHandler.php
+++ b/engine/Shopware/Bundle/CookieBundle/Services/CookieRemoveHandler.php
@@ -91,13 +91,14 @@ class CookieRemoveHandler extends CookieHandler implements CookieRemoveHandlerIn
 
         foreach ($requestCookies as $cookieKey => $cookieName) {
             if (!$validationFunction($cookieKey)) {
-                $response->headers->setCookie(new Cookie($cookieKey, null, 0));
-                $response->headers->setCookie(new Cookie($cookieKey, null, 0, $cookieBasePath));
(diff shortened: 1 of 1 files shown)


shyim:
This needs to be done in both domains. With www and without. Otherwise the Shopware Cookies itself are not removed anymore :thinking:

mkreusch:
@shyim 
Thanks for your comment!
Obviously you are right with that. I screwed it up in my test so it seemed to work for both hosts. I added the generic host=null to it to not interfere with the original functionality but only add the new stripped host.

shyim:
The CI is failing

mkreusch:
Do have any idea why?

shyim:
https://github.com/shopware/shopware/pull/2379/checks?check_run_id=1160554921#step:6:22 fix your code style

mkreusch:
Sorry - there was a space missing - after all that was probably more effort than fixing it by yourself. Sorry for that!

PR #2381, opened 2020-10-01
Added filter event to allow file name change

srusche:
It is currently not possible to customize PDF document download file names.
Shopware users with heavy use of those documents are having problems to organize locally downloaded files.
They have to rename files manually to add information like company, customer name or date.
By adding a filter event, it is possible to create an automatic solution with a plugin.

Instead of just applying the document ID as file name, this value is now emitted in a filter event. Therefore it is possible to return a custom file name when subscribing to this event. Additional information to be used in the custom file name can be taken form the document model, which is available as 'data' argument inside the event.

### 6. Checklist

### DIFF
--- a/UPGRADE-5.7.md
+++ b/UPGRADE-5.7.md
@@ -36,6 +36,7 @@ This changelog references changes done in Shopware 5.7 patch versions.
 * Added support for MySQL 8 `sql_require_primary_key`
 * Added `attribute` to users listing in API
 * Added new blocks `document_index_head_logo` and `document_index_head_wrapper` to `themes/Frontend/Bare/documents/index.tpl`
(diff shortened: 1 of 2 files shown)


shyim:

srusche:
Hey shyim! I just updated some shops to 5.7.3 and wonder why my Shopware_Controllers_Order_OpenPdf_FilterName event is missing, as it was announced to be added in the 5.7.0 changelog (three months ago)? 🤷

mitelg:
hey @srusche 

sorry for the inconvenience :slightly_frowning_face: Somehow the change was not merged back, while it was temporary on a development branch. 
But I managed to bring it back and it will be part of the next release 5.7.4

PR #2382, opened 2020-10-01
Wrap mail template preview in iframe

jkrzefski:

This helps separating the styles of the mail from the styles of the backend. Without it, the styles from the backend would also take effect in the mail preview.

Instead of including the rendered mail body in the dom directly, this generates an iframe with a `srcdoc` attribute that holds the mail body.

1. Open a mail template.
2. Click the preview button.

- [Html code as IFRAME source rather than a URL - Stack Overflow](https://stackoverflow.com/questions/6102636/html-code-as-iframe-source-rather-than-a-url/6102829#6102829)
- [html - HtmlSpecialChars equivalent in Javascript? - Stack Overflow](https://stackoverflow.com/questions/1787322/htmlspecialchars-equivalent-in-javascript/4835406#4835406)

None.

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/mail/controller/main.js
+++ b/themes/Backend/ExtJs/backend/mail/controller/main.js
@@ -383,11 +383,21 @@ Ext.define('Shopware.apps.Mail.controller.Main', {
                 var status = Ext.decode(response.responseText);
                 if (status.success) {
 
-                    if (isHtml) {
-                        htmlString =  "<div style=\"margin: 15px\">" + status.message + "</div>"
-                    } else {
-                        htmlString =  "<div style=\"margin: 15px\"><pre>" + status.message + "</pre></div>"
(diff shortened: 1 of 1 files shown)


JoshuaBehrens:
This explains a lot 🙃

hlohaus:
cool, like https://github.com/FriendsOfShopware/FroshMailArchive/pull/1/files

maybe you can add sandbox too? :)

jkrzefski:
@hlohaus Good catch. I added the sandbox attribute.

shyim:
Hey @jkrzefski ,

The window looks weird in Backend. Can you fix the double scrollbar?

image

JoshuaBehrens:
Double scrollbar just every OS things but Mac default 🙊

jkrzefski:
I'll look into it.

PR #2386, opened 2020-11-02
Fix for wrong listing of shop name in the shipping

rohithmeethal:
Currently shipping table lists wrong shop name, if there is another shop with id that has same starting pattern.
eg 1,11,  2,25, etc 

This fix adds whole string matching restriction. So 1 will match only 1 and 10 will match only 10.

Add shops with same pattern id's like 1,11,12,123 etc assign shipping method to each shop.

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/shipping/view/main/list.js
+++ b/themes/Backend/ExtJs/backend/shipping/view/main/list.js
@@ -244,7 +244,7 @@ Ext.define('Shopware.apps.Shipping.view.main.List', {
      */
     multishopIdColumn : function(value, obj, record) {
         var me = this,
-            shop = me.shopStore.findRecord('id', record.get('multiShopId'));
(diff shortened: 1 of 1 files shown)


shyim:
Can we not use `findExact`?
https://docs.sencha.com/extjs/4.1.0/#!/api/Ext.data.Store-method-findExact

rohithmeethal:
@shyim not that familiar with extjs, but the findExact return the index and then we have to again fetch the shop from the store.
If it is due to lot of parameters we could modify it this way I suppose.
```php
    multishopIdColumn : function(value, obj, record) {
        // if we have an empty or a zero value show default empty text
        if (null === record.get('multiShopId') || 0 === record.get('multiShopId')) {
            return this.emptySubshopText;
        }
        let shop = this.shopStore.find({
            property:'id',
            value: record.get('multiShopId'),
            exactMatch:true
        });
        return shop.get('name');
    },
```

PR #2388, opened 2020-11-05
Fixed Newsletter Double-opt-in bug when comparing lowercase "nocaptcha" with camelcase "noCaptcha"

yannickpschroeder:
Bug: $config->get('newsletterCaptcha') returns 'nocaptcha' and never matches 'noCaptcha'.
Solution: Make comparism case insensitive. 

indexAction in Shopware_Controllers_Frontend_Newsletter never reaches the Double-Optin procedure because
lowercase 'nocaptcha'  doesn't match camelcase 'noCaptcha' and just returns.

Makes old comparism  `$config->get('newsletterCaptcha') !== "noCaptcha"` case insensitive.

Bug happened with a client. It might happen to more people without knowing why and it won't be a problem to make this comparison case insensitive.


shyim:

PR #2389, opened 2020-11-11
NTR - fix price display in net mode

teiling88:
If you have a customer group with display net prices in the storefront the checkout confirm page will look like the following: 
image
the amount of 6,40 € is the vat value not the net price of the article. With the change this display issue is fixed: 
image

### 6. Checklist


shyim:
Hey @teiling88,

as discussed personally with you this approach is wrong, so we decided to close this pull request.

PR #2390, opened 2020-11-12
append DIRECTORY_SEPARATOR to the registerNamespace path, in the HttpCache plugin

breaker92:
If the 'ShopwarePlugins\HttpCache\CacheControl' is used the followeing error occured:
**PHP Fatal error:  Class 'ShopwarePlugins\\HttpCache\\CacheControl' not found** 
The reason is a missing DIRECTORY_SEPARATOR on the end of the registerNamespace path, given in the HttpCache plugin.

add a DIRECTORY_SEPARATOR on the end of path parameter, given to the enlight loader registerNamespace method, in the HttpCache plugin.

The CacheControl class seams to be used only if ESI-Tags are used/supported.
If the shopware is hosted behind a reverse proxy, this error may occure

nothing

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Core/HttpCache/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Core/HttpCache/Bootstrap.php
@@ -180,7 +180,7 @@ class Shopware_Plugins_Core_HttpCache_Bootstrap extends Shopware_Components_Plug
 
     public function afterInit()
     {
-        $this->get('loader')->registerNamespace('ShopwarePlugins\\HttpCache', __DIR__);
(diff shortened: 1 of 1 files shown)


shyim:
I am pretty sure you have used composer 2 on a not supported Shopware 5 version

breaker92:
Yes, I used composer 2, but if look at the `Enlight_Loader::getClassPath` method:
Without the appended `DIRECTORY_SEPARATOR `the lookup leads to **engine/Shopware/Plugins/Default/Core/HttpCacheCacheControl.php**.

Every other call of the `Enlight_Loader::registerNamespace`, which I have found so far, has a closing `DIRECTORY_SEPARATOR `on the path.


P.S.: @shyim I'm impressed by your response time 👍

shyim:
But this should be fixed on newest version 
Try again in newest Shopware version

PR #2392, opened 2020-11-20
Update OrmBacklogSubscriber.php

christopherholst:
fixed workaround

### 6. Checklist


engine/Shopware/Bundle/EsBackendBundle/Subscriber/OrmBacklogSubscriber.php
```diff
@@ -76,6 +76,42 @@ public function onFlush(OnFlushEventArgs $eventArgs)
         }
     }
 
+    private function trace(OnFlushEventArgs $eventArgs): void
```
mitelg, line 79: why did you move the method? this makes it hard to find out the actual change. please move the method back

PR #2394, opened 2020-12-02
Default behaviour for redirect-after-login looks like checkout

JoshuaBehrens:
When you want to go somewhere else after logging in - using the sTarget way of forwarding - it'll look like you are in the checkout. No!

Change the checking behaviour to check for a specific checkout target to get checkout content instead of using a condition that explicitly states not toAccount therefore we are in checkout.

1. Redirect to the homepage of the shop after login like this: https://www.shopwaredemo.eu/account?sTarget=index
2. Be visually in a checkout
3. (・_・;)
4. no `steps--container`, what are you doing? Please let me go homepage. Not there. Don't do the checkout
5. grafik
6. Be stuck in checkout

### 4. Checklist


JoshuaBehrens:
I chose 5.8 as this is quite breaking.

shyim:
We decided to pick it for 5.7. I will rebase it internally

JoshuaBehrens:
Woohoo 🎉

shyim:
Hey,

I have checked the MR and the register page looks weird. Your PR indicates that the steps should be only shown when the target is checkout but other changes also happend.

The screenshots are from /register page without parameters


## Branch 5.7

image


## Your branch

image

Can you look into it 😄 ?

PR #2396, opened 2020-12-03
fixed x-content-type-options policy with "nosniff" refreshstatics

christopherholst:
on every site in the frontend comes the mime type error text/html for refreshstatics url ->
<script async="" id="refresh-statistic" src="//www.mystore.de/widgets/index/refreshStatistic?requestPage=/store/&amp;requestController=listing&sCategoryId=410"></script>

fixed the error for chrome and edge

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Widgets/Index.php
+++ b/engine/Shopware/Controllers/Widgets/Index.php
@@ -47,6 +47,9 @@ class Shopware_Controllers_Widgets_Index extends Enlight_Controller_Action
         /** @var Shopware_Plugins_Frontend_Statistics_Bootstrap $plugin */
         $plugin = Shopware()->Plugins()->Frontend()->Statistics();
         $plugin->updateLog($request, $response);
+        
+        /** Setting mime type for jsonp request */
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Widgets/Index.php
```diff
@@ -47,6 +47,9 @@ public function refreshStatisticAction()
         /** @var Shopware_Plugins_Frontend_Statistics_Bootstrap $plugin */
         $plugin = Shopware()->Plugins()->Frontend()->Statistics();
         $plugin->updateLog($request, $response);
+        
+        /** Setting mime type for jsonp request */
```
shyim, line 51:
```suggestion
        /** Setting mime type for json request */
```
It's not jsonp

shyim:
Hey @christopherholst ,

We cannot reproduce the error in Edge, Chrome and Firefox. Can you check is that still needed?

christopherholst:
@shyim 
No, I'll close it again because the real bug was in the enterprise search plugin.

PR #2399, opened 2020-12-05
Fixes various formatting problems of numerical values in the sales widget

mayerni:
Fixes various formatting problems of numerical values in the dashboard widget "Sales Yesterday and Today"

The output in the dashboard widget "Sales Yesterday and Today" has several minor formatting and display problems. No value (empty) is displayed in the "Visitors" column if there is (yet) no entry for the day in the database. Furthermore, the output of values in the entire widget is not formatted. The "conversion rate" is always with a point as a decimal separator, although it would have to be a comma in the german backend. The other values (columns: "sales", "orders", "new customers" and "visitors") are also not properly formatted as numerical values. This is particularly unpleasant in the sales column, which, depending on the sales value, shows sometimes no, sometimes one or sometimes two decimal places.

The changes in the controller relate to the uniform return of the transferred values as numerical values (int & float). The changes to the Javascript widget aim at the correct output of the transmitted values. For this purpose, these are formatted with the appropriate ExtJs functions (which format according to the language environment set).

Enable and take a look at the "Sales Yesterday and Today" dashboard widget. All numerical values are displayed unformatted. This should be noticed at all times, regardless of whether there are values or not. Especially when it comes to conversion rate (no locale formatting) and revenue (no solid currency accuracy). If there were no visitors (or if you turn off visitor statistics), the corresponding column is simply empty.

I found a ticket with a matching case: "SW-23263 | Zahlendarstellung im Backend im Umsatz-Widget". My changes capture exactly the problem described (and a little more).

I have make a phpDoc for the function i added. But I haven't found any part of the documentation that needs to be adjusted.

### 6. Checklist

### 7. Remarks
- I have no idea what "I have squashed any insignificant commits" means.
- I didn't write a test for my changes, but I tested in my shops and a clean shopware installation, with no problems.
- I have not made an entry in upgrade.md.
- I tried to follow the coding guidelines.


themes/Backend/ExtJs/backend/index/view/widgets/sales.js
```diff
@@ -341,37 +351,43 @@ Ext.define('Shopware.apps.Index.view.widgets.Sales', {
...
-            },
-            {
-                header: me.snippets.headers.new_customers,
-                align: 'right',
-                dataIndex: 'newCustomers',
-                flex: 1
-            },
-            {
-                header: me.snippets.headers.visitors,
+        return {
```
shyim: This should only return the columns. Otherwise we break plugins which could add new columns

mayerni: Thanks for pointing this out, I didn't think about it that way. I have fixed it.

shyim:

PR #2400, opened 2020-12-07
Fix session id when switching between sessions

felixbrucker:
Currently with SW 5.7 when switching between sessions (Frontend/Backend) the session id is not updated with the correct session id for the respective session based on the cookie values. With SW <= 5.6 Zend took care of this: https://github.com/shopware/shopware/blob/cbc212ca4642878cac62193d3a2f41e08f4849a2/engine/Library/Zend/Session.php#L421-L423
Additionally any active session was closed before starting a new one with Zend: https://github.com/shopware/shopware/blob/cbc212ca4642878cac62193d3a2f41e08f4849a2/engine/Shopware/Components/DependencyInjection/Bridge/Session.php#L77-L79.

This change replicates the Zend logic to set the session id to the session cookies value if there is none set currently. As such when creating e.g. a Frontend session after using a Backend session the session id is set to the Frontend Session cookie value and thus correct, instead of still being the Backend session id. The backend session is also saved and closed before starting the frontend session to ensure clean switches between each other. The same is true vice versa.

1) Call a Backend Controller Action with a Frontend and Backend session cookie set (`SHOPWAREBACKEND` and `session-1` for a shop with id 1)
2) Close the Backend session:
```php
$backendSession = $container->get('backendsession');
$backendSession->save();
$backendSession->getId();
```
3) Register a Shop and request the Frontend session:
```php
$shop->registerResources();
$session = $container->get('session');
$session->getId();
```
4) Observe the session id from the Frontend session is the same as the session id from the Backend session.

### 4. Checklist

### DIFF
--- a/engine/Library/Enlight/Components/Session/Namespace.php
+++ b/engine/Library/Enlight/Components/Session/Namespace.php
@@ -140,4 +140,58 @@ class Enlight_Components_Session_Namespace extends Session implements ArrayAcces
         parent::clear();
         $this->set('sessionId', $this->getId());
     }
+
+    public function start()
+    {
+        // Generally `start()` is called by Shopware's session factories
+        // - Shopware_Plugins_Backend_Auth_Bootstrap::onInitResourceBackendSession()
+        // - Shopware\Components\DependencyInjection\Bridge\Session::createSession()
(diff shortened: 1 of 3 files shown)


shyim:
Thanks for your contribution! 💙

PR #2401, opened 2020-12-12
Fix PHPStan error in Acl component

myfluxi:
We must add integer to the hinted types of optional $pluginID arg.
Else PHPstan level 5 check fails with:

------ --------------------------------------------------------------------
  Line   Bootstrap/Acl.php
 ------ --------------------------------------------------------------------
  53     Parameter #4 $pluginID of method
         Shopware_Components_Acl::createResource() expects null, int given.
 ------ --------------------------------------------------------------------

PHPStan level 5 check in our plugin fails.

We just add int to the hinted types of the optional $pluginID argument to createResource().

Use createResource(), run a PHPStan check level 5 to trigger an error like:
------ --------------------------------------------------------------------
  Line   Bootstrap/Acl.php
 ------ --------------------------------------------------------------------
  53     Parameter #4 $pluginID of method
         Shopware_Components_Acl::createResource() expects null, int given.
 ------ --------------------------------------------------------------------

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Acl.php
+++ b/engine/Shopware/Components/Acl.php
@@ -73,9 +73,9 @@ class Shopware_Components_Acl extends Zend_Acl
         $repository = $this->em->getRepository(Role::class);
         $roles = $repository->findAll();
 
-        /** @var \Shopware\Models\User\Role $role */
+        /** @var Role $role */
         foreach ($roles as $role) {
-            /** @var \Shopware\Models\User\Role|null $parent */
+            /** @var Role|null $parent */
             $parent = $role->getParent();
 
             // Parent exist and not already added?
@@ -104,7 +104,7 @@ class Shopware_Components_Acl extends Zend_Acl
     {
         $rules = $this->em->getRepository(Rule::class)->findAll();
 
-        /** @var \Shopware\Models\User\Rule $rule */
+        /** @var Rule $rule */
         foreach ($rules as $rule) {
             $role = $rule->getRole();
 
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/Acl.php
```diff
@@ -144,12 +144,14 @@ public function hasResourceInDatabase($resourceName)
...
-     *
-     * @throws Enlight_Exception
+     * @param int|null   $pluginID     - optionally pluginID that implements this resource
      */
-    public function createResource($resourceName, array $privileges = null, $menuItemName = null, $pluginID = null)
-    {
+    public function createResource(
+        $resourceName,
+        array $privileges = null,
+        /* @noinspection PhpUnusedParameterInspection */ $menuItemName = null,
```
mitelg, line 152:
```suggestion
        $menuItemName = null,
```

engine/Shopware/Components/Acl.php
```diff
@@ -144,12 +144,14 @@ public function hasResourceInDatabase($resourceName)
...
+        $pluginID = null
+    ){
```
mitelg, line 154:
```suggestion
    ) {
```

engine/Shopware/Components/Acl.php
```diff
@@ -144,12 +144,14 @@ public function hasResourceInDatabase($resourceName)
      * @param string     $resourceName - unique identifier or resource key
      * @param array|null $privileges   - optionally array [a,b,c] of new privileges
      * @param null       $menuItemName - optionally s_core_menu.name item to link to this resource
```
mitelg, line 146:
```suggestion
      * @param string|null       $menuItemName - optionally s_core_menu.name item to link to this resource
```

PR #2402, opened 2020-12-15
use major version option (--1) for composer self-update

breaker92:
use major version option (--1) for composer self-update, instead of explicit using version 1, because version 1 doesn't exists.

because `composer self-update` in **ant** failed with:
`The "https://getcomposer.org/download/1/composer.phar.sig" file could not be downloaded (HTTP/1.1 404 Not Found)  
`

change argument for **ant** target **update-composer-binary**

run **ant** target **update-composer-binary**

none

nothing

### 6. Checklist

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -90,7 +90,7 @@
         <exec executable="${script.php}">
             <arg value="composer.phar"/>
             <arg value="self-update"/>
-            <arg value="1"/>
+            <arg value="--1"/>
             <arg value="--no-interaction"/>
         </exec>
     </target>


shyim:
I would like to remove that file instead of changing there something. We switched to a simple Makefile in 5.7 do not require java anymore

breaker92:
As long as it is not removed I would fix bugs.

shyim:

We have merged it into the 5.6 branch. We will remove ant configuration completly on 5.7, please use there the Makefile

PR #2406, opened 2021-01-06
Remove superfluous # from woff2 icon font path to fix font preloading

larsbo:

The `#` prevents the preloading of the font resource.

Remove a superfluous hash character.

1. Add a preload hint for the font file `shopware.woff2`:
`<link rel="preload" as="font" type="font/woff2" crossorigin href="/themes/Frontend/Responsive/frontend/_public/src/fonts/shopware.woff2" />`
2. Validate preloading, e.g. with Google Pagespeed (https://developers.google.com/speed/pagespeed/insights/).
3. The analysis shows that the preload doesn't work because the url inside the css file doesn't match the one from the preload hint. For all the other font files (without the # inside the path) preloading works as expected.

### 6. Checklist


shyim:

PR #2408, opened 2021-01-15
README: note the filter extension is required

avdv:

Symfony uses the filter_var function.

Add a note the README, and require the `ext-filter` extension in the composer file.

You use PHP without enabling the filter extension. Then, start the builtin server (for example) and request a page. You see 500 error in your browser.

### 6. Checklist


shyim:

PR #2410, opened 2021-01-27
Fix filter empty Orders

jalbersdorfer:
If there are open Orders without Details (Orderlines) - unusual, right - then the Open Orders Overview in Customer Frontend breaks.

Fix filtering out Orders without Details (Orderlines) by delaying the `unset` to after iterating the array.

Produce an Order without any Details (Orderlines) and have a look in the Open Customer Orders Frontend Listing.

n.a.

n.a.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sAdmin.php
+++ b/engine/Shopware/Core/sAdmin.php
@@ -1345,6 +1345,7 @@ class sAdmin implements \Enlight_Hook
         );
         $foundOrdersCount = (int) Shopware()->Db()->fetchOne('SELECT FOUND_ROWS()');
 
+        $orderKeysToRemove = [];
         foreach ($orders as $orderKey => $orderValue) {
             $orders[$orderKey]['invoice_amount'] = $this->moduleManager->Articles()
                 ->sFormatPrice($orderValue['invoice_amount']);
@@ -1353,8 +1354,17 @@ class sAdmin implements \Enlight_Hook
             $orders[$orderKey]['invoice_shipping'] = $this->moduleManager->Articles()
                 ->sFormatPrice($orderValue['invoice_shipping']);
 
-            $orders = $this->processOpenOrderDetails($orderValue, $orders, $orderKey);
+            $ret = $this->processOpenOrderDetails($orderValue, $orders, $orderKey);
+            if ($ret) {
+                $orders = $ret;
+            } else {
+                $orderKeysToRemove[] = $orderKey;
+            }
             $orders[$orderKey]['dispatch'] = $this->sGetPremiumDispatch($orderValue['dispatchID']);
+        }
+
+        foreach ($orderKeysToRemove as $orderKeyToRemove) {
+            unset($orders[$orderKeyToRemove]);
         }
 
         $orders = $this->eventManager->filter(
@@ -3669,7 +3679,7 @@ SQL;
      *
      * @param string $orderKey
      *
-     * @return array
+     * @return array|false
      */
     private function processOpenOrderDetails(array $orderValue, array $orders, $orderKey)
     {
@@ -3680,9 +3690,7 @@ SQL;
         );
 
         if (!count($orderDetails)) {
-            unset($orders[$orderKey]);
-
-            return $orders;
+            return false;
         }
 
         $context = $this->contextService->getShopContext();


mitelg:
hey @jalbersdorfer 
thanks for your PR :slightly_smiling_face: :+1: 

why are there orders without details in the first place? Afaik, this should not be possible in a default installation. Shouldn't it prevented, that orders could be created without details?

I think, your PR fixes only symptoms, but not the cause

jalbersdorfer:
Hi @mitelg
thanks for your quick response.

And es you are right, it should not happen in a default installation.
I observed the behavior in regards to a plugin I'm developing.

The core functionality of the plugin is to combine orders of the same customer within a certain period of time into a single order.
Therefore, I _move_ the Order Details from one to another Order. The original Order Heads get set into a _combined_ state,
so they will not be part any further processing.

The customer on the other Hand has received Order Confirmation E-Mails for the original Orders, and therefore I want them be displayed in the Customer Orders Overview with its correct states and also clearly stated that the original Orders have been merged to another Order.

So that's the root cause is myself. On the other hand, even if it might by default never happen, it is clearly a implementation fault which should be fixed.

I even wanted to first fix this and then talk about changing the behavior to not to remove Orders without details from the List. Which should not be a big thing if it nevertheless should never happen in a default environment.

mitelg:
thanks for the explanation :+1: 

But if you are the reason behind this misbehaviour by yourself, shouldn't you fix it on your site in your plugin? :slightly_smiling_face: 
I'm okay with adding some events or other extensions points, so you can take care of that problem in your plugin properly. what do you think about that?

jalbersdorfer:
In my opinion, I'm not the reason for the misbehavior (bug), I just found it and want to fix it - not to work around it.
The extension Point is already there (`Shopware_Modules_Admin_GetOpenOrderData_FilterResult`), but the array passed to this is broken on receive because of the bug I try to fix with this PR. I'm fully aware that what I do is not a default workflow, and I want to achieve my goal with as less impact to the shopware core as possible. But please understand that this is clearly a bug. Even if it is not me causing the function to run that path, it would break nevertheless. It is a bug. It is fully reproduceable by unit test. It should be fixed. While on the other hand, the implementation details of my plugin are open for any discussion.

mitelg:
hey @jalbersdorfer 

thanks again :slightly_smiling_face: :+1: 

I guess, we can go with your changes. But at least removing of the items should not happen silently. What do you think about some logging here?

I will also make some suggestions regarding your code :wink:

engine/Shopware/Core/sAdmin.php
```diff
@@ -3680,9 +3690,7 @@ private function processOpenOrderDetails(array $orderValue, array $orders, $orde
         );
 
         if (!count($orderDetails)) {
-            unset($orders[$orderKey]);
-
-            return $orders;
+            return false;
```
mitelg, line 3693: return `null` here, so we can add a proper return type hint to the method

engine/Shopware/Core/sAdmin.php
```diff
@@ -3669,7 +3679,7 @@ private function resendConfirmationMail(array $userInfo, $hash)
      *
      * @param string $orderKey
      *
-     * @return array
+     * @return array|false
      */
     private function processOpenOrderDetails(array $orderValue, array $orders, $orderKey)
```
mitelg, line 3684:
```suggestion
    private function processOpenOrderDetails(array $orderValue, array $orders, string $orderKey): ?array
```

engine/Shopware/Core/sAdmin.php
```diff
@@ -3669,7 +3679,7 @@ private function resendConfirmationMail(array $userInfo, $hash)
      *
      * @param string $orderKey
      *
-     * @return array
+     * @return array|false
```
mitelg, line 3682:
```suggestion
```

engine/Shopware/Core/sAdmin.php
```diff
@@ -1353,10 +1354,19 @@ public function sGetOpenOrderData($destinationPage = 1, $perPage = 10)
             $orders[$orderKey]['invoice_shipping'] = $this->moduleManager->Articles()
                 ->sFormatPrice($orderValue['invoice_shipping']);
 
-            $orders = $this->processOpenOrderDetails($orderValue, $orders, $orderKey);
+            $ret = $this->processOpenOrderDetails($orderValue, $orders, $orderKey);
```
mitelg, line 1357:
```suggestion
            $processedOrders = $this->processOpenOrderDetails($orderValue, $orders, $orderKey);
```

engine/Shopware/Core/sAdmin.php
```diff
@@ -1353,10 +1354,19 @@ public function sGetOpenOrderData($destinationPage = 1, $perPage = 10)
...
+            if ($ret) {
```
mitelg, line 1358:
```suggestion
            if ($processedOrders !== null) {
```

engine/Shopware/Core/sAdmin.php
```diff
@@ -1353,10 +1354,19 @@ public function sGetOpenOrderData($destinationPage = 1, $perPage = 10)
...
+                $orders = $ret;
```
mitelg, line 1359:
```suggestion
                $orders = $processedOrders;
```

mitelg (changes requested):
also adding a test would be nice :+1:

jalbersdorfer:
Hi, as you state in your comment

> But at least removing of the items should not happen silently. What do you think about some logging here?

I did not start removing the Empty Order Bodies in the first place. I'm just producing them. The silent removing is the default behavior of the Shopware Core which just has an implementation fault.

I really would like to see the empty Order Heads not getting removed silently, but just kept. 

If you agree, I would change the Implementation to just keep them. - Then most of my changes and of your suggestions would be obsolete.

What do you think?

mitelg:
okay you are right. we can leave the logging out

shyim:
Hey @jalbersdorfer,

this PR seems still incomplete due to requested changes. We would like to let it open for another week and close it if we don't get a response from you.

PR #2412, opened 2021-02-03
Fix - Set SearchBundleDBAL condition handler public

aragon999:
Otherwise one needs to manually set the condition handlers public when adding the tag `condition_handler_dbal`, see:
https://github.com/shopware/shopware/blob/5.7/engine/Shopware/Bundle/SearchBundleDBAL/QueryBuilderFactory.php#L74-L75

Automatically sets all condition handler which are added public.

Write a plugin, add a condition handler.

\-

\-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleDBAL/DependencyInjection/Compiler/DBALHandlerCompilerPass.php
+++ b/engine/Shopware/Bundle/SearchBundleDBAL/DependencyInjection/Compiler/DBALHandlerCompilerPass.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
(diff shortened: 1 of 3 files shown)


engine/Shopware/Bundle/SearchBundleDBAL/DependencyInjection/Compiler/ConditionHandlerCompilerPass.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Shopware\Bundle\SearchBundleDBAL\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class ConditionHandlerCompilerPass implements CompilerPassInterface
+{
+    /**
+     * You can modify the container here before it is dumped to PHP code.
```
shyim: Remove this phpdoc

shyim:

PR #2414, opened 2021-02-04
fix invalid datetime value error on generic entity search

philippmahlow:
The generic entity search (for example used in attribute fields with entity selection) throws exceptions regarding "General error: 1525 Incorrect DATETIME value" with MySQL 8 (MySQL 8.0.21 in my case)

The change excludes all datetime fields from beeing used as search field in generic entity search

- Add a attribute with type multi selection for entity Category
- Open a category and start typing in the multiselection field
- The triggered AJAX request produces an exception as described in 1

### 6. Checklist

- [ X ] I have read the contribution requirements and fulfil them.


mitelg:
would you mind adding a test for this? 🙂

PR #2415, opened 2021-02-20
Speed up thumbnail generation command by chunking tasks in batches

JoshuaBehrens:
When you have E6+ images in an album that you want to generate thumbnails for the underlying paginator queries every id and queries using an IN operator. This works out but the transformation from the fetch result to the ORM objects takes time and memory when you are about to load millions of images.

It offers a new `batch` option for the `sw:thumbnail:generate` command. Using a batch size changes the behaviour of this command as it uses the paginator in a different mode which steps by classical `OFFSET LIMIT` parameters and therefore is loading data only in chunks and can operate it in a way with less memory consumption and without bloating memory to prepare the Paginator without doing any work.

1. Have lots of images to generate thumbnails for
2. Lots of images
3. Force the article album -1 to rebuild every thumbnail
4. Wait for it

### 4. Checklist

### 5. Appendix
Thanks to @jschwalmi for hitting her toe on that one. Great investigation for the reason behind the scenes

### DIFF
--- a/engine/Shopware/Commands/ThumbnailGenerateCommand.php
+++ b/engine/Shopware/Commands/ThumbnailGenerateCommand.php
@@ -25,6 +25,7 @@
 namespace Shopware\Commands;
 
 use Doctrine\ORM\Query\Expr\Join;
+use Doctrine\ORM\Tools\Pagination\Paginator;
 use Exception;
 use Shopware\Components\Model\ModelManager;
 use Shopware\Components\Model\ModelRepository;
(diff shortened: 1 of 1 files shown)


shyim:
Hey @JoshuaBehrens,

I have checked out your Branch and tested the command and it's broken not working anymore. Can you check that issue?

https://user-images.githubusercontent.com/6224096/116219433-e4539200-a74b-11eb-9907-57bafe4364d8.mp4

JoshuaBehrens:
👀

PR #2416, opened 2021-02-24
Fix robots.txt setAllow() and setDisallow()

larsbo:
Calling `setAllow()` or `setDisallow()` for a route tthat is currently disallowed respectively allowed has not the expected effect.

It ensures that an allowed route is output in /robots.txt even if it was previously disallowed and vice versa.

Open /robots.txt. There should be an entry `Allow /widgets/` because of https://github.com/shopware/shopware/blob/fba270614c5007167774a9ddb64b1d71eea0dee7/themes/Frontend/Bare/frontend/robots_txt/index.tpl#L26
But it is missing because it has been previously disallowed by https://github.com/shopware/shopware/blob/fba270614c5007167774a9ddb64b1d71eea0dee7/themes/Frontend/Bare/frontend/robots_txt/index.tpl#L13

The entry `{$robotsTxt->setAllow('/widgets')}` is also a bug I think and should be fixed in a separate PR.

### 6. Checklist


shyim:

PR #2417, opened 2021-02-24
Fix robots.txt allow directive

larsbo:
The current allow entry has no effect because a disallow entry for the same route exists and is output previously.

It changes the conflicting directive `Allow: /widgets` to a more specific directive `Allow: /widgets/emotion`.

As explained here the second directive will be ignored if conflicting directives exist:
https://developers.google.com/search/reference/robots_txt#order-of-precedence-for-group-member-lines

### 6. Checklist


shyim:

PR #2421, opened 2021-03-19
Remove Webmozart dependency

mkreusch:
Since Webmozart is not included in the project's dependencies it should be replaced with the actually present library for assertions. 
This is the only place where Webmozart was referenced.

Replace the Webmozart assertion with an alternative assertion.

Any filter actions in the store frontend will fail with:
`PHP Fatal error:  Uncaught Error: Class 'Webmozart\\Assert\\Assert' not found in engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php:66`

### 6. Checklist


shyim:

PR #2422, opened 2021-04-06
Filter order number on raw data

apfelbeck:
Filtering on tokenized data will cause issues for some order numbers

When using the order number filter with ElasticSearch, the filtering is currently done on the tokenzied field.
This causes order numbers like A1234567 to be split up into multiple tokens, and the terms query will not match correctly when given the original value.
This is especially problematic when using ES in combination with the Doofinder plugin, as it sends back order numbers as a result of the async search call.

Instead of using the tokenized field, it uses the raw value of the order number. This way, the terms query will work with any kind of order numer.

- Create article with order number "A12345678"
- Set up ElasticSearch and let it index the article
- Try to filter by order number "A12345678"
- No results are returned

https://issues.shopware.com/issues/SW-25971

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleES/ConditionHandler/OrdernumberConditionHandler.php
+++ b/engine/Shopware/Bundle/SearchBundleES/ConditionHandler/OrdernumberConditionHandler.php
@@ -54,7 +54,7 @@ class OrdernumberConditionHandler implements PartialConditionHandlerInterface
     ) {
         /* @var OrdernumberCondition $criteriaPart */
         $search->addQuery(
-            new TermsQuery('number', $criteriaPart->getOrdernumbers()),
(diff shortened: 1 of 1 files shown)


shyim:
Would an mapping change to keyword do the same?

apfelbeck:
Yeah probably, I don't think one would need the tokenized value anywhere.
Can you point me to where the mapping is created?

apfelbeck:
Nvm found it.

But it looks like someone already created a special method for this:

`'number' => $this->getTextFieldWithRawData(),`

The `number.raw` value is also used in `engine\Shopware\Controllers\Backend\Order.php`.
If I change the mapping I'd have to touch this code as well.
So changing it in OrdernumberConditionHandler would be less invasive.

shyim:

PR #2424, opened 2021-05-10
Update with displayOnlySubShopBlogComments

christopherholst:
If you have a main shop in German and two other languages ​​in English and French and only activate the comments here for the respective shops, the blog posts which should only be displayed in one language shop are missing.

If the comments should only be displayed per language shop, then all contributions that were limited by shop limitation are now displayed again.

Activate the blog comments in the configuration for each language shop only. Create a blog post which should only be displayed in German (shop limitation).
Create another blog post which should be displayed in all languages ​​and translate it accordingly.

The result is that we only see the blog post which is available in all languages.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Blog/Repository.php
+++ b/engine/Shopware/Models/Blog/Repository.php
@@ -108,8 +108,8 @@ class Repository extends ModelRepository
 
         if ($shopId && Shopware()->Config()->get('displayOnlySubShopBlogComments')) {
             $builder
-                ->leftJoin('blog.comments', 'comments', Join::WITH, 'comments.active = 1 AND (comments.shopId IS NULL OR comments.shopId = :shopId)')
-                ->setParameter('shopId', $shopId);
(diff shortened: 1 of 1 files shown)


engine/Shopware/Models/Blog/Repository.php
```diff
@@ -108,8 +108,8 @@ public function getListQueryBuilder($blogCategoryIds, $filter, $shopId = null)
 
         if ($shopId && Shopware()->Config()->get('displayOnlySubShopBlogComments')) {
             $builder
-                ->leftJoin('blog.comments', 'comments', Join::WITH, 'comments.active = 1 AND (comments.shopId IS NULL OR comments.shopId = :shopId)')
-                ->setParameter('shopId', $shopId);
+                ->leftJoin('blog.comments', 'comments', Join::WITH, 'comments.active = 1 AND (comments.shopId IS NULL OR comments.shopId = :shopId2)')
```
shyim: can you name it something like commentShopId

christopherholst: sure and good idea... I couldn't think of anything better earlier

shyim:

PR #2425, opened 2021-05-12
Allow the cache to be invalided by subnets

wbm-mkopp:

If you have a dynamic environment you can't invalidate the cache. That's because AppCache will check if the request is allowed. At the moment you can only add IP addresses and you can't add subnets.

If you run a cronjob in a Kubernetes cluster then the cronjob runs in a different pod than the webserver. So if the cronjob pod sends a BAN request to the web server, the IP address is not allowed. You can't add the IP address of the pod because it's dynamic and will change.

Add the possibility to add subnets to the Shopware config file ("app/config/config.php" => purge_allowed_ips).

Setup a dynamic environment for example with Kubernetes. Run a cronjob that updates doctrine entities (e.g. Articles).

None.

https://developers.shopware.com/developers-guide/http-cache/#cluster-environments

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/HttpCache/AppCache.php
+++ b/engine/Shopware/Components/HttpCache/AppCache.php
@@ -306,6 +306,8 @@ class AppCache extends HttpCache
     }
 
     /**
+     * @deprecated Since this method does not respect subnet definitions, please use IpUtils::checkIp instead
+     *
      * Checks if $ip is allowed for Http PURGE requests
      *
      * @param string $ip
--- a/engine/Shopware/Components/HttpCache/AppCache.php
+++ b/engine/Shopware/Components/HttpCache/AppCache.php
@@ -24,6 +24,7 @@
 
 namespace Shopware\Components\HttpCache;
 
+use Symfony\Component\HttpFoundation\IpUtils;
 use Shopware\Bundle\CookieBundle;
 use Shopware\Bundle\CookieBundle\CookieGroupCollection;
 use Shopware\Components\Privacy\CookieRemoveSubscriber;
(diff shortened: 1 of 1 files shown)


shyim:
Hey,

can you change the target to `5.7`. We don't plan a new patch version for 5.6 anymore

engine/Shopware/Components/HttpCache/AppCache.php
```diff
@@ -301,7 +302,7 @@ protected function isPurgeRequestAllowed(Request $request)
             }
         }
 
-        return $this->isPurgeIPAllowed($request->getClientIp());
```
shyim, line 304: Can you mark the old method as deprecated?

cjost1988:
> Hey,
> 
> can you change the target to `5.7`. We don't plan a new patch version for 5.6 anymore

Then definitely there should be a backport for version `5.6`.

cjost1988:
> Hey,
> 
> can you change the target to `5.7`. We don't plan a new patch version for 5.6 anymore

Is it safe that there will be no update in `5.6` anymore?
Because we might not update to `5.7` and would need to keep our intermediate fork with the fix on `5.6`.

engine/Shopware/Components/HttpCache/AppCache.php
cjost1988: Yes, I'd recommend this as well.

We thought about preparing a MR for removing the method, but since the class is not `final` we can't remove it straight away and since there will be no following major version depending on this source code removing will not be a plan anyways.

mitelg: we should still mark it as deprecated. so user of this method are notified about that and could also change to a better method

wbm-mkopp: We marked the method as deprecated.

shyim:
We are releasing in a short time 5.7 and therefore we don't plan an another version

shyim:

PR #2426, opened 2021-05-19
Fix/advanced menu delay

favor-it:
The advanced menu hover effect does not work correctly. This pull request faces/solves multiple issues:
- The "Hover delay (ms)", that can be configured in the settings, is not respected on first hover of the navigation bar after loading a page. The advanced menu instantly opens. The delay is respected on following (2nd, 3rd, ...) hovers.
- When moving the cursor downwards moving over the navigation bar, the advanced menu always opens. Even when the cursor leaves the navigation bar in less time than configured as "Hover delay (ms)". This happens when the cursor stays in the area where the advanced menu is about to be displayed. This is an annoying behavior, because the advanced menu covers the first part of the content section - every time a user moves down the cursor from the header to the content section (without the intention to open the menu).
- The performance impact of the used mousemove function on body is bad. This problem has partially been addressed with this contribution: https://github.com/shopware/shopware/pull/2153 https://issues.shopware.com/issues/SW-24297 https://github.com/shopware/shopware/commit/06eb4d0b23548848da0b1ed011aee1790ab67408# But this "fix" broke the hover delay resulting in the issue described above.

This change solves the three issues described above by
- reverting the commit https://github.com/shopware/shopware/commit/06eb4d0b23548848da0b1ed011aee1790ab67408#
- replacing the mousemove listener on the body element with mouseleave listeners only on the relevant elements
- initially calling closeMenu() to fix the first issue (hover delay not working on first hover)

Description under '1. Why is this change necessary?' should contain all information to reproduce the issues.

N/A

N/A

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Views/frontend/_public/src/js/jquery.advanced-menu.js
+++ b/engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Views/frontend/_public/src/js/jquery.advanced-menu.js
@@ -156,7 +156,7 @@
                     me._on($el, 'touchstart', $.proxy(me.onTouchStart, me, i, $el));
                 }
 
-                me._on($el, 'mouseenter', $.proxy(me.onListItemEnter, me, i, $el));
+                me._on($el, 'mouseenter touchstart', $.proxy(me.onListItemEnter, me, i, $el));
                 me._on($el, 'click', $.proxy(me.onClick, me, i, $el));
                 me._on($el, 'mouseleave', $.proxy(me.onMouseLeave, me));
             });
--- a/engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Views/frontend/_public/src/js/jquery.advanced-menu.js
+++ b/engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Views/frontend/_public/src/js/jquery.advanced-menu.js
@@ -157,6 +157,8 @@
                 me._on($el, 'click', $.proxy(me.onClick, me, i, $el));
             });
 
+            $body.on('mousemove touchstart', $.proxy(me.onMouseMove, me));
+
             me._on(me._$closeButton, 'click', $.proxy(me.onCloseButtonClick, me));
         },
 
@@ -312,8 +314,6 @@
         openMenu: function () {
             var me = this;
 
-            $body.on('mousemove touchstart', $.proxy(me.onMouseMove, me));
-
             me.$el.show();
 
             $.publish('plugin/swAdvancedMenu/onOpenMenu', [ me ]);
@@ -332,8 +332,6 @@
             me._$list.find('.' + opts.itemHoverClass).removeClass(opts.itemHoverClass);
 
             me.$el.hide();
-
-            $body.off('mousemove touchstart', $.proxy(me.onMouseMove, me));
 
             me._targetIndex = -1;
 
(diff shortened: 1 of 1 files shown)


shyim:
Hey @favor-it ,

Can you push your changes in two commits? First reverting that commit and then applying your fixes?

Thanks! :blue_heart:

favor-it:
Hey @shyim, I have just updated my pull request accordingly.

xf-:
Hi,

looks interesting. I wrote the first patch from a modified version used in production. Basically the patch moved the events to better position and nowhere near from the setTimeout or delay.
In my version i use `throttle` from underscore to handle the events and delay to avoid rapid firing the event.

I currently testing your version and looks fine with my patches applied.

xf-:
@favor-it i noticed touch won't work, maybe i messed up something, but it looks like touch won't open the menu. e.g. Tablet.

Fix for me
```js
                me._on($el, 'touchstart', $.proxy(me.onTouchStart, me, i, $el));
                me._on($el, 'mouseenter touchstart', $.proxy(me.onListItemEnter, me, i, $el));
```

DennisGarding:
> @favor-it i noticed touch won't work, maybe i messed up something, but it looks like touch won't open the menu. e.g. Tablet.
> 
> Fix for me
> 
> ```js
>                 me._on($el, 'touchstart', $.proxy(me.onTouchStart, me, i, $el));
>                 me._on($el, 'mouseenter touchstart', $.proxy(me.onListItemEnter, me, i, $el));
> ```
@favor-it Can you fix this issue so we can import it?

xf-:
@swDennis What do you mean exactly?
This is a patch for this patch (not in patch format) and replaces 2 lines.
This version runs for now for 3 weeks live without any issue.

DennisGarding:
> @swDennis What do you mean exactly? This is a patch for this patch (not in patch format) and replaces 2 lines. This version runs for now for 3 weeks live without any issue.

I mean that favour-it can check the problem and fix it in this pullrequest

xf-:
Line 159 - add `touchstart` to mouseenter

I have a different version after all, that is the reason i didn't provide a patch file. e.g. i removed all the IE non-sence. Touch devices with IE as browser are uncommon (especially to be a tablet or larger for advanced menu).

_Offtopic
There also more patches in development version.
e.g. is an idea to move the menu into js/json. The issue is lighthouse or recommended amount of DOM elements. An unchanged Demostore is already nearly around 1000 Elements and 1500 shouldn't be passed. With a full menu and not 4 items per list + few standard plugins you can easily bypass the 1500 in a list few or shopping world with an article slider. A;sp new Cookie Banner settings are in source code and maybe used at first page hit, afterwards a lot of DOM elements.
https://web.dev/dom-size/_

favor-it:
@xf- / @swDennis: I just fixed the issue in an additional commit

xf-:
@mitelg this should be fine now and also work in responsive mode

engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Views/frontend/_public/src/js/jquery.advanced-menu.js
```diff
@@ -246,12 +251,12 @@
          * @event onMouseLeave
          * @param {jQuery.Event} event
          */
-        onMouseMove: function (event) {
+        onMouseLeave: function (event) {
```
mitelg, line 254: also this method is not used anymore, we cannot simply remove it, as it would break for people who would have extended this JS plugin and this method. 

please add it again as empty method

mitelg (changes requested):

DennisGarding (approved):

mitelg:
@favor-it one more last thing to do for you :see_no_evil: 

please sign the CLA, otherwise we cannot use your commits :slightly_smiling_face:

favor-it:
@mitelg CLA has just been signed

mitelg:
thanks for your contribution @favor-it   👍🎉💙

PR #2427, opened 2021-06-02
add index on s_order_basket.userID

jonashrem:
we have noticed shopware doing querries like this one on many of our customers:
```
UPDATE `s_order_basket` SET `sessionID` = ? WHERE `userID` = ? AND `modus` = ? 
```

yet, this querry is not efficieent as there is no index on userID.

We noticed that this querry can create quite a load on the database on high traffic szenarios.

Add an MySQL index to s_order_basket.userID

/

/

/

### 6. Checklist


JoshuaBehrens:
Aaah something I can drop from my bucket list 😁

_sql/migrations/966-add-userid-index-to-s-order-basket.php
```diff
@@ -0,0 +1,12 @@
+<?php
+class Migrations_Migration966 Extends Shopware\Components\Migrations\AbstractMigration
```
shyim:
```suggestion
class Migrations_Migration1710 extends Shopware\Components\Migrations\AbstractMigration
```

_sql/migrations/966-add-userid-index-to-s-order-basket.php
```diff
@@ -0,0 +1,12 @@
...
+{
+    public function up($modus)
```
shyim:
```suggestion
    public function up($modus): void
```

_sql/migrations/966-add-userid-index-to-s-order-basket.php
shyim: Don't forget to change the file name too :)

shyim:
Looks good. Can you do apply these two changes?

jonashrem:
> Looks good. Can you do apply these two changes?

Thanks. I just applied them.

_sql/migrations/966-add-userid-index-to-s-order-basket.php
jonashrem: So filename still needs to be changed to 1710-add-userid-index-to-s-order-basket.php as far as I understand. Is this correct? @shyim

shyim: Yes

_sql/migrations/966-add-userid-index-to-s-order-basket.php
```diff
@@ -0,0 +1,12 @@
+<?php
+class Migrations_Migration1710 extends Shopware\Components\Migrations\AbstractMigration
```
mitelg: the filename still needs to be adjusted :blush:

_sql/migrations/966-add-userid-index-to-s-order-basket.php
jonashrem: I just changed the filename

_sql/migrations/966-add-userid-index-to-s-order-basket.php
jonashrem: I just changed it

shyim:

PR #2428, opened 2021-06-12
initServiceMode: case-insensitive compare for `offlineip` addresses

leeps:
## Make the comparison of allowed IP addresses in `offlineip` to the client's IP address as reported by the web server(s) case-insensitive.

This will enhance working with IPv6 addresses, which are sometime reported with uppercase letters by some "whatismyip" sites.
The case-sensitive behavior has lead to confusion for at least one of our customers who could not access the frontend in ServiceMode despite entering their IPv6 address into the allow list.

IPv6 addresses contain letters, they are therefore sensitive to letter case when comparing strings.
Using an case-insensitive comparison solves this issue.

This change makes the comparison of the user configuration value for `offlineIp`, which is a string of space-separated IP addresses (IPv4 or IPv6), with the HTTP client's IP address a case-insensitive one.

When entering IPv6 addresses with upper-case letters for the configuration value of `offlineIp`, users connecting from those IP addresses would not be able to bypass ServiceMode and be denied access to the store frontend pages.

n/a

n/a

### 6. Checklist

Please notify me if any issues appear in this PR, I'll gladly work to correct them.

### DIFF
--- a/engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php
@@ -168,7 +168,7 @@ class Shopware_Plugins_Core_Router_Bootstrap extends Shopware_Components_Plugin_
     protected function initServiceMode($request)
     {
         $config = $this->Application()->Config();
-        if (!empty($config->setOffline) && strpos($config->offlineIp, $request->getClientIp()) === false) {
+        if (!empty($config->setOffline) && strpos(strtolower($config->offlineIp), $request->getClientIp()) === false) {
             if ($request->getControllerName() !== 'error') {
                 $request->setControllerName('error')->setActionName('service')->setDispatched(false);
             }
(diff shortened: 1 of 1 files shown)


engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php
```diff
@@ -168,7 +168,7 @@ public function getCapabilities()
     protected function initServiceMode($request)
     {
         $config = $this->Application()->Config();
-        if (!empty($config->setOffline) && strpos($config->offlineIp, $request->getClientIp()) === false) {
+        if (!empty($config->setOffline) && strpos(strtolower($config->offlineIp), $request->getClientIp()) === false) {
```
shyim: What about using stripos? So I would not matter if the config case is wrong or the actual ip

shyim (changes requested):

engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php
leeps: Hey @shyim , thank you for your feedback. You are, of course, absolutely right. I'll change it to `stripos()`!

shyim:

PR #2429, opened 2021-06-21
Fix empty changelog in DOM parser

xf-:
With PHP8, trim and strip_tags the changelog can be empty and 'DOM->load` needs content or it throws an exception.

Check if the changelog is empty and returns on empty

Changelog entry from a Store plugin like this
```
    <changelog version="2.3.3">
        <changes lang="de"><![CDATA[]]></changes>
        <changes lang="en"><![CDATA[]]></changes>
    </changelog>
```

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLocalService.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLocalService.php
@@ -278,7 +278,7 @@ class PluginLocalService
     {
         $html = strip_tags($html, '<br><i><b><strong><em><del><u><div><span><ul><li><ll><ol><p><a>');
 
-        if (empty($html)) {
+        if ($html === '') {
             return '';
         }
 
--- a/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLocalService.php
+++ b/engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLocalService.php
@@ -278,6 +278,10 @@ class PluginLocalService
     {
         $html = strip_tags($html, '<br><i><b><strong><em><del><u><div><span><ul><li><ll><ol><p><a>');
 
+        if (empty($html)) {
+            return '';
+        }
+
         $dom = new \DOMDocument();
         $dom->loadHTML(mb_convert_encoding($html, 'HTML-ENTITIES', 'UTF-8'));
 


engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLocalService.php
```diff
@@ -278,6 +278,10 @@ private function parseChangeLog($html)
     {
         $html = strip_tags($html, '<br><i><b><strong><em><del><u><div><span><ul><li><ll><ol><p><a>');
 
+        if (empty($html)) {
```
mitelg:
```suggestion
        if ($html === '') {
```
we should for an empty string here. otherwise texts with `0` or `false` will also be skipped :wink:

engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLocalService.php
```diff
@@ -278,6 +278,10 @@ private function parseChangeLog($html)
     {
         $html = strip_tags($html, '<br><i><b><strong><em><del><u><div><span><ul><li><ll><ol><p><a>');
```
mitelg, line 279: since SW 5.7 has now a min requirement of PHP 7.4, 
maybe we can switch to an array here for better readability -> https://www.php.net/manual/en/function.strip-tags.php

engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLocalService.php
xf-: yeah good point.
`strip_tags` can only return a string and so only 0 is left and i think nobody will have a 0 standalone in there changelog, but this is cleaner.

engine/Shopware/Bundle/PluginInstallerBundle/Service/PluginLocalService.php
xf-: @mitelg that 7.4 thing is still funny because the docu says 7.3 is minimum.
https://developers.shopware.com/developers-guide/shopware-5-upgrade-guide-for-developers/#system-requirements-changes.

Array has a nice touch. But in same step we should clean that tags. What is `ll` and tags like `strong`, `em`, `span` have no effect (no style or effect). `div` only breaks the line

shyim: Whoops I have fixed the docs

xf-: @shyim maybe a stupid question. I search news, docu, roadmap and i didn't didn't a EOL update or any information on that. Did the EOL change with delayed 5.7 and some vendors still not fixed porting to v6.

shyim: We decided to change the min to 7.4 in the release of RC2 as the rc phase was too long and it doesn't make sense to support 7.3 anymore in the last months of it.

xf-: sorry i meant Shopware 5.7 :) i would prefer if everything runs on PHP 8.0. I think the last date was set on the presentation from v6 with 5.7 released last year. Would be nice to communicate a clear EOL to a customer or Support at least until XY.

xf-:
Would love to see a release - breaks the current backend with a store extension, because of empty changelog entries in 5.7.2

mitelg:

your fix will be part of the upcoming 5.7.3 release

PR #2430, opened 2021-06-30
Remove time_limit and add rowIndex into Event

PhenomEY:
Remove the time_limit from the sExport function and add the rowIndex into the Filter_Fixed Event

Easier extension of the sExport

Customizing your feed through the Filter_Fixed Event can reach 30seconds really fast. 
Adding the rowIndex into the Event makes it easier to detect in which current row the event got fired.


mitelg:
thanks for your contribution @PhenomEY   👍🎉💙

we re-added the time-out config, but we check the max execution time before, if it has a min time of 30 seconds. so if the execution time is higher, it will not be decreased anymore

PR #2432, opened 2021-07-19
Uncaught DivisionByZeroError with PHP 8

christopherholst:
The change is necessary for PHP 8, otherwise not a single blog post will be displayed in the frontend.

Displays the blog posts in the frontend again.

Change of the PHP version from 7.4 to 8.0.3 and not a single blog post can be opened anymore.

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/blog/header.tpl
+++ b/themes/Frontend/Bare/frontend/blog/header.tpl
@@ -61,7 +61,11 @@
 {* Canonical link *}
 {block name='frontend_index_header_canonical'}
     {* Count of available product pages *}
-    {$pages = ceil($sNumberArticles / $sPerPage)}
+    {if $sPerPage != 0}
+        {$pages = ceil($sNumberArticles / $sPerPage)}
+    {else}
+        {$pages = 0}
+    {/if}
 
     {if $sArticle}
         <link rel="canonical" href="{url controller=blog action=detail sCategory=$sArticle.categoryId blogArticle=$sArticle.id}" />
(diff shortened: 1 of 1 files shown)


themes/Frontend/Bare/frontend/blog/header.tpl
```diff
@@ -61,7 +61,11 @@
 {* Canonical link *}
 {block name='frontend_index_header_canonical'}
     {* Count of available product pages *}
-    {$pages = ceil($sNumberArticles / $sPerPage)}
+    {if $sPerPage != 0}
```
soebbing: I think it would be more readable to initialize $pages with 0 and remove the `else` case

christopherholst (commented):
like this @soebbing ?

mitelg:
@christopherholst please fill out the PR template correctly. It is not there to be ignored. 
especially a description how to reproduce this error

christopherholst:
done.

mitelg:
I checked this on PHP 8.0.8 and could not reproduce it :thinking: 

looking at the code, `$sPerPage` could only be `0` if there is a plugin doing things, or the products per page setting in backend was set to `0`, which would be silly :grin:

philipreinken (changes requested):
Hi @christopherholst , would you mind moving this change out of the template, for example, into a Controller instead? This could help with testing and also decrease the amount of logic inside the template.

philipreinken:
Hello, we're closing this pull request due to inactivity. If this change is still important to you, feel free to open another pull request.

For more information about our contribution guidelines, see https://developers.shopware.com/community/contribution-guideline

PR #2433, opened 2021-07-23
fix analytics conversion export

resslinger:
Currently, only the entries from the first page are exported from the "Conversion Overview" report type.
This PR fix the csv export of this report type.

The change export all entries.

1.) Go into the backend.
2.) Open the window marketing > evaluations > evaluations.
3.) Select the type "Conversion Overview"
4.) Filter for more then 3 month, for example.
5.) Click on export

=> in the export is only the result of the first page included.

With this fix, all entries are exported as csv now.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Analytics.php
+++ b/engine/Shopware/Controllers/Backend/Analytics.php
@@ -279,11 +279,17 @@ class Shopware_Controllers_Backend_Analytics extends Shopware_Controllers_Backen
             }
         }
 
+        // Sets the correct limit
+        $limit = 25;
+        if (strtolower($this->format) === 'csv') {
+            $limit = \count($data);
+        }
+
         $values = array_values($data);
         $splice = array_splice(
             $values,
-            (int) $this->Request()->getParam('start', 0),
-            (int) $this->Request()->getParam('limit', 25)
+            $this->Request()->getParam('start', 0),
+            $this->Request()->getParam('limit', $limit)
         );
 
         $this->send($splice, \count($data));
--- a/engine/Shopware/Controllers/Backend/Analytics.php
+++ b/engine/Shopware/Controllers/Backend/Analytics.php
@@ -212,8 +212,8 @@ class Shopware_Controllers_Backend_Analytics extends Shopware_Controllers_Backen
         $values = array_values($data);
         $splice = array_splice(
             $values,
-            $this->Request()->getParam('start', 0),
-            $this->Request()->getParam('limit', $limit)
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Backend/Analytics.php
```diff
@@ -279,11 +279,17 @@ public function getRatingAction()
             }
         }
 
+        // Sets the correct limit
+        $limit = 25;
+        if (strtolower($this->format) === 'csv') {
```
shyim, line 284: You can wrap the array_splice in an if condition format != csv. I guess it is better to read.

engine/Shopware/Controllers/Backend/Analytics.php
```diff
@@ -279,11 +279,17 @@ public function getRatingAction()
...
+            $limit = \count($data);
+        }
+
         $values = array_values($data);
         $splice = array_splice(
             $values,
-            (int) $this->Request()->getParam('start', 0),
```
mitelg, line 285: why did you remove the cast?

mitelg:
would you mind adding a test for this action here? https://github.com/shopware/shopware/blob/5.7/tests/Functional/Controllers/Backend/AnalyticsTest.php

engine/Shopware/Controllers/Backend/Analytics.php
resslinger: I copied the logic from the method getOverviewAction before. I will add the cast on this method too.

engine/Shopware/Controllers/Backend/Analytics.php
resslinger: I don't think, that a wrap make it easier to read. The array_splice is needed with format == csv and format != csv. So it duplicates the code.

resslinger:
@mitelg I'll try it... :)

mitelg:
let us know, if you need help with it :blush: :+1:

philipreinken:
Hello, we're closing this pull request due to inactivity. If this change is still important to you, feel free to open another pull request.

For more information about our contribution guidelines, see https://developers.shopware.com/community/contribution-guideline

resslinger:
@philipgatzka I only have not enough time to add a test. But the bug exists furthermore.

philipreinken:
Got you, and thank you for investing the time - it's just that we don't want old pull requests to pile up over time and we, therefore, clean them up now and then. Please open another PR (or re-open this one) once you find the time to complete it 🙂

PR #2435, opened 2021-08-17
Make the Produkt-Export Cache Directory configurable

teiling88:
This change is needed, if you want to store your product_exports not in the cache directory.

I want to store the product exports outside the cache directory to save them on a nfs volume. Without this change, I have to replace the whole controller - yeah!

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Export.php
+++ b/engine/Shopware/Controllers/Backend/Export.php
@@ -184,7 +184,7 @@ class Shopware_Controllers_Backend_Export extends Enlight_Controller_Action impl
     /**
      * @return string Path to new output directory
      */
-    private function createOutputDirectory()
+    protected function createOutputDirectory()
     {
         $dirName = $this->container->getParameter('kernel.cache_dir');
         if (!\is_string($dirName)) {
--- a/engine/Shopware/Controllers/Backend/Export.php
+++ b/engine/Shopware/Controllers/Backend/Export.php
@@ -184,7 +184,7 @@ class Shopware_Controllers_Backend_Export extends Enlight_Controller_Action impl
     /**
      * @return string Path to new output directory
      */
-    protected function createOutputDirectory()
+    private function createOutputDirectory()
     {
         $dirName = $this->container->getParameter('kernel.cache_dir.product_export');
         if (!\is_string($dirName)) {
--- a/engine/Shopware/Controllers/Backend/Export.php
+++ b/engine/Shopware/Controllers/Backend/Export.php
@@ -186,12 +186,11 @@ class Shopware_Controllers_Backend_Export extends Enlight_Controller_Action impl
      */
     protected function createOutputDirectory()
     {
-        $dirName = $this->container->getParameter('kernel.cache_dir');
+        $dirName = $this->container->getParameter('kernel.cache_dir.product_export');
         if (!\is_string($dirName)) {
-            throw new \RuntimeException('Parameter kernel.cache_dir has to be an string');
+            throw new \RuntimeException('Parameter kernel.cache_dir.product_export has to be an string');
         }
 
-        $dirName .= '/productexport/';
         if (!file_exists($dirName)) {
             mkdir($dirName);
         }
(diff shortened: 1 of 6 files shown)


engine/Shopware/Controllers/Backend/Export.php
```diff
@@ -184,7 +184,7 @@ private function getExportContentType()
     /**
      * @return string Path to new output directory
      */
-    private function createOutputDirectory()
+    protected function createOutputDirectory()
     {
         $dirName = $this->container->getParameter('kernel.cache_dir');
```
philipreinken: Would you mind changing this, so that the `$dirName` is initially read from config and the `kernel.cache_dir` is used as a fallback?

philipreinken (changes requested):
Hi @teiling88 , thank you for creating this PR! 😁

engine/Shopware/Controllers/Backend/Export.php
teiling88: I will do that next week!

teiling88: I created a slightly different approach. What do you think about it?

engine/Shopware/Kernel.php
```diff
@@ -741,6 +741,7 @@ protected function getKernelParameters()
             'kernel.debug' => $this->debug,
             'kernel.name' => $this->name,
             'kernel.cache_dir' => $this->getCacheDir(),
+            'kernel.cache_dir.product_export' => $this->getCacheDir().'/productexport',
```
mitelg: the places where you replaced the code with your new DI parameter looked like this `$cacheDir .= '/productexport/';`(or similar :wink: )

so I guess
```suggestion
            'kernel.cache_dir.product_export' => $this->getCacheDir().'/productexport/',
```
a slash is missing at the end? :thinking:

engine/Shopware/Controllers/Backend/Export.php
```diff
@@ -184,14 +184,13 @@ private function getExportContentType()
     /**
      * @return string Path to new output directory
      */
-    private function createOutputDirectory()
+    protected function createOutputDirectory()
```
mitelg: I really like your new approach :+1: 

so I guess changing the visibility is not necessary anymore?

engine/Shopware/Controllers/Backend/Export.php
mitelg: nice! just two little questions :blush:

tests/Functional/Core/sExportTest.php
```diff
@@ -75,9 +75,8 @@ public function setUp(): void
         $this->template = $this->container->get('template');
 
         /** @var string $cacheDir */
```
mitelg: needs to be removed. PHPStan is complaining :wink:

engine/Shopware/Controllers/Backend/Export.php
teiling88: Yes of course - is fixed :-)

engine/Shopware/Kernel.php
teiling88: Yes good catch, bad copy and past from me. I developed it in our local docker setup and not directly in this repo 👼🏻

tests/Functional/Core/sExportTest.php
teiling88: fixed

teiling88 (commented):
Pls let me know if everything is ok so I can squash the commits for you :-)

engine/Shopware/Kernel.php
mitelg: I guess, the cs-fixer will complain about that :grin: 

```suggestion
            'kernel.cache_dir.product_export' => $this->getCacheDir() . '/productexport/',
```

mitelg:
one more thing :grin: otherwise it looks good to me :+1:

teiling88:
Fixed :-)

mitelg:
very nice! thank you very much so far :+1:

mitelg:
thanks for your contribution @teiling88   👍🎉💙

note, that we changed the name of the parameter and made it configurable via the `config.php` 
see: https://github.com/shopware/shopware/commit/54f0b29e3aa7eb8265aebeab5b8f7ee0f76e69d8

PR #2436, opened 2021-08-20
Fix backend user adds

Fayti1703:
It's currently impossible to log in as a newly created backend user, unless you change the user's `s_core_auth` entry to have a non-`NULL`  `locked_until` field.

Partially revert bdbf6210c52f30378da3265c60332e3e2185ff11. Specifically, this re-adds the `__construct` function of the `Models\User\User` class.

1. Create a new backend user via the user manager in the backend
2. Attempt to log in as that user

None. There is a comment on [SW-26142](https://issues.shopware.com/issues/SW-26142) that mentions the problem this PR fixes, but I couldn't find any standalone issue that mentions it.

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/User/User.php
+++ b/engine/Shopware/Models/User/User.php
@@ -190,6 +190,15 @@ class User extends ModelEntity
     private $role;
 
     /**
+     * Initial the date fields
+     */
+    public function __construct()
+    {
+        $this->lastLogin = new \DateTime();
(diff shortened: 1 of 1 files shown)


mitelg:
hey @Fayti1703 
we already fixed this problem here: https://github.com/shopware/shopware/commit/7e97df41f3b49d89decbdc2dc070bc961646a8fc

PR #2438, opened 2021-09-08
Properly create the _controller attribute to handle the autowiring of controller actions parameters.

xrogers:
This change allows to create the autowired services as controllers action paramters with currently used (and future) symfony version. 
See: https://github.com/symfony/http-kernel/blob/4619cd0da8b1528ee3d75e0187ac2349a523817e/Controller/ControllerResolver.php#L119

Fixes the format of the _controller attribute of the request.

1. Create controller which uses autowired service.
2. Try to use such controller action.
3. System is unable to find the service and breaks.

-

-

### 6. Checklist


mitelg:
thanks for your contribution @xrogers  👍🎉💙

PR #2439, opened 2021-09-22
OrderDetailID set to 0 instead of null for non-article items in order_details

antoineverger1:

For non-articles, the orderDetaildID column of order items (s_order_details) is to set to 0 instead of null. It is not a correct behavior as surcharges or discount have no article details and therefore should contain a null value. The issue comes from string conversion in a sprintf that cast a null value into a 0.  

This changes will properly make the articleDetailID being null when necessary. Thought it's important to notice that there is not migration involved for existing order_details and I am not sure it's even wished to do that. 

Create an order that includes a discount (or a surcharge) and go through the entire checkout

Without the current PR:
- in s_order_details, for the discount item, the column articleDetailID contains 0
With the current PR:
- in s_order_details, for the discount item, the column articleDetailID contains null :+1:

### DIFF
--- a/engine/Shopware/Core/sOrder.php
+++ b/engine/Shopware/Core/sOrder.php
@@ -696,7 +696,7 @@ class sOrder implements Enlight_Hook
                 pack_unit,
                 articleDetailID
                 )
-                VALUES (%d, %s, %d, %s, %f, %d, %s, %d, %s, %d, %d, %d, %f, %s, %s, %s, %d)
+                VALUES (%d, %s, %d, %s, %f, %d, %s, %d, %s, %d, %d, %d, %f, %s, %s, %s, %s)
             ';
 
             $sql = sprintf(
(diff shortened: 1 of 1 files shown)


engine/Shopware/Core/sOrder.php
```diff
@@ -717,7 +717,7 @@ public function sSaveOrder()
                 $this->db->quote((string) $basketRow['ean']),
                 $this->db->quote((string) $basketRow['itemUnit']),
                 $this->db->quote((string) $basketRow['packunit']),
-                $basketRow['additional_details']['articleDetailsID']
+                !\is_null($basketRow['additional_details']['articleDetailsID']) ?: 'null'
```
xrogers: It doesn't look right, it will put either boolean true or 'null' in articleDetailID

antoineverger1: OMG it is so wrong... Wanna have me fixing the fix or you are gonna take care of it?

antoineverger1: I fixed it directly

mitelg:
thanks for your contribution @antoineverger1   👍🎉💙

antoineverger1:
Well, if that helped @mitelg :laughing:

PR #2441, opened 2021-10-08
Fix ES indexing query skipping property groups

Nevay:
The ES property query relies on the order in which property groups are returned, property groups will be skipped if they are not returned in strictly ascending order.

Explicitly sets the order.

### 6. Checklist


mitelg:
thanks for your contribution @Nevay   👍🎉💙

PR #2444, opened 2021-10-12
Fix category sorting on PHP8

Nevay:
PHP8 changed the behavior of sort functions, [RFC stable sorting](https://wiki.php.net/rfc/stable_sorting) Illegal comparison functions:
> This works, because PHP currently only checks whether the comparison result is “greater than” or not, and never explicitly distinguishes the “equal” and “smaller than” cases. This breaks down with the approach proposed here, because we now do need to know whether values are equal or not, in order to use the fallback sorting criterion.

Return `-1` instead of `0` if left is smaller than right.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
+++ b/engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
@@ -120,10 +120,10 @@ class CategoryGateway implements Gateway\CategoryGatewayInterface
         //use php usort instead of running mysql order by to prevent file-sort and temporary table statement
         usort($data, function ($a, $b) {
             if ($a['__category_position'] === $b['__category_position']) {
-                return $a['__category_id'] > $b['__category_id'] ? 1 : 0;
+                return $a['__category_id'] <=> $b['__category_id'];
             }
 
-            return $a['__category_position'] > $b['__category_position'] ? 1 : 0;
+            return $a['__category_position'] <=> $b['__category_position'];
         });
 
         $categories = [];


engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
```diff
@@ -120,10 +120,10 @@ public function getList(array $ids, Struct\ShopContextInterface $context)
         //use php usort instead of running mysql order by to prevent file-sort and temporary table statement
         usort($data, function ($a, $b) {
             if ($a['__category_position'] === $b['__category_position']) {
-                return $a['__category_id'] > $b['__category_id'] ? 1 : 0;
+                return $a['__category_id'] > $b['__category_id'] ? 1 : -1;
```
mitelg: we should simply use the space ship operator here
```suggestion
                return $a['__category_id'] <=> $b['__category_id'];
```

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/CategoryGateway.php
```diff
@@ -120,10 +120,10 @@ public function getList(array $ids, Struct\ShopContextInterface $context)
...
             }
 
-            return $a['__category_position'] > $b['__category_position'] ? 1 : 0;
+            return $a['__category_position'] > $b['__category_position'] ? 1 : -1;
```
mitelg:
```suggestion
            return $a['__category_position'] <=> $b['__category_position'];
```

mitelg (changes requested):

mitelg:
hey @Nevay 

thanks for your first contribution :+1: 

I left some comments, to even improve the code more :slightly_smiling_face:

mitelg (approved):

LePrau:
The failed mysql:8.0 unit test seems to be unrelated.
Is a re-run needed?

mitelg:
I did a re-run and it works now :slightly_smiling_face:

mitelg:
thanks for your contribution @Nevay  :+1: :tada: :blue_heart:

PR #2445, opened 2021-10-12
Fix fuzzy search also looking through inactive articles

Algorithman:
Fuzzy search is looking at all articles, even deactivated ones (which would not appear on the frontend)

Added active=1 in the where clauses

Premise:
Multiple articles which react to a keyword. The first in the default sorting is inactive.

If you then in the frontend search for this keyword, the shop lists the search results. If you press enter or click Show all, shopware tries to show the deactivated article (which of course won't show up).

Dunno

None, it is a bug fix

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Search.php
+++ b/engine/Shopware/Controllers/Frontend/Search.php
@@ -135,7 +135,7 @@ class Shopware_Controllers_Frontend_Search extends Enlight_Controller_Action
                 FROM s_articles_details
                   INNER JOIN s_articles
                    ON s_articles.id = s_articles_details.articleID
-                WHERE ordernumber = ? AND s_articles.active = 1
+                WHERE ordernumber = ? AND s_articles.active = 1 AND s_articles_details.active = 1
                 GROUP BY articleID
                 LIMIT 2
             ';
--- a/engine/Shopware/Controllers/Frontend/Search.php
+++ b/engine/Shopware/Controllers/Frontend/Search.php
@@ -135,7 +135,7 @@ class Shopware_Controllers_Frontend_Search extends Enlight_Controller_Action
                 FROM s_articles_details
                   INNER JOIN s_articles
                    ON s_articles.id = s_articles_details.articleID
-                WHERE ordernumber = ?
+                WHERE ordernumber = ? AND s_articles.active = 1
                 GROUP BY articleID
                 LIMIT 2
             ';
@@ -151,8 +151,8 @@ class Shopware_Controllers_Frontend_Search extends Enlight_Controller_Action
                 $sql = '
                     SELECT DISTINCT articleID
                     FROM s_articles_details
-                    WHERE ordernumber = ?
-                    OR ordernumber LIKE ?
+                    WHERE (ordernumber = ?
+                    OR ordernumber LIKE ?) AND active = 1
                     GROUP BY articleID
                     LIMIT 2
                 ';


engine/Shopware/Controllers/Frontend/Search.php
```diff
@@ -135,7 +135,7 @@ protected function searchFuzzyCheck($search)
                 FROM s_articles_details
                   INNER JOIN s_articles
                    ON s_articles.id = s_articles_details.articleID
-                WHERE ordernumber = ?
+                WHERE ordernumber = ? AND s_articles.active = 1
```
shyim: s_articles_details.active = 1 is missing

Algorithman: Depends. I don't know what shopware shows if you link to a inactive variant. Doesn't it default to the next active one?
And if all are inactive, won't the article be inactive as well?

Algorithman: We don't work with variants here, so i don't have a quick way to check. I just stumbled over it today after a customer told us.

mitelg:
hey @Algorithman 
would you mind adding a test for this in https://github.com/shopware/shopware/blob/5.7/tests/Functional/Controllers/Frontend/SearchTest.php ?

Algorithman:
I wouldn't know where to start there, since it is a rare condition (where i'm not sure what the condition exactly is) for it to fail.
But i know that it obviously was missing there.

mitelg:
okay, we will see, what we can do :slightly_smiling_face:

mitelg:
hey @Algorithman 

after some discussion with my colleagues, we decided to not implement your PR in the current stage. The current behaviour of the search is intentional and informs customers about the deactivated product, if they search for the exact product number. 

But we would be fine, if you implement an event, which makes this behaviour changeable for you. We are thinking about a filter event in [line 179](https://github.com/shopware/shopware/blob/5.7/engine/Shopware/Controllers/Frontend/Search.php#L178-L179) right before the URL is assembled. 

We hope this is still okay for you :slightly_smiling_face: 
Thanks in advance :+1:

Algorithman:
I doubt that it is intentional to search through deactivated products, but if you want it like that, sure.
I for my part can live with the change as is, since we don't want to redirect to a deactivated product if there are 10 others who should have been shown as a list.

To reiterate:
This happens when you have multiple possible search results, but the first one is deactivated (and probably is a LIKE hit for the ordernumber while the others only hit on the article name for example). Then it will fail to produce a list of valid results.


Looking in our database here is a example what i mean:

1. Article: Ordernumber 2929.1645.35
2. Article: Ordernumber 10276.2021.10.01 (Articlename: 3052 2929 35 45-21)
3. Article: Ordernumber 10412.2021.10.04 (Articlename: 2004 2929 1634 42-24)
and so forth.

Searching for 2929, the first one will be tagged as the only one hitting the LIKE clause of the fuzzy search (but is deactivated).
The others only show up through secondary searches.

Since the first one is a single result you redirect directly to a deactivated product.

Seems the whole searchFuzzyCheck needs to be fixed/checked afterwards wether there are other search results if the product is deactivated.

mitelg:
thanks for your elaboration. I will check again tomorrow 👍

mitelg:
hey @Algorithman 
sorry for the late answer! :see_no_evil: 
We discussed this again, and we think that the search query with `LIKE` is not correct here. In our opinion this feature is for customers, who exactly know the order number of the products. so only an exact match should do a redirect to the product detail page. Having this in mind, it might become clearer, why deactivated products should still be considered :slightly_smiling_face: 

so, would removing this part, resolve your problem? 

```
            if (empty($products)) {
                $like_search = $search . '%';
                $sql = '
                    SELECT DISTINCT articleID
                    FROM s_articles_details
                    WHERE ordernumber = ?
                    OR ordernumber LIKE ?
                    GROUP BY articleID
                    LIMIT 2
                ';
                $products = $this->get('db')->fetchCol($sql, [$search, $like_search]);
            }
```
If so, we would like to ask you to remove this query :slightly_smiling_face: 

Thanks for your patience for us

mitelg:
Hey @Algorithman 

PR #2446, opened 2021-10-12
Avoid PHP8 error if the bucket in ES is null

xf-:
Avoid PHP error with v8

\Check if the value is null (array_filter needs an array)

Use ES, have an attribute without a value (my case a checkbox - filterable).

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundleES/FacetHandler/ProductAttributeFacetHandler.php
+++ b/engine/Shopware/Bundle/SearchBundleES/FacetHandler/ProductAttributeFacetHandler.php
@@ -155,9 +155,11 @@ class ProductAttributeFacetHandler implements HandlerInterface, ResultHydratorIn
 
             $type = $attribute ? $attribute->getColumnType() : null;
 
-            $aggregations[$key]['buckets'] = array_filter($aggregations[$key]['buckets'], function ($item) {
-                return $item['key'] !== '';
-            });
+            if ($aggregations[$key]['buckets']) {
+                $aggregations[$key]['buckets'] = array_filter($aggregations[$key]['buckets'], function ($item) {
+                    return $item['key'] !== '';
+                });
+            }
 
             if (\in_array($type, [TypeMappingInterface::TYPE_DATE, TypeMappingInterface::TYPE_DATETIME])) {
                 $aggregations[$key] = $this->formatDates($aggregations[$key]);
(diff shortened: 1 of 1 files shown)


engine/Shopware/Bundle/SearchBundleES/FacetHandler/ProductAttributeFacetHandler.php
```diff
@@ -155,9 +155,11 @@ public function hydrate(
 
             $type = $attribute ? $attribute->getColumnType() : null;
 
-            $aggregations[$key]['buckets'] = array_filter($aggregations[$key]['buckets'], function ($item) {
-                return $item['key'] !== '';
-            });
+            if ($aggregations[$key]['buckets']) {
```
mitelg:
```suggestion
            if (\is_array($aggregations[$key]['buckets'])) {
```
instead of an implicit check, we should explicitly check what is expected here

mitelg (changes requested):
would also be nice, if you manage to add a test for this :slightly_smiling_face:

engine/Shopware/Bundle/SearchBundleES/FacetHandler/ProductAttributeFacetHandler.php
xf-: Should work, but from my understanding it can only be `?array` and a simple & fast check should work.

xf-:
@mitelg any chance to get another php8 fix (plugin manager is broken) into next 5.7.4 release?

mitelg:
@xf- the broken plugin manager is already fixed and will be part of 5.7.4

xf-:
@mitelg is there a repo for automatic plugin check. I would love to add a regex check for templates.
https://developers.shopware.com/developers-guide/shopware-5-upgrade-guide-for-developers/#snippets

Some developers are ***** and arguing after a bug report, because of something like this `{s name=foo/bar}` isn't a bug, but it will result in '`foo' / 'bar'` and dividing strings is a bad idea.

I just saw the bug is fixed in plugin manager. I will switch development stage to PHP8 Maybe the sentry will catch something.

Would be nice to have finally a working PHP8 release 10 months into lifecycle and 2 months from PHP8.1 away

mitelg (approved):

mitelg:

PR #2447, opened 2021-10-13
Add command sw:system:config:set

jkrzefski:
This PR adds a command to configure basic settings. The signature looks like this:

```
sw:system:config:set [--shopId[=SHOPID]] [--] <name> <value>
```

In short: This command sets basic settings. You have to specify a configuration key and its new value. Optionally, you can also specify a shop-id. If you don't specify one, the default shop-id is assumed.

Internally the command creates an HttpKernel and dispatches a request to the `backend/config` controller. So this will call the exact same business logic as the backend would, including all the events that are dispatches during this controller action. This way, a maximum of compatibility is ensured.

```
bin/console sw:system:config:set shopName 'It works!'
```

None.

None.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Commands/ConfigSetCommand.php
+++ b/engine/Shopware/Commands/ConfigSetCommand.php
@@ -1,0 +1,96 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * Shopware 5
+ * Copyright (c) shopware AG
+ *
+ * According to our dual licensing model, this program can be used either
+ * under the terms of the GNU Affero General Public License, version 3,
+ * or under a proprietary license.
+ *
+ * The texts of the GNU Affero General Public License with an additional
+ * permission and of our proprietary license can be found at and
+ * in the LICENSE file you have received along with this program.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * "Shopware" is a registered trademark of shopware AG.
+ * The licensing of the program under the AGPLv3 does not imply a
+ * trademark license. Therefore any rights, title and interest in
+ * our trademarks remain entirely with us.
+ */
+
+namespace Shopware\Commands;
+
+use Exception;
+use Shopware\Components\Model\ModelManager;
+use Shopware\Models\Config\Element as ConfigElement;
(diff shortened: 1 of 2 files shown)


mitelg:
hey @jkrzefski 

We really appreciate your new feature, but we cannot accept how it was implemented. As you already might know this is very hacky :wink: 

Instead we would like to extract the need logic from the controller into a new service. This should be quite easy doable, as the concerning method is private: https://github.com/shopware/shopware/blob/5.7/engine/Shopware/Controllers/Backend/Config.php#L1392

With this done, you can also use this new service in your command and there is no need to boot a new kernel and create a dummy request and so on...

engine/Shopware/Commands/SystemConfigSetCommand.php
```diff
@@ -0,0 +1,110 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+class SystemConfigSetCommand extends ShopwareCommand
+{
+    protected function configure(): void
+    {
+        $this
+            ->setName('sw:system:config:set')
```
mitelg:
```suggestion
            ->setName('sw:config:set')
```

sounds like the SW6 command :grimacing:

jkrzefski:
@mitelg Thank you for your feedback 😄

Unfortunately, I don't see this task as that "easy". Extracting the logic from the controller into a dedicated service would normally be my go-to solution, but the way I see it, that would introduce a **breaking change**. The events `Shopware_Controllers_Backend_Config_Before_Save_Config_Element` and `Shopware_Controllers_Backend_Config_After_Save_Config_Element` both need an instance of the backend controller for their event arguments. (https://github.com/shopware/shopware/blob/5.7/engine/Shopware/Controllers/Backend/Config.php#L1450-L1466) A service could not possibly provide such an instance.

(I could probably make it work, but that would be even more "hacky" than this proposed solution.)

[I was told, we're not doing breaking changes in Shopware 5 anymore.](https://github.com/shopware/shopware/pull/2311#issuecomment-585827456) So what should we do? Is the only possible solution to not support the events when the command is used?

mitelg:
as discussed on Slack, only implement a very thin layer in the command to set the value to the element

jkrzefski:
@mitelg I changed the implementation as we discussed on Slack. There is one slight change I added: Instead of always using `json_decode` on the input value, this behavior can now be toggled with a new option `--decode`.

mitelg:
thanks for your contribution @jkrzefski  :+1: :tada: :blue_heart:

PR #2448, opened 2021-10-14
SW-26345 - fix sorting logic for categories

LePrau:
PHP8 changes the behaviour of usort() in a way that the implementation of category sorting is not working reliably.

It changes the result of the comparison from "greater than or equal" to "greater than or less than" resulting in categories always being sorted. Before PHP8 using usort, "equal" values could be shuffled (basically "luck" that it worked before), with PHP 8 equal values will always keep their relative order.

See ticket.

https://issues.shopware.com/issues/SW-26345

### 6. Checklist

- [ x ] I have not written any tests and if it gets stuck because of that:  ¯\_(ツ)_/¯
- [ x ] This change has no additional comments because obvious stuff is obvious
- [ x ] I have read the contribution requirements and am not sure if I fulfil them.


mitelg:
thanks for your contribution @LePrau :+1: 

but this is already fixed with https://github.com/shopware/shopware/pull/2444

PR #2452, opened 2021-10-19
add var sCustomPage to template paths

resslinger:
Actually, the data of a shop custom page, like attributes, will not be available in templates, if a template path (1-3) is configured.

This PR assign the var sCustomPage to the included tpls. So this templates have access to the other fields of the custom page, like the attributes.

Add a template path
Add a template var
=> The template has no access to the var sCustomPage

Add this line in this PR
Add a template path
Add a template var
=> The template has access to the var sCustomPage

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Custom.php
+++ b/engine/Shopware/Controllers/Frontend/Custom.php
@@ -56,7 +56,7 @@ class Shopware_Controllers_Frontend_Custom extends Enlight_Controller_Action
         if (!empty($staticPage['html'])) {
             $this->View()->assign('sContent', $staticPage['html']);
         }
-        
+
         $this->View()->assign('sCustomPage', $staticPage);
 
         for ($i = 1; $i <= 3; ++$i) {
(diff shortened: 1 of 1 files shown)


engine/Shopware/Controllers/Frontend/Custom.php
```diff
@@ -57,6 +57,8 @@ public function indexAction()
             $this->View()->assign('sContent', $staticPage['html']);
         }
 
+        $this->View()->assign('sCustomPage', $staticPage);
```
mitelg, line 60: the exact same thing happens in line 75

what does not work for you? :thinking:

mitelg (changes requested):

engine/Shopware/Controllers/Frontend/Custom.php
resslinger: We need the attributes of the called shop page in the included template files. The attributes are not available without this change. So I added the code in line 60, so the included templates have access to the attributes of the shop page.

mitelg: ah, now I see the problem :slightly_smiling_face: :+1: 

would it still work if we move the assignment to this place and remove it at the end of the method, so we do not have it duplicated?

resslinger: I will test it and will come back to you asap. :)

mitelg (approved):

mitelg:
thanks for your contribution @resslinger  👍🎉💙

PR #2453, opened 2021-10-19
Adapt conditional logic on cookie permission

torbenkorb:
In the jquery.cookie-permission.js a modal is opened only when cookieRemoval === 2. But on the hide method for this module the $.modal.close() method is called without any conditional logic which leads to a closing modal if you have added one from third party code and this modal uses Shopware built-in modal component. When the same logic for opening is used for closing it other initial modals can survive on the page. 

The logic for closing the modal is now exactly same for opening the modal, so only when cookieRemoval equals 2 the $.modal.close() method is called. 

Spin up an initial modal from a plugin or theme (e.g. age verification or newsletter popup) and configure Shopware to display cookie note and use cookie consent tool or only the simple note (setting equals window.cookieRemoval either 0 or 1). In this case your modal will get closed when you confirm any button in the cookie note. 

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.cookie-permission.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.cookie-permission.js
@@ -382,7 +382,9 @@
         hideElement: function() {
             this.$el.addClass(this.opts.isHiddenClass);
             $body.css('padding-bottom', 0);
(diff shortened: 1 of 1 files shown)


mitelg:
thanks for your contribution @torbenkorb  :+1: :tada: :blue_heart:

PR #2459, opened 2021-11-08
Fix: CLI product feed generation cacheDir path

xf-:
Before the property in class was null and tried to create a directory with path and  named `null`.

Set the path do existing variable.

Generate a Product feed on CLI

### 6. Checklist

### DIFF
--- a/engine/Shopware/Commands/GenerateProductFeedCommand.php
+++ b/engine/Shopware/Commands/GenerateProductFeedCommand.php
@@ -108,9 +108,9 @@ class GenerateProductFeedCommand extends ShopwareCommand implements CompletionAw
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $this->output = $output;
-        $cacheDir = $this->container->getParameter('shopware.product_export.cache_dir');
+        $this->cacheDir = $this->container->getParameter('shopware.product_export.cache_dir');
 
-        if (!\is_string($cacheDir)) {
+        if (!\is_string($this->cacheDir)) {
             throw new RuntimeException('Parameter shopware.product_export.cache_dir has to be a string');
         }
 


mitelg:
hey @xf- 

I added your PR to the existing ticket, we already have on our board: https://issues.shopware.com/issues/SW-26389

mitelg:
thanks again for your contribution @xf-  👍🎉💙

sedlareg:
This commit was wrong. I opened an issue
https://issues.shopware.com/issues/SW-26421

xf-:
@sedlareg what is wrong?!
This is an unreleased commit and your issue states you are using 5.7.6...

I think this fixes the issue....

sedlareg:
Look at the Issue. I have explained everything. You assign the dir to a local variable $cacheDir. And the class later uses a private variable $this->cacheDir. So this private variable will always be empty. That is Not correctAm 01.12.2021 17:46 schrieb xaver ***@***.***>:
@sedlareg what is wrong?!

xf-:
@sedlareg you are just unable to read code/diffs or understand git & releases.

I remove the assignment for `$cacheDir` and use ` $this->cacheDir`.

You complain about old code base
https://github.com/shopware/shopware/blob/v5.7.6/engine/Shopware/Commands/GenerateProductFeedCommand.php

and the green stuff is added in my commit and 5.7 branch
https://github.com/shopware/shopware/blob/5.7/engine/Shopware/Commands/GenerateProductFeedCommand.php

sedlareg:
Sorry. No need to be so harsh.So why is the error still in my Code base?I updated today to 5.7.6 and the error is still in the Release zip ;)Am 01.12.2021 18:31 schrieb xaver ***@***.***>:
@sedlareg you are just unable to read code/diffs or understand git & releases.
I remove the assignment for $cacheDir and use  $this->cacheDir.
You complain about old code base
https://github.com/shopware/shopware/blob/v5.7.6/engine/Shopware/Commands/GenerateProductFeedCommand.php
and the green stuff is added in my commit and 5.7 branch
https://github.com/shopware/shopware/blob/5.7/engine/Shopware/Commands/GenerateProductFeedCommand.php

mitelg:
@sedlareg this fix is not released yet 🙂
it will be part of the upcoming 5.7.7 release.

xf-:
@sedlareg I wrote in my first comment, this is unreleased and this PR fixes the issue. You wrote back, look in the issue and it describe4s 1-to-1 this issue and the diff is clearly a fix. You should always compare a tag and not a branch. If 2 versions are supported, backports will happen and a release to a later date.

There is also a second fix for feed with timeout issue on console. Console has normally 0 (unlimited) and in that case 30 seconds is used and that is far to short for a large shop. If you want to patch stuff in your system, you should apply that as well.

PR #2460, opened 2021-11-08
Fix: Support CLI exports

xf-:
A normal setup has no time limit on CLI and value is 0. This breaks larger shops or slow servers.

Support normal CLI behaviour.

Use a large Shop and export via cron.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sExport.php
+++ b/engine/Shopware/Core/sExport.php
@@ -2014,7 +2014,7 @@ class sExport implements Enlight_Hook
     private function ensurePHPTimeLimit(): void
     {
         $maxExecutionTime = (int) ini_get('max_execution_time');
-        if ($maxExecutionTime >= 30) {
+        if ($maxExecutionTime <= 0 || $maxExecutionTime >= 30) {
             return;
         }
 


xf-:
Additional information:
https://www.php.net/manual/en/info.configuration.php#ini.max-execution-time
**max_execution_time int**
> This sets the maximum time in seconds a script is allowed to run before it is terminated by the parser. This helps prevent poorly written scripts from tying up the server. The default setting is 30.  **When running PHP from the command line the default setting is 0**.

mitelg:
as in the linked PR mentioned, we should also consider `-1` as value

xf-:
@mitelg I thought about using `<= 0`,  but normally only memory uses `-1` and found it in no documentation. Looks more like something from somebody without reading the manual.

Also, I find your solution really strange with hardcoded time check. If I run it as Backend URL cron thingy, the backend timeout (configured) in Backend is applied. Why not use that value as minimum> If the value is lower, set the Backend (except for endless).

image

xf-:
OK — WTF

I checked the code and PHP Timeout is the session garbage collection time in code and has nothing todo with PHP timeout.

xf-:
@mitelg any ETA for an update (I know you have no ETAs). This breaks the Exports to Facebook, Merchant Center in front of Black Friday and holiday season.

mitelg:
hey @xf- 

your fix will definitely part of the next release, but we also want to fix some other things before releasing the next update. Most likely it will happen in December first

mitelg:
thanks for your contribution @xf-  👍🎉💙

PR #2461, opened 2021-11-08
Update sExport.php to also support no timelimit

BigAndini:
When no timelimit is set the sExport.php sets the timelimit to 30 seconds due to max_execution_time = 0 or = -1 within the php.ini results in $maxExecutionTime >= 30 to be false.

### 6. Checklist


mitelg:
hey @BigAndini 
thanks for your contribution 👍
but @xf- was a bit faster 😉 https://github.com/shopware/shopware/pull/2460
but we will consider your hint with `-1` as value

PR #2463, opened 2021-11-23
NTR - update text snippet for php8.0 compatibility

teiling88:
create a basket which triggers this error message. 

It change the argument order for implode. Since php8.0 the not official supported way of implode($array, 'string'); is not possible anymore.

Add an article to the basket
remove all category bindings
refresh the cart

### 6. Checklist


mitelg:
thanks for your contribution @teiling88 👍🎉💙

PR #2465, opened 2021-11-30
Fix analytics csv export strlen() argument

Nevay:
`TypeError: strlen(): Argument #1 ($str) must be of type string, array given`

Passes correct argument to `strlen()`.

Backend: Marketing -> Analysis -> Analysis -> Item detail viewing (/ any statistic with shop selection) -> select shop -> Export (as csv)

### 6. Checklist


mitelg:
After almost eight years, this bug is finally fixed :see_no_evil: 
thanks for your contribution @Nevay  👍🎉💙

PR #2466, opened 2021-12-02
added author url to blogposting

christopherholst:
added author url to blogposting because its missing for google

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/blog/detail.tpl
+++ b/themes/Frontend/Bare/frontend/blog/detail.tpl
@@ -45,7 +45,10 @@
                                 {* Author *}
                                 {block name='frontend_blog_detail_author'}
                                     {if $sArticle.author.name}
-                                        <span class="blog--metadata-author blog--metadata is--first" itemprop="author" itemscope itemtype="https://schema.org/Person">{s name="BlogInfoFrom"}{/s}: <span itemprop="name">{$sArticle.author.name}</span></span>
+                                        <span class="blog--metadata-author blog--metadata is--first" itemprop="author" itemscope itemtype="https://schema.org/Person">
+                                        <meta itemprop="url" content="{url controller=blog action=detail sCategory=$sArticle.categoryId blogArticle=$sArticle.id}">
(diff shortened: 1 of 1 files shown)


themes/Frontend/Bare/frontend/blog/detail.tpl
```diff
@@ -45,7 +45,10 @@
                                 {* Author *}
                                 {block name='frontend_blog_detail_author'}
                                     {if $sArticle.author.name}
-                                        <span class="blog--metadata-author blog--metadata is--first" itemprop="author" itemscope itemtype="https://schema.org/Person">{s name="BlogInfoFrom"}{/s}: <span itemprop="name">{$sArticle.author.name}</span></span>
+                                        <span class="blog--metadata-author blog--metadata is--first" itemprop="author" itemscope itemtype="https://schema.org/Person">
+                                        <meta itemprop="url" content="{url controller=blog action=detail sCategory=$sArticle.categoryId blogArticle=$sArticle.id}">
```
mitelg, line 49: hey @christopherholst 

do you have more information about that? any link to google docs or something similar? 

we are wondering why you want to add the URL of the blog post to the author. We think, that _if_ we want to add an URL here, it has to lead to a page, where the author is presented, hasn't it?

christopherholst: moin @mitelg  
that would surely be the best solution to lead directly to an author page, but Google does not make any specifications here and we would not have an area for this in the shopware core either. I tested it exactly with it and google no longer has any problems with it.

mitelg: I guess Google just checks if it is a valid URL, so it could be anything :slightly_smiling_face: 

but looking at it in a semantically way it would be wrong to point to the blog post here, right? 

so the question is: what is worse? having no URL at all, or pointing to a wrong page?

what about adding a block here, where it could individually handled what should be taken as URL. with this, one is able to create author pages with the content types e.g.

mitelg:
Hey @christopherholst 

PR #2467, opened 2021-12-18
Fix issue with selected_articles field in article slider component

dimitri-koenig:
Fetching "selected_articles" config should return a string, which will be given over to "explode". But default value is an empty array. That results in an error.


PascalThesing:
Hey @dimitri-koenig,

thank you for your pull-request.
We will merge as soon as possible.

mitelg:
thanks for your contribution @dimitri-koenig  👍🎉💙

PR #2468, opened 2021-12-29
Set empty string as default for missing password reset hash

mdjdev:
Completly removing the hash parameter from the password reset url results in a server error.

Set the default for the password hash to an empty string so it can't be NULL anymore.
An empty string is than handled by an (already existing) exception.

Call resetPasswordAction in Account frontend controller without hash parameter ("/account/resetPassword").
An empty hash ("/account/resetPassword/hash") or invalid hashes ("/account/resetPassword/hash/abc") were already handled correctly. Server error only occurs if hash parameter is missing, since getCustomerByResetHash() method does not accept NULL as parameter.

/

/

### 6. Checklist


mitelg:
thanks for your contribution @mdjdev  👍🎉💙

PR #2473, opened 2022-01-10
Fix ES with product attribute boolean

xf-:
DBAL wants `1` or `0` as value and doesn't work with `true` or `false`.
ES is the other way around. Switching between then, needs Manuel backend changes from an editor.
Without the search ends in an exception.

ES v7.10.1

Cast bool vales 1/0 to true or false

Add a product attribute and choose type boolean and switch between DBAL and ES

-

-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/SearchBundle/Condition/ProductAttributeCondition.php
+++ b/engine/Shopware/Bundle/SearchBundle/Condition/ProductAttributeCondition.php
@@ -39,7 +39,7 @@ class ProductAttributeCondition implements ConditionInterface, JsonSerializable
     protected $field;
 
     /**
-     * @var array|int|string|null
+     * @var array|boolean|int|string|null
      */
     protected $value;
 
@@ -86,7 +86,7 @@ class ProductAttributeCondition implements ConditionInterface, JsonSerializable
     }
 
     /**
-     * @return array|int|string|null $value
+     * @return array|boolean|int|string|null
      */
     public function getValue()
     {
@@ -94,7 +94,7 @@ class ProductAttributeCondition implements ConditionInterface, JsonSerializable
     }
 
     /**
-     * @param array|int|string|null $value
+     * @param array|boolean|int|string|null $value
      */
     public function setValue($value)
     {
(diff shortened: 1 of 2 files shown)


xf-:
Will push after lunch. (Used local working code, no clue why code IDE didn't complained about the type)

```
            if ($type === 'boolean') {
                $criteriaPart->setValue($criteriaPart->getValue() ? 'true' : 'false');
            }
```

mitelg:
just add bool as allowed type

xf-:
> just add bool as allowed type

OKi chnage the hits.
I wanted to do less impactful implementation (only ES gets touched), no clue what DBAL would do with boolean.

engine/Shopware/Bundle/SearchBundle/Condition/ProductAttributeCondition.php
```diff
@@ -86,15 +86,15 @@ public function setField($field)
     }
 
     /**
-     * @return array|int|string|null $value
```
xf-, line 89: return has no variable name

engine/Shopware/Bundle/SearchBundleES/ConditionHandler/ProductAttributeConditionHandler.php
```diff
@@ -97,6 +97,10 @@ private function getQuery(ProductAttributeCondition $criteriaPart)
             $attribute = $this->attributeService->get('s_articles_attributes', $criteriaPart->getField());
             if ($attribute instanceof ConfigurationStruct) {
                 $type = $attribute->getElasticSearchType()['type'];
+                
+                if ($type === 'boolean') {
+                    $criteriaPart->setValue((bool) $criteriaPart->getValue());
```
PascalThesing, line 102: Hey, first of all, thank you for taking the time for creating the pull request.
Since you already adjusted the bool value here, you still have to convert the bool value to the corresponding string in the Elasticsearch function in line 109.

Moreover, it would be wonderful if you would add a test for this case!

xf-: I think this should be fine. ES has null and default is false or set a null value. So null will be false after the cast.
https://www.elastic.co/guide/en/elasticsearch/reference/current/boolean.html

mitelg:
and thanks again for your contribution @xf-  :wink:  👍🎉💙

PR #2474, opened 2022-01-10
Update tinymce expected ,

christopherholst:
Expected , was missing.

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/base/component/Shopware.form.field.TinyMCE.js
+++ b/themes/Backend/ExtJs/backend/base/component/Shopware.form.field.TinyMCE.js
@@ -149,7 +149,7 @@ Ext.define('Shopware.form.field.TinyMCE', {
             theme_advanced_buttons1 : "save,newdocument,|,bold,italic,underline,strikethrough,|,justifyleft,justifycenter,justifyright,justifyfull,|,formatselect,fontselect,fontsizeselect",
             theme_advanced_buttons2 : "cut,copy,paste,pastetext,pasteword,|,search,replace,|,bullist,numlist,|,outdent,indent,blockquote,|,undo,redo,|,link,unlink,anchor,image,cleanup,help,code",
             theme_advanced_buttons3 : "tablecontrols,|,hr,removeformat,visualaid,|,sub,sup,|,charmap,emotions,iespell,media,advhr,ltr,rtl,|,fullscreen",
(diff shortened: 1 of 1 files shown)


themes/Backend/ExtJs/backend/base/component/Shopware.form.field.TinyMCE.js
```diff
@@ -149,7 +149,7 @@ Ext.define('Shopware.form.field.TinyMCE', {
             theme_advanced_buttons1 : "save,newdocument,|,bold,italic,underline,strikethrough,|,justifyleft,justifycenter,justifyright,justifyfull,|,formatselect,fontselect,fontsizeselect",
             theme_advanced_buttons2 : "cut,copy,paste,pastetext,pasteword,|,search,replace,|,bullist,numlist,|,outdent,indent,blockquote,|,undo,redo,|,link,unlink,anchor,image,cleanup,help,code",
             theme_advanced_buttons3 : "tablecontrols,|,hr,removeformat,visualaid,|,sub,sup,|,charmap,emotions,iespell,media,advhr,ltr,rtl,|,fullscreen",
-            theme_advanced_buttons4 : "styleprops,|,cite,abbr,acronym,del,ins,attribs,|,visualchars,nonbreaking,template,pagebreak,|,insertdate,inserttime,preview,|,forecolor,backcolor,|,media_selection"
+            theme_advanced_buttons4 : "styleprops,|,cite,abbr,acronym,del,ins,attribs,|,visualchars,nonbreaking,template,pagebreak,|,insertdate,inserttime,preview,|,forecolor,backcolor,|,media_selection",
```
mitelg, line 152: why is this necessary?

mitelg: it is surrounded by smarty and so this will be at the end of the object

christopherholst: i got this error directly from phpstorm with php 8, but you are right outside is a if and then a else comment but phpstorm doesnt recognizes this correct. in this case i think its not false. but can phpstorm be set so that it does not see it as an error?

mitelg: I guess this is not possible because it can not evaluate the smarty code :thinking: 

but as of this, I will close this PR

PR #2475, opened 2022-01-17
Add regex to naming of content types

xf-:
Chars like `-` causing errors in SQL table or `_` in ExtJS backend

Add a regex for XML

Create a content type with - or _ in the name

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/ContentTypeBundle/Resources/contenttypes.xsd
+++ b/engine/Shopware/Bundle/ContentTypeBundle/Resources/contenttypes.xsd
@@ -18,7 +18,13 @@
 
     <xsd:complexType name="singleType">
         <xsd:sequence>
-            <xsd:element name="typeName" type="xsd:string"/>
+            <xsd:element name="typeName">
+                <xsd:simpleType>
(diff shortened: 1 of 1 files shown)


mitelg:

daniel-memo-ict:
@mitelg This change is great, but we have content types with numbers in the name (not as first character) which now no longer work because of this restrictive regex.

bforbird:
@mitelg The goal to prevent - or _ is nice, but we have plugins with names in contenttypes containing numbers like "b2b".
Can you change the regex to allow numbers as well?

mitelg:
@daniel-memo-ict @bforbird please provide another PR or create a ticket for that :slightly_smiling_face:

xf-:
I tried to keep the Regex simple and easy to merge and to be on safe side regarding further problems.
SQL + PHP + JS (and maybe somebody adds a CSS class) and it works. It is only a technical name. `A-z` can be used without problems including one character technical name.

B2B is the same as bToB or simply businessToBusiness and no shortened stuff. Saving 1 character other understandable technical names.

(I have no problem if somebody creates a PR, as long I don't get any issues from possible names)

PR #2477, opened 2022-01-21
Fix issue with NULL values within implode

dimitri-koenig:
In the Newsletter Backend Controller, when sending emails, it's possible to run into an PHP Fatal error when implode is given NULL instead of an array.
Interestingly, later on in the code it's being checked for emptiness, so it's possible for that array to be empty, but it's not starting with an empty array. why I don't know.

Without this fix a Newsletter cannot be sent out.

Current Versions in use: Shopware 5.7.6 and Intelligent Newsletter 3.4.0

### 6. Checklist

### DIFF
--- a/.phpstan-baseline.neon
+++ b/.phpstan-baseline.neon
@@ -29192,11 +29192,6 @@ parameters:
 			path: engine/Shopware/Controllers/Backend/Newsletter.php
 
 		-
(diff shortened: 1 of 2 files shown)


PascalThesing:
Hey @dimitri-koenig, 

Thank you for your efforts in creating this pull request. Your changes were just merged into the 5.7 branch and will be introduced with our next version. 

Congratulations for this 🎉

PR #2478, opened 2022-01-30
refactor: Replace plain sql of getMaxTax by querybuilder and add event

aragon999:
Currently there is no way to modify the SQL which is used to retrieve the max tax. In order to extend there a replace hook is needed.

Refactor the plain SQL as query builder and supply this query builder in a new `notify` event: `Shopware_Modules_Basket_GetMaxTax_QueryBuilder`.

\-

\-

\-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -666,19 +666,37 @@ class sBasket implements \Enlight_Hook
     public function getMaxTax()
     {
         $sessionId = $this->session->get('sessionId');
+        if (!\is_string($sessionId)) {
+            return false;
+        }
 
-        $sql = <<<SQL
-SELECT a.taxID
-FROM s_order_basket b
-JOIN s_articles a ON a.id = b.articleID
-WHERE b.sessionID = ? AND b.modus = 0
-ORDER BY b.tax_rate DESC LIMIT 1;
-SQL;
-        $maxTaxId = $this->connection->fetchColumn($sql, [empty($sessionId) ? session_id() : $sessionId]);
+        $qb = $this->connection->createQueryBuilder();
+        $qb
+            ->select(['product.taxID'])
+            ->from('s_order_basket', 'basket')
+            ->join('basket', 's_articles', 'product', 'product.id = basket.articleID')
+            ->where($qb->expr()->andX(
+                $qb->expr()->eq('basket.sessionID', ':sessionId'),
+                $qb->expr()->eq('basket.modus', CartPositionsMode::PRODUCT)
+            ))
+            ->orderBy('basket.tax_rate', 'DESC')
+            ->setMaxResults(1)
+            ->setParameter('sessionId', $sessionId)
+        ;
 
+        $this->eventManager->notify(
+            'Shopware_Modules_Basket_GetMaxTax_QueryBuilder',
+            [
+                'sessionId' => $sessionId,
+                'queryBuilder' => $qb,
+            ]
+        );
+
+        $maxTaxId = $qb->execute()->fetchOne();
         if (!$maxTaxId) {
             return false;
         }
+
         $tax = $this->contextService->getShopContext()->getTaxRule($maxTaxId);
 
         return $tax->getTax();


engine/Shopware/Core/sBasket.php
```diff
@@ -666,19 +667,41 @@ public function sInsertPremium()
     public function getMaxTax()
     {
         $sessionId = $this->session->get('sessionId');
+        if (!$sessionId) {
```
mitelg:
```suggestion
        if (!is_string($sessionId)) {
```

engine/Shopware/Core/sBasket.php
```diff
@@ -666,19 +667,41 @@ public function sInsertPremium()
...
+            ))
+            ->orderBy('basket.tax_rate', 'DESC')
+            ->setMaxResults(1)
+            ->setParameter('sessionId', $sessionId)
+        ;
+
+        $this->eventManager->notify(
+            'Shopware_Modules_Basket_GetMaxTax_QueryBuilder',
+            [
+                'subject' => $this,
```
mitelg: please remove the reference to the class. If we want to move the method into an own service, we will get a problem, if we still need to reference the old core class :wink:

engine/Shopware/Core/sBasket.php
```diff
@@ -666,19 +667,41 @@ public function sInsertPremium()
...
+                'sessionId' => $sessionId,
+                'queryBuilder' => $qb,
+            ]
+        );
+
+        $statement = $qb->execute();
+        \assert($statement instanceof ResultStatement);
```
mitelg: this should not be necessary. also asserts made some troubles in the past :grimacing:

engine/Shopware/Core/sBasket.php
```diff
@@ -666,19 +667,41 @@ public function sInsertPremium()
...
 
+        $maxTaxId = $statement->fetchColumn();
```
mitelg:
```suggestion
        $maxTaxId = $statement->fetchOne();
```
`fetchColumn` is deprecated

mitelg (changes requested):

engine/Shopware/Core/sBasket.php
aragon999: I usually have a phpstan error with such things, that is why I added the assert here, since the `execute` method can return an `int`. But actually I noticed that is not very good, how I have done it, since `ResultStatement` does not has the property `fetchOne`, thus the next "issue" with this PR. Since I did not know how to replace `fetchColumn` on an object which does not has `fetchOne`.

TL;DR: I will look into it, but would be happy to get input on how to handle the ambiguous return values.

engine/Shopware/Core/sBasket.php
```diff
@@ -666,19 +667,41 @@ public function sInsertPremium()
...
-SQL;
-        $maxTaxId = $this->connection->fetchColumn($sql, [empty($sessionId) ? session_id() : $sessionId]);
+        $qb = $this->connection->createQueryBuilder();
+        $qb
+            ->select(['product.taxID'])
+            ->from('s_order_basket', 'basket')
+            ->join('basket', 's_articles', 'product', 'product.id = basket.articleID')
+            ->where($qb->expr()->andX(
+                $qb->expr()->eq('basket.sessionID', ':sessionId'),
+                $qb->expr()->eq('basket.modus', '0')
```
mitelg:
```suggestion
                $qb->expr()->eq('basket.modus', \Shopware\Bundle\CartBundle\CartPositionsMode::PRODUCT)
```

engine/Shopware/Core/sBasket.php
mitelg: this is a known issue: https://github.com/shopware/shopware/blob/5.7/.phpstan.neon#L92-L93

just omit the `$statement variable` and PHPStorm should auto-complete the right method :slightly_smiling_face:

aragon999: I see, I was not aware that there exists such a thing in the baseline :)

aragon999: Btw. I do not use PHPStorm :) but the language server works as well.

The problem was, that I forgot to run a composer install... and therefore had a too old version of `doctrine/dbal` installed. But now it should be all solved :) Thanks for the feedback.

mitelg (approved):

PascalThesing:
Hey @aragon999, 

We just merge your changes.
Thank you for taking the time to create and adapt it! 

Congratulations 🎉

PR #2484, opened 2022-02-21
SW-26162 fix possible side effects in swTabMenu

mjossdev:
When switching tabs the is--active class is removed from element that has it, because jQuery's `find` function gets all descendants (and not just direct children) which match a selector. So if there are any elements with the `is--active` class in a tabs content, it will also be removed, which is probably not the intention.

Instead of using `find`, it calls `removeClass` on just the tab and container elements, which is fine because `removeClass` does not require the class to be set on the element.

With the following HTML the `is--active` class on the innermost div will be removed once the swTabMenu is initialized:
```
<div class="js--tab-menu tab-menu--cross-selling">
  <div class="tab--navigation">
    <a class="tab--link has--content is--active">
      Test
    </a>
  </div>
  <div class="tab--container-list">
    <div class="tab--container has--content is--active">
      <div class="tab--content">
        <div class="is--active">Test</div>
      </div>
    </div>
  </div>
</div>
```

https://issues.shopware.com/issues/SW-26162

none

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.tab-menu.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.tab-menu.js
@@ -206,15 +206,11 @@
             $tab = $(me.$tabs.get(index));
             $container = $(me.$container.get(index));
 
-            me.$tabContainer
-                .find('.' + activeTabClass)
-                .removeClass(activeTabClass);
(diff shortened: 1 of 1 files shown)


mitelg:
hey @mjossdev 

I added your PR to the existing ticket you linked. :+1:

mitelg:
thanks for your contribution @mjossdev 👍🎉💙

PR #2485, opened 2022-02-28
Add mainnumber to product export

xf-:
I needed or wanted the main number in multiple instances

Google Shopping (yeas main variant ID works as workaround, but is and inaccessible ID for backend user and unfamiliar)
https://support.google.com/merchants/answer/6324507?hl=en

Shopware in/export works with that ID and export is for 3rd-Party service that enriches Product Data the easiest solution - on Demand an up-to-date export. For import the main number is needed.

Add a join on main detail. Tested with ten thousands and noticed no performance impact. The table is already used in that query.

Want an export with a mainnumber

Add field to field list

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sExport.php
+++ b/engine/Shopware/Core/sExport.php
@@ -953,6 +953,7 @@ class sExport implements Enlight_Hook
 
                 d.id as `articledetailsID`,
                 IF(v.ordernumber IS NOT NULL,v.ordernumber,d.ordernumber) as ordernumber,
+                main.ordernumber as mainnumber,
 
                 d.suppliernumber,
                 d.ean,
@@ -960,7 +961,7 @@ class sExport implements Enlight_Hook
                 d.height,
                 d.length,
                 d.kind,
-                IF(v.standard=1||kind=1,1,0) as standard,
+                IF(v.standard=1||d.kind=1,1,0) as standard,
                 d.additionaltext,
                 COALESCE(sai.impressions, 0) as impressions,
                 d.sales,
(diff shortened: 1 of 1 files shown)


engine/Shopware/Core/sExport.php
```diff
@@ -953,14 +953,15 @@ public function sCreateSql()
...
+                main.ordernumber as mainnumber,
 
                 d.suppliernumber,
                 d.ean,
                 d.width,
                 d.height,
                 d.length,
                 d.kind,
-                IF(v.standard=1||kind=1,1,0) as standard,
+                IF(v.standard=1||d.kind=1,1,0) as standard,
```
xf-, line 964: Was missing - query had an issue, because main has kind aswell

xf-:
@mitelg Any chance of getting this into core? (And I need to check out the test issue or is not worth the trouble)

No clue why tests failed. I have also a version with a subquery. Performance was the same. Should be more contained and the issue disappear

mitelg:
your changes look okay for me :+1: 

I guess the tests are failing because the output of the export now contains an additional value. I guess this is easy fixable. would be nice, if you could do that :blush:

xf-:
@mitelg is there any docu about how to run the tests. Ii have a lot of errors. Using demo Data and tried PHP 8.1 and 8.0
image
```
ERRORS!
Tests: 2979, Assertions: 16342, Errors: 171, Failures: 152, Skipped: 44, Risky: 3.
```

Output from CI is only an - and no +. Tried to look for a foreach or something like that.

mitelg:
just execute this make command: https://github.com/shopware/shopware/blob/5.7/Makefile#L70

`make test-phpunit`

xf-:
No clue, why inner join caused the different result an issue. I tested some exports locally in different projects with the patch and `diff` tool and was always the same result

mitelg:
thanks @xf- 

we will now continue :blush:

mitelg:
thanks for your contribution @xf-  👍🎉💙

PR #2486, opened 2022-03-08
explicitly allow /widgets/index/refreshStatistic in robots.txt to give crawlers the chance to read x-header-tag: noindex and remove already indexed pages

stfnschrmr:
Google Search Console is complaining that pages under `/widgets` are blocked through `robots.txt`. 
In relation to Googles documentation how to prevent pages from beeing index (https://developers.google.com/search/docs/advanced/crawling/block-indexing) `/widgets` should be removed from `robots.txt` – indeed it already has an `x-header: noindex` entry.

### DIFF
--- a/tests/Functional/Controllers/Frontend/RobotsTxtTest.php
+++ b/tests/Functional/Controllers/Frontend/RobotsTxtTest.php
@@ -60,6 +60,7 @@ SQL,
         static::assertArrayHasKey('Disallow: /de/account/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /de/address/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /de/note/', $robotsTxt);
+        static::assertArrayHasKey('Disallow: /de/widgets/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /de/listing/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /de/ticket/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /en/compare/', $robotsTxt);
@@ -68,6 +69,7 @@ SQL,
         static::assertArrayHasKey('Disallow: /en/account/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /en/address/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /en/note/', $robotsTxt);
+        static::assertArrayHasKey('Disallow: /en/widgets/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /en/listing/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /en/ticket/', $robotsTxt);
 
@@ -97,6 +99,7 @@ SQL,
         static::assertArrayHasKey('Disallow: /de/account/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /de/address/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /de/note/', $robotsTxt);
+        static::assertArrayHasKey('Disallow: /de/widgets/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /de/listing/', $robotsTxt);
         static::assertArrayHasKey('Disallow: /de/ticket/', $robotsTxt);
         static::assertArrayNotHasKey('Disallow: /en/compare/', $robotsTxt);
@@ -105,6 +108,7 @@ SQL,
         static::assertArrayNotHasKey('Disallow: /en/account/', $robotsTxt);
         static::assertArrayNotHasKey('Disallow: /en/address/', $robotsTxt);
         static::assertArrayNotHasKey('Disallow: /en/note/', $robotsTxt);
+        static::assertArrayNotHasKey('Disallow: /en/widgets/', $robotsTxt);
         static::assertArrayNotHasKey('Disallow: /en/listing/', $robotsTxt);
         static::assertArrayNotHasKey('Disallow: /en/ticket/', $robotsTxt);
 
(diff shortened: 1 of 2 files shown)


mitelg:
hey @createdwithlove 

could you please have a look at the failing tests? https://github.com/shopware/shopware/blob/5.7/tests%2FFunctional%2FControllers%2FFrontend%2FRobotsTxtTest.php

larsbo:
Which widgets modules exactly is Google complaining about? There are several modules that should explicitly not be indexed such as widgets/checkout or widgets/compare. Maybe a better approach would be to only allow specific modules like https://github.com/shopware/shopware/blob/3c3c171609f95409df9f1c86bf6d7d5d78dfbdd3/themes/Frontend/Bare/frontend/robots_txt/index.tpl#L26 instead of the whole widgets/ path?

stfnschrmr:
GSC is complaining about `/widgets/index/refreshStatistic`.

But to quote Google: 
> Important: For the noindex directive to be effective, the page or resource **must not** be blocked by a robots.txt file, and it has to be otherwise accessible to the crawler. If the page is blocked by a robots.txt file or the crawler can't access the page, the crawler will never see the noindex directive, and the page can still appear in search results, for example if other pages link to it.

That means, that if there are other `/widgets` that should not get indexed, they should have a proper `noindex` directive (`x-header` or `meta`) instead of dissallowing access for crawlers using `robots.txt`

philipreinken:
> GSC is complaining about `/widgets/index/refreshStatistic`.
> 
> But to quote Google:
> 
> > Important: For the noindex directive to be effective, the page or resource **must not** be blocked by a robots.txt file, and it has to be otherwise accessible to the crawler. If the page is blocked by a robots.txt file or the crawler can't access the page, the crawler will never see the noindex directive, and the page can still appear in search results, for example if other pages link to it.
> 
> That means, that if there are other `/widgets` that should not get indexed, they should have a proper `noindex` directive (`x-header` or `meta`) instead of dissallowing access for crawlers using `robots.txt`

Would you mind adding the header to the corresponding responses? Otherwise, the routes will still be listed by the google search engine (if I got this correctly), as they're not on the denylist anymore after this change.

stfnschrmr:
@philipgatzka as mentioned above that seems to be already done. Checked `/Shopware/Controller/Widgets/` – are there more?

https://github.com/shopware/shopware/blob/63500edcc1dbed4f8162e7f2eb0282d245eab991/engine/Shopware/Controllers/Widgets/Recommendation.php#L47

https://github.com/shopware/shopware/blob/63500edcc1dbed4f8162e7f2eb0282d245eab991/engine/Shopware/Controllers/Widgets/Listing.php#L49

https://github.com/shopware/shopware/blob/63500edcc1dbed4f8162e7f2eb0282d245eab991/engine/Shopware/Controllers/Widgets/Index.php#L34

https://github.com/shopware/shopware/blob/63500edcc1dbed4f8162e7f2eb0282d245eab991/engine/Shopware/Controllers/Widgets/Error.php#L32

https://github.com/shopware/shopware/blob/63500edcc1dbed4f8162e7f2eb0282d245eab991/engine/Shopware/Controllers/Widgets/Compare.php#L32

https://github.com/shopware/shopware/blob/63500edcc1dbed4f8162e7f2eb0282d245eab991/engine/Shopware/Controllers/Widgets/Checkout.php#L46

https://github.com/shopware/shopware/blob/63500edcc1dbed4f8162e7f2eb0282d245eab991/engine/Shopware/Controllers/Widgets/Captcha.php#L34

https://github.com/shopware/shopware/blob/63500edcc1dbed4f8162e7f2eb0282d245eab991/engine/Shopware/Controllers/Widgets/Campaign.php#L34
  https://github.com/shopware/shopware/blob/63500edcc1dbed4f8162e7f2eb0282d245eab991/engine/Shopware/Controllers/Widgets/Recommendation.php#L47

philipreinken:
Indeed, I didn't read your first comment here, my bad. I think all is fine then! 👍

philipreinken:
The tests are still failing, the error seems unrelated though, I scheduled another run.

philipreinken:
We just discussed this in the team - as it stands now, plugin widget controllers, which don't set the header themselves, would be indexed still, right? In this case, could you check, whether it's possible to apply the header to all widget controller responses, including plugin controllers?

stfnschrmr:
You're right, if the `x-header` isn't set, plugin widgets maybe get indexed.
If you're concerned that it might have unwanted side effects, feel free to decline the pull request.

For someone who deals with such little things, it shouldn't be a problem to adapt the `robots.txt` to their own needs on a project basis.

I will observe how Google deals with the adapted `robots.txt` in our shops and, if necessary, start a new pull request again later.

(Found a solved ticket https://issues.shopware.com/issues/SW-24373 – maybe someone in your team already dealt with that problem?)

larsbo:
> GSC is complaining about `/widgets/index/refreshStatistic`.
> That means, that if there are other `/widgets` that should not get indexed, they should have a proper `noindex` directive (`x-header` or `meta`) instead of dissallowing access for crawlers using `robots.txt`

I can confirm this situation. We had the same GSC issues some time agout and therefore I added
` {$robotsTxt->setAllow('/widgets/index/refreshStatistic')}`
in our theme to let the crawler found the noindex header.

I think the root problem was the [wrong](https://github.com/shopware/shopware/commit/774ff4b0358cdfd6f64665d8a38614d3213b61d9) `noindex` header as you mentioned in the issue some years ago. This resulted in wrong indexing and now the robot.txt disallow entries prevent any de-listing by the corrected noindex-header.

So the PR would fix older shops that still face this indexing issue but it would have no effect on newer shops that started with the `x-robots-tag`.

mitelg:
@createdwithlove @larsbo thank you very much for the discussion :blue_heart: :+1: 
I guess that we now have a good solution.

@createdwithlove would you mind changing the title and description of the PR so it is matching the change?

mitelg:
@createdwithlove @larsbo thank you very much for the discussion :blue_heart: :+1: 
I guess that we now have a good solution.

@createdwithlove would you mind changing the title and description of the PR so it is matching the change?

mitelg:
Hello,

thank you for creating this pull request.
I have opened an issue on our Issue Tracker for you. See the issue link: https://issues.shopware.com/issues/SW-26598

Please use this issue to track the state of your pull request.

mitelg:
thanks for your contribution @createdwithlove  👍🎉💙

PR #2488, opened 2022-03-15
Fix TypeError while using proportional taxes and tax rates with decimals (PHP8)

StVak:

It throws TypeError error while using proportional taxes and tax rates with decimals

Cast tax rate to float

- Enable Proportional taxes
- Set up a tax rate with decimal for a country (eg 5.5) 
- Assign this tax rate to a product
- Add this product to cart and go to cart

```
Uncaught TypeError: number_format(): Argument #1 ($num) must be of type float, string given
```

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Cart/ProportionalTaxCalculator.php
+++ b/engine/Shopware/Components/Cart/ProportionalTaxCalculator.php
@@ -41,7 +41,7 @@ class ProportionalTaxCalculator implements ProportionalTaxCalculatorInterface
 
         foreach ($sumByTaxes as $taxRate => $price) {
             if ((float) $price === 0.0) {
-                $taxes[] = new Price(0.0, 0.0, $taxRate, 0.0);
+                $taxes[] = new Price(0.0, 0.0, (float) $taxRate, 0.0);
                 continue;
             }
 
@@ -53,7 +53,7 @@ class ProportionalTaxCalculator implements ProportionalTaxCalculatorInterface
 
             $tax = $net * ($taxRate / 100);
 
-            $taxes[] = new Price($isNetPrice ? $net : $priceForTax, $net, $taxRate, $tax);
+            $taxes[] = new Price($isNetPrice ? $net : $priceForTax, $net, (float) $taxRate, $tax);
         }
 
         return $taxes;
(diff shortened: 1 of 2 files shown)


engine/Shopware/Components/Cart/TaxAggregator.php
```diff
@@ -119,7 +119,7 @@ public function shippingCostsTaxSum(array $cart): ?array
             /** @var callable(array<numeric-string, float> $carry, Price $shippingTax): array<numeric-string, float> $callback */
             $callback = static function (array $carry, Price $shippingTax) {
                 /** @var numeric-string $taxRate */
-                $taxRate = number_format($shippingTax->getTaxRate(), 2);
+                $taxRate = number_format((float) $shippingTax->getTaxRate(), 2);
```
mitelg: instead of fixing it here, we should fix the creation of the `\Shopware\Components\Cart\Struct\Price` struct.

please check where `\Shopware\Components\Cart\Struct\Price::__construct` is called and filled with wrong values :blush:

mitelg:
thanks for your first contribution @StVak :+1: 

I left you a comment :wink:

engine/Shopware/Components/Cart/TaxAggregator.php
StVak: ok!

engine/Shopware/Components/Cart/ProportionalTaxCalculator.php
```diff
@@ -53,7 +53,7 @@ public function calculate($discount, $prices, $isNetPrice)
 
             $tax = $net * ($taxRate / 100);
 
-            $taxes[] = new Price($isNetPrice ? $net : $priceForTax, $net, $taxRate, $tax);
+            $taxes[] = new Price($isNetPrice ? $net : $priceForTax, $net, (float) $taxRate, $tax);
```
mitelg, line 56: I can't comment there. but does it also work, if you only cast on line 122 in this class?

StVak: Nope, can't use float values as array keys...

mitelg: yeah, I know.
I mean after the array key stuff

StVak: hm, could you give me an example how you think about it? because `$key` must be string, the next step is to change the data inside the foreach loop

mitelg: I will check tomorrow again 👍

mitelg: sorry, I checked again, and your solution is okay :+1: 

was a bit late and not so clear on only mobile view of GitHub :sweat_smile:

mitelg:
merged with aaf0acf42b3566068071e56fd68bf7f335085671

thanks for your contribution @StVak  👍🎉💙

PR #2489, opened 2022-04-05
Skip Element from Media Normalisation if path key not exists

wrongspot:
At the Moment if an Emotion Element has a json Field, it is assumed that there always exists a key with the value path.

It checks before Media normalization if Array Key `path` exists, otherwise skip Media normalization. 

- Create Custom Emotion Element with a hidden json Field
- Try to Save Emotion World with this custom Element 
- Then the following StackTrace appears
```
2022/04/05 10:51:14 [error] 15898#15898: *70241 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught TypeError: Cannot access offset of type string on string in /var/webroot/vendor/shopware/shopware/engine/Shopware/Controllers/Backend/Emotion.php:1333
Stack trace:
#0 /var/webroot/vendor/shopware/shopware/engine/Shopware/Controllers/Backend/Emotion.php(1309): Shopware_Controllers_Backend_Emotion->processDataFieldValue()
#1 /var/webroot/vendor/shopware/shopware/engine/Shopware/Controllers/Backend/Emotion.php(1269): Shopware_Controllers_Backend_Emotion->createElementData()
#2 /var/webroot/vendor/shopware/shopware/engine/Shopware/Controllers/Backend/Emotion.php(1210): Shopware_Controllers_Backend_Emotion->createElements()
#3 /var/webroot/vendor/shopware/shopware/engine/Shopware/Controllers/Backend/Emotion.php(492): Shopware_Controllers_Backend_Emotio" while reading response header from upstream
```

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Emotion.php
+++ b/engine/Shopware/Controllers/Backend/Emotion.php
@@ -1325,6 +1325,9 @@ class Shopware_Controllers_Backend_Emotion extends Shopware_Controllers_Backend_
         if ($valueType === 'json') {
             if (\is_array($value)) {
                 foreach ($value as &$val) {
+                    if (!isset($val['path'])) {
+                        continue;
+                    }
                     $val['path'] = $mediaService->normalize($val['path']);
                 }
                 unset($val);


mitelg:
thanks for your contribution @wrongspot  👍🎉💙

we merged it with https://github.com/shopware/shopware/commit/fa39691bf02e374e390171f7e8db080cb32e845c on our `next` branch, which will be merged back to the main branch a while after the release :slightly_smiling_face:

PR #2490, opened 2022-04-14
update contact

christopherholst:
Der Rückgabe Typ ist im neuen nur als string erlaubt. 
Ohne Standardwert ist es aber 'null' aber es ist nur 'string' erlaubt, er versucht halt null zurück zugeben.
=====
The return type is only allowed as a string in the new one.
With no default it's 'null' but only 'string' is allowed, it just tries to return null.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Models/Mail/Contact.php
+++ b/engine/Shopware/Models/Mail/Contact.php
@@ -47,7 +47,7 @@ class Contact extends ModelEntity
      *
      * @ORM\Column(name="mail_address", type="string", nullable=false)
      */
-    protected $mailAddress;
+    protected $mailAddress = '';
 
     public function getId(): ?int
     {


mitelg:
hey @christopherholst 

thanks for your PR.

when did such an error occur on your side? the mail address is required and should therefore not be an empty string. :thinking: 
Could you elaborate a bit more? 
either we missed a thing here, or you are using it wrong :wink:

christopherholst:
The problem was that in the s_mail_log_contact there was an ID that no longer had a mail address, since the customer no longer existed but mail entries. After deleting the mail entries in the log and removing the ID with the empty entry, it worked again.

mitelg:
so we should fix the removing of customers then? :thinking: 
maybe you can investigate further how those invalid entries could occur and fix that instead

mitelg:
as discussed on slack, we would like to fix the root cause of the problem and not implementing a workaround. 
I will close this PR for now, please provide another one, if you want to contribute another solution :slightly_smiling_face: :+1:

PR #2491, opened 2022-04-15
SW-26654 Use interface typehints for listing controller

TobiasGraml11:

- Using the original repository class, instead of the interface, removes the possibility to extend the repository
- There will be an exception, because PHP expects the original repository, instead of the extended one (or just the interface)
  image

It changes the typehint for the `ProductStream\Repository` to it's corresponding `ProductStream\RepositoryInterface`

Class from our plugin: https://github.com/findologic/plugin-shopware-5/blob/develop/FinSearchUnified/Components/ProductStream/Repository.php
Reproduction URL: https://shopware5.plugins.findologic.dev/food/

1. Create a custom product stream repository
2. Extend the default repository with your own
3. Open a navigation page

Change was done here: https://github.com/shopware/shopware/commit/7caa7080a8fd82a039e977af143fe628026cdb11

### 6. Checklist


mitelg:
thanks for your contribution @TobiasGraml11  👍🎉💙

PR #2492, opened 2022-04-21
 Add event to advanced menu plugin

larsbo:

Because there is no good entry point to extend the advanced menu data.

Adds a new event `Shopware_Plugins_AdvancedMenu_ConvertCategory`. 

Try to add thumbnails to media entries and realize there is no simple ways to append the thumbnail details.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Bootstrap.php
@@ -351,8 +351,9 @@ class Shopware_Plugins_Frontend_AdvancedMenu_Bootstrap extends Shopware_Componen
     private function convertCategories(array $categories): array
     {
         $converter = Shopware()->Container()->get(LegacyStructConverter::class);
+        $eventManager = Shopware()->Container()->get('events');
 
-        return array_map(function (Category $category) use ($converter) {
+        return array_map(function (Category $category) use ($converter, $eventManager) {
             $data = $converter->convertCategoryStruct($category);
 
             $data['flag'] = false;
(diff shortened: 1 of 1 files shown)


engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Bootstrap.php
```diff
@@ -363,6 +363,11 @@ private function convertCategories(array $categories): array
                 $data['link'] = $category->getExternalLink();
             }
 
+            $eventManager = Shopware()->Container()->get('events');
```
mitelg: please move this out of the loop

engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Bootstrap.php
```diff
@@ -363,6 +363,11 @@ private function convertCategories(array $categories): array
...
+            $data = $eventManager->filter('Shopware_Plugins_AdvancedMenu_ConvertCategory', $data, [
```
mitelg: maybe one event after the loop is enough for your purpose? I don't know if this could have performance impacts, if we start an event for each category

mitelg:
hey @larsbo 

your PR looks good so far, but I have some comments :wink:

larsbo:
@mitelg what comments? 😃

mitelg (changes requested):

mitelg:
wow, they were still pending 🤦‍♂️ 
sorry for that!

engine/Shopware/Plugins/Default/Frontend/AdvancedMenu/Bootstrap.php
larsbo: sure, one event is sufficient.

larsbo: I think about that, too. But the main reason for an event inside the loop is to reuse the Media instance. If we have just an event after the loop all the (legacy) struct converting stuff has to be done again or you have to query the db again for additional media details that are lost after the loop.

mitelg: okay, I guess it would be okay. we will see :grin:

mitelg:

the changes were pushed on the next branch and will be merged back, as soon we are stable again :wink:

PR #2493, opened 2022-04-25
SW-26672 - Fix contact reference loading in mail log entry builder

lacknere:
PHP throws an error if a mail is sent to an address which has an existing contact record because the mail address is null in the entity object.

Screenshot 2022-04-25 113221

Changing the way of getting the referenced contact from partial to full because the mail address property is needed later on.
Out of the Doctrine ORM comment on EntityManagerInterface getPartialReference method:
\* The returned reference may be a partial object if the entity is not yet loaded/managed.
\* If it is a partial object it will not initialize the rest of the entity state on access.
\* Thus you can only ever safely access the identifier of an entity obtained through
\* this method.

1. Go to Settings > E-mail management > Email templates and select a mail template (e.g. sORDER).
2. Use the "Send test mail to shop owner" to send the mail.
3. Redo step 2 because this happens only if the mail contact already exists in the database.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/MailBundle/Service/LogEntryBuilder.php
+++ b/engine/Shopware/Bundle/MailBundle/Service/LogEntryBuilder.php
@@ -227,7 +227,7 @@ class LogEntryBuilder implements LogEntryBuilderInterface
         foreach ($knownRecipients as $recipient) {
             unset($unknownRecipients[$recipient['mail_address']]);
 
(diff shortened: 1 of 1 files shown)


mitelg:
will fix https://github.com/shopware/shopware/pull/2490

mitelg:
thanks for your contribution @lacknere  👍🎉💙

it is part of the recent 5.7.9 release

PR #2494, opened 2022-05-06
Fixed listing redirects

tvswe:
In case of a category has an external link, the method "getRedirectLocation" will return a string. So the return type declaration of this method is wrong and the shop will crash.

I rewrote the method to return only strings for redirects or null if there is no reditect.

Add an external link to a category and call it in the frontend. The shop will crash.

### 4. Checklist


PascalThesing:
Dear @tvswe,

Instead of removing just the return type.
Please fix the problem properly by adjusting the function call to return a common type and not mixed.

tvswe:
Hi @PascalThesing , I made the changes. Please review again.

mitelg:
thanks for your contribution @tvswe  👍🎉💙

PR #2499, opened 2022-06-07
Ignore inactive surcharge dispatches in exports

larsbo:
The product export module does not consider the status of surcharge dispatches and therefore product exports may be wrong or may fail completely because of wrong configured inactive dispatches.

Append an active column check to the surcharge dispatch sql query.

- Create a new dispatch wih type "Aufschlag-Versandregel (in Versandkosten eingerechnet)" and add an invalid custom condition like `foo=bar` to it. Let the dispatch inactive and save it.
- Start a new product export and the export will fail with an error caused by the invalid condition inside the inactive dispatch.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sExport.php
+++ b/engine/Shopware/Core/sExport.php
@@ -1709,7 +1709,7 @@ class sExport implements Enlight_Hook
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


mitelg:

PR #2500, opened 2022-06-09
SW-26453 - Fix translation update api

lacknere:
Because translation update api is not working as expected if a translation already exists. Updating an article translation via PUT for example: if a translation already exists, the existing translation is deleted but the new translation is not written to the database, so the article has no translation anymore.

Removing the unnecessary delete of the previous translation because the insert statement uses "ON DUPLICATE KEY UPDATE 'objectdata'=VALUES('objectdata')" anyways.

1. PUT bulk request of multiple article translations => translations are fine.
2. Do another PUT bulk request of the same article translations => translations are gone.

https://issues.shopware.com/issues/SW-26453

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Api/Resource/Translation.php
+++ b/engine/Shopware/Components/Api/Resource/Translation.php
@@ -432,7 +432,6 @@ class Translation extends Resource implements BatchInterface
             $existing = [];
         } else {
             $existing['data'] = unserialize($existing['data'], ['allowed_classes' => false]);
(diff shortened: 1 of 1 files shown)


mitelg:
hey @lacknere 

thanks for your PR. I assigned it to the mentioned ticket

mitelg:
thanks for your contribution @lacknere and fixing this nasty bug 👍🎉💙

it will be part of the upcoming 5.7.12 release

PR #2501, opened 2022-06-22
Change mPDF default fontDir to array

larsbo:
Because the fontDir variable is set to a string you cannot add additional font directories using Mpdf::AddFontDirectory().
This will cause an `[] operator not supported for strings` error.

Replace the fontDir string with an array containing the font dir.

Call `AddFontDirectory()` on the Shopware mPDF instance.

### 6. Checklist


mitelg:
merged with d27a42668ce3f3a49e9af51f33b9920c9bb57067
will be merged back into the main branch, in a few days after the release

PR #2503, opened 2022-07-29
SW-26876: History back paging fix

Ivan-Rud:
The wrong behaviour with product paging and browser history back button. Data in the listing doesn't change when you use the history back button in the browser and the browser history is duplicated after a few times click 

Additional window event to reload the listing products data when you go back/forward using the browser history button
Prevent browser history duplication

Disable infinite scrolling (theme manager)
Enable automatic reload for the filter (performance module)
Go to the second or next page and then try to go back with the browser history back button

https://issues.shopware.com/issues/SW-26876

-

### 6. Checklist

### DIFF
--- a/themes/Frontend/Responsive/frontend/_public/src/js/jquery.listing-actions.js
+++ b/themes/Frontend/Responsive/frontend/_public/src/js/jquery.listing-actions.js
@@ -329,6 +329,8 @@
             me.bufferTimeout = 0;
             me.closeFilterOffCanvasBtnText = me.$closeFilterOffCanvasBtn.html();
             me.closeFilterOffCanvasBtnTextWithProducts = me.$closeFilterOffCanvasBtn.attr('data-show-products-text');
+            
+            me.isBrowserHistoryButton = false;
 
             me.getPropertyFieldNames();
             me.setCategoryParamsFromTopLocation();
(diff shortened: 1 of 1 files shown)


themes/Frontend/Responsive/frontend/_public/src/js/jquery.listing-actions.js
```diff
@@ -1169,7 +1200,10 @@
 
             listing.removeClass(this.opts.isLoadingCls);
 
-            window.history.pushState('data', '', window.location.href.split('?')[0] + this.urlParams);
+            //prevent history item duplication
+            if(!this.isBrowserHistoryButton) {
```
mitelg, line 1204: we should not use a global state. instead a parameter should be passed down to here, from the methods called before.
```suggestion
            if(pushToHistory) {
```

mitelg: the function can then look like this:

```
         * @param {Object} response
         * @param {boolean} pushToBrowserHistory
         */
        updateListing: function (response, pushToBrowserHistory) {
            var pushToHistory = pushToBrowserHistory !== undefined ? pushToBrowserHistory : true,
                html,
```

this then needs to be applied to all calling methods up to `onActionLink`

mitelg (changes requested):
hey @Ivan-Rud 

thanks again for your contribution :+1: 
We finally had the chance to have a deeper look into it. Overall your solution works for the pagination. 
But if a filter is applied and the the browser back button is triggered, the view of the listing is broken. 
Would you mind having a look at this, too?

mitelg:
Hey @Ivan-Rud 

This PR was closed due to inactivity. If you find a proper solution, feel free to create a new pull request. 

PR #2505, opened 2022-08-14
fix: Type error in Exception code due to PDOException

aragon999:
Currently one might run into an error 
```
Fatal error: Uncaught TypeError: Exception::__construct(): Argument #2 ($code) must be of type int, string given in /srv/http/.../engine/Library/Zend/Db/Adapter/Exception.php on line 43
```
Which is due to an type incompatibility of the exceptions, as described here:
https://www.php.net/manual/en/class.pdoexception.php#95812
As the methods of `\Exception` are `final` one cannot change the return type. PHP apparently handles this problem internally as the \PDOException::getCode() method should return a wrong type as well.

Set the error code to `0` in case of a string error code, as this code is already encoded in the Message and a wrong code (e.g. converted to `int`) probably creates more confusion as it helps.

Install a plugin where the attribute columns are already created in the database but not in the attribute table.

\-

\-

### 6. Checklist

### DIFF
--- a/engine/Library/Zend/Db/Adapter/Exception.php
+++ b/engine/Library/Zend/Db/Adapter/Exception.php
@@ -40,6 +40,15 @@ class Zend_Db_Adapter_Exception extends Zend_Db_Exception
         if ($e && (0 === $code)) {
             $code = $e->getCode();
         }
+
+        /**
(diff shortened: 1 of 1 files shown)


mitelg:
thanks for your contribution @aragon999  👍🎉💙

PR #2507, opened 2022-08-17
SW-24530 - multi-edit backup can now be imported again

christopherholst:
The problem here is, that a type check for int takes place, but the value is coming as a string and the zip cant be open.

open the backup zip again and imports all data that have been changed before.

change the suppliertime for many articles
restore your changes, backend says all good but the zip are not be open and all changes are not correct re imported.

https://issues.shopware.com/issues/SW-24530

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/MultiEdit/Resource/Product/Backup.php
+++ b/engine/Shopware/Components/MultiEdit/Resource/Product/Backup.php
@@ -236,7 +236,7 @@ class Backup
         $path = $backup->getPath();
         $dir = \dirname($path);
 
-        if ($offset == 0) {
+        if ($offset === 0) {
             $zip = new ZipArchive();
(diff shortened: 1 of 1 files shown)


engine/Shopware/Components/MultiEdit/Resource/Product/Backup.php
```diff
@@ -236,7 +236,7 @@ public function restore($id, $offset = 0)
         $path = $backup->getPath();
         $dir = \dirname($path);
 
-        if ($offset === 0) {
+        if ($offset == 0) {
             $zip = new ZipArchive();
```
mitelg: please make sure, that `$offset` is an integer, instead of using weak comparison

mitelg (changes requested):

PR #2512, opened 2022-10-18
Move deprecation notice to correct service

larsbo:
Because the deprecation tag is inside the new service instead of the deprecated one every usage of the right service causes a (wrong) deprecation log entry.

Move the deprecation tag from the new service `shopware_elastic_search.client` to the deprecated service `shopware_elastic_search.client.logger`.

Use Shopware 5 with elastic search and debug log level, open up the article list in the backend and look into the logfile. You will find entries with `The "shopware_elastic_search.client.logger" service is deprecated. Use "shopware_elastic_search.client" instead`.

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/ESIndexingBundle/DependencyInjection/services.xml
+++ b/engine/Shopware/Bundle/ESIndexingBundle/DependencyInjection/services.xml
@@ -15,7 +15,7 @@
         </service>
 
         <service id="shopware_elastic_search.client.logger" alias="shopware_elastic_search.client">
-            <deprecated>The "%alias_id%" service is deprecated. Use "%service_id%" instead</deprecated>
+            <deprecated>The "%service_id%" service is deprecated. Use "%alias_id%" instead</deprecated>
         </service>
 
         <service id="shopware_elastic_search.shop_indexer_factory"
--- a/engine/Shopware/Bundle/ESIndexingBundle/DependencyInjection/services.xml
+++ b/engine/Shopware/Bundle/ESIndexingBundle/DependencyInjection/services.xml
@@ -12,10 +12,11 @@
             <argument>%shopware.es.client%</argument>
             <argument type="service" id="eslogger"/>
             <argument type="service" id="Shopware\Bundle\ESIndexingBundle\Console\EvaluationHelperInterface"/>
-            <deprecated>The "shopware_elastic_search.client.logger" service is deprecated. Use "%service_id%" instead</deprecated>
         </service>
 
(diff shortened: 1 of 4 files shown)


engine/Shopware/Bundle/ESIndexingBundle/DependencyInjection/services.xml
```diff
@@ -12,10 +12,11 @@
             <argument>%shopware.es.client%</argument>
             <argument type="service" id="eslogger"/>
             <argument type="service" id="Shopware\Bundle\ESIndexingBundle\Console\EvaluationHelperInterface"/>
-            <deprecated>The "shopware_elastic_search.client.logger" service is deprecated. Use "%service_id%" instead</deprecated>
         </service>
 
-        <service id="shopware_elastic_search.client.logger" alias="shopware_elastic_search.client"/>
+        <service id="shopware_elastic_search.client.logger" alias="shopware_elastic_search.client">
+            <deprecated>The "%alias_id%" service is deprecated. Use "%service_id%" instead</deprecated>
```
mitelg: shouldn't it be the other way arround? :thinking: 

```suggestion
            <deprecated>The "%service_id%" service is deprecated. Use "%alias_id%" instead</deprecated>
```

mitelg (changes requested):

engine/Shopware/Bundle/ESIndexingBundle/DependencyInjection/services.xml
larsbo: good point! 😄 
I just copied it from https://github.com/shopware/shopware/blob/67d2985a7f84bd7c5cb55b4efd6af72d0468a975/engine/Shopware/Components/DependencyInjection/api.xml#L104-L115 😉

mitelg: now I am confused :smile: 

would you mind checking, if the message is now correct? maybe we need to fix the other ones, too? :thinking:

larsbo: The other ones are wrong and I copied it from there. 😸 
`%alias_id%` will output the service name defined in the `alias` attribute of course.
This means the api deprecations currently output
```
shopware.api.customer_stream instead is deprecated, use shopware.api.customer_stream instead
```
I will fix them too. 😉

mitelg (approved):

mitelg:

PR #2513, opened 2022-10-27
Update Application.php

teiling88:
We (@migo315 and I) run into a strange issue with the shopware update process from 5.6 to 5.7 with php 7.4.23. Not all Commands showed up after the update to 5.7. After the php update process everything worked fine. So we digged deeper cause shopware and php update at once is to risky for us. 

So we found a hot fix to add to all missing Commands the "protected static $defaultName" parameter and then they showed up again. But with this fix we are not happy so we digged deeper. So we created this pull request. This change make it work again for us. It seems, that we run in a rare situation from in_array. See this: https://www.php.net/manual/en/function.in-array.php and our running example with example data:
https://3v4l.org/T9bA8
Without the strict no service is registered. With strict mode: 
https://3v4l.org/9N4HM

We think, through the symfony update  in sw 5.7 we run with our heavily extended shopware installation in a rare php issue with in_array.

"Prior to PHP 8.0.0, a string needle will match an array value of 0 in non-strict mode, and vice versa. That may lead to undesireable results. Similar edge cases exist for other types, as well. If not absolutely certain of the types of values involved, always use the strict flag to avoid unexpected behavior."

What do you think about that?

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Console/Application.php
+++ b/engine/Shopware/Components/Console/Application.php
@@ -200,17 +200,9 @@ class Application extends BaseApplication
      */
     protected function registerTaggedServiceIds()
     {
-        $ids = (array) $this->kernel->getContainer()->getParameter('console.command.ids');
-        $lazyServices = array_keys($ids);
-
         if ($this->kernel->getContainer()->hasParameter('console.command.ids')) {
-            /** @var array<string, string> $commandIds */
-            $commandIds = $this->kernel->getContainer()->getParameter('console.command.ids');
-
-            foreach ($commandIds as $id) {
-                if (!\in_array($id, $lazyServices)) {
-                    $this->add($this->kernel->getContainer()->get($id));
-                }
+            foreach ($this->kernel->getContainer()->getParameter('console.command.ids') as $id) {
+                $this->add($this->kernel->getContainer()->get($id));
             }
         }
     }


engine/Shopware/Components/Console/Application.php
```diff
@@ -208,7 +208,7 @@ protected function registerTaggedServiceIds()
             $commandIds = $this->kernel->getContainer()->getParameter('console.command.ids');
 
             foreach ($commandIds as $id) {
-                if (!\in_array($id, $lazyServices)) {
+                if (!\in_array($id, $lazyServices, true)) {
```
mitelg: after debugging with @shyim the whole method can simplified to this:

```php
    protected function registerTaggedServiceIds()
    {
        if ($this->kernel->getContainer()->hasParameter('console.command.ids')) {
            foreach ($this->kernel->getContainer()->getParameter('console.command.ids') as $id) {
                $this->add($this->kernel->getContainer()->get($id));
            }
        }
    }
```

mitelg (changes requested):

engine/Shopware/Components/Console/Application.php
teiling88: Done! :-)

mitelg (approved):

mitelg:
merged with [1ce9651](https://github.com/shopware/shopware/commit/1ce96514f6d675fe5bd855f1e7a4dece4fd462d3) and will be part auf 5.7.17

thanks for your contribution @teiling88  👍🎉💙

PR #2517, opened 2023-01-16
Update sOrder.php

gevorgcode:
private function getAvailableSerialsOfEsd($esdId) starts working slowly when the number of the license keys in DB increases and that causes the clients to wait about 1 to 2 minutes in order to receive the order. (the number of purchased product keys in DB is about 20,000). These all is caused by LEFT JOIN query in SQL.

The query was changed from a left join to an IN statement.
A new line is added in the WHERE clause, which checks if serialID exists in s_order_esd table or not.

### DIFF
--- a/engine/Shopware/Core/sOrder.php
+++ b/engine/Shopware/Core/sOrder.php
@@ -1784,10 +1784,8 @@ EOT;
         return $this->db->fetchAll(
             'SELECT s_articles_esd_serials.id AS id, s_articles_esd_serials.serialnumber AS serialnumber
             FROM s_articles_esd_serials
-            LEFT JOIN s_order_esd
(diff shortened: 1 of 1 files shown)


teiling88:
Instead of changing the query, why not adding an index on s_order_esd.serialID? This should to the job in a much better way :-)

gevorgcode:
@teiling88 
Can you show your code version?

teiling88:
I made a EXLPAIN SELECT Query and the result was, there is an index missing. I created nothing for that. You can create a migraiton which is adding the missing index there. Something like that:
 ```create unique index s_order_esd_serialID_uindex
    on s_order_esd (serialID);
```
should. do the job :-)

PR #2518, opened 2023-01-21
Change default Value for `Shopware.form.field.PagingComboBox` element

Dominikrt:
It's very annoying that the searchable combobox field, has a default value, that you have to clear before you can use the search. Specially for long list, like the profile select in the ImportExport plugin.

If no item is selected, the placeholder text will only be displayed as placeholder and not as value.

Open a Backend module that use a `Shopware.form.field.PagingComboBox` field e.g. ImportExport Plugin > Export > Select profile.

You have to clear the text "Please choose" before you can search for a profile.

### 6. Checklist


rathmerdominik:
thanks for your contribution @Dominikrt   :+1::tada::blue_heart:

PR #2520, opened 2023-02-20
Update config detail.js with the correct default store

christopherholst:
with check on default store and not the id (1) from the store

if you need to change the default store

check the default property under s_core_shops

change the default store in the database under s_core_shops

### 6. Checklist


mitelg:
thanks for your contribution @christopherholst  👍🎉💙

It was merged on the "next" branch, as we are currently in the release phase of 5.7.17

ennasus4sun:
Hello @christopherholst,
I just wanted to let you know that the change in this pull request, unfortunately, caused changes that were done in the settings of subshops are not saved anymore. So we decide to revert these. 
It is important to know that the main shop, created on installation, must have id 1. A lot of the functionality of Shopware-5 is based on this.

PR #2521, opened 2023-02-21
feat: Add variable fonts for OpenSans

aragon999:
I already implemented the same changes for Shopware 6 and thought Shopware 5 could also need some love. So this PR should potentially reduce transferred bytes, by combining multiple separate font files into one. E.g. currently if one uses two different font weights (e.g. `normal` and `bold`) on the same page two font files are downloaded, with roughly 85kb. This can be reduced, e.g. if one uses only the latin charset, only 48kb font files are downloaded. This reduction increases if more font weights are used.

Add an variable subsetted font to Shopware 5.

\-

\-

\-

### 6. Checklist


mitelg:
thanks for your contribution @aragon999  👍🎉💙

It was merged on the "next" branch, as we are currently in the release phase of 5.7.17

PR #2522, opened 2023-02-23
Fix cart instock info message output

lacknere:
There is no message when adding an article by articlenumber fails because of unfulfilled instock conditions.
Problem: Assigned view variables are lost on forward/redirect.

Add instock info as session error message which is then assigned to the view as basket info in cart action.

Go to the cart page and add an out of stock article by articlenumber > no message is shown.

### 6. Checklist


mitelg:
thanks for your contribution @lacknere 👍🎉💙

It was merged on the "next" branch, as we are recently released 5.7.17

PR #2525, opened 2023-03-29
Update createMessageId

christopherholst:
This change will make $hostName an empty string if $messageIdHostName is not initialized. If $hostName is still empty after this line, it will try to use the server name or fall back to the value of php_uname('n').

Fatal error: Uncaught Error: Typed property Zend_Mail::$messageIdHostName must not be accessed before initialization

This change will make $hostName an empty string if $messageIdHostName is not initialized. If $hostName is still empty after this line, it will try to use the server name or fall back to the value of php_uname('n').

If no sender email address is required by a third-party plugin.

### 6. Checklist

### DIFF
--- a/engine/Library/Zend/Mail.php
+++ b/engine/Library/Zend/Mail.php
@@ -1092,7 +1092,7 @@ class Zend_Mail extends Zend_Mime_Message
             $recipient = 'unknown';
         }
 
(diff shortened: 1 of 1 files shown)


mitelg:
hey @christopherholst 

as discussed, this is not a problem of Shopware itself. The error was in a plugin

PR #2526, opened 2023-03-30
Use ListingLinkRewriteServiceInterface instead of ListingLinkRewriteService

mkreusch:
The type should be the interface to not break extensions

Use ListingLinkRewriteServiceInterface instead of ListingLinkRewriteService

Install our Plugin https://store.shopware.com/onco740284870863/power-ajax-filter.html ;-)

### 6. Checklist


mitelg:
thanks for your contribution @mkreusch  👍🎉💙

PR #2527, opened 2023-04-08
Fix 'slt' tag not added to nocache cookie

dot-frost:
This change is necessary to fix an issue when new customers register or login the response deleting the "no-cache" cookie and the next request does not have the "no-cache" cookie then "AppCache::checkSltCookie" will not add the "no-cache" cookie with the tag "slt". and the response of the checkout widget will return from the cache and show the wrong user info. 

I change the  "AppCache::checkSltCookie" when getting the value of the "no-cache" cookie will be an empty string if not exist.

First login with a user to cache the checkout request and then Send a request for a logged-in another user session with "slt" cookie and without the "no-cache" cookie and check the response the user info in the checkout is wrong

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/HttpCache/AppCache.php
+++ b/engine/Shopware/Components/HttpCache/AppCache.php
@@ -347,10 +347,7 @@ class AppCache extends HttpCache
             return;
         }
 
-        $noCache = $request->cookies->get('nocache');
-        if (!\is_string($noCache)) {
-            return;
-        }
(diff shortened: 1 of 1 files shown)


mitelg:
hey @dot-frost 

thanks for you contribution! please sign our CLA, otherwise we can not merge your contribution :slightly_smiling_face:

mitelg:
thanks for your contribution @dot-frost 👍🎉💙

PR #2529, opened 2023-06-13
ci(sw-27129): add danger php

PascalThesing:
Currently, all types of commit messages are possible.
We should be strict about this.

This change introduces a check for the commit messages.
These should follow the https://www.conventionalcommits.org/ specification.

### DIFF
--- a/.danger.php
+++ b/.danger.php
@@ -1,0 +1,50 @@
+<?php declare(strict_types=1);
+
+use Danger\Config;
+use Danger\Context;
+use Danger\Platform\Github\Github;
+use Danger\Platform\Gitlab\Gitlab;
+use Danger\Rule\CommitRegex;
+use Danger\Rule\Condition;
+use Danger\Rule\DisallowRepeatedCommits;
+use Danger\Struct\File;
+
+return (new Config())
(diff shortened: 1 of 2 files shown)


mitelg (approved):

.danger.php
```diff
@@ -0,0 +1,50 @@
...
+use Danger\Struct\File;
+
+return (new Config())
+    ->useThreadOn(Config::REPORT_LEVEL_WARNING)
+    ->useRule(new DisallowRepeatedCommits())
+    ->useRule(function (Context $context): void {
+        $files = $context->platform->pullRequest->getFiles();
+
+        if ($files->matches('UPGRADE-*.md')->count() === 0) {
+            $context->warning('The Pull Request doesn\'t contain any changelog file');
```
mitelg: this message should be adjusted

mitelg (changes requested):

mitelg (approved):

mitelg (changes requested):

mitelg (approved):

DennisGarding (approved):

PR #2533, opened 2023-06-16
SW-27062 - update dependencies

mitelg:


.phpstan-baseline.neon
```diff
@@ -18855,6 +18815,16 @@ parameters:
 			count: 1
 			path: engine/Shopware/Components/Plugin/ResourceSubscriber.php
 
+		-
+			message: "#^Cannot assign offset non\\-falsy\\-string to array\\<string, array\\<int, string\\|null\\>\\>\\|string\\|null\\.$#"
```
mitelg, line 18819: I have no clue how to fix those :disappointed:

ennasus4sun (approved):

engine/Library/Enlight/Components/Session/Namespace.php
```diff
@@ -45,7 +45,7 @@ public function __get($name)
      */
     public function __set($name, $value)
     {
-        return $this->set($name, $value);
```
DennisGarding, line 48: Theoretically, this is a break, but set return void. So i think it is ok

mitelg: exactly. the called method does not return, so this return will return nothing :see_no_evil:

DennisGarding (approved):

PR #2544, opened 2023-06-27
ci(SW-27140): Tests are executed in dependence on each other

DennisGarding:


aragon999:
Just a minor remark, not related to the content of this PR, I guess I would name the commit something like `ci(SW-27140): Execute tests in dependence of each other` :) 

The reasoning behind this tag would be, that `ci` commits can be omitted from changelogs, as they are not really relevant for downstream :-)

composer.lock
```diff
@@ -2405,39 +2405,36 @@
         },
         {
             "name": "laminas/laminas-code",
-            "version": "4.7.1",
+            "version": "4.11.0",
```
mitelg: please revert this. otherwise PHP7.4 can no longer used

mitelg: this will probably cause a fail in the pipeline once a PHP7.4 job is executed. at least I hope so :grimacing:

mitelg (changes requested):

ennasus4sun (approved):

composer.lock
DennisGarding: ohh no.. i forgot to revert it. Thnaks

mitelg:
looks good so far, but you need a rebase on "next" branch and then resolving the conflicts

aragon999:
And the scope of the commit (i.e. the ticket number) should be before the colon :-) 
https://www.conventionalcommits.org/en/v1.0.0/#summary

Not sure why danger did not ran.

mitelg (changes requested):
needs a rebase due to conflicts

mitelg (approved):

PR #2546, opened 2023-07-04
ci(SW-27133): Cancel previous pipelines

DennisGarding:


mitelg (changes requested):
branch needs a rebase. the tests are failing because of a test, that is already fixed

rathmerdominik (approved):

mitelg (approved):

mitelg (changes requested):
needs a rebase again, because the other PR was merged first :grimacing:

mitelg (approved):

.github/workflows/check-code.yml
mitelg, line 1: what is this for an empty file?

mitelg (changes requested):

.github/workflows/create-release.yml
```diff
@@ -7,6 +7,10 @@ on:
         tags:
             - "*"
 
+concurrency:
```
mitelg: I don't think we need that on create release and create update, as they are only triggered manually or when a tag is pushed

.github/workflows/create-update.yml
```diff
@@ -7,6 +7,9 @@ on:
     push:
         tags:
             - "*"
+concurrency:
```
mitelg: see above

mitelg (changes requested):

.github/workflows/check-code.yml
mitelg: I think you wanted to split the JS and PHP part, right? :thinking:

mitelg (changes requested):

.github/workflows/check-code.yml
DennisGarding: it cames from the rebase to main... there it was recreated

mitelg (approved):

mitelg (approved):

ennasus4sun (approved):

PR #2547, opened 2023-07-05
ci(SW-27149) - Create code coverage

DennisGarding:


composer.lock
```diff
@@ -2405,39 +2405,36 @@
         },
         {
             "name": "laminas/laminas-code",
-            "version": "4.7.1",
```
mitelg: please revert

PR #2549, opened 2023-07-06
fix(SW-27151): split emails before validation

ennasus4sun:
Separates the transmitted addresses before passing them to the validation.
Otherwise they will always be validated to false.

### DIFF
--- a/engine/Shopware/Controllers/Backend/Base.php
+++ b/engine/Shopware/Controllers/Backend/Base.php
@@ -1043,15 +1043,20 @@ class Shopware_Controllers_Backend_Base extends Shopware_Controllers_Backend_Ext
         $this->Front()->Plugins()->ViewRenderer()->setNoRender();
         $this->Front()->Plugins()->Json()->setRenderer(false);
 
-        $email = $this->Request()->getParam('value');
+        $emails = explode(',', (string) $this->Request()->getParam('value', ''));
 
         /** @var EmailValidatorInterface $emailValidator */
         $emailValidator = $this->container->get(EmailValidator::class);
-        if ($emailValidator->isValid($email)) {
-            $this->Response()->setContent(1);
-        } else {
-            $this->Response()->setContent('');
+
+        foreach ($emails as $email) {
+            if (!$emailValidator->isValid($email)) {
+                $this->Response()->setContent('');
+
+                return;
+            }
         }
+
+        $this->Response()->setContent(1);
     }
 
     public function getSalutationsAction()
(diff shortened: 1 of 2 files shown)


aragon999:
Could it break, for example when one creates a new user, with multiple mails? This was invalid before, but now would become valid:
https://github.com/shopware5/shopware/blob/95e77156a9c8e0f2c3b731d7e4835cb26921c36d/themes/Backend/ExtJs/backend/user_manager/view/user/create.js#L424-L435

engine/Shopware/Controllers/Backend/Base.php
```diff
@@ -1043,14 +1043,18 @@ public function validateEmailAction()
...
         /** @var EmailValidatorInterface $emailValidator */
         $emailValidator = $this->container->get(EmailValidator::class);
-        if ($emailValidator->isValid($email)) {
-            $this->Response()->setContent(1);
-        } else {
-            $this->Response()->setContent('');
+
+        foreach ($emails as $email) {
+            if ($emailValidator->isValid($email)) {
+                $this->Response()->setContent(1);
```
mitelg: we should set this content only at the end once. if multiple emails are provided, and they are all valid, we would have multiple `1` in the response, which could be misinterpreted

engine/Shopware/Controllers/Backend/Base.php
```diff
@@ -1043,14 +1043,18 @@ public function validateEmailAction()
...
+            } else {
+                $this->Response()->setContent('');
+                break;
```
mitelg: instead of a break, we could directly return, because we found some invalid mail already

mitelg (changes requested):

mitelg:
@aragon999 it was possible before :thinking: the old email validation was basically "is there an "@" then it is valid" 

so it didn't matter, if the string provided was multiple emails divided with a comma or just one :grin:

mitelg (approved):

engine/Shopware/Controllers/Backend/Base.php
```diff
@@ -1043,15 +1043,20 @@ public function validateEmailAction()
         $this->Front()->Plugins()->ViewRenderer()->setNoRender();
         $this->Front()->Plugins()->Json()->setRenderer(false);
 
-        $email = $this->Request()->getParam('value');
+        $emails = explode(',', $this->Request()->getParam('value', ''));
```
mitelg:
```suggestion
        $emails = explode(',', (string) $this->Request()->getParam('value', ''));
```

just to make sure, explode gets always a string. I think it would otherwise give some notices on newer PHP versions :thinking:

mitelg (approved):

DennisGarding (approved):

PR #2564, opened 2023-07-13
ci: Add PHP 8.3 to the test matrix

mitelg:

PHP 8.3 will be released end of the year. We should already try to test it

Adds PHP 8.3 to the test matrix


mitelg:
the tests fails due to an issue in mpdf. it is already fixed: https://github.com/mpdf/mpdf/commit/938651fca46e9556324ba4d93bb106bd7b07380c

ennasus4sun (approved):

PR #2566, opened 2023-08-02
ci(): check for mysql availbality

PascalThesing:
The CI pipeline sometimes fails because MySQL is not yet started.

Wait for MySQL.

### DIFF
--- a/.github/workflows/check-js-code.yml
+++ b/.github/workflows/check-js-code.yml
@@ -33,17 +33,6 @@ jobs:
             run: nix profile install github:cachix/devenv/v0.5
             shell: sh
 
-          - name: Get Composer Cache Directory
-            id: composer-cache
-            run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/check-js-code.yml
```diff
@@ -45,5 +45,6 @@ jobs:
                     ${{ runner.os }}-composer-
 
           - run: devenv up &
+          - run: devenv shell ./.github/check-mysql.sh
```
mitelg: why do we need a mysql server at all here? and composer cache should also not be necessary :thinking:

mitelg (approved):

mitelg:
the commit message seems also to be wrong

.github/workflows/check-js-code.yml
PascalThesing: You are right.

ennasus4sun (approved):

PR #2567, opened 2023-08-06
fix: PHP 8 error if plugin bootstrap class is missing

aragon999:
If the plugin bootstrap class was used as a subscriber, but the plugin has been deleted on the filesystem but is still present in the plugin database table, and if it has a subscriber, starting from PHP8.0 this will yield an error, since the `$plugin` variable is `null` but `method_exists` does not accept `null`.

Fixes the typehint, and the `if` condition. Furthermore I noticed that at least currently if the `plugin` was an instance of `Enlight_Plugin_Bootstrap` also `null` was returned. I replicated this behavior such that it can be more easily removed for Shopware 5.8.

There currently exist no tests, as far as I can tell for this area. But in general it might also be more favorable to refactor this area completely, hence I am not sure if I should add a test case for this.

See above.

\-

\-

### 6. Checklist

### DIFF
--- a/.phpstan-baseline.neon
+++ b/.phpstan-baseline.neon
@@ -2801,11 +2801,6 @@ parameters:
 			path: engine/Library/Enlight/Event/Handler/Default.php
 
 		-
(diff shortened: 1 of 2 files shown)


mitelg:
hey @aragon999

thanks for your PR, it looks good to me 👍

please rebase your branch

mitelg (approved):

ennasus4sun (approved):

mitelg:
thanks for your contribution @aragon999 👍🎉💙

PR #2568, opened 2023-08-06
Refactor/tell a friend

aragon999:
I wanted to remove occurrences where `\sSystem` is used, and found the `Tellafriend` controller. I refactored a bit and thereby removed the usage of `\sSystem`. 

While at it, I noticed that there is an unused parameter in `\sArticles::sGetPromotionById`, which I therefore deprecated. I was not sure where I should document this deprecation.

Furthermore I am not sure, if the `Tellafriend` functionality should be removed (and therefore deprecated) altogether. As the mails probably will be categorized as spam anyways.
Before adding a test for this controller, I would rather remove it :-)

See above.

See above.

\-

\-

### 6. Checklist

### DIFF
--- a/.phpstan-baseline.neon
+++ b/.phpstan-baseline.neon
@@ -24646,31 +24646,6 @@ parameters:
 			path: engine/Shopware/Controllers/Frontend/Sitemap.php
 
 		-
-			message: "#^Access to an undefined property Enlight_Controller_Request_RequestHttp\\:\\:\\$sArticle\\.$#"
-			count: 1
-			path: engine/Shopware/Controllers/Frontend/Tellafriend.php
-
-		-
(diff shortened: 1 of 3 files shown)


mitelg:
hey @aragon999 

thanks for your PR :+1: 

new deprecations should be documented here: https://github.com/shopware5/shopware/blob/5.7/UPGRADE-5.7.md (under the 5.7.19 section)

IMO the whole tell-a-friend thing could be removed with 5.8 :man_shrugging: 
I don't think this is a feature which is still used nowadays. so we could just leave the controller as it is (to make sure, we do not break anything unintentionally :grin: ) and deprecate the whole stuff, including templates and configs and so on.

mitelg:
hey @aragon999 any progress here?

mitelg:
Hey @aragon999 

For more information about our contribution guidelines, see https://github.com/shopware5/shopware/wiki/Contribute

PR #2578, opened 2023-08-26
refactor: Prepare Enlight\Event for Shopware 5.8 and fix the phpstan issues

aragon999:
This is a first attempt to cleanup the Enlight event system, on the one hand fixing the types and on the other hand updating the deprecations. The complete Enlight event system probably should be replaced completely by the Symfony Event system, but not sure how the exact migration path should look like.

I updated the deprecations and added some new ones and furthermore fixed the phpstan issues of the non deprecated classes. What do you think of the deprecations and the other changes? Are there any objections?
Locally I have also prepared a branch, where I applied the deprecations and changes for Shopware 5.8 (for the Event system), which I will try to regularly rebase.

\-

\-

Once everything is in shape I will add the necessary information to the `UPGRADE.md`. 

### 6. Checklist

### DIFF
--- a/.phpstan-baseline.neon
+++ b/.phpstan-baseline.neon
@@ -2661,36 +2661,6 @@ parameters:
 			path: engine/Library/Enlight/Controller/Router.php
 
 		-
-			message: "#^Method Enlight_Event_EventArgs\\:\\:__construct\\(\\) has parameter \\$args with no value type specified in iterable type array\\.$#"
-			count: 1
-			path: engine/Library/Enlight/Event/EventArgs.php
-
-		-
-			message: "#^Method Enlight_Event_EventArgs\\:\\:getReturn\\(\\) has no return type specified\\.$#"
-			count: 1
-			path: engine/Library/Enlight/Event/EventArgs.php
-
-		-
-			message: "#^Method Enlight_Event_EventArgs\\:\\:setName\\(\\) has parameter \\$name with no type specified\\.$#"
-			count: 1
-			path: engine/Library/Enlight/Event/EventArgs.php
-
-		-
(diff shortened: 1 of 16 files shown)


mitelg:
hey @aragon999 

at first glance, this looks quite good :+1: 

if we want to deprecate things, we should also make sure, that those things could be removed without problems and only if we have proper replacements for them. beside that, we can remove what ever works :grin: 

Exchanging the event system with the Symfony component could work within a 5.8 release. the good thing with Enlight is, that we have a layer between the framework and our application. so simply said, we can call the Symfony component in the Enlight code and remove our "custom" event logic and in best case nothing changes for Shopware and the plugins.

aragon999:
hey @mitelg very good :-)

Currently everything which is deprecated for removal, is unused. As far as I can tell, only the `Enlight_Event_Subscriber_Plugin` is still in use here: https://github.com/shopware5/shopware/blob/f56e358f200efaaf0e1f0227c0127c69b8da7195/engine/Library/Enlight/Plugin/Namespace/Config.php#L148-L155

But, this currently is integrated deep in plugin system (and also in the Shopware default plugins). Therefore I decided to make this class "only" internal, such that we can later safely deprecate and remove it and give plugin developers time to adopt. Maybe we should also deprecate some more methods in the plugin system, or as suggested use the Symfony event system for that. But I think this is somewhat of a different task, when cleaning up the `Enlight_Plugin_*` stuff.

mitelg:
hey @aragon999 any progress here?

mitelg:
as this PR is in draft mode, I will leave it open for now.

mitelg:
Hey @aragon999 

For more information about our contribution guidelines, see https://github.com/shopware5/shopware/wiki/Contribute

PR #2585, opened 2023-09-12
Fix paged listing with emotion

lacknere:
Because the product listing is not shown if the page parameter is defined and an emotion is shown.

Remove the sPage parameter check when collecting devices with listing.

Create an emotion > assign it to a category with products > activate product listing > select "Category start page and listing" as visibility > go to the second page of the product listing > emotion is shown, product listing is not

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Frontend/Listing.php
+++ b/engine/Shopware/Controllers/Frontend/Listing.php
@@ -495,10 +495,6 @@ class Shopware_Controllers_Frontend_Listing extends Enlight_Controller_Action
      */
     private function getDevicesWithListing(array $emotions): array
     {
-        if ($this->Request()->getParam('sPage')) {
-            return [];
-        }
-
         $visibleDevices = [0, 1, 2, 3, 4];
         $permanentVisibleDevices = [];
 


aragon999:
Mh, I always thought this is a feature and not a bug, i.e. that the emotion is only shown on the first page.

mitelg:
this is indeed intentional and the if condition should not be simply removed. 

what about an event where you are able to adjust if the method returns data or an empty array?

lacknere:
Just to be clear, if I have an emotion with visibility "start page and listing" and the option "product listing" active, I would expect the emotion AND the product listing to show up when I click on the next page button of the pagination. That is not the case, which is why i thought this was a bug rather than intention.

mitelg:
I guess, then the change is too general and will affected other settings and options as well. 

we should find a more particular solution and make sure that other settings are not affected. (by writing or extending tests)

mitelg:
hey @lacknere any progress here?

lacknere:
@mitelg well, this turned out to work as expected with a clean shopware 5 setup. That was just a bug caused by css from the custom theme 🤦

PR #2612, opened 2023-09-21
SW-27165: Change consideration of graduated prices in order module

mitelg:
While adding new positions to an existing order in the backend order module, graduated prices are still considered.

If an existing position is changed, the graduated price will only be checked and considered if the quantity was changed.

With this change it is possible again to also set a custom price for a position that contains a product with graduated prices.

### DIFF
--- a/engine/Shopware/Controllers/Backend/Order.php
+++ b/engine/Shopware/Controllers/Backend/Order.php
@@ -117,7 +117,7 @@ class Shopware_Controllers_Backend_Order extends Shopware_Controllers_Backend_Ex
 
     /**
      * @deprecated - Will be removed in Shopware 5.8 without a replacement
-     * Contains the dynamic receipt repository
+     * Contains the order document repository
      *
      * @var OrderDocumentRepository
      */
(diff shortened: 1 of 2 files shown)


engine/Shopware/Controllers/Backend/Order.php
```diff
@@ -1950,7 +1950,17 @@ private function getPositionAssociatedData(array $data, Order $order): array
...
-        if ($this->hasProductGraduatedPrices($data['articleNumber'], $order)) {
+        $orderPosition = null;
+        foreach ($order->getDetails() as $position) {
+            if ($position->getId() === (int) $data['id']) {
+                $orderPosition = $position;
+                break;
+            }
+        }
+        // Only get graduated price if a new position should be added or quantity has changed
+        if (($orderPosition === null || $orderPosition->getQuantity() !== (int) $data['quantity'])
```
mitelg, line 1961: I am not sure, if we should only allow new positions at all :man_shrugging:

rathmerdominik (approved):

engine/Shopware/Controllers/Backend/Order.php
DennisGarding: Good question.. maybe we make a demo, to get the whole picture

mitelg: we will leave this, as implemented here

DennisGarding (approved):

PR #2621, opened 2023-11-16
fix: doc path for moved repository

jschwalmi:
Wrong DocPath means that the path or class cannot be downloaded.
Bildschirmfoto 2023-11-16 um 10 36 22
Bildschirmfoto 2023-11-16 um 10 37 08

### DIFF
--- a/engine/Shopware/Components/ReflectionHelper.php
+++ b/engine/Shopware/Components/ReflectionHelper.php
@@ -63,7 +63,7 @@ class ReflectionHelper
         $folders = Shopware()->Container()->getParameter('shopware.plugin_directories');
 
         $folders[] = Shopware()->DocPath('engine_Shopware');
-        $folders[] = Shopware()->DocPath('vendor_shopware_shopware');
+        $folders[] = Shopware()->DocPath('vendor_shopware5_shopware');
 
         foreach ($folders as $folder) {
             $directories[] = substr($folder, \strlen($docPath));


mitelg:
hey @jschwalmi 

thanks for your contribution! Good catch indeed :grin: :+1: 

Please make sure, that you commit the changes with your GitHub user and sign the CLA with this user :slightly_smiling_face: 

The commit message should look like described here: https://github.com/shopware5/shopware/wiki/Contribute#commit-message

engine/Shopware/Components/ReflectionHelper.php
```diff
@@ -63,7 +63,7 @@ public function createInstanceFromNamedArguments($className, $arguments, $secure
         $folders = Shopware()->Container()->getParameter('shopware.plugin_directories');
 
         $folders[] = Shopware()->DocPath('engine_Shopware');
-        $folders[] = Shopware()->DocPath('vendor_shopware_shopware');
+        $folders[] = Shopware()->DocPath('vendor_shopware5_shopware');
```
shyim, line 66: isn't the name computed by the composer package name, and that is still shopware/shopware 🤔

mitelg: @shyim is right. @jschwalmi with your change, the path would look like this: `/home/mt/www/5.7/vendor/shopware5/shopware/`, but that path does not exist. 

Could you explain, how the problem could be reproduced?

mitelg (changes requested):

engine/Shopware/Components/ReflectionHelper.php
jschwalmi: @mitelg @shyim 
I discovered this problem because it was reported that the product stream with Peudo prices no longer worked. All products were displayed in the stream and not only the discounted products.
In the [\Shopware_Controllers_Backend_ProductStream::loadPreviewAction](https://github.com/shopware5/shopware/blob/15919fda5f13a3b74c870456a3548981c6035da6/engine/Shopware/Controllers/Backend/ProductStream.php#L74) the conditions could not be unserialized. [vendor/shopware5/shopware/engine/Shopware/Controllers/Backend/ProductStream.php:99](https://github.com/shopware5/shopware/blob/15919fda5f13a3b74c870456a3548981c6035da6/engine/Shopware/Controllers/Backend/ProductStream.php#L99).
I think all product streams that have a condition are affected by this.

shyim: How did you installed Shopware? 🤔

jschwalmi: @shyim I install Shopware over composer. Do you think that this is not the case in a ZIP installation?

shyim:
```sh
composer create-project shopware/composer-project:@dev test
cd test
ls -la vendor/shopware/shopware
```

seems fine for me 🤔

jschwalmi: I agree that it could work

mitelg:
hey @jschwalmi what is the status here? :blush:

jschwalmi:
> hey @jschwalmi what is the status here? 😊

The comment from @shyim is fine for me  :) 
https://github.com/shopware5/shopware/pull/2621#discussion_r1399177921

mitelg:
alright, then I will close your PR :slightly_smiling_face: :+1:

PR #2627, opened 2024-01-09
fix NotifyAlreadyRegistered is never set caused by fetchColumn

kayaArvatis:
`$NotifyAlreadyRegistered` is never set `fetchColumn` is still used but it is an old method and is deprecated with new library version of doctrine. `$ordernumbers` is currently an string but it should be an array since it is iterated.

Fixes the variable `$NotifyAlreadyRegistered` to be set and available in the storefront.

Choose a product that is out of stock and has notification feature enabled:

- Enter an email address
- Confirm the email address
- Check the product
- You should see "You have already subscribed to notifications on this item!" 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
@@ -109,13 +109,16 @@ class Shopware_Plugins_Frontend_Notification_Bootstrap extends Shopware_Componen
 
         if (!empty($notificationProducts)) {
             $sql = 'SELECT `ordernumber` FROM `s_articles_details` WHERE `articleID`=?';
-            $ordernumbers = $this->get(Connection::class)->fetchColumn($sql, [$id]);
+            $ordernumbers = $this->get(Connection::class)->fetchAllAssociative($sql, [$id]);
+            $ordernumbers = array_map(function (array $product) {
+                return $product['ordernumber'];
+            }, $ordernumbers);
 
             if (!empty($ordernumbers)) {
                 foreach ($ordernumbers as $ordernumber) {
                     if (\in_array($ordernumber, $notificationProducts)) {
                         $notificationVariants[] = $ordernumber;
-                        if ($ordernumber === $sArticle) {
+                        if ($ordernumber === $sArticle['ordernumber']) {
                             $view->assign('NotifyAlreadyRegistered', true);
                         }
                     }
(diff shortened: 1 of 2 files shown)


engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
```diff
@@ -109,13 +109,16 @@ public function onPostDispatch(Enlight_Event_EventArgs $args)
 
         if (!empty($notificationProducts)) {
             $sql = 'SELECT `ordernumber` FROM `s_articles_details` WHERE `articleID`=?';
-            $ordernumbers = $this->get(Connection::class)->fetchColumn($sql, [$id]);
+            $ordernumbers = $this->get(Connection::class)->fetchAllAssociative($sql, [$id]);
```
mitelg:
```suggestion
            $ordernumbers = $this->get(Connection::class)->fetchAssociative($sql, [$id]);
```

that should give you a list of order numbers directly. so you do not need to iterate again, for picking the order numbers

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
```diff
@@ -109,13 +109,16 @@ public function onPostDispatch(Enlight_Event_EventArgs $args)
...
+            $ordernumbers = array_map(function (array $product) {
+                return $product['ordernumber'];
+            }, $ordernumbers);
 
             if (!empty($ordernumbers)) {
                 foreach ($ordernumbers as $ordernumber) {
                     if (\in_array($ordernumber, $notificationProducts)) {
                         $notificationVariants[] = $ordernumber;
-                        if ($ordernumber === $sArticle) {
+                        if ($ordernumber === $sArticle['ordernumber']) {
```
mitelg, line 118: good catch :ok_hand:

mitelg (changes requested):
thanks for your contribution @kayaArvatis :+1: 

your changes look good so far. Would you mind adding a test for this? 
there is already a test, setting up the plugin here: https://github.com/shopware5/shopware/blob/5.7/tests/Functional/Plugins/Frontend/NotificationTest.php

engine/Shopware/Plugins/Default/Frontend/Notification/Bootstrap.php
kayaArvatis: @mitelg `fetchAssociative` returns only one order number but `fetchAllAssociative` returns all order numbers for the product it contains all variant order numbers, not sure what exactly is expected here

kayaArvatis:
@mitelg Thanks for your feedback, i will take a look at the tests

kayaArvatis:
Hi @mitelg I have applied the changes and added an test

mitelg:
thanks for your contribution @kayaArvatis 👍🎉💙

PR #2631, opened 2024-02-19
fix: media name and description usage

lacknere:
Media name and description are not correctly mapped.

* Fix the hydration and conversion of media name and description.
* Add media name fallbacks if no description is available.

### 6. Checklist


UPGRADE-5.7.md
```diff
@@ -10,11 +10,18 @@ This changelog references changes done in Shopware 5.7 patch versions.
 
 * Added compatibility with PHP 8.3
 * Added new polyfill `symfony/polyfill-php83` to be able to use PHP 8.3 features
+* Added image name fallbacks if no image description is available to:
```
mitelg:
```suggestion
* Added fallback to image name for `alt` and `title` attributes if image description is not available in following templates:
```

and then please list the tpl files and then the block names

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/Hydrator/MediaHydrator.php
```diff
@@ -126,8 +126,9 @@ public function hydrateProductImage(array $data)
         $translation = $this->getTranslation($data, '__image');
         $data = array_merge($data, $translation);
 
-        $media->setName($data['__image_description']);
+        $media->setName($data['__image_img']);
```
mitelg, line 129: would you mind extending the `\Shopware\Tests\Functional\Bundle\StoreFrontBundle\ProductMediaTest`? so we can ensure, this behaviour stays in the future

engine/Shopware/Components/Compatibility/LegacyStructConverter.php
```diff
@@ -641,7 +641,8 @@ public function convertMediaStruct(?Media $media = null)
             'id' => $media->getId(),
             'position' => null,
             'source' => $media->getFile(),
-            'description' => $media->getName(),
+            'name' => $media->getName(),
```
mitelg, line 644: please also extend the `\Shopware\Tests\Unit\Components\Compatibility\LegacyStructConverterTest`

themes/Frontend/Bare/frontend/index/index.tpl
```diff
@@ -206,7 +206,10 @@
             {if $sArticle.additionaltext}
                 {$lastSeenProductsConfig.currentArticle.articleName = $lastSeenProductsConfig.currentArticle.articleName|cat:' ':$sArticle.additionaltext}
             {/if}
-            {$lastSeenProductsConfig.currentArticle.imageTitle = $sArticle.image.description}
+            {$lastSeenProductsConfig.currentArticle.imageTitle = $sArticle.image.name}
```
mitelg, line 209: don't we need to escape here as well? :thinking:

mitelg (changes requested):
thanks for your contribution, nice fix! 
would you mind adding a few tests?

themes/Frontend/Bare/frontend/index/index.tpl
lacknere: `$lastSeenProductsConfig`  is json encoded [here](https://github.com/shopware5/shopware/blob/40b94b7027d717cc2b9c9f93a0e1d7a73928232f/themes/Frontend/Bare/frontend/index/index.tpl#L250) so it should be escaped anyways, I think 🤔

ennasus4sun (approved):

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/Hydrator/MediaHydrator.php
mitelg, line 129: hey @lacknere 

I had another thought about that. This change is a break. People and code are assuming, that the "name" contains the image description, if this hydration method is used. If we now change this, their code will break. So unfortunately we cannot merge it like that

mitelg (changes requested):

engine/Shopware/Bundle/StoreFrontBundle/Gateway/DBAL/Hydrator/MediaHydrator.php
lacknere: What about using description primarily and name as a fallback there, if no description is available? Something like `$media->setName($data['__image_description'] ?: $data['__image_img']);`

mitelg: sorry for the late answer :see_no_evil: 

I am not sure, if this is also a desired state :thinking: could also lead to unexpected behaviour, where not values where before. 

what about deprecating the old behaviour and trigger the new one with a parameter? :thinking: where the default ensures the old behaviour. 

to not break the public method, we would need to add the parameter commented at first, and get the value with https://www.php.net/manual/en/function.func-get-args.php :thinking: 

maybe you can find a way to build it like that, that you have the possibility to get the hydrated media with "correct" values, but the default way stays at it is.

mitelg:
hey @lacknere 
are you still on this? what are your thoughts?

mitelg:
Hey @lacknere 

I will close this PR due to inactivity. Let me know, if you want to discuss this further :slightly_smiling_face:

PR #2649, opened 2024-03-30
fix: backend statistics notification pagination

Dominikrt:
The pagination of the notification module is broken. The first page has no limit and the second page shows wrong total results.

The `start` property can now be 0.

The total results get now count from SQL Server and not in php code.

signup for over 21 article notification and the pagination will not work

### 6. Checklist

### DIFF
--- a/engine/Shopware/Controllers/Backend/Notification.php
+++ b/engine/Shopware/Controllers/Backend/Notification.php
@@ -28,26 +28,24 @@ class Shopware_Controllers_Backend_Notification extends Shopware_Controllers_Bac
 {
     /**
      * returns a JSON string to with all found articles for the backend listing
+     *
+     * @return void
      */
     public function getArticleListAction()
     {
         try {
-            $limit = (int) $this->Request()->limit;
-            $offset = (int) $this->Request()->start;
-
-            /** @var array $filter */
-            $filter = $this->Request()->getParam('filter', []);
+            $limit = (int) $this->Request()->get('limit', 20);
+            $offset = (int) $this->Request()->get('start', 0);
+            $filter = (array) $this->Request()->getParam('filter', []);
 
             // order data
             $order = (array) $this->Request()->getParam('sort', []);
 
             $repository = $this->get('models')->getRepository(Article::class);
             $dataQuery = $repository->getArticlesWithRegisteredNotificationsQuery($filter, $offset, $limit, $order);
-            $data = $dataQuery->getArrayResult();
 
-            // manually calc the totalAmount cause the paginate($this->get('models')->getQueryCount) doesn't work with this query
-            $dataQuery->setFirstResult(0);
-            $totalCount = \count($dataQuery->getArrayResult());
+            $data = $dataQuery->getResult(AbstractQuery::HYDRATE_ARRAY);
+            $totalCount = (int) $repository->getProductsWithNotificationsCountQuery()->getOneOrNullResult(AbstractQuery::HYDRATE_SINGLE_SCALAR);
 
             $summaryQuery = $repository->getArticlesWithRegisteredNotificationsQuery($filter, $offset, $limit, $order, true);
             $summaryData = $summaryQuery->getOneOrNullResult(AbstractQuery::HYDRATE_ARRAY);
@@ -68,16 +66,17 @@ class Shopware_Controllers_Backend_Notification extends Shopware_Controllers_Bac
 
     /**
      * returns a JSON string to with all found customers for the backend listing
+     *
+     * @return void
      */
     public function getCustomerListAction()
     {
         try {
-            $limit = (int) $this->Request()->limit;
-            $offset = (int) $this->Request()->start;
-            $productOrderNumber = $this->Request()->orderNumber;
+            $limit = (int) $this->Request()->get('limit', 20);
+            $offset = (int) $this->Request()->get('start', 0);
+            $productOrderNumber = $this->Request()->get('orderNumber');
 
-            /** @var array $filter */
-            $filter = $this->Request()->getParam('filter', []);
+            $filter = (array) $this->Request()->getParam('filter', []);
 
             // order data
             $order = (array) $this->Request()->getParam('sort', []);
--- a/engine/Shopware/Models/Article/Repository.php
+++ b/engine/Shopware/Models/Article/Repository.php
@@ -2086,7 +2086,7 @@ class Repository extends ModelRepository
         $summarize = null
     ) {
         $builder = $this->getArticlesWithRegisteredNotificationsBuilder($filter, $order, $summarize);
-        if (empty($summarize) && !empty($limit) && !empty($offset)) {
+        if (empty($summarize) && !empty($limit) && $offset !== null) {
             $builder->setFirstResult($offset)
                 ->setMaxResults($limit);
         }
@@ -2141,6 +2141,19 @@ class Repository extends ModelRepository
     }
 
     /**
+     * @return Query<Notification>
+     */
+    public function getProductsWithNotificationsCountQuery(): Query
+    {
+        $builder = $this->getEntityManager()->createQueryBuilder();
+        $builder->select('COUNT(DISTINCT notification.articleNumber)');
+
+        $builder->from(Notification::class, 'notification');
+
+        return $builder->getQuery();
+    }
+
+    /**
      * Returns an instance of the \Doctrine\ORM\Query object which selects all notification customers by the given articleOrderNumber
      *
      * @param string $articleOrderNumber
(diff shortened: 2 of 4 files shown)


mitelg:
hey @Dominikrt 

thanks again for your contribution! :tada: :+1: 

unfortunately we did not manage to get the pipeline running again yet. but we are on it

engine/Shopware/Controllers/Backend/Notification.php
```diff
@@ -32,22 +33,21 @@ class Shopware_Controllers_Backend_Notification extends Shopware_Controllers_Bac
...
+            /** @var ArticleRepository $repository */
             $repository = $this->get('models')->getRepository(Article::class);
             $dataQuery = $repository->getArticlesWithRegisteredNotificationsQuery($filter, $offset, $limit, $order);
-            $data = $dataQuery->getArrayResult();
 
-            // manually calc the totalAmount cause the paginate($this->get('models')->getQueryCount) doesn't work with this query
-            $dataQuery->setFirstResult(0);
-            $totalCount = \count($dataQuery->getArrayResult());
+            $data = $dataQuery->getResult(AbstractQuery::HYDRATE_ARRAY);
+            $totalCount = $repository->getArticleWithNotificationCountQuery()->getOneOrNullResult(AbstractQuery::HYDRATE_SINGLE_SCALAR);
```
mitelg:
```suggestion
            $totalCount = (int) $repository->getArticleWithNotificationCountQuery()->getOneOrNullResult(AbstractQuery::HYDRATE_SINGLE_SCALAR);
```

do the cast here and not where the variable is assigned to another one

engine/Shopware/Controllers/Backend/Notification.php
```diff
@@ -32,22 +33,21 @@ class Shopware_Controllers_Backend_Notification extends Shopware_Controllers_Bac
...
+            $limit = (int) $this->Request()->get('limit', 20);
+            $offset = (int) $this->Request()->get('start', 0);
 
             /** @var array $filter */
             $filter = $this->Request()->getParam('filter', []);
 
             // order data
             $order = (array) $this->Request()->getParam('sort', []);
 
+            /** @var ArticleRepository $repository */
```
mitelg: please do not add var annotations if not necessary

engine/Shopware/Controllers/Backend/Notification.php
```diff
@@ -72,16 +72,17 @@ public function getArticleListAction()
...
+            $offset = (int) $this->Request()->get('start', 0);
+            $productOrderNumber = $this->Request()->get('orderNumber');
 
             /** @var array $filter */
             $filter = $this->Request()->getParam('filter', []);
 
             // order data
             $order = (array) $this->Request()->getParam('sort', []);
 
+            /** @var ArticleRepository $repository */
```
mitelg: please do not add var annotations if not necessary

tests/Functional/Controllers/Backend/NotificationTest.php
```diff
@@ -50,20 +59,28 @@ public function setUp(): void
...
+        $getParameter = http_build_query($parameter);
+
+        $this->dispatch('backend/Notification/getArticleList?' . $getParameter);
         static::assertTrue($this->View()->success);
         $returnData = $this->View()->data;
         static::assertNotEmpty($returnData);
-        static::assertCount(2, $returnData);
+
+        static::assertCount(self::NOTIFICATION_LIST_PAGINATION_ITEMS_COUNT_1, $returnData);
+        static::assertEquals(self::NOTIFICATION_LIST_TOTAL_RESULTS, $this->View()->totalCount);
```
mitelg:
```suggestion
        static::assertEquals(self::NOTIFICATION_LIST_TOTAL_RESULTS, $this->View()->getAssign('totalCount'));
```

do not use magic property access, but this should also be comment on by PHPStan if the pipeline would run :grimacing:

tests/Functional/Controllers/Backend/NotificationTest.php
```diff
@@ -27,6 +27,13 @@
 
 class NotificationTest extends Enlight_Components_Test_Controller_TestCase
 {
+    private const NOTIFICATION_LIST_PAGINATION_LIMIT = 2;
+    private const NOTIFICATION_LIST_PAGINATION_START_1 = 0;
```
mitelg: you should name the constants after the usage in the respective test. and not just 1, 2 and so on.

tests/Functional/Controllers/Backend/NotificationTest.php
mitelg: please use `assertSame` as this ensures that the type is also the same

mitelg (changes requested):

engine/Shopware/Controllers/Backend/Notification.php
Dominikrt: Hi, thank you for the feedback. 

How to decide if a var annotations is necessary? autocomplete isn't working without the annotation 🤔

mitelg: adding a var annotation should only be the last resort, if the variable type could not be determined by PHPStan. in this case, PHPStan knows the variable. you may want to install the PHPStorm Symfony extension, which has some support for services directly received from the DI container

mitelg:
after a rebase the CI should work again

Dominikrt:
Hi @mitelg,
thank you for you're recommendations

I have also add an additional test to check the second page with the customer list.

engine/Shopware/Controllers/Backend/Notification.php
```diff
@@ -29,25 +29,21 @@ class Shopware_Controllers_Backend_Notification extends Shopware_Controllers_Bac
     /**
      * returns a JSON string to with all found articles for the backend listing
      */
-    public function getArticleListAction()
+    public function getArticleListAction(): void
```
mitelg: please revert this, as this is considered as a breaking change.

to fix PHPStan you can add a return annotation to the PHPdoc

engine/Shopware/Controllers/Backend/Notification.php
```diff
@@ -69,15 +65,14 @@ public function getArticleListAction()
     /**
      * returns a JSON string to with all found customers for the backend listing
      */
-    public function getCustomerListAction()
+    public function getCustomerListAction(): void
```
mitelg: revert

engine/Shopware/Models/Article/Repository.php
```diff
@@ -2140,6 +2140,19 @@ public function getArticlesWithRegisteredNotificationsBuilder($filter, $order, $
         return $builder;
     }
 
+    /**
+     * @return Query<Notification>
+     */
+    public function getArticleWithNotificationCountQuery(): Query
```
mitelg: please do not use "article" if a product is meant. this is just a wrong translation and should be avoided. especially if new things are added

tests/Functional/Controllers/Backend/NotificationTest.php
```diff
@@ -27,6 +27,20 @@
 
 class NotificationTest extends Enlight_Components_Test_Controller_TestCase
 {
+    private const TEST_ARTICLE_LIST_LIMIT = 2;
```
mitelg: sames as above, please exchange "article" with "product"

tests/Functional/Controllers/Backend/NotificationTest.php
```diff
@@ -50,49 +67,90 @@ public function setUp(): void
...
-        $sql = 'DELETE FROM s_articles_notification WHERE id IN (1111111111, 1111111112, 1111111113)';
+        $sql = 'DELETE FROM s_articles_notification WHERE id IN (1111111111, 1111111112, 1111111113, 1111111114, 1111111115)';
         Shopware()->Db()->query($sql);
     }
 
     /**
      * test getList controller action
      */
-    public function testGetArticleList()
+    public function testGetArticleListFirstPage(): void
```
mitelg, line 77: just FYI: no need to revert the void return type in the tests, as this is not part of the BC promise :wink:

mitelg (changes requested):
just a few more tiny things :blush: 

and a rebase of your branch would be nice

engine/Shopware/Models/Article/Repository.php
Dominikrt: I think it makes things more confusing if everything is called `article` and I rename my new method to `product`.

But no problem I can rename it.

mitelg: the thing is "article" is already confusing everybody who is not coming from German or Netherlands language :wink: 

so naming it `getProductsWithNotifictionsCountQuery` is quite clear, imo

Dominikrt:
Hi @mitelg,
I committed the recommended changes, but it looks like we have pipeline problems again...

mitelg (approved):

mitelg:
after an update of a GitHub action, the e2e tests didn't work anymore. but they should run now again. please try to rebase :slightly_smiling_face: See https://github.com/shopware5/shopware/actions/runs/8611563619

Otherwise the changes look good :+1:

mitelg:
thanks for your contribution @Dominikrt 👍🎉💙

PR #2671, opened 2024-04-24
fix: backend voucher individual code search

Dominikrt:
Hi,

the individual voucher code search is broken, after a reload the store lose its filter

the last filter get cleared bevor and not after the search method.

open a individual voucher and search for a code, after a list reload the filter gets cleared.

### 6. Checklist

### DIFF
--- a/themes/Backend/ExtJs/backend/voucher/controller/code.js
+++ b/themes/Backend/ExtJs/backend/voucher/controller/code.js
@@ -264,8 +264,15 @@ Ext.define('Shopware.apps.Voucher.controller.Code', {
         var me = this,
             searchString = Ext.String.trim(value),
             store = me.subApplication.getStore('Code');
-        store.filter('filter', searchString);
+
         store.filters.clear();
+
+        if (searchString.length <= 0) {
+            store.load();
+            return;
+        }
+
+        store.filter('filter', searchString);
     },
 
     /**
--- a/themes/Backend/ExtJs/backend/voucher/controller/voucher.js
+++ b/themes/Backend/ExtJs/backend/voucher/controller/voucher.js
@@ -169,6 +169,7 @@ Ext.define('Shopware.apps.Voucher.controller.Voucher', {
                     mode = record.data.modus,
                     codeStore = me.getStore('Code');
 
+                codeStore.filters.clear();
                 codeStore.clearData();
                 codeStore.currentPage = 1;
 


themes/Backend/ExtJs/backend/voucher/controller/voucher.js
```diff
@@ -166,11 +166,14 @@ Ext.define('Shopware.apps.Voucher.controller.Voucher', {
             scope: this,
             callback: function (records, operation, success) {
                 var record = records[0],
-                    mode = record.data.modus;
+                    mode = record.data.modus,
+                    codeStore = me.getStore('Code');
+
+                codeStore.filters.clear();
```
mitelg, line 172: I just merged https://github.com/shopware5/shopware/pull/2649

So I guess we need a rebase here

themes/Backend/ExtJs/backend/voucher/controller/code.js
```diff
@@ -264,8 +264,15 @@ Ext.define('Shopware.apps.Voucher.controller.Code', {
         var me = this,
             searchString = Ext.String.trim(value),
             store = me.subApplication.getStore('Code');
-        store.filter('filter', searchString);
+
         store.filters.clear();
+
+        if (searchString.length <= 0) {
+            store.load();
```
mitelg, line 271: why are you loading the store only, when the search string is not given? :thinking:

mitelg (changes requested):

themes/Backend/ExtJs/backend/voucher/controller/voucher.js
mitelg: ah no, I mixed it up with https://github.com/shopware5/shopware/pull/2677 :see_no_evil:

mitelg: but that will also be merged in a few minutes, so the rebase should really be necessary :grin:

themes/Backend/ExtJs/backend/voucher/controller/code.js
Dominikrt: Hi, to load the entire list with no input (show all voucher codes). 
The filter functions do a store reload, its not needed to reload the store manually.

mitelg (approved):

mitelg:
thanks again for your contribution @Dominikrt 👍🎉💙

PR #2677, opened 2024-04-30
fix: backend voucher codes stuck on last page and shows old entry's

Dominikrt:
the voucher individual code list keep the page of the last voucher. If you open a voucher with less codes, the list will be empty, if you stay on a not longer existing page.

Every time opening a voucher, the page will set to 1.
To not showing outdated results, the store get cleared.

### 6. Checklist


mitelg (approved):

mitelg:
thanks for your contribution @Dominikrt 👍🎉💙

PR #2688, opened 2024-05-24
Fix/version compare deprecation

aragon999:
This PR fixes two minor issues, on the one hand a typo in a command. On the other hand a PHP deprecation:
```
PHP Deprecated:  version_compare(): Passing null to parameter #2 ($version2) of type string is deprecated in .../engine/Shopware/Components/Plugin/Bootstrap.php on line 112
```
which occurs when a plugin from the legacy plugin structure is not present in the `s_core_plugins` table.

Fix those issues.

\-

\-

\-

### 6. Checklist

### DIFF
--- a/engine/Shopware/Components/Plugin/Bootstrap.php
+++ b/engine/Shopware/Components/Plugin/Bootstrap.php
@@ -98,8 +98,11 @@ abstract class Shopware_Components_Plugin_Bootstrap extends Enlight_Plugin_Boots
     public function hasInfoNewerVersion(Enlight_Config $updatePluginInfo, Enlight_Config $currentPluginInfo)
     {
         $currentVersion = $currentPluginInfo->get('version');
-        $updateVersion = $updatePluginInfo->get('version');
+        if ($currentVersion === null) {
+            return true;
+        }
 
+        $updateVersion = $updatePluginInfo->get('version');
         if (empty($updateVersion)) {
             return false;
         }
--- a/tests/Functional/Components/Plugin/BootstrapTest.php
+++ b/tests/Functional/Components/Plugin/BootstrapTest.php
@@ -25,6 +25,7 @@ declare(strict_types=1);
 
 namespace Shopware\Tests\Functional\Components\Plugin;
 
+use Enlight_Config;
 use PHPUnit\Framework\TestCase;
 use Shopware\Tests\Functional\Components\Plugin\fixtures\TestPlugin;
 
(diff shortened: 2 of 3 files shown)


engine/Shopware/Components/Plugin/Bootstrap.php
```diff
@@ -98,8 +98,11 @@ public function __construct($name, $info = null)
     public function hasInfoNewerVersion(Enlight_Config $updatePluginInfo, Enlight_Config $currentPluginInfo)
     {
         $currentVersion = $currentPluginInfo->get('version');
-        $updateVersion = $updatePluginInfo->get('version');
+        if (empty($currentVersion)) {
+            return true;
```
mitelg, line 102: why don't you return false here? :thinking: 

and is there not a serious problem with the plugin, if the current version is not set? how could that happen?

mitelg (changes requested):

engine/Shopware/Components/Plugin/Bootstrap.php
aragon999: It only happens when a plugin is added, and has not yet been put into the `s_core_plugin` table, i.e. on the first plugin refresh, after the plugin has been added on the filesystem. Hence I thought, okay in that case always a new version is present. But maybe false is also more reasonable, and probably I should add a comment. 

What do you think?

mitelg: I guess we should give the same value back, as the call of `version_compare` with `null` is giving back.

maybe you can add a small test case here https://github.com/shopware5/shopware/blob/5.7/tests/Functional/Components/Plugin/BootstrapTest.php so it is ensured, that the behaviour stays the same

aragon999: Yes the value is always true, if the second argument is `null`:
image

I will look into the tests though :-)

mitelg: so returning true is okay? :grin:

aragon999: Yes, `true` is okay :-) I also added a test.

tests/Functional/Components/Plugin/BootstrapTest.php
```diff
@@ -37,4 +38,32 @@ public function testInstantiation(): void
...
+
+        static::assertSame($expected, $pluginBootstrap->hasInfoNewerVersion($updatePluginConfig, $currentPluginConfig));
+    }
+
+    /**
+     * @return list<array{0: ?string, 1: ?string, 2: bool}>
+     */
+    public function pluginVersionProvider(): array
+    {
+        // current version, update version, expected
```
mitelg: instead of writing a comment, you can also make the arrays associative with a speaking array key

```
        return [
            [
                'currentVersion' => '2.0.0',
                'updateVersion' => '2.0.0',
                'isNewer' => false
            ],
            [
                'currentVersion' => '1.0.0',
                'updateVersion' => '2.0.0',
                'isNewer' => true
            ],
            [
                'currentVersion' => null,
                'updateVersion' => '2.0.0',
                'isNewer' => true
            ],
            [
                'currentVersion' => '2.0.0',
                'updateVersion' => null,
                'isNewer' => false
            ],
            [
                'currentVersion' => '1',
                'updateVersion' => '1.0.0',
                'isNewer' => false
            ],
        ];
```

tests/Functional/Components/Plugin/BootstrapTest.php
```diff
@@ -37,4 +38,32 @@ public function testInstantiation(): void
         $pluginName = $pluginInstance->getName();
         static::assertSame($expectedPluginName, $pluginName);
     }
+
+    /**
+     * @dataProvider pluginVersionProvider
+     */
+    public function testHasNewerVersion(?string $currentVersion, ?string $updateVersion, bool $expected): void
```
mitelg:
```suggestion
    public function testHasNewerVersion(?string $currentVersion, ?string $updateVersion, bool $isNewer): void
```

engine/Shopware/Components/Plugin/Bootstrap.php
```diff
@@ -98,8 +98,11 @@ public function __construct($name, $info = null)
     public function hasInfoNewerVersion(Enlight_Config $updatePluginInfo, Enlight_Config $currentPluginInfo)
     {
         $currentVersion = $currentPluginInfo->get('version');
-        $updateVersion = $updatePluginInfo->get('version');
+        if (empty($currentVersion)) {
```
mitelg: as you are using `empty` here, we should also check how it behaves with empty string or `string(0)`

mitelg (changes requested):

mitelg:
thanks for the update! just a few smaller improvement suggestions

engine/Shopware/Components/Plugin/Bootstrap.php
aragon999: Ah yes, that would have been a "breaking" change :-) I adjusted the comparison to strictly compare against `null`. Furthermore I added two more test cases.

mitelg (approved):

mitelg:
thank you very much for your contribution @aragon999 👍🎉💙

it will be merged in a minute, it is set to auto-merge once the test have passed

PR #2695, opened 2024-05-29
fix: backend notification customer list stuck on last seen page

Dominikrt:
Hi,

The backend voucher customer list stay on last page by switching to another product.
The result is an empty page if the new product has less subscriber.

Every-time switching a product, the subscriber list will set to page 1

1. open a product with more than 21 subscriber
2. switch to the next customer page
3. switch to another product with less subscriber

### 6. Checklist


mitelg (approved):

mitelg:
thanks again for your contribution @Dominikrt 👍🎉💙

PR #2702, opened 2024-06-09
implement fix for wrong sAddresses data issue. consider userId to get…

seriewe:
… unique order if there are multiple orders with same orderid

If multiple shops has not correctly setup the order nuber ranges and they overlap customer see billing data from other customer on /checkout/finish page. 

the change consider the userId and orderId to get the unique oder for the customer which does the checkout

If you have two stores with same odernumber ranges and for example Store A hast already the order 45500 and the second store now also has a new order with 45500, because of same order number ranges, the user from Store B will see the address data of user from Store A because the Method: getOrderAddress from Controller Shopware_Controllers_Frontend_Checkout just search for oder ID and does not consider Store oder User!

### 6. Checklist


engine/Shopware/Controllers/Frontend/Checkout.php
```diff
@@ -373,8 +373,8 @@ public function finishAction()
                 $orderVariables = $this->session['sOrderVariables']->getArrayCopy();
 
                 if (!empty($orderVariables['sOrderNumber'])) {
-                    $orderVariables['sAddresses']['billing'] = $this->getOrderAddress($orderVariables['sOrderNumber'], 'billing');
-                    $orderVariables['sAddresses']['shipping'] = $this->getOrderAddress($orderVariables['sOrderNumber'], 'shipping');
+                    $orderVariables['sAddresses']['billing'] = $this->getOrderAddress($orderVariables['sOrderNumber'], Shopware()->Session()->get('sUserId'), 'billing');
```
mitelg: the customer ID is now used in many places. would be nice, if you could introduce a local variable for it. 
also please avoid using the `Shopware()` singleton. 
the session is also available as class property.

engine/Shopware/Controllers/Frontend/Checkout.php
```diff
@@ -2131,7 +2131,7 @@ private function isValidAddress(int $addressId, bool $isShippingAddress = false)
     /**
      * @return array<string, mixed>
      */
-    private function getOrderAddress(string $orderNumber, string $source): array
+    private function getOrderAddress(string $orderNumber, int $userId, string $source): array
```
mitelg: I know it is called like this, but to be clear, we should name it `customerId`

mitelg (changes requested):
thanks for your contribution! some test would be nice

engine/Shopware/Controllers/Frontend/Checkout.php
seriewe: changed

engine/Shopware/Controllers/Frontend/Checkout.php
seriewe: changed

mitelg:
hey @seriewe 

are you still on this one? would be nice to see some test validating your changes :slightly_smiling_face:

seriewe:
@mitelg I'm sorry had still not time. Does it still make sense for me to submit this in the coming weeks?

mitelg:
no worries, we can leave it open for now. maybe someone else would like to jump in :blush: or when you have the time

PR #2737, opened 2024-07-21
fix: backend article preview 'shop' cookie

Dominikrt:
If you have a subshop with a different domain and click on the preview button in the backend, you are no longer able to visit the main shop on the main domain. You get redirected to the subshop all the time.

The `shop` cookie get a ttl from 30 seconds. after this time you can visit the main shop again.

Setup a subshop with a different domain and show the preview of an article in this subshop. you are no longer able to visit the main shop.

This bug works only over https!

### 6. Checklist

### DIFF
--- a/engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php
+++ b/engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php
@@ -29,6 +29,7 @@ use Shopware\Models\Customer\Group as CustomerGroup;
 use Shopware\Models\Shop\Shop;
 use Shopware\Models\Shop\Template;
 use Symfony\Component\HttpFoundation\Cookie;
+use Symfony\Component\HttpFoundation\Response;
 
 class Shopware_Plugins_Core_Router_Bootstrap extends Shopware_Components_Plugin_Bootstrap
 {
@@ -53,6 +54,7 @@ class Shopware_Plugins_Core_Router_Bootstrap extends Shopware_Components_Plugin_
     public function onRouteStartup(Enlight_Controller_EventArgs $args)
     {
         $request = $args->getRequest();
+        $response = $args->getResponse();
 
         if (str_starts_with($request->getPathInfo(), '/backend')
             || str_starts_with($request->getPathInfo(), '/api/')
@@ -60,7 +62,7 @@ class Shopware_Plugins_Core_Router_Bootstrap extends Shopware_Components_Plugin_
             return;
         }
 
-        $shop = $this->getShopByRequest($request);
+        $shop = $this->getShopByRequest($request, $response);
 
         if (!$shop->getHost()) {
             $shop->setHost($request->getHttpHost());
@@ -306,7 +308,7 @@ class Shopware_Plugins_Core_Router_Bootstrap extends Shopware_Components_Plugin_
      *
      * @return Shop
      */
-    protected function getShopByRequest(Request $request)
+    protected function getShopByRequest(Request $request, Response $response)
     {
         $repository = $this->get(ModelManager::class)->getRepository(Shop::class);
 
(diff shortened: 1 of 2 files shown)


mitelg:
this looks like a "hacky" workaround, and not really like a solution to the problem :grimacing: 
Is there no other possibility? couldn't this cookie be deleted once it is read? :thinking:

Dominikrt:
Hi @mitelg ,
I do it this way, to keep compatibility for 3rd party solutions. I don't know, if this cookie is needed somewhere else.

[here](https://github.com/shopware5/shopware/compare/5.7...Dominikrt:shopware5:fix-shop-cookie-core) is a solution with clearing the cookie after reading it. But i think this is a breaking change.

mitelg:
I would have to dig deeper into the code, but if I remember correctly the cookie is only used, to select the current shop from the cookie. I hope, that no one is relying on it :see_no_evil: 

your change is also breaking, just after 30 seconds :grin:

mitelg:
@Dominikrt any news on this?

Dominikrt:
Hi @mitelg,
i pushed now the other branch in this, **I think this is a breaking change!**

But I don't know if there is a better way doing this.

engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php
```diff
@@ -317,6 +319,7 @@ protected function getShopByRequest(Request $request)
 
         if ($shop === null && $request->getCookie('shop') !== null) {
             $shop = $repository->getActiveById($request->getCookie('shop'));
+            $response->headers->clearCookie('shop');
```
mitelg, line 322: clearing the cookie here, would also affect the next if condition, if I see that correctly. 

Before your change, the if condition could also be met, if this condition was met before. 

With your change this is no longer possible. 

A test of the `\Shopware_Plugins_Core_Router_Bootstrap::onRouteStartup` method in `\Shopware\Tests\Functional\Plugins\Core\Router\BootstrapTest` would be really nice, to ensure, that we do not break stuff

mitelg (changes requested):

mitelg:
@Dominikrt thanks! this looks better to me, but still a little question mark :grin:

tests/Functional/Plugins/Core/Router/BootstrapTest.php
```diff
@@ -52,4 +53,32 @@ public function testOnRouteShutdown(): void
...
+        $request->setCookie('shop', 2);
+        $response = new Enlight_Controller_Response_ResponseTestCase();
+        $args = new Enlight_Controller_EventArgs([
+            'request' => $request,
+            'response' => $response,
+        ]);
+        $pluginBootstrap->onRouteStartup($args);
+
+        static::assertSame(Response::HTTP_OK, $response->getStatusCode());
+
```
Dominikrt, line 72: @mitelg, here 2 options for testing this, I'm not happy with both solutions

engine/Shopware/Plugins/Default/Core/Router/Bootstrap.php
Dominikrt: Hi, the main problem is, I don't understand the reason of this cookie. 

I tested this change on my server and I still get redirect after the first attempt, trying to visit the start page. Only on the second attempt is the Cookie gone and I can visit the primary shop again.

But I don't know, if there is a case that the cookie is needed to redirect to the right shop.

mitelg: do we need the cookie at all then? :thinking: 

unfortunately I cannot reproduce this locally. you wrote this only occurs via HTTPS? :thinking: If I am using only HTTP, the cookie is not set at all :thinking:

PR #2749, opened 2024-08-19
Update phpstan to 1.11.11

aragon999:
Supersedes https://github.com/shopware5/shopware/pull/2744

### DIFF
--- a/engine/Shopware/Controllers/Backend/Export.php
+++ b/engine/Shopware/Controllers/Backend/Export.php
@@ -123,9 +123,8 @@ class Shopware_Controllers_Backend_Export extends Enlight_Controller_Action impl
     private function generateExport($output)
     {
         $outputHandle = fopen($output, 'w');
-
-        if (!\is_resource($outputHandle)) {
-            throw new RuntimeException(sprintf('Output file %s can not be opened', $outputHandle));
+        if ($outputHandle === false) {
+            throw new RuntimeException(sprintf('Output file %s can not be opened', $output));
         }
 
         $this->export->sSmarty = $this->View()->Engine();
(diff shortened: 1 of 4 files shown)


engine/Shopware/Controllers/Backend/Export.php
```diff
@@ -123,9 +123,8 @@ public function getWhitelistedCSRFActions()
     private function generateExport($output)
     {
         $outputHandle = fopen($output, 'w');
-
-        if (!\is_resource($outputHandle)) {
-            throw new RuntimeException(sprintf('Output file %s can not be opened', $outputHandle));
+        if ($outputHandle === false) {
```
mitelg, line 126: why did you change this? I like it more, if we explicitly check for what we need.

mitelg (changes requested):

engine/Shopware/Controllers/Backend/Export.php
aragon999: So maybe this is a more philosophical question, but in this case the docs for [`fopen`](https://www.php.net/manual/en/function.fopen.php) state that
> Returns a file pointer resource on success, or [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) on failure

So the return type is either a `resource` or `false`. As this `if` condition should only catch the error case, we should explicitly check for `false` as other cases (which currently do not exist) should be handled differently.

I can also revert the check in the `if` condition.

mitelg: is fine for me, the check for is_resource seems to be kind of tricky anyway. see https://github.com/phpstan/phpstan/issues/11498#issuecomment-2285536288

mitelg (approved):

PR #2757, opened 2024-09-08
fix: customer group discount tax

### LINKED ISSUE #2608: Static fallback to 19% in basket
Instructions for reproduction:

* Basic settings > Create a shopping cart discount for customer groups, e.g. 10%.
* Create a tax rule for this customer group and define 0%-> Shopping cart discount is still calculated with 19%.

The cause is this query: [https://github.com/shopware5/shopware/blob/v5.7.18/engine/Shopware/Core/sBasket.php#L425-L427](https://github.com/shopware5/shopware/blob/v5.7.18/engine/Shopware/Core/sBasket.php#L425-L427)

If you change this e.g. to (!$tax && $tax != 0) then it also works with 0% tax.

Dominikrt:
close #2608 

Tax for Customer Group Discount can be 0%, currently there is a fallback for 19%

check issue

check issue

#2608 

### 6. Checklist

### DIFF
--- a/engine/Shopware/Core/sBasket.php
+++ b/engine/Shopware/Core/sBasket.php
@@ -421,7 +421,7 @@ class sBasket implements Enlight_Hook
             $tax = $this->config->get('sDISCOUNTTAX');
         }
 
-        if (!$tax) {
+        if (!$tax && $tax != 0) {
             $tax = 19;
         }
 
--- a/tests/Functional/Bundle/StoreFrontBundle/Helper.php
+++ b/tests/Functional/Bundle/StoreFrontBundle/Helper.php
@@ -27,6 +27,8 @@ namespace Shopware\Tests\Functional\Bundle\StoreFrontBundle;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\DBAL\Connection;
+use Doctrine\ORM\Exception\ORMException;
+use Doctrine\ORM\OptimisticLockException;
 use Enlight_Components_Db_Adapter_Pdo_Mysql;
 use Shopware\Bundle\StoreFrontBundle\Service\ListProductServiceInterface;
 use Shopware\Bundle\StoreFrontBundle\Struct\Customer\Group as CustomerGroupStruct;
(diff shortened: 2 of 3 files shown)


engine/Shopware/Core/sBasket.php
```diff
@@ -421,7 +421,7 @@ public function sInsertDiscount()
             $tax = $this->config->get('sDISCOUNTTAX');
```
aragon999, line 421:
```suggestion
            $tax = (int) $this->config->get('sDISCOUNTTAX');
```

engine/Shopware/Core/sBasket.php
```diff
@@ -421,7 +421,7 @@ public function sInsertDiscount()
...
         }
 
-        if (!$tax) {
+        if (!$tax && $tax != 0) {
```
aragon999, line 424:
```suggestion
        if ($tax === false) {
```

aragon999 (commented):
I think I would make the check more explicit by explicitly checking that the tax is `false`. Maybe the `if` can also be moved into the `if` branch of the `getMaxTax()`.

engine/Shopware/Core/sBasket.php
Dominikrt: Hi @aragon999 ,
this way looks much nicer, I wasn't sure, if there is another datatype that can trigger the fallback.

aragon999: As far as I can tell, the only case which needs to be handled is that `getMaxTax()` returns false.

It would still change the logic slightly, as when previously when the tax `0` has been configured in `sDISCOUNTTAX` the tax wont be `19` anymore. But that is I think in all cases like this.

engine/Shopware/Core/sBasket.php
mitelg: is this the tax value? if so, it should definitely not be an integer, but a float instead

tests/Functional/Bundle/StoreFrontBundle/Helper.php
```diff
@@ -361,6 +367,40 @@ public function createTax(array $data = []): TaxModel
         return $tax;
     }
 
+    /**
+     * @param array<string, string|int|Customer|Group|bool> $data
```
mitelg, line 371: an object shape array annotation would make more sense in this case

`array{foo: string, bar: float}`

tests/Functional/Bundle/StoreFrontBundle/Helper.php
```diff
@@ -361,6 +367,40 @@ public function createTax(array $data = []): TaxModel
...
+        $taxRule->fromArray($data);
+
+        $this->entityManager->persist($taxRule);
+        $this->entityManager->flush();
+
+        return $taxRule;
+    }
+
+    /**
+     * @param array<string, string|int|Customer|Group|bool> $data
```
mitelg, line 388: same as above

tests/Functional/Core/BasketTest.php
```diff
@@ -156,6 +159,44 @@ public function testsCheckBasketQuantitiesWithLowerQuantityThanAvailable(): void
...
+        $discount = $this->connection->fetchAssociative(
+            'SELECT * FROM s_order_basket WHERE sessionID = :sessionId AND ordernumber = :ordernumber',
+            [
+                'sessionId' => $this->getSessionId(),
+                'ordernumber' => 'sw-discount',
+            ]
+        );
+
+        static::assertIsArray($discount);
+        static::assertSame(0, (int) $discount['tax_rate']);
```
mitelg, line 185: tax rates should be floats!

engine/Shopware/Core/sBasket.php
mitelg: I would also prefer a type safe comparison. but then we need to max sure, that `$tax` is really a `float` or `false` at this point. 

additionally to that, I would like to remove the fallback to a hardcoded value :see_no_evil: 

what do you think? throwing an exception or getting a tax value from somewhere else?

mitelg (changes requested):

PR #2762, opened 2024-09-25
chore: Remove system default fonts due to license mismatches

mitelg:
### Why is this change necessary?

Remove providing of system fonts, which should be available everywhere anyway

### What does this change do, exactly?

Remove font files


shyim (approved):

PascalThesing (approved):

ennasus4sun (approved):

build/composer-post-install-cmd.sh
```diff
@@ -7,7 +7,6 @@ root_dir="$script_dir"/..
 
 set -x
 
-rm -rf -- "$root_dir"/vendor/mpdf/mpdf/ttfonts
```
mitelg, line 10: not removing those fonts, will let the package size explode :grimacing: 

any ideas @shyim @PascalThesing ?

shyim: you can still delete it whats the problem?

mitelg: if they are not present, this will happen https://github.com/shopware5/shopware/actions/runs/11034760852/job/30649668548

I guess, because there are no fonts anywhere anymore

shyim: ahh, then I would keep them 😅  or you delete all files except we had in Shopware itself

mitelg: those are not there :sweat_smile: https://github.com/mpdf/mpdf/tree/development/ttfonts

mitelg: I am thinking about only keeping the `DejaVu*` once. they seem to be not that big

mitelg: I changed it like written above. we have now 5MB more. I think that is okay. at least the tests are still running :shrug:

PR #2763, opened 2024-09-27
NTR - fix es backend index

teiling88:
Without this changes you cannot use it with elasticsearch 8

### 6. Checklist

### DIFF
--- a/engine/Shopware/Bundle/EsBackendBundle/EsBackendIndexer.php
+++ b/engine/Shopware/Bundle/EsBackendBundle/EsBackendIndexer.php
@@ -128,7 +128,6 @@ class EsBackendIndexer
         $this->client->bulk(
             [
                 'index' => $index,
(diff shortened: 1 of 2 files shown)


mitelg:
hey @teiling88 

thanks for you fix! Can we adjust our test pipeline, to make sure we stay compatible with ES8?

teiling88:
At the moment, you are not compatible to es 8 with the es backend. So I don't get your question. I see no test for the es backend wich is testing the integration to es 8.

mitelg:
there are a few :grimacing: https://github.com/shopware5/shopware/tree/5.7/tests/Functional/Bundle/EsBackendBundle

my idea was to change the ES version here https://github.com/shopware5/shopware/blob/5.7/.github/workflows/phpunit.yml#L161
I don't know, if this would be sufficient

PR #2765, opened 2024-10-01
fix: change z-index of element for compatibility with iOS 18

tappert-it:
Changed z-index to unset to avoid overlapping of blocks on mobile displays in iOS 18

Compatibility with iOS 18. Without the change overlapping of blocks can occur on Safari, so the cart button is not clickable anymore.

The z-index of the div with class "product--image-container" is before set to 1000. This change unsets the z-index.

Occurs after the update to iOS 18 on Safari and Chrome (and maybe more browsers). The image container overlaps the cart button.

https://forum.shopware.com/t/ios-safari-oder-chrome-detailseite-buttons-ohne-funktion/105237/

### 6. Checklist

### DIFF
--- a/themes/Frontend/Bare/frontend/detail/content.tpl
+++ b/themes/Frontend/Bare/frontend/detail/content.tpl
@@ -32,7 +32,7 @@
                     data-image-gallery="true"
                     data-maxZoom="{$theme.lightboxZoomFactor}"
                     data-thumbnails=".image--thumbnails"
-                    {/if}>
+                    {/if} style="z-index: unset;">
                     {block name="frontend_detail_index_image"}
                         {include file="frontend/detail/image.tpl"}
                     {/block}


aragon999:
I would like to avoid inline CSS. Is this also fixed, if the `z-index` is defined here: https://github.com/shopware5/shopware/blob/8520a8d742e55d664cf0d1dd6c009ec6074231dc/themes/Frontend/Responsive/frontend/_public/src/less/_modules/detail.less#L95-L103

To me it still sounds like a issue of floating elements in WebKit. But probably the fix is fine. But it should be added as a comment, that it has been added due to a bug in WebKit, otherwise someone may remove it, as it has no effect in other browser engines.

mitelg:
I also would like to avoid specific OS fixes 🤔 for me it still sounds like a bug like @aragon999 already mentioned. 

Also inline styles should not be the solution 🙂

tappert-it:
Defining the z-index in the detail.less should do the trick, i think.

As we can see in the forum, it's not the perfect solution, as long as iOS 18 seems a bit uncalculatable in this issue. But at least customers can reach the cart button.

mitelg:
Hey @tappert-it 

This PR was closed due to inactivity. In the meantime some other fixes were proposed in the forum, which sound more applicable. Maybe one of those can be provided instead.

PR #2829, opened 2025-01-26
fix: Looping of data by reference with null coalescing

aragon999:
When looping through an array, where a reference is used, the original data is not the same reference, when null coalescing is used, see: https://3v4l.org/oU3is

Hence the data is not properly modified.

Check if the values are a non-empty array and remove the null coalescing.

\-

\-

\-

### 6. Checklist


mitelg (approved):

mitelg:
thanks for your contribution @aragon999 👍🎉💙

PR #2858, opened 2025-06-25
fix: removes unused polyfills from replace section

### LINKED ISSUE #2853: Drop polyfill replace composer.json
### Description

I want to depend on shopware/shopware as a dev requirement in my unit testing, however my package requires another dependency that conflicts with the polyfills shopware replaces.

More details can be found in the discussion at https://github.com/composer/composer/issues/12412

While i get that usually no one would pull in shopware as a dependency and those polyfilly should be fine, i really just want to have type definitions in my IDE in a dev setup without needing to create an elaborated setup. It would be cool if the replace entries could be dropped, because as just mentioned: Usually projects don't pull in shopware as a dependency.

### PHP Version

8.2

### Shopware Version

5.7.18

### How to reproduce

See https://github.com/composer/composer/issues/12412

### Expected behaviour

I can install my project with `composer install --no-dev` and no shopware is installed

### Actual behaviour

I install my project with `composer install --no-dev` and shopware is installed

kevin-kortum-trustedshops:
shopware5 has chosen to add polyfill replacements for libraries that are not used or not relevant to be replaced. While there is nothing wrong with it skipping polyfills, if there a package requiring it, there will be a conflict.

Considering #2853 there might be use cases where someone would like to add shopware5 as a dev requirement, but will fail if there is conflicts due to the polyfill. Although time has told that this is very unlikely (or was never reported) the docs of symfony/polyfill https://github.com/symfony/polyfill?tab=readme-ov-file#design recommend only to replace them if provided by the package or if they can safely be skipped e.g. of an unsupported php version.

Instead of not installing polyfills, it will not install those into the vendor directory. For PHP 7.4 it will increase the vendor directory by ~33% (42 MB) as for PHP 8.2 only by ~0,12% (2 MB).

Please refer to https://github.com/composer/composer/issues/12412 . 

#2853

None required

### 6. Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -93,11 +93,8 @@
     },
     "replace": {
         "paragonie/random_compat": "*",
-        "symfony/polyfill-ctype": "*",
         "symfony/polyfill-php72": "*",
-        "symfony/polyfill-php73": "*",
-        "symfony/polyfill-iconv": "*",
-        "symfony/polyfill-mbstring": "*"
+        "symfony/polyfill-php73": "*"
     },
     "suggest": {
         "ext-apcu": "*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "dce98e28d81b7c9aa07575f30a7e65bb",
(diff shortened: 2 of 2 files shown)


mitelg (approved):

mitelg:
thanks for your contribution @kevin-kortum-trustedshops 👍🎉💙

PR #2863, opened 2025-07-17
fix: Update devenv configuration for current devenv versions in 2025

jhit:
Remember to run `devenv update` once before running `devenv up`!

The current devenv.nix will not work in updated Nix/devenv environment or a new install of devenv.

It pins the Nix versions to the last version that includes NodeJS v18.

just checkout and run `devenv up`

### 6. Checklist

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -1,10 +1,16 @@
 { pkgs, lib, inputs, config, ... }:
-
+let
+  # Use an older nixpkgs-version, containing Node 18.
+  oldPkgs = import (fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/nixos-23.11.tar.gz";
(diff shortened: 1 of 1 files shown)


devenv.nix
```diff
@@ -1,10 +1,16 @@
 { pkgs, lib, inputs, config, ... }:
-
+let
+  # Verwende eine ältere nixpkgs-Version, die noch Node 18 enthält
```
mitelg: please do not use German comments

mitelg (changes requested):
thanks for your PR :+1: 
please have a look at your commit message. we are using conventional commits. adjust the message accordingly :wink:

mitelg (approved):

mitelg:
thanks for your contribution @jhit 👍🎉💙

